PR #18, opened 2013-07-08
Fix index query so it is slightly less worse

mp15:
Fix index query so it is slightly less worse and doesn't bail out when first or last target has no entries (it will still bail if no reads are mapped)


jmarshall:
Let's add a test case to the test suite to exercise this with some unpopulated reference chromosomes.

jkbonfield:
Is this related to the issue @pd3 fixed recently when iterators on an empty BCF file could fail?

This PR is rather dormant, so we should check if it's still an issue and close if not (and fix if so).

mp15:
5c695a174f74f5785bb5a6b27713d234147dbdb8 does the same thing but better. Closed

PR #51, opened 2014-01-03
Improve concurrency in multithreaded BGZF writer (bgzf_mt)

mlin:
This PR includes two commits that significantly increase compression throughput with the multithreaded BGZF writer (bgzf_mt), improving the performance of samtools sort, merge, and view -b in multithreaded mode. 

The first commit yields most of the performance benefit by continuing to ingest data while multithreaded compression proceeds in the background. This change does not complicate things very much and hopefully should be fairly uncontroversial.

The second commit, which yields a smaller improvement, is a bit riskier because it introduces writing to the file handle from different threads. It currently works out that the writes are serialized without needing an explicit mutex, but this could be a maintenance concern going forward.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -2,6 +2,7 @@
 
    Copyright (c) 2008 Broad Institute / Massachusetts Institute of Technology
                  2011 Attractive Chaos <attractor@live.co.uk>
+                 2014 DNAnexus, Inc.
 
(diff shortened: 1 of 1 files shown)


jkbonfield:
Obsoleted by the thread pool interface.  Sorry this hung around for so long without commenting.

Issue #98, opened 2014-05-28
htslib fails to build on mips/mipsel


azlicic:
Hello,
Your package htslib fails to build on Debian for mips/mipsel architectures.

I have already reported a bug on Debian bug tracking system at:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=749367,
with proposed patches attached.

Build logs for mips and mipsel are available at
https://buildd.debian.org/status/fetch.php?pkg=htslib&arch=mips&ver=0.2.0~rc3-1&stamp=1386050228
and
https://buildd.debian.org/status/fetch.php?pkg=htslib&arch=mipsel&ver=0.2.0~rc3-1&stamp=1386204701

Current version of htslib on Debian is 0.2.0~rc3-1.

I am currently adapting the patch to fit the latest htslib version.

I will send you a pull request as soon as I'm done.

Best Regards
Aleksandar Zlicic

xnox:
Hi, these failures affect armhf, powerpc and s390x builds too. W.r.t. build and testing, all debian developers have access to such machines and can run occasional builds for you. If you want access yourself you can request access to debian porter boxes, after agreeing to the Debian Machine Usage Policy (?! can't remember exact name) and somebody sponsoring (a DD recommending you) your access. Can we please resolve this? At the moment there is a chain of dependencies that fails to build due to this. blasr in debian stopped using bundled libs, and thus now depends on pbseqlib and htslib to be available on more arches. Specifically armhf and big-endian systems (powerpc, s390x). What's outstanding to resolve with this merge proposal?

daviesrob:
There are some forks that should work on big-endian platforms in:

https://github.com/daviesrob/samtools/tree/SPARC
https://github.com/daviesrob/htslib/tree/SPARC

but I'm afraid they're now well out of date.  Unfortunately more pressing 
things got in the way so it's been ignored for a while and would now need 
a bit of rework to get it merged.

How much demand is there for this?  As it's not something we need 
ourselves it gets a low priority at the moment.



-- 
 The Wellcome Trust Sanger Institute is operated by Genome Research 
 Limited, a charity registered in England with number 1021457 and a 
 company registered in England with number 2742969, whose registered 
 office is 215 Euston Road, London, NW1 2BE.

mr-c:
`mipsel` and `mips64el` have been successfully building for many years on Debian, so this issue can be closed:

https://buildd.debian.org/status/logs.php?pkg=htslib&arch=mips64el
[...]

PR #99, opened 2014-05-28
Support for mips/mipsel

azlicic:
Fix for issue #98

Avoid unaligned memory access on architectures that don't support it.
Fixed endianness related issues.

### DIFF
--- a/htslib/hts.h
+++ b/htslib/hts.h
@@ -4,6 +4,8 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#include "cram/os.h"
+
 #ifndef HTS_BGZF_TYPEDEF
 typedef struct BGZF BGZF;
 #define HTS_BGZF_TYPEDEF
@@ -280,7 +282,17 @@ static inline uint16_t ed_swap_2(uint16_t v)
 }
 static inline void *ed_swap_2p(void *x)
 {
+#ifdef ALLOW_UAC
 	*(uint16_t*)x = ed_swap_2(*(uint16_t*)x);
+#else
+	uint8_t tmpData[2];
+	uint16_t *ptmpData = (uint16_t*)&tmpData;
+	uint8_t *px = (uint8_t*)x;
+	int j;
+	for(j=0;j<2;j++) tmpData[j] = px[j];
+        *ptmpData = ed_swap_2(*ptmpData);
+	for(j=0;j<2;j++) px[j] = tmpData[j];
+#endif
 	return x;
 }
 static inline uint32_t ed_swap_4(uint32_t v)
@@ -290,7 +302,17 @@ static inline uint32_t ed_swap_4(uint32_t v)
 }
 static inline void *ed_swap_4p(void *x)
 {
+#ifdef ALLOW_UAC
 	*(uint32_t*)x = ed_swap_4(*(uint32_t*)x);
+#else
+	uint8_t tmpData[4];
+	uint32_t *ptmpData = (uint32_t*)&tmpData;
+	uint8_t *px = (uint8_t*)x;
+	int j;
+	for(j=0;j<4;j++) tmpData[j] = px[j];
+	*ptmpData = ed_swap_4(*ptmpData);
+	for(j=0;j<4;j++) px[j] = tmpData[j];
+#endif
 	return x;
 }
 static inline uint64_t ed_swap_8(uint64_t v)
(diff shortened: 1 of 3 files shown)


mp15:
Where are we with solving this issue?

jkbonfield:
It was rewritten by myself and then furthered by Rob:

https://github.com/daviesrob/htslib/tree/SPARC

However I don't think there was any inclination to incorporate it, so it got dropped and is now way out of date.

daviesrob:
Looking at https://buildd.debian.org/status/package.php?p=htslib, 1.3.1 is fine on little-endian platforms (including mipsel), but fails on big-endian.  I think big-endian support is partly present in htslib, so it might be possible to revive it.  Not sure how much work it would be though.

tillea:
Hi,
the initialy suggested patch for the problem was not accepted since more than two years.  Do you have any other solution to fix the build issue on armhf?
Kind regards, Andreas.

jkbonfield:
On Sun, Dec 11, 2016 at 12:16:13AM -0800, Andreas Tille wrote:
> the initialy suggested patch for the problem was not accepted since more than two years.  Do you have any other solution to fix the build issue on armhf?

I think for now it's fair to say there has been little appetite in
supporting less well used architectures, in favour of spending time on
more serious issues and wish-list items (of which there are still
plenty to work through!) and I don't really see that changing any time
soon.

It's unfortunate that we didn't create a PR of the rewrite of this
(https://github.com/daviesrob/htslib/tree/SPARC by Rob and myself) and
merge it promptly as that too is now so old it would be a large amount
of work to integrate (and so unlikely to happen).

Can you make a real case for these platforms being important to the
bioinformatics community?  If so it may be possible to reevaluate the
priorities.

James
-- 


-- 

tillea:
On Mon, Dec 12, 2016 at 01:36:56AM -0800, James Bonfield wrote:
> 
> Can you make a real case for these platforms being important to the
> bioinformatics community?  If so it may be possible to reevaluate the
> priorities.

No, I can't except that in Debian we try to get all software that is
possible released on all architectures.  Considering that there was a
patch that should enable this I was simply checking the status of this
issue.  If you explicitly state that htslib is not and will not be
supported on armhf I would add this to the package metadata and close
the bug in the Debian bug tracking system.  Its just not sensible to
leave a bug with a patch open for years and do nothing. :-)

Kind regards

      Andreas.

-- 
http://fam-tille.de

daviesrob:
Actually armhf is little-endian, so it should in theory work fine.  And in fact when I last tried it using qemu (so not a _real_ armhf machine) it worked for me.  See https://github.com/samtools/samtools/issues/612

If we can get into your build system and can have a poke around we might be able to find out what's going on.  Without that, we're a bit stuck until we can get something to test on.

plugwash:
When it comes to alignment issues qemu favours performance over correctness. So code with alignment issues that will fail on real hardware will often work in qemu.

tillea:
On Mon, Dec 12, 2016 at 03:13:28PM -0800, daviesrob wrote:
> Actually armhf is little-endian, so it should in theory work fine.  And in fact when I last tried it using qemu (so not a _real_ armhf machine) it worked for me.  See https://github.com/samtools/samtools/issues/612
> 
> If we can get into your build system and can have a poke around we might be able to find out what's going on.  Without that, we're a bit stuck until we can get something to test on.

Hi,

I have asked how to gain access to a real armhf box.  Hope this helps:

On Tue, Dec 13, 2016 at 7:12 PM, peter green <plugwash@p10link.net> wrote:
> On 13/12/16 07:37, Andreas Tille wrote:
>> when discussing bug #800469 upstream thinks that armhf should work and
>> the question came up whether there is some chance to access an armhf
>> machine that is comparable to our autobuilders.
>>
>
> The porterbox abel.debian.org runs the same hardware as the buildds and has
> both armel and armhf chroots available.
>
> There is a procedure for guest accounts on porterboxes, though I haven't
> used it myself so I don't know how well it works in practice.

The process has simplified, and just visit: https://nm.debian.org/wizard/
and click "Request guest account".

Just curious that this issue cannot be reproduced by pbuilder or
similar on self arm boxes?

Cheers,
-- 
Roger Shimizu, GMT +9 Tokyo
PGP/GPG: 4096R/6C6ACD6417B3ACB1

-- 
http://fam-tille.de

daviesrob:
I'm not in the office at the moment, but I can try digging out a local Debian developer when I get back if that helps.  Assuming there's any left one week before Christmas...

I thought modern arm cpus could manage unaligned access.  This could be a problem if not.

plugwash:
On 13/12/16 23:41, daviesrob wrote:
>
> I thought modern arm cpus could manage unaligned access. This could be 
> a problem if not.
>
They can for integer but not for floating point.

domibel:
The issue can be reproduced on a Raspberry PI.
See also the patch mentioned here:
    https://github.com/samtools/samtools/issues/612#issuecomment-266706309

PR #115, opened 2014-07-23
Integrated igzip library to enable fast level 1 compression of BAM files.

mishalinaik:
New pull request to resolve issues with Travis CI

igzip enables faster level 1 compression on BAM files compared to zlib while providing similar compression ratio as zlib. (https://software.intel.com/en-us/articles/igzip-a-high-performance-deflate-compressor-with-optimizations-for-genomic-data).
Response to jmarshall:

1) Compiler reverted to gcc by including the stdc++ library.
2) Before this compression library was customized to Genomics it has been used in storage acceleration https://01.org/storage-acceleration-library, we don't expect the API to change.
3) Happy to create another branch, but would be great if the latest commit is first tested and reviewed.


dalexander:
will this be merged?

armintoepfer:
A merge would be more than welcomed. Thanks!

wookietreiber:
Is it really necessary to include the `*.a` archive, can't this be built from source and integrated into `Makefile`?

jkbonfield:
We cannot just accept .a files clearly, but would instead need some Makefile or autoconf equivalent to link against a pre-installed igzip package. 

However I am unsure of what this really buys us.  What are the things we are trying to address by using this library?

I tried the Intel optimised deflate (ie not BAM specific) and on a 1 billion byte BAM (not bgzf, just raw compression so startup costs may differ) and it's faster than the igzip code for -1. It doesn't compress as well (553508373 vs 430955563) but is perfectly adequate for fast temporary output.  The round-trip time using /dev/shm took 13.1s vs 16.9s.  If size matters, then intel zlib -2 had 414335449 sizwe for 22.4s, so the igzip is somewhere inbetween the two basically.  The advantage of using Intel's full zlib or CloudFlare's optimised version is it needs nothing more than an adjustment to LD_LIBRARY_PATH or LD_PRELOAD.

Alternatively, if we're looking at *internal* temporary file usage for things like samtools sort, then it doesn't have to be zlib compatible at all. We could use a fast generic format like LZ4 or Snappy, or an intermediate fast compressor like ZSTD.  On the same 1Gb file it ZSTD compressed it to 405947800 (smaller than the classic zlib -1) using 10.8s round trip, so smallest and fastest of all 3 tested.

jkbonfield:
Actually I take that back - that was izgip1c_BAM.  The 0c version is round trip of 11.3s and 446580855 bytes so clearly better than the standard intel optimised zlib.  Is this the version you envisage using?

igzip_042/include/igzip_lib.h
```diff
@@ -0,0 +1,50 @@
...
+    UINT8 *next_out;  // Next output byte
+    UINT32 avail_out; // number of bytes available at next_out
+    UINT32 total_out; // total number of bytes written so far
+    UINT32 end_of_stream; // non-zero if this is the last input buffer
+
+    LZ_State2 internal_state;
+} LZ_Stream2;
+
+
+void init_stream(LZ_Stream2 *stream);
```
mp15: This should probably have a:
```
extern "C" {
```
Wrapper

igzip_042/include/types.h
```diff
@@ -0,0 +1,46 @@
...
+
+#ifndef __TYPES_H
+#define __TYPES_H
+
+#include <stdint.h> // For standard integer types
+
+typedef struct {
+    uint64_t low;
+    uint64_t high;
+} uint128_t;
```
mp15: Could use __uint128_t on clang or gcc? Has ICC caught up with this yet?

avilella:
@jkbonfield any news on this PR? can it be turned into a branch from the main github.com/samtools/htslib repo?

igzip_042/include/types.h
mp15: ICC supports __uint128_t as of version 13.0.0 apparently.

igzip_042/include/types.h
```diff
@@ -0,0 +1,46 @@
+/**********************************************************************
```
mp15: Come to think of it given all of these types are now supported by modern compilers is this file even necessary?

mp15:
@avilella Couple of issues that are barring integration:

- As @jkbonfield said trying to lump .a binary blob into htslib is a non-starter, htslib needs to be cross platform (mainly linux of many flavours and os x, though we will probably put some more effort into windows again soon). If there's no way to release the source maybe we could work out a way of integrating this as a plugin?
- Adding libstdc++ to a C library seems a bit of a bad practice. There are better ways of cleanly exposing a c++ library to a c library.
- Finally have made some comments on the branch but nobody responded to my reviews.

igzip_042/include/types.h
jkbonfield: Possibly for 32-bit systems.  You'll find it hard to do 128bit integer maths on those as they don't have hardware support for a 128 bit register.

jkbonfield:
Firstly, it's not going to happen before 1.3.  Just no time.

Secondly as Martin points out it needs some work I think, but I haven't had time to look at this myself.  I don't know how much Martin played with it.  I think the correct thing to do is simply to unC++ify the entire lot (if it's possible, I haven't looked how extensively they used it) so it can then build sensibly without causing a major restructuring of htslib.

jkbonfield:
Is this still something you wish to exist in samtools?  If so we need more work, if not then please close it.

Currently I think it has the following issues.
- We cannot accept binary object files in pull requests.  We need either the source or a configure system for linking against an "external" dependency installed via a separate package.
- We need a way to enable / disable this component so it is easy to build on other systems.  You have X86_64 guards around your asm, but I'm not sure every system we'll build on may have the emmintrin.h. (Eg would ARM?) 
- It needs rebasing.  I attempted to do this, but it's messy as there are all sorts of tab related issues making the code look different and causing lots of conflicts.  I think I managed it though.  See: https://github.com/jkbonfield/htslib/tree/intel_igzip   Maybe you'd want to check I got it right.
- It needs tests.  As you have added a different code path for compression level 1 specifically then we need to have a set of tests that explicitly check level 1 works fine.
- Consider using a .so file for the igzip library (maybe fetched from elsewhere if you MUST go down the binary only release) and using load_plugin to fetch it from libexecdir ($HTS_PATH I think).  This was designed for registering new URI schemes and access methods (irods:, http via curl, etc).  Perhaps it could also be utilised for loading a custom zlib layer with a bit of tweaking.  This will also make samtools linking against an htslib.a work more robustly as the samtools Makefile doesn't need tweaking to add additional htslib dependencies. (It does right now it seems.)  It also neatly sidesteps the whole C vs C++ thing as the libigzip.so library itself has the C++ dependencies while htslib can just use it as-is.  (The fly in the ointment is if we wish to keep libhtslib.a, but that's sometihng we have to fix up via pkgconfig at some point.)

Finally, if you have a highly optimised level 1 zlib, then does this only work for BAM or is it applicable to other formats too (eg bcf, cram?).  I suspect BAM only due to custom huffman tables, unless there is something generic in there too.

NB: I haven't attempted to build this on any exotic hardware so I have zero testing there, but rest assured we WILL get bug reports from all the OS vendors who support these platforms if it doens't work.

cielavenir:
Well, C++ dependencies can be removed with small fix...
https://gist.github.com/cielavenir/2cb983fb7e1795de06f84aeb8b161481 (besides please rename cpp to c; this is a limitation of diff utility).
This will eliminate the requirement of .so or binary object files.

However, utilizing Makefile guard is still required (`We need a way to enable / disable this component`). This should be @mishalinaik's task (of course other people can do, but...)

jmarshall:
There is now [Genomics Kernel Library](https://github.com/Intel-HLS/GKL) which wraps igzip etc up into an externally-supported shared library, as we hoped for.  As noted in comments here and in the previous PR #112, we are not in a position to have igzip become part of htslib or maintained by htslib maintainers, but there is scope for enabling the functionality by calling your separate library.

We would probably enable this by providing a compression plugin mechanism and a plugin that used GKL.  As a step towards this, please convert this PR to a new PR that calls functions in an external _libIntelGKL.so_.  (Don't worry about configuration, as this will be taken care of by the plugin arrangement.)  We would then refactor that into a suitable plugin.

PR #133, opened 2014-09-24
remove code duplication from thread_pool.c

lomereiter:
`t_pool_dispatch` function is merely `t_pool_dispatch2` called with 'block' option


jmarshall:
@jkbonfield to reduce this to the desired change and apply.  Thanks!

jkbonfield:
Deduplication manually merged into 3fac82bc743e57bf34d3e8d1954e850db6cf2a8e without the other changes.

Shout if you want to make a case for the removal of the bit fields or changes to function visibility.

lomereiter:
Thanks, I should have used a feature branch.

The case against bitfields is that their physical allocation is unspecified, which makes it impossible to access them through FFI, for example. Real-life example of minor headache caused can be found in [pysam](https://github.com/pysam-developers/pysam/blob/0cef5f512fd9873714bf96bae200f123267be795/pysam/htslib_util.h#L94-L133).

PR #134, opened 2014-09-26
adding back support for printing the flag field as a string when

nh13:
outputting to SAM

[Review on Reviewable](https://reviewable.io/reviews/samtools/htslib/134)


nh13:
@jmarshall closing this, though I really do want samtools to once again print out the flags, as I use this feature regularly.

PR #146, opened 2014-11-10
RFC:   iRODS support with autoconf

keithj:
An initial pass at adding configuration support for iRODS using autoconf. The intention
is that aside from running autoconf, all previous Makefile behaviour is preserved.

* The original Makefile becomes Makefile.in, preserving the custom
   build behaviour. The package version and SO version are configured
   in configure.ac.
    
* iRODS is off by default and enabled using --with-irods=/path/to/irods
  or just --with-irods where a default path is used.
    
* A -O option has been added to test_view and
   "test_view.pl irods" will run test_view.pl as normal, but writing
   its results directly into iRODS.
    
* The Travis config is updated to run configure.


jmarshall:
Thanks for pushing this, Keith!

I've now merged the iRODS backend.  The eventual configury, ermm..., doesn't bear a lot of resemblance to the configury in your pull request, but yours was very useful as a proof-of-concept and to crib parts from.

I haven't merged the test harness stuff, as that seems somewhat separate (and perhaps bound up with  our local environment?).  Please resurrect it as a new pull request if you wish, and we can merge it (more quickly this time!) post-1.2.

keithj:
Thanks for merging.

Regarding the tests - they are independent of our local environment. Of course, they could be destructive if run somewhere else without care (the post-test cleanup target will irm files matching a pattern in the iRODS cwd). All they do is allow the existing suite of test_view tests to be run while writing the data into iRODS. It's somewhat ugly, but I wanted some tests for the pull request while still focussing mainly on the autoconf part.

PR #157, opened 2014-12-15
Cigar P support

jkbonfield:
Adds a bam_plp_insertion() function that queries the cigar string to cope with multiple I and P operators, producing an aligned insertion sequence.


jmarshall:
Replaced by #226

PR #160, opened 2015-01-06
Misc CRAM fixes (mainly unsorted data)

jkbonfield:
CRAM supported unsorted data OK, but wasn't so well optimised.  Sometimes it deremented the reference sequence ref-counters too early, causing it to have to reload the same sequence file multiple times.  Other times it didn't decrement them early enough, causing it to fail to free memory at an appropriate time (it did on file close instead).

In the process of testing it all under valgrind, I also found and fixed a couple small but genuine memory leaks.

### DIFF
--- a/cram/rANS_static.c
+++ b/cram/rANS_static.c
@@ -356,7 +356,7 @@ unsigned char *rans_uncompress_O0(unsigned char *in, unsigned int in_size,
 
 unsigned char *rans_compress_O1(unsigned char *in, unsigned int in_size,
 				unsigned int *out_size) {
-    unsigned char *out_buf = malloc(1.05*in_size + 257*257*3 + 9);
+    unsigned char *out_buf;
     unsigned char *cp = out_buf, *out_end;
     unsigned int last_i, tab_size, rle_i, rle_j;
     RansEncSymbol syms[256][256];
@@ -364,6 +364,7 @@ unsigned char *rans_compress_O1(unsigned char *in, unsigned int in_size,
     if (in_size < 4)
 	return rans_compress_O0(in, in_size, out_size);
 
+    out_buf = malloc(1.05*in_size + 257*257*3 + 9);
     if (!out_buf)
 	return NULL;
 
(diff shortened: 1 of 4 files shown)


cram/rANS_static.c
```diff
@@ -356,7 +356,7 @@ unsigned char *rans_uncompress_O0(unsigned char *in, unsigned int in_size,
 
 unsigned char *rans_compress_O1(unsigned char *in, unsigned int in_size,
 				unsigned int *out_size) {
-    unsigned char *out_buf = malloc(1.05*in_size + 257*257*3 + 9);
+    unsigned char *out_buf;
     unsigned char *cp = out_buf, *out_end;
```
mp15: Doesn't this move of initialisation make cp initialised from uninit'd data?  clang is complaining about it?

jkbonfield: On Wed, Jan 07, 2015 at 07:16:56AM -0800, Martin Pollard wrote:
> > @@ -356,7 +356,7 @@ unsigned char *rans_uncompress_O0(unsigned char *in, unsigned int in_size,
> >  
> >  unsigned char *rans_compress_O1(unsigned char *in, unsigned int in_size,
> >  				unsigned int *out_size) {
> > -    unsigned char *out_buf = malloc(1.05*in_size + 257*257*3 + 9);
> > +    unsigned char *out_buf;
> >      unsigned char *cp = out_buf, *out_end;
> 
> Doesn't this move of initialisation make cp initialised from uninit'd data?  clang is complaining about it?

Um, so it does. Odd gcc didn't spot this (and even make test
worked).

[...pause, during which frantic keyboard hammering is heard...]

Looking into it further though it's just braindead code! It promptly
resets cp a couple lines later:

    unsigned char *out_buf;
    unsigned char *cp = out_buf, *out_end;
    unsigned int last_i, tab_size, rle_i, rle_j;
    RansEncSymbol syms[256][256];

    if (in_size < 4)
        return rans_compress_O0(in, in_size, out_size);

    out_buf = malloc(1.05*in_size + 257*257*3 + 9);
    if (!out_buf)
        return NULL;

    out_end = out_buf + (int)(1.05*in_size) + 257*257*3 + 9;
    cp = out_buf+9;

I'll just remove the initial assignment.


-- 

PR #164, opened 2015-01-26
Optionally using libcurl for network access

lh3:
Just typing `make` still compiles libhts with knetfile. Typing `make KURL=1` uses libcurl instead. If libcurl is compiled with openssl and libssh2, we will be able to achieve random access to files on `https://` and `sftp://`.

The key component of this commit is the kurl library, an stdio.h-like interface to libcurl. It was heavily modified from the [fopen.c example][1]. I have not made kurl the default because: 1) kurl has not been heavily used and might be buggy (of course I have run basic tests); 2) libcurl is not a very common library; 3) kurl requires relatively recent libcurl because it has to use some new APIs. In the long run, when kurl becomes more stable and when we use autoconf, we may consider to make kurl the default.

For testing, a precompiled samtools-1.1 with statically linked https and sftp support is [available here][2]. To see how it works, you can:
```sh
samtools tview https://s3.amazonaws.com/lh3test/44X.bam NC_000913.fa
samtools tview sftp://username@host/~/aln-in-home-dir.bam
```

[1]: http://curl.haxx.se/libcurl/c/fopen.html
[2]: http://sourceforge.net/projects/biobin/files/xyztools/samtools%2Bhttps-1.1_x64-linux.tar.bz2/download


jmarshall:
:-1: For htslib, it will be better to implement this as an _hfile_curl.c_ backend.

A configure script is about to land, so in any case this should wait for that so that it can take advantage of `--with-libcurl`.

lh3:
Adding hfile_curl.c would be beneficial if we want knetfile and kurl coexist. However, conceptually they are mutually exclusive. When one approach is in use, we'd better not use the other at the same time. I would rather make a compile time decision in hfile_net.c. In addition, faidx.c and hts.c assume we are using knetfile. They will not take advantage of kurl if we only add a hfile_curl.c backend.

jmarshall:
hts.c does not assume knetfile since babaaf3381719904d328a7624c3d17de78633347.  faidx.c still does but that is a bug to be fixed :smile:

Nonetheless it would be good for htslib to have improved networking (in particular, to support https), and so we'll merge something based on libcurl and this fairly soon.

lh3:
hts.c does. It only accepts protocol "http://" and "ftp://".

jmarshall:
That's not an assumption of knetfile; it's an assumption that the only remote protocols are named `ftp://` or `http://` -- see the FIXME on the line above.

This is not an academic distinction.  An `irods://` backend is about to land, and when it does the PR's proposed `strstr(fn, "://")` tests (especially in hopen()) won't suffice.

Makefile
```diff
@@ -28,11 +28,21 @@ RANLIB = ranlib
 
 CPPFLAGS = -I.
 # TODO: probably update cram code to make it compile cleanly with -Wc++-compat
-CFLAGS   = -g -Wall -O2
+CFLAGS   = -g -Wall -Wno-unused-function -O2
```
mp15: I wonder whether instead of doing this it might be better to add __attribute((unused)) to the #defined functions from k*

lh3:
I have merged the latest develop branch into this PR and added autoconf support. I have moved part of `hisremote()` into `hfile_net.c` such that `hfile_net.c` is the only file depending on kurl (otherwise, hfile.c would also depend on kurl). Note that I am not making a new backend as kurl and knetfile are mutually exclusive. By default, htslib is still using knetfile even if libcurl development files are present on the default system path.

`configure.ac` and `config.mk.in` probably need some improvements. It should be easy, but I am not sure how.

jmarshall:
Superseded by PR #292, which has now been merged.  Thanks for the proof-of-concept, especially of the S3 stuff.

PR #165, opened 2015-01-27
Fixes imported from Staden io_lib revisions 3792/3795.

jkbonfield:
Bug fix the TLEN decoding sign.  It wasn't correctly handling the case
when two reads start at the same coordinate.

Overhauled the pnext/tlen/flags for detached reads.

Now when we find a second (or more) copy of a read within a slice, we
do not automatically mark is as non-detached.  We check whether the
derived fields match those found in the file and if not we emit a
"detached" read causing these fields to be written verbatim.  We also
do this if the data is supplementary, as the meaning of that flag is
poorly defined.

This is particularly useful on bwa-mem output where there is currently
a disparity between bwa and io_lib on the interpretation of
supplementary/primary.

PS. Sorry about the gotos! However they are pretty clear.

Also in time this may be better done by a complete reorganisation of the process_one_read() function, breaking it into two parts, gathering data and then encoding the data.  This could allow for shared use of the decoding (cram_decode_xref or similar?) function that resolves tlen, pnext, etc so we can check with the actual same code rather than a quick and dirty replication of it here.  Time is of the essence though!

### DIFF
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -1689,10 +1689,13 @@ static void cram_decode_slice_xref(cram_slice *s) {
 		    int aleft = cr->apos, aright = cr->aend;
 		    int tlen;
 		    int ref = cr->ref_id;
+		    int left_cnt = 0;
 
 		    do {
 			if (aleft > s->crecs[id2].apos)
-			    aleft = s->crecs[id2].apos;
+			    aleft = s->crecs[id2].apos, left_cnt = 1;
+			else if (aleft == s->crecs[id2].apos)
+			    left_cnt++;
 			if (aright < s->crecs[id2].aend)
 			    aright = s->crecs[id2].aend;
 			if (s->crecs[id2].mate_line == -1) {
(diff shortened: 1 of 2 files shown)


cram/cram_decode.c
```diff
@@ -1689,10 +1689,13 @@ static void cram_decode_slice_xref(cram_slice *s) {
 		    int aleft = cr->apos, aright = cr->aend;
 		    int tlen;
 		    int ref = cr->ref_id;
+		    int left_cnt = 0;
```
mp15: Could we get a comment explaining what this variable is doing?

jkbonfield: There's a comment in the code beneath it where it is used, but I will make it more explicit in the left_cnt variable too.

cram/cram_encode.c
```diff
@@ -2827,8 +2878,8 @@ static int process_one_read(cram_fd *fd, cram_container *c,
 	    cram_stats_add(c->stats[DS_MF], p->mate_flags);
 
 	    cram_stats_del(c->stats[DS_TS], p->tlen);
-	    p->tlen = p->apos - cr->aend;
-	    cram_stats_add(c->stats[DS_TS], p->tlen);
+	    //p->tlen = p->apos - cr->aend;
+	    //cram_stats_add(c->stats[DS_TS], p->tlen);
```
jmarshall: why?

jkbonfield: Why indeed! The io_lib code even has a "//why?" at the end of those lines. :-)

The answer is simply that I forgot to go back and cross all the 'T's and dot all the 'i's to fully justify my decision of those lines being dead. (Hence commented out rather than removed.)  Sorry, I'll have a further analysis of this.

jmarshall:
Thanks, merged now (via a little cherry-picking).

PR #167, opened 2015-01-27
CSIv2

pd3:
This pull request adds support for:
- CSIv2
- scientific notation for querying regions (30000000 can be written as 3e7)
- adds tests for tabix


jmarshall:
This pull request is essentially unreviewable because:
* The merge from develop introduces complexity for reviewers, and has discarded parts of existing **develop** commits without explanation (e.g. ce1a547f0e01d32fbffb62efbcca51c6ddc77d34)
* The separate commits reflect the initial false start in which record counts were added to chunks instead of bins
* CSIv2 and the scientific notation thing are two separate pieces of functionality

IMHO this should be proposed and reviewed as two separate pull requests: one for the scientific notation thing, and one for CSIv2 that has been squashed and rebased.

----

To the extent that the CSIv2 code can be reviewed, my initial comments are:

* it changes the existing libhts.so binary interface (e.g. `hts_idx_save()`)

so :-1: in its existing form.

htslib/vcf.h
```diff
@@ -723,9 +723,15 @@ extern "C" {
     #define bcf_itr_queryi(idx, tid, beg, end) hts_itr_query((idx), (tid), (beg), (end), bcf_readrec)
     #define bcf_itr_querys(idx, hdr, s) hts_itr_querys((idx), (s), (hts_name2id_f)(bcf_hdr_name2id), (hdr), hts_itr_query, bcf_readrec)
     #define bcf_itr_next(htsfp, itr, r) hts_itr_next((htsfp)->fp.bgzf, (itr), (r), 0)
-    #define bcf_index_load(fn) hts_idx_load(fn, HTS_FMT_CSI)
+    #define bcf_index_load(fn) hts_idx_load(fn, csi)
     #define bcf_index_seqnames(idx, hdr, nptr) hts_idx_seqnames((idx),(nptr),(hts_id2name_f)(bcf_hdr_id2name),(hdr))
 
+    /** 
+     *  tbx_index_build() - create .csi or .tbi index
```
mp15: Judging by the position of this this is a typo and it should read bcf_index_build not tbx_index_build

mp15:
The enums do look like the right thing to do, but this would break stuff that links to us on certain platforms  most notably android, though whether these platforms matter is highly questionable (see https://lists.gnu.org/archive/html/bug-gnulib/2011-10/msg00196.html). The usual way to deal with this in libraries is to add a numerical suffix to the function name with the new interface and make the old interface a wrapper or declare ourselves incompatible and bump the interface version. Any other opinions on this? There are quite a few non-samtools folks using htslib now and we probably should be kind to them and not break this before we release 1.2. Looking at the interface for hts_idx_save, I really don't like the fact it returns void anyway, perhaps it needs breaking, but when we do this we should make the next release of htslib 2.0 (and break a few of the other interfaces too).

Also we need at least one version where the HTS_FMT_CSI interfaces are marked as deprecated before removing them.

pd3:
On Tue, 2015-01-27 at 07:04 -0800, John Marshall wrote:
> This pull request is essentially unreviewable because:
> 
>       * The merge from develop introduces complexity for reviewers,
>         and has discarded parts of existing develop commits without
>         explanation (e.g. ce1a547)

This is a funny coincidence. MAX_CSI_COOR was present in my code before
INT_MAX which came with the commit ce1a547.


>       * The separate commits reflect the initial false start in which
>         record counts were added to chunks instead of bins
>       * CSIv2 and the scientific notation thing are two separate
>         pieces of functionality
> 
> IMHO this should be proposed and reviewed as two separate pull
> requests: one for the scientific notation thing, and one for CSIv2
> that has been squashed and rebased.
> 
> 
> ______________________________________________________________________
> 
> To the extent that the CSIv2 code can be reviewed, my initial comments
> are:
> 
>       * it changes the existing libhts.so binary interface (e.g.
>         hts_idx_save())

It makes sense to use the new fmt in the API, why should it stop
evolving?





-- 

mp15:
I think John is referring to the fact it changes the meaning of the inputs, meaning you'd have to recompile stuff to make them work with this.  If you used this with an existing package where the change to enum didn't break ABI compat, the existing package would call with the old HTS_FMT_CSI as an integer of value 0 but htslib would interpret it as htsExactFormat.unknown_format.  Also the insertion in the middle of the enum means that old code would think it was calling with tbi and htslib would interpret it as calling with gzi.

pd3:
I don't understand: Don't dynamic libraries come with a version?

mp15:
They do it's the SOVERSION in https://github.com/samtools/htslib/blob/develop/Makefile#L105 with the rules for linux libraries defined by this publication https://www.usenix.org/legacy/publications/library/proceedings/als00/2000papers/papers/full_papers/browndavid/browndavid_html/index.html. It's not that we can't change the version, it's just considered mean to your downstream users to break it if you don't have to, especially for a minor release of the library.

I've just looked more carefully at the enums and htsExactFormat wasn't present in 1.1 so we haven't released an ABI which contains these which means we can reorder it as much as we like.  However the HTS_FMT_CSI to htsExactFormat.unknown_format issue still stands. A good start would be to reorder it so that the position and value of csi corresponds to HTS_FMT_CSI, bai to HTS_FMT_BAM etc.

pd3:
OK, I am happy with any good solution you propose. How do we technically do it: do you want to make a pull request against the CSI branch and then recreate the request? Or merge this first and then make the change at the top of it?

pd3:
@mp15 OK, it seems that the problem does not have a good solution:
1) use the format enums and reorder them to completely match the old HTS_FMT_* macros (`unknown` will not be first anymore)
2) do not use the fmt enums at all and add new HTS_FMT_CSIv1instead
3) break compatibility and do/do not increase version number

Which do you pick?

jkbonfield:
On Wed, Jan 28, 2015 at 08:16:54AM -0800, pd3 wrote:
> @mp15 OK, it seems that the problem does not have a good solution:
> 1) use the format enums and reorder them to completely match the old HTS_FMT_* macros (`unknown` will not be first anymore)

You can assign a numerical ID when declaring an enum, so the order
doesn't necessarily have to imply 0, 1, 2, 3, ...

James

-- 


-- 

lh3:
I am slowly going over the CSIv2 topic. I don't think we should use `enum htsExactFormat` because it changes APIs. As to solution 2), we should add a new macro `HTS_FMT_CSIv2` with a new value, instead of changing `HTS_FMT_CSI`. This is also about backward compatibility.

mcshane:
We're going to close these CSI2 PRs to clean up old PRs. I think we left with CSI2 needing to start from scratch anyway to get desired behaviour.

PR #169, opened 2015-01-28
Make change to htsExactFormat not break the ABI

mp15:
This is just an example.  It packs the enum and set the max value to force it into an int, this preserves the old binary interface.


mp15:
This needs to be reworked, given that version exists on the file type structure now.

jmarshall:
This was an example; will be revisited along with CSIv2.

PR #178, opened 2015-02-13
Fixed .crai loading of local irods:... caches.

jkbonfield:
For .bai files on remote protocols we download the .bai locally and
use the local copy.

The previous .crai implementation just used hopen so it opened the
remote end and never cached it.  We now follow the same procedure used
in hts_idx_load for bai.

### DIFF
--- a/htslib/hts.h
+++ b/htslib/hts.h
@@ -366,7 +366,6 @@ extern "C" {
 
     int hts_idx_get_stat(const hts_idx_t* idx, int tid, uint64_t* mapped, uint64_t* unmapped);
     uint64_t hts_idx_get_n_no_coor(const hts_idx_t* idx);
-    char *hts_idx_getfn(const char *fn, const char *ext);
 
     const char *hts_parse_reg(const char *s, int *beg, int *end);
     hts_itr_t *hts_itr_query(const hts_idx_t *idx, int tid, int beg, int end, hts_readrec_func *readrec);
--- a/htslib/hts.h
+++ b/htslib/hts.h
@@ -366,6 +366,7 @@ extern "C" {
 
     int hts_idx_get_stat(const hts_idx_t* idx, int tid, uint64_t* mapped, uint64_t* unmapped);
     uint64_t hts_idx_get_n_no_coor(const hts_idx_t* idx);
+    char *hts_idx_getfn(const char *fn, const char *ext);
 
     const char *hts_parse_reg(const char *s, int *beg, int *end);
     hts_itr_t *hts_itr_query(const hts_idx_t *idx, int tid, int beg, int end, hts_readrec_func *readrec);
--- a/Makefile
+++ b/Makefile
@@ -189,7 +189,7 @@ cram_samtools_h = cram/cram_samtools.h $(htslib_sam_h) $(cram_sam_header_h)
 cram_structs_h = cram/cram_structs.h cram/thread_pool.h cram/string_alloc.h $(htslib_khash_h)
 cram_open_trace_file_h = cram/open_trace_file.h cram/mFILE.h
 hfile_internal_h = hfile_internal.h $(htslib_hfile_h)
-
+hts_internal_h = hts_internal.h $(htslib_hts_h)
 
 # To be effective, config.mk needs to appear after most Makefile variables are
 # set but before most rules appear, so that it can both use previously-set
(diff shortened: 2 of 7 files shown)


htslib/hts.h
```diff
@@ -366,6 +366,7 @@ extern "C" {
 
     int hts_idx_get_stat(const hts_idx_t* idx, int tid, uint64_t* mapped, uint64_t* unmapped);
     uint64_t hts_idx_get_n_no_coor(const hts_idx_t* idx);
+    char *hts_idx_getfn(const char *fn, const char *ext);
```
jmarshall: This is an htslib-internal function -- it has weird side-effects due to calling `test_and_fetch()`.  So I think we need to come up with a better solution than declaring it in a public header...

jkbonfield: On Fri, Feb 13, 2015 at 03:47:25AM -0800, John Marshall wrote:
> > @@ -366,6 +366,7 @@ extern "C" {
> >  
> >      int hts_idx_get_stat(const hts_idx_t* idx, int tid, uint64_t* mapped, uint64_t* unmapped);
> >      uint64_t hts_idx_get_n_no_coor(const hts_idx_t* idx);
> > +    char *hts_idx_getfn(const char *fn, const char *ext);
> 
> This is an htslib-internal function -- it has weird side-effects due to calling `test_and_fetch()`.  So I think we need to come up with a better solution than declaring it in a public header...

I'm not really happy with the alternative solution of moving all the
cram index code into hts.c either.  I was sort of assuming all of the
hts_idx_* functions weren't really designed to be properly external
anyway as none of them have the usual comments describing the
functions that most (well some!) of our public API does.

Maybe we need an hts_internal.h or some such?

James

-- 


-- 

jkbonfield:
I cannot just split my cram_index_load() function in two parts, with one part in hts.c and the other in cram_index.c, as although hts_idx_getfn becomes local I then have to somehow define the prototype for the part of cram_index_load() which moved.  It just shuffles the problem from one function name to the next.

The real crux of the problem is no clear separation of internal and external API components.  I really don't believe we intended for half the stuff to be external. Does anyone else really want to call hts_idx_set_meta?  Not likely.  My assumption was the doxygen marked up functions are intended to be public and everything else is just an internal thing, but I could be wrong here.

There are usually two solutions to controlling public access.

1) Add them into the public header files but guarded by #ifdefs.  This is the sort of thing you see in the linux header files and it's extensible to supporting multiple API versions too.  Eg:

    #ifdef HTSLIB_INTERNAL
    char *hts_idx_getfn(const char *fn, const char *ext);
    #endif

2) Use a completely separate hts_int.h or similar name containing internal only definitions.  These can even be omitted from installing if we wish.

Any preferences?  I think the first is better if we can restrict the #ifdef region down to a single portion.

mp15:
Mmm I prefer option 2, given that we're already coding that way. If you look at hfile_internal.h in the htslib root directory this is pretty much what you've described.  The other thing we could do is do a visibility pragma on the functions to hide them from the automatic ABI checks such as https://gcc.gnu.org/onlinedocs/gcc/Visibility-Pragmas.html#Visibility-Pragmas.

jmarshall:
> Ah. I hadn't realised all [our] dependencies were manually generated. Could we not auto-generate them?
> [...]
>  Because it automatically edits the Makefile it does not require users to have makedepend installed.  It's just something we can run periodically and before an official release to ensure the Makefile dependencies are up to date.

Me, I tend to think trying to generate dependencies with users running makedepend on each build themselves (even if they don't have it or a compiler with -MD or whatever) is more trouble than it's worth.  OTOH a maintainer-side script to help is more palatable and we have the basis for it already.

However, dependency generation infrastructure is a separate question from this PR :smile:

jkbonfield:
On Tue, Feb 17, 2015 at 09:18:45AM -0800, John Marshall wrote:
> Me, I tend to think trying to generate dependencies with users
> running makedepend on each build themselves (even if they don't have
> it or a compiler with -MD or whatever) is more trouble than it's
> worth.

That's not my intention.  Users should never have to make dependencies
themselves.

> OTOH a maintainer-side script to help is more palatable and we
> have the basis for it already. 

This is the right solution.  I shouldn't have to edit two different
Makefiles in a total of 4 separate places to add one single header file.
It's just too error prone.  So the maintainers having an auto-update
script is fine.

> However, dependency generation infrastructure is a separate question from this PR :smile:

Agreed. Please review my latest commit and see if it ticks all the
boxes now.

Thanks for the pointers.


-- 

PR #180, opened 2015-02-23
Merge of samtools/misc and htslib/cram MD5 implementations

jkbonfield:
Htslib now exposes its own internal MD5 implementation, in a similar (but not exact) vein to OpenSSL.
This is based around the Solar Designer implementation, as previously used within the cram subdirectory.


htslib/hts.h
```diff
@@ -394,6 +394,48 @@ extern "C" {
...
+
+    /*! @abstract Resets an md5_context to the initial state, as returned
+     *            by hts_md5_init().
+     */
+    void hts_md5_reset(hts_md5_context *ctx);
+
+    /*! @abstract Converts a 128-bit MD5 hash into a 33-byte nul-termninated
+     *            hex string.
+     */
+    void hts_md5_hex(unsigned char *digest, char *hex);
```
jmarshall: Apologies for the late review... other functions here follow the usual C `foo(dest, src)` convention.  So could this be `void hts_md5_hex(char *hex, const unsigned char *digest)` and also the **const**?

jmarshall:
Merged the unrebased version with _changed hts_md5_hex args_ cherry-picked on top, as some erroneous merging went on in 5a270f644a9af33679c961a915239b6c60fb24c0.

PR #182, opened 2015-02-25
Index tests

keithj:
I've added some basic tests for index creation and use, both with and without iRODS (default is without). Included is a utility script to make very simple, regular SAM files.

My changes to the makefile are conservative - no auto detection of iRODS is done WRT tests. To run tests
with iRODS you need to explicitly run the test_index.pl script with an argument `--with-irods`.

Everything passes for me using iRODS 3.3.1


keithj:
By the way, I can squash onto a new branch and resubmit if the merge commit noise is a probelm.

keithj:
2 years later now, so I'm closing this.

PR #184, opened 2015-02-25
Pass CPPFLAGS from "./configure" to "make" invocation.

ghuls:
"make" never got the CPPFLAGS value passed to "./configure" as CPPFLAGS was
missing in "config.mk.in", so make kept using the hard coded CPPFLAGS value
in the Makefile.

Closes samtools/htslib/#183


ekg:
Why not use `$(make)`? Does this resolve the issue?

jmarshall:
This is incomplete as it stands, as `./configure; make` fails due to loss of `-I.`.  I'm fixing that up now...

PR #189, opened 2015-03-15
Fixes bam_index/bam_index_build to handle truncated or corrupted BAMs

jrandall:
Adds an EOF check to bam_index_build .

Adds a check of error status from bam_read1 to bam_index and bails out on the index if it has failed (i.e. due to corruption). 

These fixes address samtools issue: samtools/samtools#362

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -407,7 +407,8 @@ static hts_idx_t *bam_index(BGZF *fp, int min_shift)
     idx = hts_idx_init(h->n_targets, fmt, bgzf_tell(fp), min_shift, n_lvls);
     bam_hdr_destroy(h);
     b = bam_init1();
-    while (bam_read1(fp, b) >= 0) {
+    int r;
+    while ((r = bam_read1(fp, b)) >= 0) {
         int l, ret;
         l = bam_cigar2rlen(b->core.n_cigar, bam_get_cigar(b));
         if (l == 0) l = 1; // no zero-length records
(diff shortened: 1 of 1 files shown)


jrandall:
If you don't think the library should print to stderr, perhaps you should also remove the `perror` on `has_EOF < 0` ?

jmarshall:
I've merged be6bd1e92c2d8ddb4f1aef4fcb49466c85a7674d (changing it a bit as noted) and cherry-picked 17fd0cc47b6ce7ee1e9b8130a930e4e79b3c5456.

I'm not convinced that we want `samtools index` to return a non-zero status when indexing a BAM file that merely has no EOF trailer, so I've left out 29955e5d02b93bc79dbcc79a2c01eabf50198e19 at least for now.

jrandall:
@jmarshall the behaviour in which `samtools index` returns an error when the EOF block is missing is the primary reason for this pull request. Thomas and I (at least, probably others) have repeatedly told students on courses that running `samtools index` is a good way to tell if your BAM file is truncated. We also rely on this behaviour in our own pipelines, and have had issues ever since we upgraded to samtools 1.0 (samtools 0.1.19 did exit with an error in that case).

jrandall:
The SAM specification also encourages implementations to "trigger a warning or error soon after opening a BGZF file..." in which the EOF marker is absent (https://github.com/samtools/hts-specs/blob/master/SAMv1.tex#L837-844).

jkbonfield:
Maybe the htsfile command should become the standard sanity checker,
perhaps with a -c option to ask it to check basic validity, or to
create a standalone htscheck tool.  I feel a bit uneasy with claiming
this is the role of samtools index.

That said, I agree with Josh that samtools index should report this as
an error (unless we add an explicit command line option to accept
legacy BAM files).

James
-- 


-- 

jmarshall:
Presented with a BAM file with no EOF trailer, `samtools-0.1.19 index` emits a warning, generates a _.bai_ index, and exits with a successful status.

I agree that samtools should be emitting a warning.  I disagree that it's obvious that it should be exiting with failure: the spec recommends but does not require an EOF trailer, and BAM files without them **are** valid BAM files.

PR #197, opened 2015-04-09
Avoid __clang_major__ undefined warnings

noporpoise:
Only clang sets __clang_major__ preprocessor macro. To avoid warnings about undefined macros being used, we should check if it is set before using it.

### DIFF
--- a/htslib/khash.h
+++ b/htslib/khash.h
@@ -152,7 +152,7 @@ typedef unsigned long long khint64_t;
 #endif /* kh_inline */
 
 #ifndef klib_unused
-#if __clang_major__ >= 3 || __GNUC__ >= 3
+#if (defined(__clang__) && __clang_major__ >= 3) || __GNUC__ >= 3
 #define klib_unused __attribute__ ((__unused__))
 #else
 #define klib_unused


jmarshall:
Presumably you are compiling with GCC and `-Wundef`.  (Is there a particular reason why you are using `‑Wundef`?  Is this when compiling htslib itself, or compiling your program that uses htslib?)

Always checking with `defined(foo)` first is pretty tedious, which is why undefined identifiers are defined to decay to `0` without error.  So pandering to this warning is a bit annoying, but I suppose not so bad if we take advantage of the fact that everybody (e.g. clang) lies and defines `__GNUC__` even though they're not actually GCC.

noporpoise:
I'm compiling our program that uses htslib, using GCC and `-Wundef`. This is the only warning we get from compiling against htslib.

We set `-Wundef` to catch typos/bugs in macros. This formal checking saves us man-hours in the long run. It also identifies preprocessor values (in our code) that should have been set during configuration.

jmarshall:
Fair enough.  There's a couple of other such warnings when compiling htslib itself, and I don't think it's a priority to fix those.

noporpoise:
Thank you, that's great.

PR #199, opened 2015-04-15
Fix tabix handling of IO errors

mp15:
Add additional check to end of tabix loop to catch termination due to IO errors and exit accordingly.


mp15:
Urgh, I just realised the expected solution is more complicated. We may have to break binary compat :(.

PR #201, opened 2015-04-28
allow underscore and dot in keys for structured header lines

mcshane:
We were forcing alphanumeric-only characters in the keys for structured
header lines. From the spec, INFO and FORMAT ID keys should match the regex
`^[A-Za-z_][0-9A-Za-z_.]*$`. It is unclear about keys for other
structured header lines, but this would apply the same restictions.


jmarshall:
Merged as 30fb9eee41953958923c56f7ea0af5a5b0376b94.

PR #203, opened 2015-04-30
Allow synced reader for multiple streams without index.

johanneskoester:
From my inspection, the synced reader code should also work without and index and multiple streams.
I have changed the error in these cases into a warning, in order to allow an --omit-index functionality for bcftools merge.

This allows to use bcftools merge with multiple named pipes, which can be very handy with large, sorted bcf files that shall not be stored on disk.

I don't know if my changes are appropriate and I am open to any suggestions. Please see the corresponding pull request in bcftools:
https://github.com/samtools/bcftools/pull/252


mcshane:
Closing. See conversation in samtools/bcftools#252

PR #204, opened 2015-05-01
Increase size of max format string size from 2^16 to 2^32.

ryan10X:
Fixes the issue of integer overflow when a VCF format string
contains a large string (in our case, nucleotide sequence).
2^32 should be a large enough number of characters (> 1 GB) such
that its an acceptable upper bound.

Note that bcftools will now crash with format field values > 2^16 characters (~65 KB). With this patch, that value is increased to 2^32 characters (~4.2 GB).

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -1478,7 +1478,7 @@ uint8_t *bcf_fmt_sized_array(kstring_t *s, uint8_t *ptr)
 
 typedef struct {
     int key, max_m, size, offset;
-    uint32_t is_gt:1, max_g:15, max_l:16;
+    uint64_t is_gt:1, max_g:31, max_l:32;
     uint32_t y;
     uint8_t *buf;
 } fmt_aux_t;


tk2:
Hi Ryan - we were just talking about this pull request and curious to know what is the use case for this? It sounds like quite an curious use of the format string. We are pretty sure this will break on existing BCF files.

Thomas

mcshane:
@pd3 thinks this might me ok since it is an aux structure only created when creating and parsing BCFs. Would be good to know the use case to justify before checking further though.

ryan10X:
Hey Thomas + Shane,

In some (admittedly pathologic) VCF files, we have format fields with very long strings > 2^16 characters (~65 KB). This 'fmt_aux_t' struct parses the format fields and has a max format field length of 2^16. However, because of this, bcftools crashes with format field values > 2^16 characters due to an integer overflow. So this patch basically 'fixes' the issue by increasing max format field length from 2^16 to 2^32.

I use 'fixes' in quotes because if you create a format field of length 2^32 (~4.2 GB), you could still crash bcftools. But 4.2 GB lines are probably outside the scope of a reasonable VCF file while 65 KB lines are definitely manageable in some applications. A better fix might be check for integer overflow (i.e. if format field length > 2^32) and if so, exit nicely saying bcftools does not support format field values of this length.

PR #206, opened 2015-05-14
Fix bin calculation

dpryan79:
According to the spec (see footnote 17 on page 13), unmapped reads should be treated as having length 1 for the purposes of bin calculation. This makes sense, as it ensures that they're assigned to the lowest level of bins, regardless of their cigar string. This part of the spec. is currently not adhered to, but doing so prevents picard from complaining as much.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -798,6 +798,7 @@ int sam_parse1(kstring_t *s, bam_hdr_t *h, bam1_t *b)
         q = _read_token(p);
         i = 1;
     }
+    if(c->flag & BAM_FUNMAP) i = 1;
     c->bin = hts_reg2bin(c->pos, c->pos + i, 14, 5);
     // mate chr
     q = _read_token(p);


dpryan79:
Here's an example alignment that samtools would currently assign to an incorrect bin: https://gist.github.com/dpryan79/1efd64bfc8e82334626f
That would be assigned to bin 669, but should instead be in 5354.

jmarshall:
That read has the UNMAPPED flag bit and also a CIGAR string, which is an invalid corner case that samtools and Picard have long disagreed on.  We fixed old samtools to the lowest common denominator (interpret either as unmapped for bin calculations), but sadly hadn't fixed htslib completely.

Thanks for pointing this out, Devon.  There's other code to fix in sam.c as well — I've updated this as PR #208.

dpryan79:
Thanks, I'm of the opinion that Picard is a bit too pedantic on many such things (though in this case it's what alerted us to this bug). The other PR looks like a better route to take, thanks!

PR #207, opened 2015-05-14
CRAM fuzz testing bug fixes

daviesrob:
These commits fix various problems found by the "american fuzzy lop" tool.  It was actually run on scram_flagstats from the Staden io_lib.  The fixes put in there are all ported to htslib in this branch.

### DIFF
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -1624,9 +1624,6 @@ cram_codec *cram_byte_array_stop_decode_init(char *data, int size,
         free(c);
         return NULL;
     }
-    c->decode = (option == E_BYTE_ARRAY_BLOCK)
-	? cram_byte_array_stop_decode_block
-	: cram_byte_array_stop_decode_char;
     c->free   = cram_byte_array_stop_decode_free;
     
     c->byte_array_stop.stop = *cp++;
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -1401,7 +1401,7 @@ cram_codec *cram_byte_array_len_decode_init(char *data, int size,
     char *cp   = data;
     char *endp = data + size;
     int32_t encoding = 0;
-    int32_t sub_size = 0;
+    int32_t sub_size = -1;
 
     if (!(c = malloc(sizeof(*c))))
 	return NULL;
@@ -1420,6 +1420,7 @@ cram_codec *cram_byte_array_len_decode_init(char *data, int size,
         goto no_codec;
     cp += sub_size;
 
+    sub_size = -1;
     cp += safe_itf8_get(cp, endp, &encoding);
     cp += safe_itf8_get(cp, endp, &sub_size);
     if (sub_size < 0 || endp - cp < sub_size)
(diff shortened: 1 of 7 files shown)


cram/cram_codecs.c
```diff
@@ -750,11 +728,12 @@ cram_codec *cram_subexp_decode_init(char *data, int size,
     c->codec  = E_SUBEXP;
     c->decode = cram_subexp_decode;
     c->free   = cram_subexp_decode_free;
-    
+    c->subexp.k = -1;
+
     cp += itf8_get(cp, &c->subexp.offset);
```
jkbonfield: Should this be the safe itf8_get?  itf8_get always sets &c->subexp.offset therefore the c->subexp.k = -1 doesn't do anything.

cram/cram_codecs.c
```diff
@@ -1416,25 +1410,36 @@ cram_codec *cram_byte_array_len_decode_init(char *data, int size,
     c->decode = cram_byte_array_len_decode;
     c->free   = cram_byte_array_len_decode_free;
     
-    cp += itf8_get(cp, &encoding);
-    cp += itf8_get(cp, &sub_size);
+    cp += safe_itf8_get(cp, endp, &encoding);
+    cp += safe_itf8_get(cp, endp, &sub_size);
+    if (sub_size < 0 || endp - cp < sub_size)
```
jkbonfield: Should this be sub_size < 0 || encoding < 0 and setting both of these to -1 instead of 0 above?

Right now it looks like a failure to decode isn't caught here, instead accepted 0,0 as the defaults . This doesn't sound right.

cram/cram_codecs.c
```diff
@@ -1416,25 +1410,36 @@ cram_codec *cram_byte_array_len_decode_init(char *data, int size,
...
 						    E_INT, version);
+    if (c->byte_array_len.len_codec == NULL)
+        goto no_codec;
     cp += sub_size;
 
-    cp += itf8_get(cp, &encoding);
-    cp += itf8_get(cp, &sub_size);
+    cp += safe_itf8_get(cp, endp, &encoding);
+    cp += safe_itf8_get(cp, endp, &sub_size);
+    if (sub_size < 0 || endp - cp < sub_size)
```
jkbonfield: Similarly here there is no reset for the default values of encoding and sub_size, so they may just come from the ones above in the case of running out of 'cp'.

cram/cram_codecs.c
```diff
@@ -1630,6 +1611,18 @@ cram_codec *cram_byte_array_stop_decode_init(char *data, int size,
...
+	break;
+    case E_BYTE_ARRAY:
+        c->decode = cram_byte_array_stop_decode_char;
+	break;
+    default:
+        fprintf(stderr, "byte_array_stop codec only supports BYTE_ARRAYs.\n");
+        free(c);
+        return NULL;
+    }
     c->decode = (option == E_BYTE_ARRAY_BLOCK)
```
jkbonfield: Duplicate code to the above new switch statement. Please delete one or the other.

daviesrob:
Issues should all be fixed in the latest version.

PR #208, opened 2015-05-15
Use bam_endpos() to fix bins for unmapped reads

jmarshall:
Alas, htslib calculates alignment stop positions and bin values via direct CIGAR calculations, without checking the UNMAPPED flag bit.  We did a bunch of work around this in old samtools and added `bam_endpos()` to htslib in 037860be80f321e19e78c2ac2a3b925aa9addf62, but sadly we didn't start using it in htslib at that time.

So htslib has been writing BAM files with occasionally incorrect bin fields for unmapped reads with non-`*` CIGARs (as in #206; also we were giving unmapped CIGARred POS=0 reads bin 0 rather than bin 4680), and constructing indices with similarly not quite correct bins for such reads.

See also PR samtools/samtools#391 which updates test *.bam files accordingly.

Any concerns with this change?  Please check implications for `overlap_push()` in particular.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -410,9 +410,7 @@ static hts_idx_t *bam_index(BGZF *fp, int min_shift)
     bam_hdr_destroy(h);
     b = bam_init1();
     while ((ret = bam_read1(fp, b)) >= 0) {
-        int l = bam_cigar2rlen(b->core.n_cigar, bam_get_cigar(b));
-        if (l == 0) l = 1; // no zero-length records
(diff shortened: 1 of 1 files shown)


jkbonfield:
On Fri, May 15, 2015 at 04:21:10AM -0700, John Marshall wrote:
> Any concerns with this change?  Please check implications for `overlap_push()` in particular.
> You can view, comment on, or merge this pull request online at:

I haven't reviewed it, but it's worth noting that this will bring BAM
into agreement with CRAM.  By design, CRAM cannot store CIGAR strings
for unmapped data.  (You can argue whether this is a feature or a
bug.)  Hence converting from bam to cram to bam could change bin.


-- 

PR #209, opened 2015-05-19
Port S3 support from samtools-legacy

jpiper:
I've ported the S3 functionality introduced in the samtools-legacy repository up to here, with a few changes (noted below).

* Works alongside http:// and ftp:// now
* Checks for a HTTP 200 response from S3 before reporting a file has been successfully read, as S3 fulfills non-existent file requests with an XML file, which htslib would gladly accept, and then complain when it's not a real file. 

For samtools and bcftools to be built with this, you'll need to add `-lcurl` to `LDLIBS`.

One thing that I haven't changed is the reading of credentials from `~/.awssecret`. It could potentially read from the default credential in `~/.aws/credentials` instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ CPPFLAGS =
 CFLAGS   = -g -Wall -O2
 EXTRA_CFLAGS_PIC = -fpic
 LDFLAGS  =
-LDLIBS   =
+LDLIBS   = -lcurl
 
 # For now these don't work too well as samtools also needs to know to
 # add -lbz2 and -llzma if linking against the static libhts.a library.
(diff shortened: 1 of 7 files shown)


lh3:
@jmarshall has a new libcurl backend on the libcurl branch of this repo. If you want to add S3 support, it would be better to work from that branch.

jpiper:
Ah ok! I'll close this request and work on it from that branch :)

jmarshall:
Porting Heng's S3 support is already mostly finished.  The thing about idiotic S3 200 responses is interesting, and something that should be remembered for later.

I'm sure I could expedite pushing it and merging the libcurl branch if @lh3 would like to merge some **klib** pull requests :smile:

jpiper:
It's a very quick fix - you can see it here :)

https://github.com/humanlongevity/htslib/blob/s3-support/ks3file.c#L195-L201

jpiper:
By the way, don't know how you're progressing on this, but you'll also need to handle the S3 reauthentication, as when seeking within a file more than 5 minutes after the first token generation the HTTP request will be rejected as the auth goes stale - I've implemented this on our code [here](https://github.com/humanlongevity/htslib/commit/bee8b607780c717b894fc0d977eb4d93878fe555).

PR #211, opened 2015-05-27
Report zlib errors in bgzf.c

kirkmcclure:
Propose corrections for Issue samtools/samtools#111 (Samtools ignores specific zlib error codes in bgzf.c)


jmarshall:
I believe this has been addressed by PR #271.

PR #214, opened 2015-06-01
allow spaces in vcf header lines

mcshane:
* allow spaces between keys and values when parsing in header lines
* these spaces will be dropped when writing out the header

e.g. `##reference=<ID=hs37d5 , Source=blah>` and `##reference=<ID=hs37d5, Source = blah >`
     will become `##reference=<ID=hs37d5,Source=blah>`

Fixes samtools/bcftools#266

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -307,6 +307,7 @@ bcf_hrec_t *bcf_hdr_parse_line(const bcf_hdr_t *h, const char *line, int *len)
     while ( *q && *q!='\n' && nopen )
     {
         p = ++q;
+        while ( *q && *q==' ' ) { p++; q++; }
         // ^[A-Za-z_][0-9A-Za-z_.]*$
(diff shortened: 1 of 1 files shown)


jmarshall:
What's the reasoning for also allowing spaces around the `=`?  The spec is not particularly clear about how these things are delimited.  There would be be an interoperability discouragement problem if we accepted such spaces and Picard did not, but on brief testing it seems (surprisingly!) that htsjdk does in fact accept such spaces.  So I suppose this is :+1:.

Spaces before the comma still fail, but htsjdk accepts them:

    ##reference=<ID=hs37d5 , Source = blah>

mcshane:
Have updated this now to take care of a few more space cases.

jmarshall:
Rebased and merged as 72cb247f0da6fff117a56b3a18cc3b4d7c2a8e68.

PR #215, opened 2015-06-05
Extra checks in bam_hdr_read.

daviesrob:
Makes bam_hdr_read more resistant to broken bam files.  Fixes valgrind failures in several of my American Fuzzy Lop test cases.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -113,7 +113,8 @@ bam_hdr_t *bam_hdr_read(BGZF *fp)
     bam_hdr_t *h;
     char buf[4];
     int magic_len, has_EOF;
-    int32_t i = 1, name_len;
+    int32_t i, name_len, num_names = 0;
+    ssize_t bytes;
     // check EOF
     has_EOF = bgzf_check_EOF(fp);
     if (has_EOF < 0) {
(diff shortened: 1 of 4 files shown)


daviesrob:
I've added the view_vcf changes that were requested.

An alternative way of doing the changes in htsfile would be to make view_sam and view_vcf take ```hFILE **hfp```, which would allow them to set it to NULL themselves if necessary.  That would allow the functions to return the error status, and *status would no longer have to be passed in.  Do you think that would be a better way of arranging them?

I.e. something like this:

```
static int view_sam(hFILE **hfp, const char *filename) {
   int retval = 1;
   samFile *in = hts_hopen(*hfp, filename, "r");
   if (in == NULL) return 1;
   // ....
   if (bad_thing) goto clean;
   // ...
   retval = 0;
clean:
  if (hdr != NULL) bam_hdr_destroy(hdr);
  if (out != NULL) hts_close(out);
  hts_close(in);
  *hfp = NULL;  // *hfp will have been closed by hts_close(in)
  return retval;
}

  // ...
  if (view_sam(*fp, argv[i])) status = EXIT_FAILURE;
}
```

jmarshall:
Thanks.  It's a simple little self-contained program, we should just make `status` a global alongside `mode` and `show_headers` and these functions can set it to `EXIT_FAILURE` themselves.

sam.c
```diff
@@ -127,26 +128,95 @@ bam_hdr_t *bam_hdr_read(BGZF *fp)
         return 0;
     }
     h = bam_hdr_init();
+    if (!h) return 0;
```
mp15: isn't this a nomem error?

daviesrob: Strictly, yes, although as you note the clean-up code is written with the assumption that *h is valid.  Hence it just bails at this point, and with the EOF check above.  It's a fairly easy change to make it print the message, though.

PR #217, opened 2015-06-09
A couple more bam file input checks.

daviesrob:
Add a missing test that there's enough data when converting B tags in sam_format1.

Ensure query name length >= 1 in bam_read1.  It has to be at least 1 due to the NUL byte.


jkbonfield:
Reviewed by jkbonfield.

jmarshall:
Variously rebased, squashed, and reimagined as d341aaf886ba4b9be8af06be50f6e93184a6c52f.  Thanks!

PR #218, opened 2015-06-11
Ensure sam_hdr_parse doesn't run off the end of the string it's reading.

daviesrob:
When looking for @SQ header lines, search for "@SQ\t" instead of just @SQ.
This ensures the term q = p + 4 points to either a NUL byte or more string
and not past the end.  The spec. says there should always be a tab after
@SQ so this should be OK on valid files.

Add extra checks for NUL to the loops that iterate through the string.

Fixes a crash found by the American Fuzzy Lop fuzz tester.


jmarshall:
Merged along with #223

PR #221, opened 2015-06-11
check for insanely long VCF lines

mcshane:
check for insanely long VCF lines. hattip @ryan10X

Resolves #204


daviesrob:
```if ( v->indiv.l > 0xffffffff )``` is fine on 64-bit platforms, but of course won't do anything on 32-bit ones.  Although it's pretty certain that something else will have gone wrong first before this becomes an issue...

PR #225, opened 2015-06-22
Fix (some) signed one-bit bitfields

nsubtil:
There are several places in the htslib tree where signed one-bit bitfields are defined. This triggers a lot of compile-time warnings on compilers that actually check for this (unfortunately it seems gcc is not one of them).

This pull request fixes the ones I've run into so far. I can extend this to cover the entire source tree if that is deemed necessary (though that could be a bit riskier than the current patch set).

### DIFF
--- a/htslib/bgzf.h
+++ b/htslib/bgzf.h
@@ -47,7 +47,7 @@ struct bgzf_mtaux_t;
 typedef struct __bgzidx_t bgzidx_t;
 
 struct BGZF {
-    int errcode:16, is_write:2, is_be:2, compress_level:9, is_compressed:2, is_gzip:1;
+    unsigned int errcode:16, is_write:2, is_be:2, compress_level:9, is_compressed:2, is_gzip:1;
     int cache_size;
     int block_length, block_offset;
     int64_t block_address, uncompressed_address;
--- a/htslib/hfile.h
+++ b/htslib/hfile.h
@@ -43,7 +43,7 @@ typedef struct hFILE {
     char *buffer, *begin, *end, *limit;
     const struct hFILE_backend *backend;
     off_t offset;
-    int at_eof:1;
+    unsigned int at_eof:1;
     int has_errno;
 } hFILE;
 


jmarshall:
A signed single-bit bit-field is fine, containing the values 0 and -1 in two's complement.

To be sure, there is a problem where we've forgotten this and code like `fp->is_gzip = 1;` produces a warning (with e.g. `gcc -Wconversion`) while giving `is_gzip` the (perhaps unexpected) value -1.  However all is otherwise well, as warnings are not produced elsewhere (with gcc; in particular, no warnings in API client code) and the field is later tested with `if (fp->is_gzip)` and is zero or non-zero as that code expects.

What compiler and warning options are you using?  Are warnings produced anywhere other than on assignments like `fp->is_gzip = 1;`?

We should fix this (in particular, we shouldn't be using plain `int` with bit-fields), but I don't think it's too pressing; it is certainly outweighed by ABI concerns.  In particular, your patch is incomplete, as we signal a special case with a negative `compress_level`, so that one in particular can't be unsigned.

nsubtil:
Thanks for the reply. Yeah, Travis also seems to be saying the change isn't good.

The compiler in question is NVIDIA's nvcc with default warning options. It triggers warnings on the structure definition itself, not on assignments --- it's enough to #include one of the headers I modified to see warnings. The problem gets a bit worse due to nvcc's quirky compilation process, where each file goes through the frontend multiple times, leading to multiple identical warnings for each compilation unit.

I would also note that the layout of a structure with bitfields is implementation dependent (and quite volatile, based on my past experience). The htslib ABI is thus not well defined and is only guaranteed to work if the same revision of the same compiler is used for client and library code. This is particularly concerning as the API includes code that gets compiled into the client itself (inline function implementations in the headers which manipulate structures with bitfields, e.g., herrno).

Either way, it's clear that the current patch is not acceptable. Would you be willing to consider a more complete version, or would you prefer a different course of action?

jmarshall:
I guess what I said about “signed single-bit bit-field is fine” pertained to _host_ architectures, and your target architectures may be weird enough that this isn't true.  And warnings from headers are always annoying.  So fair cop.

Layout of structs and bit-fields is defined by the platform ABI.  So on platforms like [x86-64](http://www.x86-64.org/documentation/abi.pdf) the htslib API/ABI is just fine even containing bit-fields, as these things are defined by the platform calling conventions and so on.

If your platform doesn't define these things, then I guess any third-party library ABI would only have guarantees when the whole thing is compiled with the same revision of the same compiler.  That's something that htslib users would have to be aware of on such platforms.  Sorry, but we're probably not going to stop using bit-fields in this code, which is now quite some years old.

PR #226, opened 2015-06-22
Cigar P support

jmarshall:
Rebased and updated version of @jkbonfield's PR #157, which:

> Adds a bam_plp_insertion() function that queries the cigar string to cope with multiple I and P operators, producing an aligned insertion sequence.

However the addition of `cigar_ind` is a binary compatibility issue.  We may be able to fix this by completely reworking the API addition, or…

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ lib*.so.*
 /tabix
 /test/fieldarith
 /test/hfile
+/test/pileup
 /test/sam
 /test/test-regidx
 /test/test-vcf-api
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,7 @@ BUILT_PROGRAMS = \
 BUILT_TEST_PROGRAMS = \
 	test/fieldarith \
 	test/hfile \
+	test/pileup \
 	test/sam \
 	test/test-regidx \
 	test/test_view \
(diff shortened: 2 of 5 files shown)


jkbonfield:
Am I missing something? The rebased version is more or less identical so the "or..." is something still to come rather than something implemented here?

One alternative I can think of is to not cache the information and recompute it, but that's ugly!   The only other idea I had is to have a completely new interface and an option in mpileup to use new or old methods (ie whether to pretend "P" operator doesn't exist).  Also ugly.  I wouldn't start from here ideally. :-)

jmarshall:
The “or…” was supposed to mean “We may be able to fix this [by some other means, fill in the blank here]”.

The “reworking the API addition” approach I had in mind was:
* `cigar_ind` was added to communicate some information from `resolve_cigar2()` (called only by `bam_plp_next()`) to the new `bam_plp_insertion()`
* so instead of adding `bam_plp_insertion()`, we could add an augmented `bam_plp_next()` that also returns the padded insertion sequence.  So no communication via `bam_pileup1_t` would be needed

However the invocations in PR samtools/samtools#331 are more complicated than just `bam_plp_insertion()` closely following `bam_plp_next()`, so perhaps this approach would cause more complexity on the samtools side…

jkbonfield:
I thought a reworking of bam_pileup1_t struct to allow constructor/destructor style callback functions and a generic void *client_data argument would be able to replace this.

However it cannot.  This PR contains modifications to resolve_cigar2 to copy out some of the contents of the cstate struct, in essence making the internal information public.  This is still a useful thing to do and it would appear to be an orthogonal change. Therefore this ABI breakage ought to be merged (in some form or other) at the same time.

jkbonfield:
I've rechecked this PR and it is still necessary in some form. See also samtools/samtools#331.  As the original author, it needs someone else to evaluate it more fully.

PR #230, opened 2015-06-25
Fix compilation with non-Kerberos iRODS

dkj:


jmarshall:
At the time, test/hfile didn't need -pthread; but really it should be regularised to be just like the others.  So good catch.

But what does this have to do with non-Kerberos?  Doesn't that need [this TODO](https://github.com/samtools/htslib/blob/develop/configure.ac#L73) fixed instead?

dkj:
I only experienced the link fail when linking with an iRODS compiled without Kerberos:
```
dj3@gs01-nfs:/nfs/gs01/tmp/dj3/htslib$ gcc  -L/usr/local/iRODS/lib/core/obj -o test/hfile test/hfile.o libhts.a  -lRodsAPIs -lgssapi_krb5 -lz
/usr/local/iRODS/lib/core/obj/libRodsAPIs.a(rcConnect.o): In function `_rcConnect':
/usr/local/iRODS/lib/core/src/rcConnect.c:148: undefined reference to `pthread_create'
/usr/local/iRODS/lib/core/obj/libRodsAPIs.a(rcConnect.o): In function `rcDisconnect':
/usr/local/iRODS/lib/core/src/rcConnect.c:296: undefined reference to `pthread_cancel'
/usr/local/iRODS/lib/core/src/rcConnect.c:297: undefined reference to `pthread_detach'
collect2: ld returned 1 exit status
```

The Kerberos libraries were installed on that system, but the iRODS install did not link against them, so the lack of AC_CHECK_LIB for libgssapi_krb5 wasn't a problem for me. Note, I've tried tweaking the config.in and configure.ac to remove libgssapi_krb5 and hit the same problem:
```
gcc  -L/usr/local/iRODS/lib/core/obj -o test/hfile test/hfile.o libhts.a  -lRodsAPIs  -lz
/usr/local/iRODS/lib/core/obj/libRodsAPIs.a(rcConnect.o): In function `_rcConnect':
/usr/local/iRODS/lib/core/src/rcConnect.c:148: undefined reference to `pthread_create'
/usr/local/iRODS/lib/core/obj/libRodsAPIs.a(rcConnect.o): In function `rcDisconnect':
/usr/local/iRODS/lib/core/src/rcConnect.c:296: undefined reference to `pthread_cancel'
/usr/local/iRODS/lib/core/src/rcConnect.c:297: undefined reference to `pthread_detach'
collect2: ld returned 1 exit status
make: *** [test/hfile] Error 1
```
I don't understand (yet) why linking to such an iRODS exposes a problem with missing pthread. Maybe building with iRODS should force the -pthread ?

jkbonfield:
Given htslib has a non-optional dependency on pthreads, it's just fluke that test/hfile happens to only pull in the .o files from the libhts.a that don't utilise pthreads.  Adding hfile_irods.o obviously brings in a dependency on libRodsAPIs.a, and adding that then *sometimes* appears to add a dependency on pthread.  I've no idea why one irods build needs it and another not, unless it happens to be the kerberos implementation itself that uses (directly or indirectly) something with a multi-threading capability.

The lack of a pkgconfig obviously makes it a bit guess work as to what we need, but pthread is used in all other tests so just adding it seems a no brainer.

jmarshall:
Thanks.  Merged with a more boring commit message as 5c51f8401462a507a090978a529dc94e08280bf6.

PR #232, opened 2015-06-26
Updated support for s3 urls.

DonFreed:
Added parsing for additional file types for hisremote.

Added code to incorporate an 'AWS_SESSION_TOKEN' environmental variable. Added this token as a CononicalizedAmzHeader to the signature message and a header incorporating the token. The token is used for granting temporary access.

Changed the request style from virtual hosted to path. The bucket I was accessing had a name that was [not DNS compliant] (http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html).

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -5,8 +5,17 @@
    This template file can be updated with autoheader, but do so carefully
    as autoheader adds #defines such as PACKAGE_* that we don't want.  */
 
+/* Define if you have the Common Crypto library. */
+#undef HAVE_COMMONCRYPTO
+
+/* Define to 1 if you have the `gmtime_r' function. */
+#undef HAVE_GMTIME_R
+
 /* Define to 1 if iRODS file access is enabled. */
 #undef HAVE_IRODS
+
+/* Define to 1 if you have the `crypto' library (-lcrypto). */
+#undef HAVE_LIBCRYPTO
 
 /* Define if libcurl file access is enabled. */
 #undef HAVE_LIBCURL
--- a/configure.ac
+++ b/configure.ac
@@ -38,6 +38,8 @@ redistribute it.  There is NO WARRANTY, to the extent permitted by law.])
 AC_PROG_CC
 AC_PROG_RANLIB
 
+need_crypto=no
+
 AC_ARG_WITH([irods],
   [AS_HELP_STRING([[--with-irods[=DIR]]],
                   [use RodsAPIs library (in DIR) to support iRODS URLs])],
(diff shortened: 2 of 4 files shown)


jmarshall:
Thanks for testing this out.  There's three parts to this pull request:

1. `hisremote()` additions — this is taken care of by the imminent plugin reorganisation

2. AWS_SESSION_TOKEN and temporary security credentials.  Okay, we should probably support this.  Is there any documentation around the interaction of the usual AWS_ACCESS_KEY_ID credentials and the temporary ones beyond what's at [this page](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html)?

3. Using path-style access.  I'd rather use the virtual hosted-style as it appears to be the new modern and to have better redirection possibilities for non-US regions, but we can fall back to path-style where needed.  Does 9344717b68594d44b781d0f4972b65c9c74d02f5 let you access your old-style-named bucket (and does its `is_dns_compliant()` look right to you?)?

DonFreed:
1. Great, I am glad to hear this will be taken care of.

2. Yes, there is a bit of documentation. These temporary credentials are created by a call to [`GetSessionToken`](http://docs.aws.amazon.com/STS/latest/UsingSTS/CreatingSessionTokens.html) or [`GetFederationToken`](http://docs.aws.amazon.com/STS/latest/UsingSTS/CreatingFedTokens.html) using the AWS [Security Token Service](http://docs.aws.amazon.com/STS/latest/UsingSTS/Welcome.html). For cloud access to the National Database for Autism Research ([NDAR](https://ndar.nih.gov/cloud_overview.html)), for example, researchers can use their username and password to generate these credentials for access to the database in S3.

3. Virtual hosted-style access is probably preferred and 1fec5f31a6c89f0b3e2da42053b2c8a58973be0d worked in my test case. `is_dns_compliant()` looked great except for the handling of buckets starting or ending with hyphens.

jmarshall:
> `is_dns_compliant()` looked great except for the handling of buckets starting or ending with hyphens.

D'oh!  Thanks for spotting this.

jmarshall:
The **libcurl** branch has (finally!) been merged to **develop** (PR #292).

1 and 3 have been taken care of.  2, AWS_SESSION_TOKEN and temporary security credentials, is still on the list to get added post-merge.

jmarshall:
I've now extended the AWS_SESSION_TOKEN stuff into PR #303, which looks for credentials (including temporary credentials) in the conventional AWS configuration files.  Apologies for the delay.

Please test PR #303 if you've got a few moments.  Thanks!

PR #236, opened 2015-06-30
Add index build/load functions with explicit index filenames

jmarshall:
Please review proposed API (naming, usage of `fnidx==NULL`, etc)…


jkbonfield:
Still curious on the errno bit, but it's completely harmless on all our systems (and ma

PR #239, opened 2015-07-06
bug for 8-bit arrays

pb-cdunn:
Reported by Derek Barnett of Pacific Biosciences:

> There was (& still is) an off-by-one error in htslib's handling of 8-bit array
> tags. The change in htslib that triggered it was recent-ish, and I don't know
> how many people are using 8-bit arrays, instead of strings. And the kicker is
> that it *only* appears if that tag is the last one on the record. Hence why it
> may not have been noticed yet.


pb-cdunn:
Nvm. Only a bug in the context of other changes. No issue.

PR #241, opened 2015-07-15
Orodeh ubcf overhaul

orodeh:
Here are my changes that allow using uncompressed BCF.


orodeh:
Sorry, this was supposed to be an internal DNAnexus merge request. We are working on a fork of the htslib library. However, since I had already made the initial mistake of sending it to the core maintainers, we might as well proceed with it. 

The core idea in my changes are the allow for uncompressed BCF. This turns out to require typing changes to some central macros and function. This is unfortunate, I think you will agree. The reason we want to use uncompressed BCF, is because a database is doing the compression for us. We do not want to do "double compression". 

Please review it, and let me know what the issues are. 

Thanks,
       Ohad.

atks:
https://github.com/samtools/htslib/blob/develop/htslib/hts.h#L237

htslib already allows for uncompressed bcf, just use hts_open with "bu"  mode.

orodeh:
Well, it doesn't work though. This is what all the code is for.

atks:
I have been using it and it is working fine.  Did you use the most recent version of htslib?

orodeh:
I looked at the most recent code base:
https://github.com/samtools/htslib/blob/develop/hts.c

The hts_close routine, for the BCF case, assumes the file is in BGZF format.
https://github.com/samtools/htslib/blob/develop/hts.c#L493

The hts_hopenroutine always opens the file in BGZF mode, if it is in BCF
format.
https://github.com/samtools/htslib/blob/develop/hts.c#L432

Our pull request extends the htslib functionality to uncompressed BCF
files. For example, the hts_open/close routines can work with BGZF
compression as well as without it.

I hope this sheds light on the issue.

Thanks,
        Ohad.

On Wed, Jul 15, 2015 at 8:08 PM, Adrian Tan <notifications@github.com>
wrote:

> I have been using it and it is working fine. Did you use the most recent
> version of htslib?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/samtools/htslib/pull/241#issuecomment-121811821>.
>

-- 
---
The contents of this e-mail and any attachments are confidential and only 
for use by the intended recipient. Any unauthorized use, distribution or 
copying of this message is strictly prohibited. If you are not the intended 
recipient please inform the sender immediately by reply e-mail and delete 
this message from your system. Thank you for your co-operation.

mlin:
@orodeh Note this PR currently includes some of our previous htslib patches to speed up bgzf_mt and such. There's a separate open PR for those (#51) though it needs updating. I think those changes are orthogonal, so when we're ready, we should make a branch with your new patches but not those old ones. Maintainers: your discretion whether to keep this PR open as a sneak preview in the meantime!

atks:
Hi Ohad,

Thanks for explaining it.  I think bgzf_hopen() and bgzf_close() are aware of uncompressed files too.  

https://github.com/samtools/htslib/blob/develop/bgzf.c#L215 
https://github.com/samtools/htslib/blob/develop/bgzf.c#L111
https://github.com/samtools/htslib/blob/develop/bgzf.c#L797

For me, when I generate an uncompressed file using htslib via vt.
1. vt view test.bcf -o + > a.ubcf
and then uncompress the original bcf file.
2. zcat test.bcf  > b.ubcf

diff a.ubcf b.ubcf shows that both files are identical. 

But I might be misunderstanding something.  I'll let the developers clarify this.

mcshane:
This is not my area, but I think the uncompressed bcf used by htslib is actually uncompressed, but written as uncompressed bgzf blocks. This allows random access on an indexed bgzf uncompressed bcf, but I guess is blocking your particular use case?

orodeh:
I will have to chew on this. I am starting to wonder if uncompressed BCF,
as we need it (without an index, and no BGZF blocks), is going to require a
major change to htslib.

Thanks,
         Ohad.

On Thu, Jul 16, 2015 at 1:49 PM, Shane McCarthy <notifications@github.com>
wrote:

> This is not my area, but I think the uncompressed bcf used by htslib is
> actually uncompressed, but written as uncompressed bgzf blocks. This allows
> random access on an indexed bgzf uncompressed bcf, but I guess is blocking
> your particular use case?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/samtools/htslib/pull/241#issuecomment-122089153>.
>

-- 
---
The contents of this e-mail and any attachments are confidential and only 
for use by the intended recipient. Any unauthorized use, distribution or 
copying of this message is strictly prohibited. If you are not the intended 
recipient please inform the sender immediately by reply e-mail and delete 
this message from your system. Thank you for your co-operation.

pd3:
The bgzf library detects plain files. It can also read/write compressed blocks, uncompressed blocks wrapped in zlib, or plain uncompressed files. I believe this is propagated to htslib
https://github.com/samtools/htslib/blob/develop/htslib/hts.h#L228
https://github.com/samtools/htslib/blob/develop/htslib/bgzf.h#L93

orodeh:
Very well, it looks like the code does support uncompressed BCF. 

I tried it, by changing line:
https://github.com/samtools/htslib/blob/develop/test/test-vcf-api.c#L34
to 
     htsFile *fp    = hts_open(fname,"wbu");
This causes the test to fail, requiring a fix to line:
https://github.com/samtools/htslib/blob/develop/test/test-vcf-api.c#L256
    idx = bcf_index_load(fname);
    if (idx == NULL)
        return;

I believe this pull request is no longer relevant, so I am closing it. 

Thanks you, 
   Ohad.

PR #242, opened 2015-07-16
Improved error checking and reporting for htslib

daviesrob:
This is an updated version of pull request #211.

The error message printing is simplified a bit by making bgzf_zerr return strings instead of writing to a file descriptor itself.  That means the error reporting can be done with a single printf instead of a printf, a call to bgzf_err and a puts.

Removed calls to abort().  Instead function prototypes have had HTS_RESULT_USED added to them (where useful) so that gcc warns when the return value is not being checked.  This has been done in bgzf.h, sam.h, hts.h and faidx.h.  Fixes for all of the warnings are included, apart from some that should be sorted by other pull requests.  At the same time, a few other cases of missing error checks have been fixed.

The error handling code in a few functions has been improved so that they shouldn't leak memory or keep file handles open when they return due to detecting an error.


bgzf.c
```diff
@@ -182,6 +210,7 @@ BGZF *bgzf_open(const char *path, const char *mode)
         hFILE *fpw;
         if ((fpw = hopen(path, mode)) == 0) return 0;
```
mp15: For readability could we make these == NULL and return NULL too? Makes it easier to read.

bgzf.c
```diff
@@ -182,6 +210,7 @@ BGZF *bgzf_open(const char *path, const char *mode)
...
         fp = bgzf_write_init(mode);
+        if (fp == NULL) return NULL;
```
mp15: Should we set errno=ENOMEM here and in similar calls?

bgzf.c
```diff
@@ -994,25 +1147,41 @@ int bgzf_index_dump(BGZF *fp, const char *bname, const char *suffix)
...
-            fwrite(&fp->idx->offs[i].caddr, 1, sizeof(fp->idx->offs[i].caddr), idx);
-            fwrite(&fp->idx->offs[i].uaddr, 1, sizeof(fp->idx->offs[i].uaddr), idx);
+            if (fwrite(&fp->idx->offs[i].caddr, 1, sizeof(fp->idx->offs[i].caddr), idx) < sizeof(fp->idx->offs[i].caddr)) goto fail;
+            if (fwrite(&fp->idx->offs[i].uaddr, 1, sizeof(fp->idx->offs[i].uaddr), idx) < sizeof(fp->idx->offs[i].uaddr)) goto fail;
         }
     }
-    fclose(idx);
+    if (fclose(idx) < 0)
+    {
+        fprintf(stderr, "[E::%s] Error on closing %s%s : %s\n",
```
mp15: Shouldn't this be wrapped in hts_verbose?

bgzf.c
```diff
@@ -1123,3 +1319,27 @@ long bgzf_utell(BGZF *fp)
...
 }
 
+static const char * bgzf_zerr(int errnum)
+{
+    static char buffer[32];
+    // gzerror OF((gzFile file, int *errnum)
+    switch (errnum) {
+    case Z_ERRNO:
+        return strerror(errno);
+    case Z_STREAM_ERROR:
```
mp15: This can also be a CRC error.

mp15:
I've reviewed the changes and they look good to me, added a few changes but other than that looks good to go.

bgzf.c
daviesrob: I could, but it would make the patch bigger, and I don't really want to change anything I don't have to for the sake of making it look more pretty.

bgzf.c
daviesrob: errno should be ENOMEM already if malloc failed in bgzf_write_init.  bgzf_write_init can also fail in deflateInit2.  In that case setting errno=ENOMEM might not be correct.  Setting errno=EINVAL for an invalid mode string (in the existing code for bgzf_open) does make sense though.

bgzf.c
daviesrob: Yes, will update.

bgzf.c
daviesrob: Probably covered under "inconsistent stream state"?  The text for this one is a bit of a mess as Z_STREAM_ERROR can have different meanings depending on which function returned it.

PR #251, opened 2015-08-07
Ensure minimum stack size for cram thread pool threads.

daviesrob:
MacOS X pthreads by default has a thread stack size of 512K (see
https://developer.apple.com/library/mac/qa/qa1419/_index.html).
rans_compress_O1() needs a bit over 1.25Mb, so this alters t_pool_init()
to create threads with a stack of at least 4Mb, which hopefully is plenty.

Also take the opportunity to add better error checking and clean-up to
t_pool_init().

### DIFF
--- a/cram/thread_pool.c
+++ b/cram/thread_pool.c
@@ -41,6 +41,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "cram/thread_pool.h"
 
+#define MIN_THREAD_STACK 4194304  /* Require at least 4Mb stack per thread */
+
 //#define DEBUG
 //#define DEBUG_TIME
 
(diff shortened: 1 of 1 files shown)


cram/thread_pool.c
```diff
@@ -434,17 +449,33 @@ t_pool *t_pool_init(int qsize, int tsize) {
...
 #endif
 
-    p->t = malloc(tsize * sizeof(p->t[0]));
+    if (!(p->t = malloc(tsize * sizeof(p->t[0]))))
+        goto fail;
 
-    pthread_mutex_init(&p->pool_m, NULL);
-    pthread_cond_init(&p->empty_c, NULL);
-    pthread_cond_init(&p->full_c, NULL);
+    if (0 != pthread_mutex_init(&p->pool_m, NULL)) goto fail;
```
jmarshall: For the record, Yoda conditionals are not the convention in extant samtools/htslib source code, and probably some of the maintainers hate them.  And they're kind of pointless when the operator is `!=`…

jmarshall:
> `rans_compress_O1()` needs a bit over 1.25Mb [of stack]

Hmmm… given that `rans_compress_O1()` is already calling `malloc()` for something else (of the order of 200K), perhaps there's not too much reason not to convert `syms` to be malloced too?

daviesrob:
On Mon, 10 Aug 2015, John Marshall wrote:

>> `rans_compress_O1()` needs a bit over 1.25Mb [of stack]
>
> Hmmm… given that `rans_compress_O1()` is already calling `malloc()` for 
>something else (of the order of 200K), perhaps there's not too much 
>reason not to convert `syms` to be malloced too?

It mallocs the output buffer, which of course needs to be passed back so 
can't be on the stack.

You're probably right, though.  I think it gets called once per block, so 
the overhead of doing a malloc would probably be fairly low.


-- 

PR #252, opened 2015-08-09
Export bgzf_compress() in bgzf.h

noporpoise:
I'd like to write bgzf files in a current project. Access to the bgzf block compress function without having to use BGZF file handles would be very useful.


mp15:
Agreed however, before we make this a public function slen and dlen should at least be made unsigned. Want to add that to your patch?

jmarshall:
Thanks.  Squashed and applied as 82bdc9f02d905916dd70774ee98770e7d425a489.

PR #255, opened 2015-08-17
Use -lpthread instead of -pthread when only linking; Fixes clang warning

noporpoise:
`-pthread` is equivalent to `-D_REENTRANT=1 -D__USE_REENTRANT=1` when compiling and `-lpthread` when linking, although the exact behaviour may be plaform specific. 

Clang gives a warning if `-pthread` is used when only linking `.o` and `.a` files:

```
clang: warning: argument unused during compilation: '-pthread'
```

This is fixed by using `-lpthread` instead if a compiler call is only linking.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,13 +270,13 @@ cram/zfio.o cram/zfio.pico: cram/zfio.c config.h cram/os.h cram/zfio.h
 
 
 bgzip: bgzip.o libhts.a
-	$(CC) -lpthread $(LDFLAGS) -o $@ bgzip.o libhts.a $(LDLIBS) -lz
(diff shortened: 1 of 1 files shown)


jmarshall:
Those warnings are indeed annoying :smile:

The makefile currently uses `-pthread` precisely because the exact behaviour is platform-specific.

In fact, you [don't need any `-pthread`/`-lpthread` at all on OS X](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man3/pthread.3.html) and it can be [a good idea](https://bugzilla.mozilla.org/show_bug.cgi?id=299661#c2) to avoid it entirely.  So I would prefer to fix this via configure so that on OS X no pthreads-related options are used.

jkbonfield:
Squashed and merged via e2a1f187e2fa73aaed94ca75285397d9a63e5440

PR #256, opened 2015-08-17
Provide updated compressBound() for bgzip blocks

noporpoise:
bgzip blocks have a higher bound on compressed size due to slightly larger headers.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -108,6 +108,11 @@ static inline void packInt32(uint8_t *buffer, uint32_t value)
     buffer[3] = value >> 24;
 }
 
+static size_t bgzf_compressBound(size_t srclen)
+{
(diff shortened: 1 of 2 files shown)


daviesrob:
Closing as the original asserts are correct.  The size extra bytes in question are accounted for in `BLOCK_HEADER_LENGTH` (see the diagram near the top of `bgzf.c`).  They don't need to be included in the calculation made by `compressBound()`, which is only relevant to the data being compressed.

As an aside, the assert will never trigger in either case.  `BGZF_BLOCK_SIZE` is defined as `0xff00` in `htslib/bgzf.h`.  `compressBound(0xff00)` is `0xff1f`, which is always less that `BGZF_MAX_BLOCK_SIZE (0x10000)`, even if you add another 6 to it.

PR #258, opened 2015-08-23
allow 0-length fasta label

mtmorgan:
A fasta file with no sequence name segfaults during index creation.

    ~/src/samtools$ cat > noname.fa
    >
    ATCG
    ^D
    ~/src/samtools$ ./samtools faidx noname.fa
    Segmentation fault (core dumped)

because the null terminator is written to a null pointer. The patch allocates enough space for the terminating null. An alternative (if sequence names are supposed to be required) would be to exit more gracefully.

### DIFF
--- a/faidx.c
+++ b/faidx.c
@@ -35,6 +35,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include "htslib/faidx.h"
 #include "htslib/hfile.h"
 #include "htslib/khash.h"
+#include "htslib/kstring.h"
 
 typedef struct {
     int32_t line_len, line_blen;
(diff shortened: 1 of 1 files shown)


jmarshall:
Thanks for the suggestion.  An unnamed fasta sequence seems barely valid or usable, although I see the [closest thing we have to a FASTA spec](https://en.wikipedia.org/wiki/FASTA_format#Format) says this identifier is optional (citation needed!).  Rather to my surprise, `samtools faidx noname.fa :1-4` retrieves it successfully, so I suppose there is no particular reason for samtools/htslib to disallow an unnamed sequence if someone really wants to do that.

The suggested patch isn't quite right as it produces the wrong output given an empty input file.  The better fix would be to redo the header parsing (within `if (c == '>')`) with `name` as a kstring and fixing samtools/samtools#449 at the same time.

mtmorgan:
I don't have that quite right yet.

jmarshall:
Thanks for your efforts, Martin.  Unfortunately I wound up taking a look at it too…

mtmorgan:
No 'unfortunately' needed; thanks for addressing the issue, and always interesting to see how someone who actually knows the code solves the problem.

Issue #248, opened 2015-07-31
Header parsing bug leads to missing VCF columns in output


pd3:
Please see the error description here https://github.com/atks/vt/issues/33

cyenyxe:
There are two possible solutions for this, depending on how error tolerant the library is expected to be:

* VCF v4.1 and v4.2 specifications don't flag trailing spaces as errors, so they could be accepted (just by trimming the input line).
* VCF v4.3 won't allow them, so an error should be displayed.

Both are easy to implement. If #214 is acceptable for merging, then probably trimming the line would be the way to go, just for the sake of consistency.

And the implementation question is, in which of the *vcf_hdr_read* and *bcf_hdr_parse_line* functions the changes should be done? *bcf_hdr_parse_line* seems better to me, but then the way it notifies errors needs to be improved too.

jmarshall:
Fixed by PR #267.

PR #267, opened 2015-09-11
Function bcf_hdr_parse_line now ignores trailing spaces

cyenyxe:
Making the function bcf_header_parse ignore trailing spaces fixes issue #248.

A test file is included, but I haven't been able to integrate it with the rest of the VCF test suite (test/test.pl). For some reason, it won't copy the input file into the temporary folder.


jmarshall:
Thanks Cristina.  I realised that htslib has `htsfile -c` that can be used to view VCF and exercise the parser, so I took the liberty of using your test data files with that instead.

PR #268, opened 2015-09-14
Fixed for handling of BCF string dictionaries

mcshane:
Gaps in IDX blocks were causing incorrect tags on output.

Fixes samtools/bcftools#317.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -56,8 +56,6 @@ static bcf_idinfo_t bcf_idinfo_def = { .info = { 15, 15, 15 }, .hrec = { NULL, N
  *** VCF header parser ***
  *************************/
 
-int bcf_hdr_sync(bcf_hdr_t *h);
-
 int bcf_hdr_add_sample(bcf_hdr_t *h, const char *s)
 {
     if ( !s ) return 0;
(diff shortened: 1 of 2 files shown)


vcf.c
```diff
@@ -118,31 +116,19 @@ int bcf_hdr_parse_sample_line(bcf_hdr_t *h, const char *str)
 int bcf_hdr_sync(bcf_hdr_t *h)
 {
     int i;
-    for (i = 0; i < 3; i++)
+    for (i=0; i<3; i++)
```
jmarshall: This line is unchanged, so let's not change it.

jmarshall:
This changes the layout of `struct bcf_hdr_t`.  Does this have binary compatibility implications?

jmarshall:
Indeed, there are public inline functions and macros that directly access the fields of this structure.  However allocating it is encapsulated in `bcf_hdr_init()` so adding extra fields at the end of the struct would probably not be a binary compatibility concern.

mcshane:
I've moved `m` to the end of the struct now.

mp15:
Looks like this no longer breaks ABI provided bcf_hdr_init and bcf_hdr_destroy are used correctly.

PR #271, opened 2015-09-16
Improved error checking and reporting for htslib (rebased)

daviesrob:
This is a rebased version of #242, which was an updated version of pull request #211.

The error message printing is simplified a bit by making bgzf_zerr return strings instead of writing to a file descriptor itself. That means the error reporting can be done with a single printf instead of a printf, a call to bgzf_err and a puts.

Removed calls to abort(). Instead function prototypes have had HTS_RESULT_USED added to them (where useful) so that gcc warns when the return value is not being checked. This has been done in bgzf.h, sam.h, hts.h and faidx.h. Fixes for all of the warnings are included, apart from some that should be sorted by other pull requests. At the same time, a few other cases of missing error checks have been fixed.

The error handling code in a few functions has been improved so that they shouldn't leak memory or keep file handles open when they return due to detecting an error.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -88,7 +88,7 @@ struct __bgzidx_t
 
 void bgzf_index_destroy(BGZF *fp);
 int bgzf_index_add_block(BGZF *fp);
-static const char * bgzf_zerr(int errnum);
+static const char * bgzf_zerr(int errnum, z_stream *zs);
 
 static inline void packInt16(uint8_t *buffer, uint16_t value)
 {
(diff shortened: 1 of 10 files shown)


daviesrob:
I've added a patch to output the z_stream msg.  For example:
```
# Make a dud bam file
cp test/ce#unmap1.tmp.bam /tmp/b.bam
echo foo | dd of=/tmp/b.bam bs=1 seek=300 conv=notrunc
# Try to read it
./htsfile -c /tmp/b.bam
@HD	VN:1.0	SO:unsorted
@SQ	SN:CHROMOSOME_I	LN:1009800
@SQ	SN:CHROMOSOME_II	LN:5000
@SQ	SN:CHROMOSOME_III	LN:5000
@SQ	SN:CHROMOSOME_IV	LN:5000
@SQ	SN:CHROMOSOME_V	LN:5000
@SQ	SN:CHROMOSOME_X	LN:5000
@SQ	SN:CHROMOSOME_MtDNA	LN:5000
@RG	ID:UNKNOWN	SM:UNKNOWN
@PG	ID:bowtie2	PN:bowtie2	VN:2.0.0-beta5
[E::inflate_block] inflate failed: invalid distance too far back
[E::bgzf_read_block] inflate_block error -1
[E::bgzf_read] bgzf_read_block error -1 after 0 of 4 bytes
```

I think the reason I ignored it before is because it's not set all the time, which makes it a bit tricky to use.  Hopefully what I have here does the right thing in most cases.

What would you want hts_verbose to be to get these messages?  I tried 4 but that made the test above just stopped silently, with a non-zero exit as the only indication that anything went wrong.  Samtools view says:
```
[main_samview] truncated file.
```
which is a bit better.  How chatty do we want to be when we fail?  The messages only appear on errors, so they shouldn't trouble anyone reading valid files.

jkbonfield:
I think this should be 1.3.1 milestone?

jmarshall:
Thanks Rob.  Rebased again, added to some old `hts_idx_save_core()` patches of mine, generally messed around with, and merged as 5285dc03725e152d0e4db8b8fefe27bdd80520d2.

PR #272, opened 2015-09-18
Enable multithreaded compression in bgzip

jpiper:
I noticed that there is support for multithreaded bgzip compression in `htslib` but this wasn't being surfaced to the user in the `bgzip` executable. I added it with the same command line flag `-@` to match the same as the multithreaded compression on `samtools sort`.

### DIFF
--- a/bgzip.c
+++ b/bgzip.c
@@ -79,6 +79,7 @@ static int bgzip_main_usage(void)
     fprintf(stderr, "   -b, --offset INT        decompress at virtual file pointer (0-based uncompressed offset)\n");
     fprintf(stderr, "   -c, --stdout            write on standard output, keep original files unchanged\n");
     fprintf(stderr, "   -d, --decompress        decompress\n");
+    fprintf(stderr, "   -@, --threads           number of compression threads to use\n");
     fprintf(stderr, "   -f, --force             overwrite files without asking\n");
     fprintf(stderr, "   -h, --help              give this help\n");
     fprintf(stderr, "   -i, --index             compress and create BGZF index\n");
@@ -96,12 +97,14 @@ int main(int argc, char **argv)
     void *buffer;
     long start, end, size;
     char *index_fname = NULL;
+    int threads = 1;
 
     static struct option loptions[] =
     {
         {"help",0,0,'h'},
         {"offset",1,0,'b'},
         {"stdout",0,0,'c'},
+        {"threads",0,0,'@'},
         {"decompress",0,0,'d'},
         {"force",0,0,'f'},
         {"index",0,0,'i'},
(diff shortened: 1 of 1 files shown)


bgzip.c
```diff
@@ -96,12 +97,14 @@ int main(int argc, char **argv)
...
     long start, end, size;
     char *index_fname = NULL;
+    int threads = 1;
 
     static struct option loptions[] =
     {
         {"help",0,0,'h'},
         {"offset",1,0,'b'},
         {"stdout",0,0,'c'},
+        {"threads",0,0,'@'},
```
jmarshall: Should be `required_argument`

jpiper:
I'd think it would be better if instead of using 1 thread by default and asking the user to set more, it would make more sense to set n_threads to the total number available, and have the user state when they want less.

This could be achieved, for example, by using

    #include <unistd.h>
    int n_threads = (int)sysconf(_SC_NPROCESSORS_ONLN);

Thoughts? Likewise, I would be happy to go through all of the tools that could be threaded and enable this and submit a pull request.

jkbonfield:
On Mon, Sep 21, 2015 at 08:41:52AM -0700, Jason Piper wrote:
> I'd think it would be better if instead of using 1 thread by default
> and asking the user to set more, it would make more sense to set
> n_threads to the total number available, and have the user state when
> they want less. 

I think it would be good to have a synonum for this, but making it the
default will change behaviour of a lot of pipeline code.

Eg I could see --threads=0 to mean _SC_NPROCESSORS_ONLN, but default
of --threads=1 is, IMO, the path of least surprise.  (I'd agree to
your suggestion if we were starting again with a totally new tool.)

> Thoughts? Likewise, I would be happy to go through all of the tools
> that could be threaded and enable this and submit a pull request. 

Most tools can now handle encoding and decoding threads, but not in an
obvious way.

I went through modifying all tools to support --input-fmt-options and
--output-fmt-options to specify extra arguments that may be useful for
CRAM.  Nominally this is a reference sequence, although it's often
known, but also because when creating a cram file it may be things
like how many sequences per slice to encode or whether the reference
should be embedded.  At this stage it also has a nthreads parameter.

So now even tools that don't have an explicit -@ or --threads param
have gained one via another route. Eg:

samtools split --output-fmt-option nthreads=16 foo.bam

Looking more closely I think I've found some subcommands that aren't
as complete as I originally expected, oops.  However the theory is
there.

Ideally we would make this available more generically as --threads,
but this is complicate as we don't know what those threads are for.
Long term we want to overhaul this a bit more to make use of thread
pools so that input, output and any computation stages can all make
use of a single pool of threads and auto-balance between them.  (There
is code for this in CRAM encoding/decoding only at the moment, and
that needs some revision too.)


-- 

jmarshall:
Thanks, now squashed and merged as 2b2239a8525e3f86912142df27e105f31b713b82.

PR #274, opened 2015-09-22
vcf.h: Remove const specifiers that are not honoured

noporpoise:
`const` specifiers were being ignored in the header, resulting in many warnings in third party code compiling against htslib.

```
libs/htslib/htslib/vcf.h:880:24: warning: cast from 'const unsigned char *' to 'unsigned char *' drops const qualifier [-Wcast-qual]
    *q = (uint8_t*)p + 1;
```

### DIFF
--- a/htslib/vcf.h
+++ b/htslib/vcf.h
@@ -874,7 +874,7 @@ static inline void bcf_enc_int1(kstring_t *s, int32_t x)
     }
 }
 
-static inline int32_t bcf_dec_int1(const uint8_t *p, int type, uint8_t **q)
+static inline int32_t bcf_dec_int1(uint8_t *p, int type, uint8_t **q)
 {
     if (type == BCF_BT_INT8) {
         *q = (uint8_t*)p + 1;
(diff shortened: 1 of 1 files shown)


mp15:
Surely it would be better just to fix the cast to:
    *q = (const uint8_t*)p + 1;
It's just pointer arithmetic so it's not modifying the value.

noporpoise:
Then you need to change the type of parameter `q` to `const uint8_t **q`, which gives warnings about each time it's called in htslib. I haven't check each of them to see if they can need non-const `q`.

The problem is the functions `bcf_dec_int1()`, `bcf_dec_typed_int1()` and `bcf_dec_size()` take parameters `const uint8_t *p` and `uint8_t **q`, where `p` is const and `q` is not. They then set `*q` to point to `p` with some offset. This violates `const` correctness. If the const is not being obeyed it should be removed.

jmarshall:
This is the classic C-function-wants-to-work-on-either-const-or-non-const-strings problem as exemplified by `strchr`: [in C `strchr`](http://en.cppreference.com/w/c/string/byte/strchr) provides the same escape hatch from `const char *` to `char *` as these functions do (and is similarly not pedantically-const-correct), whereas [in C++](http://en.cppreference.com/w/c/string/byte/strchr) there are two separate functions overloaded.

It's made worse here by having the output string as passed by reference as a parameter rather than returned.

There are various solutions:

1. Don't compile htslib headers with -Wcast-qual.  If htslib headers are being #included as system headers then this is (probably) trivial; otherwise please don't tell us why your coding standards say you can't turn -Wcast-qual off as we already understand the arguments :smile:

2. If HTSlib did not have these functions inline in the public header file, then you could compile your own application code with -Wcast-qual without triggering the diagnostic on these functions.

3. HTSlib could provide separate `const` and non-`const` versions of these functions.  As we don't have function overloading in C, they would need to have different names.  (We might find that we only currently need one of each pair… currently (at a glance) most uses in vcf.c are non-const, so this PR would suffice — at the cost of being a pain for future code that tries to be const-correct.)

noporpoise:
Thanks @jmarshall that all makes sense. Solution 2 seems like the best solution: it won't change the API and we don't have to adjust our compile options to satisfy htslib.

noporpoise:
Closing in favour of #275.

noporpoise:
Reopening, inlining is required so removing `const` hopefully satisfies everyone. If anyone needs to use with `const uint8_t*`, they can cast to `uint8_t*`.

PR #275, opened 2015-09-23
Move bcf_dec_int1() bcf_dec_size() defs from vcf.h -> vcf.c

noporpoise:
`bcf_dec_int1()` and `bcf_dec_size()` cast `const` pointers to non-`const`, which gives
multiple warnings to those compiling third party code which uses `vcf.h` with
`-Wcast-qual` compile flags. Moving function definitions into `vcf.c` hides these
details from sensitive third parties.

This PR is instead of PR #274.


daviesrob:
Closing in favour of #274

PR #276, opened 2015-09-25
Expose cram_get_refs() function to return the opaque refs_t data type.

jkbonfield:
This permits explicit sharing of references between file descriptors.


mp15:
This looks like an easy merge, did we agree how desirable this is?

jmarshall:
Used by samtools/samtools#471.

PR #277, opened 2015-09-25
Fix bcf_empty1() function declaration typo in vcf.h

noporpoise:
vcf.h declares bcf_empty()
vcf.c defines bcf_empty1()

I've renamed the header declaration to bcf_empty1()

### DIFF
--- a/htslib/vcf.h
+++ b/htslib/vcf.h
@@ -261,7 +261,7 @@ typedef struct {
      *  Same as bcf_destroy() but frees only the memory allocated by bcf1_t,
      *  not the bcf1_t object itself.
      */
-    void bcf_empty(bcf1_t *v);
+    void bcf_empty1(bcf1_t *v);
 
     /**
      *  Make the bcf1_t object ready for next read. Intended mostly for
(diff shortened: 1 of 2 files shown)


jmarshall:
Actually following 3f4ecdf1fdcc41319d8d68313a6bbd4924312fcc the intended spelling would be `bcf_empty()`, with a `bcf_empty1()` macro provided for compatibility consistency.

Were you trying to use `bcf_empty()` and ran into problems, or did you just happen to notice this in passing?

noporpoise:
I was trying to use it. 

I was worried renaming an exported function would break ABI compatibility. Currently the way to use `bcf_empty()`/`bcf_empty1()` is to manually declare `bcf_empty1()` before calling it, as is done in `vcf_sweep.c`.

jmarshall:
There are indeed ABI considerations, however as there is currently no declaration of a working `bcf_empty[anything]` function in the public header file, API and ABI considerations are moot.

If you were trying to use it, I guess we had better fix it :smile:

noporpoise:
I've created a new PR #281.

PR #281, opened 2015-09-29
Fix bcf_empty(), bcf_init(), bcf_destroy() definitions

noporpoise:
vcf.h declares:

    bcf1_t *bcf_init(void);
    void bcf_empty(bcf1_t *v);
    void bcf_destroy(bcf1_t *v);

but vcf.c defines:

    bcf1_t *bcf_init1()
    void bcf_empty1(bcf1_t *v)
    void bcf_destroy1(bcf1_t *v)

So I've changed the vcf.c definitions to match the vcf.h declarations and added
the missing macro for bcf_empty().

I've also updated calls in htslib to init/empty/destroy to use the functions
rather than macros.

Note: I've got no idea how/why htslib compiled before, since `bcf_init()` was called (or called via macro `bcf_init1()` -> `bcf_init()`) but never declared anywhere. Can anyone explain?

### DIFF
--- a/htslib/vcf.h
+++ b/htslib/vcf.h
@@ -238,6 +238,7 @@ typedef struct {
     #define vcf_parse1(s,h,v)   vcf_parse((s),(h),(v))
     #define bcf_clear1(v)       bcf_clear(v)
     #define vcf_format1(h,v,s)  vcf_format((h),(v),(s))
+    #define bcf_empty1(v)       bcf_empty(v)
 
     /**
      *  bcf_hdr_init() - create an empty BCF header.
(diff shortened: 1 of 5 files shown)


jmarshall:
Thanks.  I've fixed this via d4482eb578aada0080eb30239ff57e9f6a6de93b as I prefer not to alter the calls to use the functions rather than the macros, as there's already a mishmash of calls of the `foo1` versions of various of the other functions too.

The macros also act on the function definitions, so `bcf1_t *bcf_init1() { … }` in _vcf.c_ was in fact defining a function with a different name.  This is technically known as obfuscation :smile:

noporpoise:
Ah thanks, makes sense.

PR #285, opened 2015-10-08
vcfutils: add bcf_remove_alleles_set

mcshane:
* new bcf_remove_alleles_set function to the new kbitset.h api to indicated alleles
  to be removed. This avoids overflowing the old integer bitmask in bcf_remove_alleles
  when there are more than 31 alleles
* bcf_remove_alleles updated to use the new bcf_remove_alleles_set
* Fixes samtools/bcftools#319

### DIFF
--- a/htslib/vcfutils.h
+++ b/htslib/vcfutils.h
@@ -1,6 +1,6 @@
 /*  vcfutils.h -- allele-related utility functions.
 
-    Copyright (C) 2012, 2013 Genome Research Ltd.
+    Copyright (C) 2012, 2013, 2015 Genome Research Ltd.
 
     Author: Petr Danecek <pd3@sanger.ac.uk>
 
@@ -31,6 +31,8 @@ DEALINGS IN THE SOFTWARE.  */
 extern "C" {
 #endif
 
+struct kbitset_t;
+
 /**
  *  bcf_trim_alleles() - remove ALT alleles unused in genotype fields
  *  @header:  for access to BCF_DT_ID dictionary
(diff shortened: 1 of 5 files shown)


htslib/vcfutils.h
```diff
@@ -1,6 +1,6 @@
 /*  vcfutils.h -- allele-related utility functions.
 
-    Copyright (C) 2012, 2013 Genome Research Ltd.
+    Copyright (C) 2012-2015 Genome Research Ltd.
```
jmarshall: Not sure there were copyright-significant changes in 2014

vcfutils.c
```diff
@@ -198,19 +198,32 @@ int bcf_trim_alleles(const bcf_hdr_t *header, bcf1_t *line)
...
+    int i;
+    kbitset_t *bset = kbs_init(line->n_allele);
+    for (i=1; i<line->n_allele; i++)
+        if ( rm_mask & 1<<i ) kbs_insert(bset, i);
+
+    bcf_remove_alleles_set(header, line, bset);
+    kbs_destroy(bset);
+}
+
+void bcf_remove_alleles_set(const bcf_hdr_t *header, bcf1_t *line, kbitset_t *bset)
```
jmarshall: const kbitset_t *

htslib/vcfutils.h
```diff
@@ -26,6 +26,7 @@ DEALINGS IN THE SOFTWARE.  */
 #define HTSLIB_VCFUTILS_H
 
 #include "vcf.h"
+#include "htslib/kbitset.h"
```
jmarshall: Changing includes ⇒ changing dependencies in the makefile.  Also what I was suggesting in _Example of declaring…_ in kbitset.h was to not include the kbitset header here, but for this header to have

    struct kbitset;

here (actually, just below the `extern "C"` bit) and

    void bcf_remove_alleles_set(const bcf_hdr_t *header, bcf1_t *line, const struct kbitset_t *rm_set);
                                                                             ^^^^^^^^^^^^^^^^

below.  Only _vcfutils.c_ actually needs to `#include <htslib/kbitset.h>` and correspondingly have `$(htslib_kbitset_h)` in its objects' dependencies.

mcshane:
Updated PR after @jmarshall comments.

jmarshall:
:+1:

(Is it `bcf_remove_alleles_set` or `bcf_remove_allele_set`?)

mcshane:
I'm not sure of the *set context you and James have in mind, so I'm reading it as remove alleles (plural) that have the bit set.

jmarshall:
I was thinking of a _(mathematical) set of alleles_… which I would write as `allele_set`, but YMMV.

mcshane:
You have a supporter in @daviesrob, so changed to `bcf_remove_allele_set`

PR #286, opened 2015-10-08
Switch some bcf1_t members to 'unsigned int'

gatoravi:
I think these variables could be unsigned, this came up while comparing with positions from another library.


mcshane:
The BCF spec has these fields as int32_t, so this would be a change to the spec. https://github.com/samtools/hts-specs

mcshane:
Closing. This should be brought up on hts-specs if you can make the case for having these as unsigned ints.

PR #287, opened 2015-10-09
bcf_write: sync dirty header before writing

mcshane:
* const has been removed... is that okay?
* Fixes samtools/bcftools#332


jkbonfield:
Looks fine to me.  As mentioned it removes the const, but this is necessary in order to fix the header.  The change also fixes a FIXME and given the already existing check for h->dirty it doesn't add any extra code paths other than the fixup itself.

PR #290, opened 2015-10-14
Bcf strerror

adamspargo:


mcshane:
Not really the right solution, as discussed in the samtools meeting a while ago. Closing this PR now.

PR #300, opened 2015-11-11
Improved memory freeing in bam_plp_destroy().

### LINKED ISSUE #299: Breaking out of a bam_plp_auto() loop leaks memory
It knows it too...

```
    while ((plp = bam_plp_auto(iter, &tid, &pos, &n_plp))) {
         //...
        if (h_iter) {
            if (pos < h_iter->beg)
                continue;
            if (pos >= h_iter->end)
                break; //htslib bug: causes memory leak in bam_plp_auto.
        }
        putchar(base);
    }
```

The break statement there causes this warning "***[bam_plp_destroy] memory leak: 1. Continue anyway***".  Removing the break (esides giving incorrect output in my code) avoids the warning.  It also doesn't occur if the break happens after having ended all in-flight sequences, eg in a zero coverage region.

If we know it leaks, we ought to fix it! It looks a bit amateurish even admitting we leak memory to the user.

jkbonfield:
Fixes #299, but to consider for 1.4.

Some comments.  The removal of dummy out of __bam_plp structure, or more specifically the need to mp_alloc it, decrements the pool size by one which coincidentally fixes samtools/samtools#126 too.  So that's how max depth is implemented!

This does mean the __bam_plp structure is now 8 bytes smaller, however the struct itself is defined in sam.c and although memory offsets change, it is internal.  bam_plp is an external typedef to a pointer to struct __bam_plp, so I think it is safe to say that we're not going to be caring about the sizeof the structure given it's only available in pointer form unless you jump through hoops.

(If it's an issue, we can add back a dummy pointer to keep struct size identical.)

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1584,7 +1584,7 @@ typedef khash_t(olap_hash) olap_hash_t;
 
 struct __bam_plp_t {
     mempool_t *mp;
-    lbnode_t *head, *tail, *dummy;
+    lbnode_t *head, *tail;
     int32_t tid, pos, max_tid, max_pos;
     int is_eof, max_plp, error, maxcnt;
     uint64_t id;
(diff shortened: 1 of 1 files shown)


jmarshall:
During our discussion in November, I was looking at refactoring this code too.  As I demonstrated in the comments on #299 (_Apologies, I misread `iter->head = iter->dummy->next` as prepending dummy to the list_), the gymnastics with `dummy` obfuscate the fact that all that is going on here is that nodes may be being spliced out of the list.  Looking at refactoring this loop led to df4a80e9c387bb1d84cdb47c02e44e77b0f57770, which I have rudely applied rather than this more conservative PR.

PR #308, opened 2015-12-16
Correct dst and ndst size bug in bcf_get_format_values (vcf.c).

dlaehnemann:
Make dst and ndst size adjustment in bcf_get_format_values also work, when the required size decreases from one parsed line to the next.
So far, dst buffer size and ndst were only upwards adjusted. If a buffer
size smaller than a previous one was needed, only the values at the
start of dst were updated, old values in later positions of dst were
carried over and ndst was kept at the larger value.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -3350,10 +3350,10 @@ int bcf_get_format_values(const bcf_hdr_t *hdr, bcf1_t *line, const char *tag, v
         return n;
     }
 
-    // Make sure the buffer is big enough
+    // Make sure the buffer has the right size
     int nsmpl = bcf_hdr_nsamples(hdr);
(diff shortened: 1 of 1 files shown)


pd3:
Hi, thank you for looking at this. However, it is not a bug! The`ndst` value is the size of the allocated array and the number of values actually written is returned by the function. This is to avoid excessive malloc calls. 

The documentation for `bcf_get_format_values()` in vcf.h says

     bcf_get_format_*() - same as bcf_get_info*() above

where the usage is explained in detail. If this is a matter of improving the documentation, we'll gladly accept pull requests which improve it.

dlaehnemann:
Thanks for the clarification. See pull request #309.

PR #312, opened 2016-01-14
Fixed a CRAM encoding assertion failure on repeated templates.

jkbonfield:
Passes all the tests and doesn't seem to change output in normal behaviour, but PR for review if anyone wishes prior to merge.

If we have a complex template where more than 2 reads exist and we
have valid pnext/tlen fields between each successive pairs, the code
was erroneously removing the statistics for the tlen(etc) values
multiple times instead of just once.  Test data included.


daviesrob:
Have checked this over and it looks fine.

PR #317, opened 2016-01-19
Add read error check to fai_load

### LINKED ISSUE #315: Failure to read from an fai file is not detected and results in huge volumes of mpileup stderr output
fai_read does not check ferror() and fai_load does not check fai_read success, so failures to read from the fai file result in an fai structure that is empty except for the bgzf and hash entries. 

For example, I recently ran into this problem with an mpileup. It ended up printing errors for every base of the format: 

```
[mplp_func] Skipping because 93561197 is outside of -2 [ref:10]
[mplp_func] Skipping because 93561197 is outside of -2 [ref:10]
[mplp_func] Skipping because 93561198 is outside of -2 [ref:10]
[mplp_func] Skipping because 93561202 is outside of -2 [ref:10]
```

I had this problem with the experimental `bcftools mpileup` but it appears that `samtools mpileup` would be effected as well. The error is printed by: https://github.com/samtools/samtools/blob/834bf7e9c9d4bbf7d60160138f728e7968373e3a/bam_plcmd.c#L248-L253

I attached to the process with `gdb` to track down this issue. 

The value of fai (up the stack in bam_mpileup) was:
```
#6  0x0000000000464895 in bam_mpileup (argc=15, argv=0x7ffd2d13ced0) at mpileup.c:945
(gdb) print *mplp.fai
$59 = {bgzf = 0x1b2f4d0, n = 0, m = 0, name = 0x0, hash = 0x1b2f4a0}
```

In this particular case, the failure was a transient read failure that would have been fine on retry (if I do another fai_load() it works just fine). htslib should at least check for read errors, and probably should retry depending on which error is returned.

mp15:
fai_load was invoking fai_read which did not check for ferror() on read. (Fixes #315)
Also convert == 0 to NULL for consistency.

### DIFF
--- a/faidx.c
+++ b/faidx.c
@@ -30,6 +30,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <stdlib.h>
 #include <stdio.h>
 #include <inttypes.h>
+#include <errno.h>
 
 #include "htslib/bgzf.h"
 #include "htslib/faidx.h"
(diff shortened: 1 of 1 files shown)


mp15:
Revised to separate out whitespace and NULL changes for future separate pull request.

jmarshall:
Thanks.  Rebased, tweaked, and merged as 601f7e525183c24482afe4141c8e20659d996bcf.

PR #320, opened 2016-01-20
Populating CRAM cache is now much less likely to have a race condition

jkbonfield:
On filesystems that don't support exclusive fopen, it was possible
totrigger a race condition causing the downloaded data to become
corrupt.  We now randomise the filenames to make this much less
likely.

The data is also md5sumed before writing.  (Logically we'd check after
writing, but if we've hit an NFS issue with multiple clients writing
to the same file then we're unlikely to detect that anyway as each
system will just read back what it wrote from its own local cache.)

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -2324,23 +2324,6 @@ static int cram_populate_ref(cram_fd *fd, int id, ref_entry *r) {
 	static int rand_init = 0;
 	FILE *fp;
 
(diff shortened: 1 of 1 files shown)


jmarshall:
Please rebase to current **develop**.  50% of this by weight appears to have been already applied in similar form in your d7ecf685f686541c6e2badd100594f2134206c92 on February 12th.

jkbonfield:
Rebase can't automatically handle this, but manually ditched the old code.

PR #332, opened 2016-02-05
Propagate vcf errors from synced reader. Fixes #318

### LINKED ISSUE #318: VCF parsing errors not propagated from the synced reader
When the reader encounters an error while parsing VCF or BCF (https://github.com/samtools/htslib/blob/develop/synced_bcf_reader.c#L457) it merely stops processing the stream but has no way of propagating the error.

Ideally the bcf_sr_next_line() should return a negative value on errors but this is backward incompatible, see the recommended usage https://github.com/samtools/htslib/blob/develop/htslib/synced_bcf_reader.h#L41.

Alternatively a new function could be added that checks for any errors encountered while processing after the loop is done.

mcshane:


synced_bcf_reader.c
```diff
@@ -75,7 +75,11 @@ char *bcf_sr_strerror(int errnum)
             return "no BGZF EOF marker; file may be truncated"; break;
         case no_memory:
             return "Out of memory"; break;
-        default: return "";
+        case vcf_parse_error:
+            return "VCF parse error"; break;
+        case bcf_read_error:
+            return "BCF read error"; break;
+        default: return "";
```
jmarshall: Trailing spaces added on the `default` line.  I guess with this rebasing this is otherwise ready to be merged?

mcshane:
Fixed up whitespace. Yes. Just cleaning up old PRs and noticed this needed a rebase after some of James and Rob's changes.

PR #335, opened 2016-02-08
Add prefix based on libdir in pkg-config

brainstorm:
For easy referencing from Makefiles, for example:

```
HTS_LIB=`pkg-config --variable=prefix htslib`
```

Instead of going for hardcoded, relative and manually pre-downloaded libraries:

https://github.com/ryanlayer/gqt/blob/master/src/Makefile#L3

@ryanlayer

### DIFF
--- a/htslib.pc.in
+++ b/htslib.pc.in
@@ -1,5 +1,6 @@
 includedir=@includedir@
 libdir=@libdir@
+prefix: ${libdir}
 
 Name: htslib
 Description: C library for high-throughput sequencing data formats


jmarshall:
Patch as it stands is not correct, as it should be `@prefix@`… if indeed there were a reason to export `prefix`.

If you `make htslib-uninstalled.pc` in htslib and add it to the pkg-config path in some way, then gqt's makefile would be able to use

```make
$(OBJ)/%.o: %.c 
    …`pkg-config --cflags htslib`…

$(BIN)/%: %.c $(OBJS)
    …`pkg-config --libs htslib`…
```

or so, in the usual way.

brainstorm:
Very much true indeed. Thanks! Closing WONTFIX.

PR #338, opened 2016-02-12
Adds HTS_OPT_BLOCK_SIZE support for SAM/BAM/CRAM.

### LINKED ISSUE #216: Add hts_opt for filesystem block size
hfile_init takes a blocksize which for filesystems is returned via struct stat.st_blksize.   Unfortunately some filesystems are just broken.  E.g. lustre sometimes recommends 4Kb as its blocksize, which is (was?) totally wrong as it prefers around 1Mb.  Maybe this has changed, but it warrants testing.

We could add an option to use via the --input-fmt-option and --output-fmt-option parameters that allows manual control over the block size.  This would also allow for a way to override the hacked 32768 limit in hfile_init(), put in place to prevent large scale merging and mpileup from consuming too much memory.

Whether it really makes any difference I don't know.

... edit - tested! ...

32K block size on lustre: 49.6sec to read ~3Gb via samtools.
2Mb block size: 47.7 sec. 

Minimal, but still a difference.

jkbonfield:
Note: this exposes new API bgzf_hfile(), cram_hfile() and hts_set_blksize()
functions.

They could all be internal, but I don't know if there is an official
htslib way of having inter-file function exports that don't
externalise to the library.  Just comments to say as much?

It is unknown whether hts_set_blksize() works on vcf/bcf (there is no
test_view equivalent so I didn't bother trying), but there is also no
equivalent -I / -O options for vcf/bcf either so it doesn't matter for
now.

Fixes #216

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -296,7 +296,7 @@ hfile_gcs.o hfile_gcs.pico: hfile_gcs.c config.h $(htslib_hts_h) $(htslib_kstrin
 hfile_libcurl.o hfile_libcurl.pico: hfile_libcurl.c config.h $(hfile_internal_h) $(htslib_hts_h) $(htslib_kstring_h)
 hfile_net.o hfile_net.pico: hfile_net.c config.h $(hfile_internal_h) $(htslib_knetfile_h)
 hfile_s3.o hfile_s3.pico: hfile_s3.c config.h $(hts_internal_h) $(hfile_internal_h) $(htslib_hts_h) $(htslib_kstring_h)
-hts.o hts.pico: hts.c config.h $(htslib_hts_h) $(htslib_bgzf_h) $(cram_h) $(htslib_hfile_h) version.h $(hts_internal_h) $(htslib_khash_h) $(htslib_kseq_h) $(htslib_ksort_h)
+hts.o hts.pico: hts.c config.h $(htslib_hts_h) $(htslib_bgzf_h) $(cram_h) $(hfile_internal_h) $(htslib_hfile_h) version.h $(hts_internal_h) $(htslib_khash_h) $(htslib_kseq_h) $(htslib_ksort_h)
 vcf.o vcf.pico: vcf.c config.h $(htslib_vcf_h) $(htslib_bgzf_h) $(htslib_tbx_h) $(htslib_hfile_h) $(hts_internal_h) $(htslib_khash_str2int_h) $(htslib_kstring_h) $(htslib_khash_h) $(htslib_kseq_h) $(htslib_hts_endian_h)
 sam.o sam.pico: sam.c config.h $(htslib_sam_h) $(htslib_bgzf_h) $(cram_h) $(hts_internal_h) $(htslib_hfile_h) $(htslib_khash_h) $(htslib_kseq_h) $(htslib_kstring_h) $(htslib_hts_endian_h)
 tbx.o tbx.pico: tbx.c config.h $(htslib_tbx_h) $(htslib_bgzf_h) $(hts_internal_h) $(htslib_khash_h)
--- a/bgzf.c
+++ b/bgzf.c
@@ -1885,3 +1885,8 @@ long bgzf_utell(BGZF *fp)
 {
     return fp->uncompressed_address;    // currently maintained only when reading
 }
+
+/* prototype is in hfile_internal.h */
+struct hFILE *bgzf_hfile(struct BGZF *fp) {
+    return fp->fp;
+}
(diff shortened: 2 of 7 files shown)


jkbonfield:
As an example of how to use it:

```
strace ./test/test_view -i block_size=123456 -o block_size=234567 /tmp/a.bam 2>&1 > /dev/null | egrep 'read|write'|less
...
read(3, "N\256\365\217\305/]Y%`\365\3625O\7\353\365\331\335\346\214]K\265\300\273\221\372\373`\254\334"..., 123456) = 123456
write(1, "FHFEGFHHEHGIGIHGGCIFHGGGCIHFHFFF"..., 234567) = 234567
read(3, "a\252\3079\367\206\253\247e\20:\344^*\250\312\260\206\203*g\221D\36\21\301\324\4d\325:)"..., 123456) = 123456
write(1, "A:U\nHS25_09827:2:2312:3457:73925"..., 234567) = 234567
...
```
We don't have any explicit tests to test these options (or any other except cram version).  Other than "does it fail" it's non-trivial to test the options affect the I/O anyway.

jkbonfield:
Rebased and fixed humorous new "switch-else" construct :-)

Pure read tests on lustre (abridged):

```
# Default 32k block size
$ perf stat ./samtools view  -f 0xffff ~/lustre/9827_2#49.bam; # best of 3 times
      78.402190173 seconds time elapsed

$ perf stat ./samtools view -@ 8 -f 0xffff ~/lustre/9827_2#49.bam
       9.731309733 seconds time elapsed

# Manually specified 1Mb block size
$ perf stat ./samtools view --input-fmt-option block_size=1000000 -f 0xffff ~/lustre/9827_2#49.bam
      78.524177464 seconds time elapsed

$ perf stat ./samtools view -@8 --input-fmt-option block_size=1000000 -f 0xffff ~/lustre/9827_2#49.bam
       9.696564556 seconds time elapsed
```

So apparently little difference now - perhaps our newer lustre versions fixed some of the bad performance with small reads.   Similarly rewrite loops showed very little change.  Still, it's a tiny change and may benefit other file systems, such as s3, irods, etc. (Untested.)

Note, once added we will need to commit to samtools to edit samtools.1 as the only relevant documentation is over there.

daviesrob:
I've taken the liberty of rebasing this, and added a commit on top to remove hfile_set_blksize() from the public API and make hts_hfile() static.  This would leave bgzf_hfile() as the only addition to the public API, which is mostly harmless.  Keeping the interfaces private gives us more room to change the implementation should we want to.  We can easily make them public in a future commit.

If we decide to add this, we should squash the two commits together before merging.  Comments, anyone? Ideally, I'd like to merge this before the end of the week.

bioinformed:
It would be very helpful for `pysam` to have `hts_hfile` as part of the public API.  If not, we're going to have to implement it ourself using essentially identical code.

daviesrob:
Yes, hts_hfile() is probably more useful than bgzf_hfile() and cram_hfile().  I'm just a bit wary about what people may do with the hFILE that's sitting behind a BGZF or CRAM file.  What were you planning to do with them?

jmarshall:
> They could all be internal, but I don't know if there is an official htslib way of having inter-file function exports that don't externalise to the library.

Since then there is a way: add them to _hts_internal.h_ rather than a public header.

There should be a high bar before adding functions to access privates like `hts_file()`, as they really constrain future htslib changes.

jmarshall:
@bioinformed: That's the canonical example of why you DO NOT want a function like `hts_hfile()`.  The high-level `htsFile` does its own buffering, and peeking at its privates and seeking them from under it produces incorrect results as you've subverted any hope of the `htsFile` adjusting its buffered data.

The correct answer to that use case is for HTSlib to finally get around to implementing `hts_seek` and `hts_tell` to seek correctly on `htsFiles`, as discussed in [this samtools-devel thread](https://sourceforge.net/p/samtools/mailman/message/34432974/).

My expectation would be that for almost any use case for an `hts_hfile()` or similar function, the appropriate answer would be to provide direct `htsFile *`-based access to the functionality instead.

bioinformed:
@jmarshall: Unfortunately, we're already doing the wrong thing in pysam.  e.g.
```python
    def seek(self, uint64_t offset):
        """move file pointer to position *offset*, see :meth:`pysam.HTSFile.tell`."""
        if not self.is_open:
            raise ValueError('I/O operation on closed file')
        if self.is_stream:
            raise OSError('seek not available in streams')

        cdef int64_t ret
        if self.htsfile.format.compression == bgzf:
            with nogil:
                ret = bgzf_seek(hts_get_bgzfp(self.htsfile), offset, SEEK_SET)
        elif self.htsfile.format.compression == no_compression:
            with nogil:
                ret = hts_useek(self.htsfile, <int>offset, SEEK_SET)
        else:
            raise NotImplementedError("seek not implemented in files compressed by method {}".format(
                self.htsfile.format.compression))
        return ret```

PR #339, opened 2016-02-17
Create API to check EOF on all htsFile that support EOF block

mp15:
Create EOF checking support for CRAM.
Create htsFile API call to check EOF on BAM and CRAM

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -4575,3 +4575,55 @@ int cram_set_voption(cram_fd *fd, enum hts_fmt_option opt, va_list args) {
 
     return 0;
 }
+
+#define TEMPLATE_2_1_LEN 30
+#define TEMPLATE_3_LEN 38
+int cram_check_EOF(cram_fd *fd)
+{
+    // Byte 9 in these templates is & with 0x0f to resolve differences
+    // between ITF-8 interpretations between early Java and C
+    // implementations of CRAM
+    const uint8_t TEMPLATE_2_1[TEMPLATE_2_1_LEN] = {
+        0x0b, 0x00, 0x00, 0x00, 0xff, 0xff, 0xff, 0xff, 0x0f, 0xe0,
+        0x45, 0x4f, 0x46, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00,
+        0x01, 0x00, 0x06, 0x06, 0x01, 0x00, 0x01, 0x00, 0x01, 0x00
+    };
+    const uint8_t TEMPLATE_3[TEMPLATE_3_LEN] = {
+        0x0f, 0x00, 0x00, 0x00, 0xff, 0xff, 0xff, 0xff, 0x0f, 0xe0,
+        0x45, 0x4f, 0x46, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x05,
+        0xbd, 0xd9, 0x4f, 0x00, 0x01, 0x00, 0x06, 0x06, 0x01, 0x00,
+        0x01, 0x00, 0x01, 0x00, 0xee, 0x63, 0x01, 0x4b
+    };
+    uint8_t major = CRAM_MAJOR_VERS(fd->version);
+    uint8_t minor = CRAM_MINOR_VERS(fd->version);
+
+    const uint8_t *template;
+    ssize_t template_len;
+    if ((major < 2) ||
+        (major == 2 && minor == 0)) {
+        return 3; // No EOF support in cram versions less than 2.1
+    } else if (major == 2 && minor == 1) {
+        template = TEMPLATE_2_1;
+        template_len = TEMPLATE_2_1_LEN;
+    } else {
+        template = TEMPLATE_3;
+        template_len = TEMPLATE_3_LEN;
+    }
+    
+    uint8_t* buf = (uint8_t*) malloc(sizeof(uint8_t)*template_len);
+    off_t offset = htell(fd->fp);
+    if (hseek(fd->fp, -template_len, SEEK_END) < 0) {
+        if (errno == ESPIPE) {
+            hclearerr(fd->fp);
+            return 2;
+        }
+        else {
+            return -1;
+        }
+    }
+    if (hread(fd->fp, buf, template_len) != template_len) return -1;
+    if (hseek(fd->fp, offset, SEEK_SET) < 0) return -1;
+    buf[8] &= 0x0f;
+    int result = (memcmp(template, buf, template_len) == 0)? 1 : 0;
+    return result;
+}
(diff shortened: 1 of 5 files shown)


mp15:
Tagging @jkbonfield for review of CRAM code

cram/cram_io.c
```diff
@@ -4575,3 +4575,27 @@ int cram_set_voption(cram_fd *fd, enum hts_fmt_option opt, va_list args) {
 
     return 0;
 }
+
+#define TEMPLATE_LEN 38
+int cram_check_EOF(cram_fd *fd)
+{
+    uint8_t buf[TEMPLATE_LEN];
+    const uint8_t template[TEMPLATE_LEN] = {
+        0x0f, 0x00, 0x00, 0x00, 0xff, 0xff, 0xff, 0xff, 0x0f, 0xe0,
```
jkbonfield: The data here is length (0x0f 00 00 00) and ref_seq_id (-1 == 0xff ff ff ff ff 0f).
The last 4 bits of that ref_seq_id are undefined due to the wording in the ITF8 description.  Traditionally this meant 0s for my code and 1s in the Java code, leading to 0x0f vs 0xff issues.

It looks like we now agree on this interpretation, but for a while it meant EOF blocks were being generated in two different ways.

The safest option is to AND off those 4 bits first before doing the check.

jkbonfield:
Another thought, EOF blocks existed in version 2.1 too.  The specification stated:

```
When compiled together the EOF marker is exactly 30 bytes long and in hex representation 
is:

0b 00 00 00 ff ff ff ff ff e0 45 4f 46 00 00 00 00 01 00  00 01 00 06 06 01 00 
01 00 01 00
```

Although my implementation had:

```
            if (30 != hwrite(fd->fp,
                             "\x0b\x00\x00\x00\xff\xff\xff\xff"
                             "\x0f\xe0\x45\x4f\x46\x00\x00\x00"
                             "\x00\x01\x00\x00\x01\x00\x06\x06"
                             "\x01\x00\x01\x00\x01\x00", 30))
                return -1;
```

(Note the 0x0f vs 0xff again.)

For the curious, the way this is specified is a container with zero records, ref_id -1 (unmapped) and a nonsensical start coordinate into the unmapped reference of 0x454f46 (EOF in ASCII).  Hence the C cram code to check for EOF is actually just the above:

```
    fd->empty_container =
        (c->num_records == 0 &&
         c->ref_seq_id == -1 &&
         c->ref_seq_start == 0x454f46 /* EOF */) ? 1 : 0;
```

It's a bit of a pain to check in an external block of code however as it requires too many gubbins to build the cram_contrainer structure.  Checking byte stream is saner, barring the undefined bits.

jkbonfield:
The rest of it looks fine and I agree with the sentiment of creating a single exported hts_check_EOF function.

For ease of testing, I suggest something like:

```
./samtools view -t test/dat/mpileup.ref.fa -O cram,version=3.0 -o _cram30.cram test/dat/mpileup.1.sam
./samtools view -t test/dat/mpileup.ref.fa -O cram,version=2.1 -o _cram21.cram test/dat/mpileup.1.sam
head -c -30 _cram21.cram > _cram21_noeof.cram
head -c -38 _cram30.cram > _cram30_noeof.cram
```

mcshane:
Any reason it should just be for BAM/CRAM? If you check for compression=bgzf rather then type=bam, then the `bgzf_check_EOF` call would also work fro VCF/BCF?

jkbonfield:
Good point, but I'm a bit unsure what the format codes are.  Sometimes I find it's "binary" rather than some named format, but maybe that's when creating new files.  What does format hold for vcf.gz?

mcshane:
I was thinking this:
```
int hts_check_EOF(htsFile *fp)
{
    if (fp->compression == bgzf)
            return bgzf_check_EOF(fp->fp.bgzf);
   else if (fp->format == cram)
            return cram_check_EOF(fp->fp.cram);
    else
            return 3;
}
```

jmarshall:
Basic idea looks good, though perhaps this is not one to go into the 1.3.1 bug fix release.  Some further comments:

1. Perhaps `cram_check_EOF()` would like to be a public function in _htslib/cram.h_?

2. Please format the `hts_check_EOF()` doxygen doc more like that of `bgzf_check_EOF()`.  At the moment, it's too abbreviated to be clear.  Is this only for files opened for reading, similarly to the bgzf one?

3. Definitely trigger on `fp->compression == bgzf` as Shane suggests.

4. Grepping for `text_format`/`binary_format` shows that these exist because after `hts_open("foo", "wb")` the code doesn't yet know whether it's going to be `bam` or `bcf`; and similarly for `"w"` and `sam` vs. `vcf`.  `sam_hdr_write()` fills in the exact format now that it's known; `bcf_hdr_write()` perhaps currently doesn't…

cram/cram_io.c
```diff
@@ -4575,3 +4575,27 @@ int cram_set_voption(cram_fd *fd, enum hts_fmt_option opt, va_list args) {
...
+        (CRAM_MAJOR_VERS(fd->version) == 2 && CRAM_MINOR_VERS(fd->version) == 0)) {
+        return 3; // No EOF support in cram versions less than 2.1
+    }
+    off_t offset = htell(fd->fp);
+    if (hseek(fd->fp, -TEMPLATE_LEN, SEEK_END) < 0) {
+        if (errno == ESPIPE) { hclearerr(fd->fp); return 2; }
+        else return -1;
+    }
+    if ( hread(fd->fp, buf, TEMPLATE_LEN) != TEMPLATE_LEN ) return -1;
+    if ( hseek(fd->fp, offset, SEEK_SET) < 0 ) return -1;
```
jmarshall: This `( foo(…) )` parenthesis style is not used in nearby code.  So don't start :smile:

cram/cram_io.c
```diff
@@ -4575,3 +4575,27 @@ int cram_set_voption(cram_fd *fd, enum hts_fmt_option opt, va_list args) {
 
     return 0;
 }
+
+#define TEMPLATE_LEN 38
+int cram_check_EOF(cram_fd *fd)
+{
+    uint8_t buf[TEMPLATE_LEN];
```
jmarshall: `uint8_t` is only used in this code for interfacing with the lzma library.  Just use `char`, or `unsigned char` if it makes you feel better.

cram/cram_io.c
mp15: Yeah I stole this from the BGZF version. Could you fix it there too?

cram/cram_io.c
mp15: This also came from BGZF implementation. I guess it depends on how likely it is that someone compiles on a DSP where CHAR_BIT > 8 (<8 is ruled out in C90). How much does this bug you will change if lots?

mp15:
Changes made.

jmarshall:
Rebased, tweaked (fix `cram_check_EOF` memory leak; tidy up `hts_check_EOF` and its half-written comment; for data read by `hread()`, _cram_io.c_ uses `unsigned char`, which is preferable to `uint8_t` for this purpose even on 9-bit DSP straw men), and merged as 6d70a0e829062470dd30e5034d201e682de60ca5.

PR #340, opened 2016-02-18
Support for quoted contig names

mcshane:
Support for quoted contig names, such as `##contig=<ID="CrazyContigName,1">`
- fixes a parsing problem when comma in the contig name (closes samtools/bcftools#266)
- when injecting contigs in the header from the index, use quoted contig IDs
- some bcftools test output files will need to be updated if this is pulled in
- there are places in at least bcftools/vcfconvert.c and samtools/bam_plcmd.c
  where contig header lines are created. These should also be changed to have quoted IDs
- still and issue is what happens if there is a `"` in the contig name

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -289,6 +289,7 @@ bcf_hrec_t *bcf_hdr_parse_line(const bcf_hdr_t *h, const char *line, int *len)
     // structured line, e.g.
     // ##INFO=<ID=PV1,Number=1,Type=Float,Description="P-value for baseQ bias">
     // ##PEDIGREE=<Name_0=G0-ID,Name_1=G1-ID,Name_3=GN-ID>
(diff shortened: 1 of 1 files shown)


jmarshall:
Quoting these contig names is one approach to handling weird contig names with commas, etc.  Percent-encoding might be another.  Does this need an update to the VCF spec mentioning that these might be quoted?  I couldn't find any clear hints in the 4.3 spec as to whether either quoting or percent-encoding would be applicable to CONTIG IDs.

ldiao:
Any workaround for contig names with commas at the moment? Looking at hundreds of samples that have been aligned to a large reference db unfortunately with many commas in contig names. Would prefer not to have to start over, but will if no choice...

daviesrob:
Need to check htsjdk compatibility.  Possible spec update too...?

PR #341, opened 2016-02-24
Wrap load_hfile_plugins() in a critical section

jmarshall:
Finally implementing this TODO, as this has caused trouble out in the real world.

Re `hfile_exit()`, seems like all bets are off for `atexit` functions in multithreaded programs (cf this [libguestfs bug report](https://bugzilla.redhat.com/show_bug.cgi?id=790837)), but if other threads are still opening files when someone calls `exit`… well, what can you do…

@daviesrob @jkbonfield **other-pthreads-experts**: does this look sufficient / about right?

### DIFF
--- a/hfile.c
+++ b/hfile.c
@@ -1,6 +1,6 @@
 /*  hfile.c -- buffered low-level input/output streams.
 
-    Copyright (C) 2013-2015 Genome Research Ltd.
+    Copyright (C) 2013-2016 Genome Research Ltd.
 
     Author: John Marshall <jm18@sanger.ac.uk>
 
@@ -28,6 +28,8 @@ DEALINGS IN THE SOFTWARE.  */
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
+
+#include <pthread.h>
 
 #include "htslib/hfile.h"
 #include "hfile_internal.h"
@@ -589,9 +591,12 @@ struct hFILE_plugin_list {
 };
 
 static struct hFILE_plugin_list *plugins = NULL;
+static pthread_mutex_t plugins_lock = PTHREAD_MUTEX_INITIALIZER;
 
 static void hfile_exit()
 {
+    pthread_mutex_lock(&plugins_lock);
+
     kh_destroy(scheme_string, schemes);
 
     while (plugins != NULL) {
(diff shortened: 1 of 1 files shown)


jkbonfield:
Provided hfile_exit() can only ever be called once (likely from the main thread) then it looks OK.  However if it can't be called more than once then what is the need for the mutex check there?  To be sure that we're not trying to open our first file (triggering the plugin loading) in one thread while simultaneously exiting in another?  I guess it *could* happen in you're really dedicated to the cause!

There is also the pthread_once() function (see http://linux.die.net/man/3/pthread_once) which sounds perfect for this task, but I haven't used it before.

jmarshall:
Yes, two lines of code to give some hope in the unlikely event one thread's opening a file and another's calling `exit()` seemed like no great overhead.  But everything's already doomed in that scenario anyway.

Thanks for the review!

daviesrob:
The bit in find_scheme_handler looks OK, as long as the schemes hash is constant after load_hfile_plugins has returned.

hfile_exit() is a bit of a problem.  atexit handlers can be called when more than one thread is still running.  In this case it would cause use-after-free bugs if any of the remaining threads used one of the interfaces that had been destroyed.

As far as I can tell to be safe:

- Threads should never call exit(), they should use pthread_exit() instead.

- The main thread should only call exit() when no child threads are running.

But as htslib can't guarantee either of these as it may be used with external code that does not play nicely.

I think the best solution may be to move the clean-up code into functions inside the plug-in shared libraries marked with `__attribute__((destructor))`, and let the OS unload the shared libraries instead of trying to dlclose them ourselves.  We'd have to check, but I think such functions would only be called when the last thread has finished, which would be much more in line with what we want.  Unfortunately they're not terribly portable, but I don't think there is a good portable solution to this problem.

An alternative would be to ensure hfile_exit() can't run if any of the htslib functions that may call into the plug-ins are in use.  That would need reference counting and conditionals to get the right behaviour, so would likely be a bit complicated.  We'd basically want to allow functions like hread2 to run in parallel, but lock out hfile_exit until they'd finished.  hfile_exit would also need to set a flag to tell hread2 etc. that we're shutting down so they should just return a failure.  Between those two it should be possible to make the whole thing safe.

jmarshall:
Especially the latter sounds like a lot of fragile complexity to improve an unlikely the-ship-is-already-going-down corner case.  I'd prefer just to document your two bullet points as constraints that external code has to satisfy to use htslib with 100% safety.

Anyway, thanks for the analysis!  I propose to cross this bridge when/if we come to it.

PR #342, opened 2016-02-26
Avoid unnecessary warning on unknown sort order

jenniferliddle:
A sort order of "unknown" is not an unknown sort order, it's a perfectly reasonable sort order defined in the BAM specification. This change will avoid unnecessary warnings in our log files when using samtools.


jmarshall:
Thanks.  I would be tempted to just nuke the warning (we're not Picard!), but there's several such in the cram code so nevermind.

PR #343, opened 2016-02-26
Add errmod_cal(), probaln_glocal(), sam_cap_mapq(), and sam_prob_realn() to HTSlib API

jmarshall:
These functions were previously in samtools; add them to htslib as they will soon be used by bcftools as well.

* Declarations added to existing _hts.h_ & _sam.h_ headers rather than adding 2 or 3 tiny new headers

* `kpa_glocal()` etc and `bam_cap_mapQ()`/`bam_prob_realn_core()` renamed for the  public API

See individual commit message for details.


pd3:
+1 from me

PR #349, opened 2016-03-09
If no CC value set then use GCC, otherwise take the value from the env

JervenBolleman:
Currently the Makefile sets CC to GCC even if a CC is set by the user.
This is now changed to set GCC if no CC is present in the user environment.
I think this was the intended behavior as the .travis.yml file wants to test compilation using both gcc and clang.

The same change is there for the AR and RANLIB setting.


jmarshall:
No.  There is no need for our makefile to use GNU Make-specific syntax here.

See samtools/samtools#69 and samtools/samtools#533 for supported alternatives.  _.travis.yml_ uses `make -e` and works as intended.

PR #350, opened 2016-03-09
Added bam_aux_update()

jenniferliddle:
I needed a 'change in place' function rather than a 'delete and append' operation to help with regression testing.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1282,11 +1282,12 @@ int bam_aux_del(bam1_t *b, uint8_t *s)
     return 0;
 }
 
-int bam_aux_update(bam1_t *b, const char tag[2], int len, uint8_t *data)
+int bam_aux_update_str(bam1_t *b, const char tag[2], int len, uint8_t *data)
 {
     uint8_t *s = bam_aux_get(b,tag);
     if (!s) return -1;
     char type = *s;
+    if (type != 'Z') { fprintf(stderr,"bam_aux_update_str() called for type '%c' instead of 'Z'\n", type); abort(); }
     bam_aux_del(b,s);
     s -= 2;
     int l_aux = bam_get_l_aux(b);
@@ -1302,7 +1303,7 @@ int bam_aux_update(bam1_t *b, const char tag[2], int len, uint8_t *data)
     s[0] = tag[0];
     s[1] = tag[1];
     s[2] = type;
-    memcpy(s+3,data,len);
+    memmove(s+3,data,len);
     return 0;
 }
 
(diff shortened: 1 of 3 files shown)


jenniferliddle:
Hmm. You're right Joshua. Changed to memmove() just to play safe.

sam.c
```diff
@@ -1282,6 +1282,31 @@ int bam_aux_del(bam1_t *b, uint8_t *s)
...
+    bam_aux_del(b,s);
+    s -= 2;
+    int l_aux = bam_get_l_aux(b);
+    uint8_t *aux = bam_get_aux(b);
+
+    b->l_data += 3 + len;
+    if (b->m_data < b->l_data) {
+        b->m_data = b->l_data;
+        kroundup32(b->m_data);
+        b->data = (uint8_t*)realloc(b->data, b->m_data);
```
jkbonfield: This should check if b->data is NULL and fail if so.

*Ideally* it's be something like:

```
uint8_t *tmp = (uint8_t *)realloc(b->data, b->m_data);
if (!tmp) return -1;
b->data = tmp;
```

Otherwise on failure to realloc it's trashed your existing value.

jkbonfield:
Essentially your bam_aux_update_str is written as a combination of delete and insert.  Maybe it's worth adding a bam_aux_insert function and making that explicit, which permits any other type of data to be added.  Bam_aux_append then is simply a special case of bam_aux_insert at the end.

sam.c
jenniferliddle: bam_aux_append doesn't check, which is why bam_aux_insert doesn't check. I was following precedent!

jenniferliddle:
A bam_aux_insert function sounds like not a bad idea. Perhaps that's something we should consider for version 1.5

jmarshall:
Now that it's explicitly `bam_aux_update_str` with **_str**, should it take `char *data` rather than `uint8_t *data`?  Currently we have

```c
#define NEW_HELLO "Yo, dude"
…
bam_aux_update_str(aln,"XZ",strlen(NEW_HELLO)+1,NEW_HELLO);
// test/sam.c:115:57: warning: passing 'char [9]' to parameter of type 'uint8_t *' (aka 'unsigned char *') converts between pointers to integer types with different sign [-Wpointer-sign]
```

(And actually for this and `bam_aux_append` it should really be `const whatever *data`.)

jmarshall:
Thanks; merged as bb03b0287bc587c3cbdc399f49f0498eef86b44a.

PR #357, opened 2016-03-18
 Fixed memory allocation issue when decoding CRAM.

jkbonfield:
This bug has existed a long time, but is now more likely to triggered by 6d2810c due to making it more likely to overrun the buffer.

The issue is caused by zero length sequences ("*") with a CIGAR alignment.  A full bullet-proof fix will require more work to use the ...decode_block() codec variants, but this fixes the common case of length 0.


jmarshall:
Thanks — merged with some additions.  Did you notice this warning in the new tests?

    ./test_view -i reference=xx.fa xx#large_aux_java.cram > xx#large_aux_java.tmp.sam
    WARNING: Header @SQ length mismatch for ref xx, 30 vs 20
    ./compare_sam.pl -nomd xx#large_aux.sam xx#large_aux_java.tmp.sam
    .. ok

jkbonfield:
Actually no I didn't spot that - it was lost in the mass of output.  It's caused by comparing to an older variant of the test ref seqs, although fortunately harmless.

Thanks.

PR #360, opened 2016-03-20
vcf.h:vcf_parse() add documentation

noporpoise:
Add a comment to vcf_parse() to document that the input line must not end with a newline character.


jmarshall:
Thanks.  We may eventually make `vcf_parse()` a bit more forgiving, but this text reflects the current reality.

PR #361, opened 2016-03-23
CRAM: aux tags in own blocks; bases_per_slice param; sync with io_lib

jkbonfield:
A mix of changes (sorry, it's non trivial to untangle) pulled over from io_lib - and some in reverse where io_lib was outdated.

The primary changes here are putting each auxiliary tag into its own block and added a bases_per_slice hts option to make PacBio and ONT data more efficient to deal with.  A few minor bug fixes were picked up on the way, although not all were triggerable here.  There are some minor code reshuffles too, but these are kept to a minimum level, to help future diffing between the two code bases.


jkbonfield:
I know some of this isn't strictly cram "split tags", but it'll just create a tangled mess of conflicts if I add yet more branches and ultimately I feel they all need merging once 1.4 work starts.  So I think if it's not urgent and needs to go in 1.3.1 then adding here is OK.

Comments on that philosophy?

jmarshall:
After discussion with @daviesrob, who would like to base some rANS fixes on top of this, I've merged all of this apart from the last commit, 7c5489106bab2f30e0ef9a17f704b47b70e31189/CRC32c.  It might be best to re-present that as a new PR…

jkbonfield:
Closing in favour of new CRC issue (to be figured out) in #467

Issue #344, opened 2016-03-02
Assorted AWS S3 issues [tracking]


jmarshall:
This is a tracking bug collecting together assorted issues with _hfile_libcurl.c_'s S3 support.  Several of these have been reported as asides on otherwise-closed PRs etc, so are easily lost.

jmarshall:
I see botocore also falls back to contacting the IAM well-known endpoint.  HTSlib could also do this, although I would have thought there would be some common script for copying these into environment variables so that such code was not multiplied across clients and SDKs…? e.g. `eval` something like

```sh
baseurl='http://169.254.169.254/latest/meta-data/iam/security-credentials'
role=`curl $baseurl`
curl $baseurl/$role | jq -r "\"export AWS_ACCESS_KEY_ID='\(.AccessKeyId)' AWS_SECRET_ACCESS_KEY='\(.SecretAccessKey)' AWS_SESSION_TOKEN='\(.Token)'\""
```

HTSlib-based tools (i.e. samtools) are used on EC2 and especially outside EC2.  If you're not on EC2, contacting random link-local IP addresses and asking them for credentials would be rather unfortunate (to put it mildly), so I'd really rather not do this by default.  I would have thought there would be a convention of only doing this when $AWS_PROFILE is set to a special `:IAM` value or something like that?

obenshaindw:
Any script that reads the instance metadata for security credentials and sets environment variables for those would need to be run on a schedule or immediately prior to running samtools (at least every hour according to the [previously linked page AWS documetation](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html) because credentials are automatically rotated.

The advantage here is not having to leave any credentials on the instance, and if using an application that reads directly from instance metadata, not having to set anything environment-specific.

I agree that this would seem like a special case for samtools (i.e., a piece of code that is environment-specific).  It makes sense for boto to have this functionality since it is an SDK for Amazon Web Services.  As you say, if one is using samtools on EC2 they could write a script to read from the instance metadata endpoint and place the security credentials in a location where samtools will find them.

delagoya:
The [AWS C++ SDK](https://aws.amazon.com/sdk-for-cpp/) instead of the current customer request signing code would solve this. But that is a big project and would likely be better done by pulling out the `htsfile_s3.c` plugin from the root source tree and  into it's own contrib project.   
  
[...]

Issue #346, opened 2016-03-02
Support Requester Pays S3 buckets


jmarshall:
[Requester Pays buckets](http://docs.aws.amazon.com/AmazonS3/latest/dev/RequesterPaysBuckets.html) need an extra `X-Amz-Request-Payer: requester` header.

Clearly it would not be appropriate for htslib/samtools to set it all the time (as it represents explicit acknowledgement from the **user** that they will be charged).  So we could set it if some flag was present in the URL or perhaps via an extra config file key on the profile used. @DonFreed or anyone else: are you aware of any existing practice in this area?

[As [reported at biostars.org](https://www.biostars.org/p/147772/#176690).]

DonFreed:
Sorry, I have no experience with requester pays buckets. It seems like a flag would work. A config key might be a little more dangerous as users could potentially make large numbers of requests while forgetting about the profile setting.

jmarshall:
Some more details about my ideas for how to do this, and the sort of existing practice I'm looking for…

Looking at their source code, it seems that e.g. [Tim Kay's aws](http://timkay.com/aws/) has a `--requester` command-line option and [s3cmd](http://s3tools.org/s3cmd) has `--requester-pays`   similarly.

However because samtools, htslib, and htslib's libcurl plugin are all separate, it would be inconvenient to have this as a samtools/bcftools/other-client-software command-line option that would need to be implemented separately for each client program and communicated to htslib and thence to the plugin.  What would be more workable for us would be to encode this flag into the URL (which is of course already the main parameter in the interface to the plugin) and/or into the S3 configuration files that the plugin consults.

There's a few places in an S3 pseudo-URL where such a flag could be bolted on:

    s3+requesterpays://bucket/foo/bar.bam
    s3://bucket/foo/bar.bam?request_payer=requester
    s3://requesterpays@bucket/foo/bar.bam

We already have schemes like `s3+http://bucket…` but using this for non-transport flags too seems unfortunate.  Using the `?query` or `#fragment` parts seems like usurping them from their real uses.  We're already using the authority authentication part for `[ID:SECRET:TOKEN@]BUCKET`, but extending this to encode this flag too seems the most workable.  At present we parse this as either `[PROFILE@]BUCKET` or `[ID:SECRET[:TOKEN]@]BUCKET`; this could be extended with `:requesterpays` or so, or even better it could piggy-back off our profile handling.

[...]

PR #362, opened 2016-03-28
IAM role support

jpiper:
I noticed #344 and thought I would get the ball rolling. Here's an implementation that will fall back to using an IAM role when no other credentials are found.


jmarshall:
Apologies for not getting back to you on this sooner after getting the 1.3.1 release out the door.

As discussed on #344, if we do include this functionality, it needs to be something that is actively enabled by the user rather than something that might take non-EC2 users by surprise.  Also it would be preferable to avoid burning Amazon IP numbers and paths (`http://169.254.169.254/latest/meta-data/iam/security-credentials/`) into htslib.

I've dusted off and committed to **develop** (see 1cc7d126189ae3d4d5b291af76995545e1322c21) a `curl_kput()` function that does the same thing as your `curl_as_string()`, but writes to a _kstring_t_ as those already have the code to manage a memory buffer.

I would accept a pull request that looked for a new key **iam_role_url** (we can workshop the key name further…) in the _.aws/credentials_ file:

```
iam_role_url = http://169.254.169.254/latest/meta-data/iam/security-credentials/
```

and, in the absence of other credentials, used the presence of that to trigger code to `curl_kput`-fetch those two URLs and extract `AccessKeyId`/`SecretAccessKey`/`Token` from the resulting JSON.

I'd rather the JSON was parsed using string functions rather than regexps (htslib currently doesn't `#include <regex.h>` and would like to stay that way).  But actually htslib may eventually need basic JSON reading capabilities for some other things too, so don't worry about that too much: I may be adding something akin to `parse_ini`'s interface for simple JSON reading needs, which would suffice for parsing this.

jpiper:
To me, rather than conflating the `.aws/credentials` file, using IAM to retrieve the security credentials could be handled via an env var (likewise, we could bikeshed on the name, but something like `HTSLIB_IAM_FALLBACK` or something ).

With regards to burning in this IP address - this shouldn't be a problem as this URL has always been used for AWS metadata, and the `169.254`* IP block is reserved for link local [see RFC3927](http://tools.ietf.org/html/rfc3927) and [here](https://en.wikipedia.org/wiki/Link-local_address) so will never traverse past the local network.

I completely agree on the regex parsing though - this was a very dirty hack and a better solution could definitely be found.

jmarshall:
People in coffee shops do not have the liberty of trusting their local network.

jpiper:
True - I guess my point here is that given that the user will always have to manually enable IAM support, is there a difference in potential security vulnerabilities between using a simple `TRUE/FALSE` flag, or having to copy and paste in an IP address that's always going to be the same?

It results in the same behaviour, it's just one is easier to activate than the other.

jmarshall:
True, the explicit switch is what protects people with miscreants on their local networks.  So the reason for specifying the URL in the switch is (1) somewhat tenuously, there may be a different equivalent for other S3-alike providers; (2) **we are not going to burn Amazon-specific IP numbers and paths into HTSlib**.

I would prefer a _.aws/credentials_ key to a random environment variable, as that is more self-contained.  However I have not verified that other _.aws/credentials_ parsing tools (notably Amazon's!) quietly accept keys that they do not recognise.  But cf #346.

jpiper:
Noted. How about using KSON (from klib) for JSON parsing? I'd be happy to cook up a new PR based on your develop branch (with `curl_kput`) with your aforementioned requirements (not storing the AWS metadata url in the code).

PR #364, opened 2016-04-04
Catch negative IDX tags; treat like other IDX parse failures.

daviesrob:
Fixes an invalid read error triggered by negative IDX values.  Error found by AFL after it had been left to run for 33 days.  bcf_hdr_set_idx tries to use the negative value in an array look-up, leading to undefined behaviour.

Currently IDX parse errors are handled by printing a warning and then just ignoring the IDX tag.  This makes the same happen for negative IDX values.  Whether or not we should actually fail on bad IDX tags is left as a subject for discussion.


jmarshall:
Thanks, rebased and merged as 653f45ef31d3bc485ff00aa67f3154904ee0a5ad.  I see the VCF spec doesn't give many details about what the value of this field should look like…

PR #365, opened 2016-04-05
Correct constant passed to bcf_hdr_id2int in bcf_hdr_check_sanity.

daviesrob:
Change BCF_HL_FMT to BCF_DT_ID when looking up GL header id.
The GL check can't have worked correctly, and using the wrong dictionary
caused odd things to happen if you used GL as a sample name.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -601,7 +601,7 @@ void bcf_hdr_check_sanity(bcf_hdr_t *hdr)
     }
     if ( !GL_warned )
     {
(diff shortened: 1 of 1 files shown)


jmarshall:
Thanks.  I also checked the bcftools test suite isn't affected due to a new _GL should be declared as Number=G_ message… which it isn't, so unsurprisingly I guess it wasn't testing this.

If we had moved past the 70's and these were enums rather than \#defines, this would have been a compile-time warning…\</snark>

PR #370, opened 2016-04-18
Ensure vcf_parse_format always fills out end of vector values.

daviesrob:
This test vcf file:

```
##fileformat=VCF
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	A	B
1	0	.	.	.	.	.	.	x:y:z	.:show_me_some_random_memory:.	.:
```
can make hts_file -c write out some uninitialized memory.  This is because  vcf_parse_format skips filling out end of vector values if the last sample record on a line finishes early without including all formats.  Fix by moving the code that fills end of vector values outside the loop that reads a sample record, to ensure it always gets run.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -1766,33 +1766,11 @@ static int vcf_parse_format(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v, char *p
                 for (; l < z->size>>2; ++l) bcf_float_set_vector_end(x[l]);
             } else abort();
             if (*t == '\0') {
-                for (++j; j < v->n_fmt; ++j) { // fill end-of-vector values
-                    z = &fmt[j];
-                    if ((z->y>>4&0xf) == BCF_HT_STR) {
-                        if (z->is_gt) {
-                            int32_t *x = (int32_t*)(z->buf + z->size * m);
-                            if (z->size) x[0] = bcf_int32_missing;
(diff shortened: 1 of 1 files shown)


vcf.c
```diff
@@ -1766,33 +1766,11 @@ static int vcf_parse_format(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v, char *p
...
-                        bcf_float_set_missing(x[0]);
-                        for (l = 1; l < z->size>>2; ++l) bcf_float_set_vector_end(x[l]);
-                    }
-                }
                 break;
             }
             else if (*t == ':') {
-                j++;
                 t++;
+                if (t < end) j++;
```
jmarshall: How come you needed this?  (Isn't it tautologically true?  `*end` is the NUL at the end of the kstring, so because `*t` here is `:`, it must be that `t < end`.  Or are you saying that `t` has maybe moved **past** `end`, which is a worse bug surely?)

daviesrob: Yes it is needed, if the line ends in a colon (see the test case).  In that case we don't want to increment `j` so that the "fill end-of-vector values" bit starts in the right place.

jmarshall: Oh, I missed that we're now _after_ the `t++` line.  Never mind then…!

PR #373, opened 2016-05-03
Add bcf_hdr_format(), handle BCF headers larger than 2GiB

jmarshall:
`bcf_hdr_read()` and `bcf_hdr_write()` need to consider BCF's `l_text` (header text length) as unsigned, so that huge headers can be handled.

Fixes samtools/samtools#567.


daviesrob:
It looks OK to me, although sadly I see it only doubles the maximum size allowed due to limitations of the BCF specification.

jmarshall:
@daviesrob: “640K should be enough for anyone” :smile:

Anyone want to comment on the addition of `bcf_hdr_format()` to the API before I merge this?

PR #376, opened 2016-05-13
Make hts_itr_query find no-coor reads when last reference is unused.

daviesrob:
When searching for HTS_IDX_NOCOOR reads, hts_itr_query would return the
wrong result for indexes where the last reference was unused.  Such
references have no entries in the binning index so it it not possible
to return an end offset for them.  Add a loop so that hts_itr_query finds
the last reference with mapped reads.  If found, its ending offset is
returned as the location where the no-coor reads start.

Fixes samtools/samtools#568

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -1752,11 +1752,17 @@ hts_itr_t *hts_itr_query(const hts_idx_t *idx, int tid, int beg, int end, hts_re
             break;
 
         case HTS_IDX_NOCOOR:
-            if ( idx->n>0 )
-            {
-                bidx = idx->bidx[idx->n - 1];
+            /* No-coor reads sort after all of the mapped reads.  The position is not stored
+               in the index itself, so need to find the end offset for the last mapped read.
+               A loop is needed here in case references at the end of the file have no mapped
+               reads.  See issue samtools#568. */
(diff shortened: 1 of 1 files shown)


daviesrob:
I noticed that the `HTS_IDX_START` case of hts_iter_query includes the line:
```
// Find the smallest offset, note that sequence ids may not be ordered sequentially
```
and then it iterates through all of the references instead of bailing out as soon as it finds something.  Can what is described in the comment actually happen for a properly sorted file?  

This pull request stops searching as soon as it finds a used reference for the `HTS_IDX_NOCOOR` case.  Is that OK, or should it be more paranoid and carefully check all of the references in case they're in a strange order?

jmarshall:
Re funky reference sequence ordering, relevant commits are b2aab8b7ed78bb0b351375fe664fdaf52b3fe570, 60c22d4849a80d47144d315966d3e5f9675e05d2, and cc207d82c83fc051c0d549d9f2e3d75448963e1c.  First one discloses some motivation; not sure I buy it, but nonetheless maybe yes you too should be more paranoid.

daviesrob:
I've made it more paranoid.

PR #379, opened 2016-05-19
Samtools#312

jkbonfield:
The htslib part of the solution for issue samtools/samtools#312.

This adds an additional return value to sam_index_build / sam_index_build2.

### DIFF
--- a/cram/cram_index.c
+++ b/cram/cram_index.c
@@ -494,8 +494,8 @@ static int cram_index_build_multiref(cram_fd *fd,
  * fn_idx is the filename of the index file to be written;
  * if NULL, we add ".crai" to fn_base to get the index filename.
  *
- * Returns 0 on success
- *        -1 on failure
+ * Returns 0 on success,
+ *         negative on failure (-1 for read failure, -4 for write failure)
  */
 int cram_index_build(cram_fd *fd, const char *fn_base, const char *fn_idx) {
     cram_container *c;
@@ -512,7 +512,7 @@ int cram_index_build(cram_fd *fd, const char *fn_base, const char *fn_idx) {
     if (!(fp = zfopen(fn_idx, "wz"))) {
         perror(fn_idx);
         free(fn_idx_str.s);
-        return -1;
+        return -4;
     }
 
     free(fn_idx_str.s);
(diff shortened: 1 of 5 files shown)


cram/cram_index.c
```diff
@@ -494,8 +494,8 @@ static int cram_index_build_multiref(cram_fd *fd,
  * fn_idx is the filename of the index file to be written;
  * if NULL, we add ".crai" to fn_base to get the index filename.
  *
- * Returns 0 on success
- *        -1 on failure
+ * Returns 0 on success,
+ *         negative on failure
```
mp15: It might be better to enumerate the list of different failures here.

jkbonfield: Done and force updated the commit.

sam.c
```diff
@@ -522,6 +522,7 @@ int sam_index_build2(const char *fn, const char *fnidx, int min_shift)
         idx = bam_index(fp->fp.bgzf, min_shift);
         if (idx) {
             ret = hts_idx_save_as(idx, fn, fnidx, (min_shift > 0)? HTS_FMT_CSI : HTS_FMT_BAI);
+            ret = ret ? -4 : 0;
```
jmarshall: Kinda incomprehensible.  Is the intent `if (ret < 0) ret = -4;` and is that clearer?

jkbonfield: Essentially yes.  It's the analogue of if (ret != 0) ret = -4, but basically it's the same thing as the only working return value is zero.

I don't particularly find ?: incomprehensible, but if you feel strongly about it I can resubmit the patch with the change you suggest.

jmarshall: Please do 😄

Me, I'm pro-`?:` in general, but in this case `ret = ret ? -4 : 0` bashes positive return values over the head, and to validate the change readers have to think about whether positive return values should be bashed to -4 (surely not!) or 0, and then go off and check exactly what `hts_idx_save_as` can return this week, and only then realise that positive return values (currently) won't occur.

Which is all a lot more work than just applying the general htslib rule that `retcode<0` ⇒ error.

jkbonfield: Agreed.  At the very least it ought to ave been ret < 0 ? -4 : 0, but an "if" is easy enough too.

jmarshall:
Thanks.  Squashed that last commit and merged as e1dd6353da4a93543674c447a12fccc95c167623.

PR #382, opened 2016-05-20
Fixed CRAM MD/NM generation to follow the same logic as calmd.

jkbonfield:
The changes here are:

- N and P cigar operations no longer count towards the edit distance.
- When seq and ref match but both are "N", this is still considered to be a difference.

Also see https://sourceforge.net/p/samtools/mailman/message/35098288/


jmarshall:
Thanks.  Merged as 4eef9a7fe5a2429550682e31c1935d5e4ef73555.

PR #383, opened 2016-05-20
Make kstring detect more errors and work better on 64 bit systems

daviesrob:
I was annoyed by the lack of an error return from k[v]sprintf, so I added one, plus:

Add kroundup_size_t which works with both 64 bit and 32 bit size_t.  **Not sure if we want to make this available as part of the API - if not it could easily be put into ks_resize as it's only used there.**

Make ks_resize preserve the value of s->m if realloc fails.

Always use ks_resize to resize the buffer.

Make ksprintf/kvsprintf and kgetline return an error code if ks_resize fails.


jmarshall:
Thanks.  Reformatted to use tabs (as the _k*.[ch]_ files do, sigh…), and merged as 67234cb0115001eb92909726fc7259bf1806377a.

PR #384, opened 2016-05-26
Add interfaces to hfile for delimited string input.

daviesrob:
There are three versions:
  hgetdelim - Reads up to a given delimiter.
  hgetln    - Specialization of hgetdelim for '\n'
  hfgets    - Wrapper for hgetln that provides the same interface as fgets.

Time to argue about if the `hFILE` pointer should be at the start or the end, and if `hfgets` should be `hgets`.

### DIFF
--- a/hfile.c
+++ b/hfile.c
@@ -28,6 +28,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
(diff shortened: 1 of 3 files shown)


jmarshall:
Thanks, applied as e87ae87d661bbdfee700517672f4ffd0054b4dfc, with `hgets()` spelt thus.  Apologies for the delay…

PR #386, opened 2016-06-06
Cygwin Support

dvklopfenstein:
Please consider this pull request which adds make targets to create a Cygwin DLL.

Added a test which checks that the Cygwin DLL contents match the Linux .so contents.
The tests compiles both:
  1. Linux .so shared lib
  2. Cygwin .dll shared lib

And uses "nm" to compare the contents of the two to ensure the contents match.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ lib*.a
 lib*.dylib
 lib*.so
 lib*.so.*
+cyghtsdll.dll*
 
 /bgzip
 /htsfile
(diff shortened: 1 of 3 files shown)


jmarshall:
Thanks for the pull request.  As it happens, I have also been looking into this lately: see 2cddfb3c75887ca02d58708e5519c6d62992634c, now on **develop**.  It's pretty similar to yours, but also builds plugins (but see the caveats in the commit message — advice welcome) and I think more closely follows Cygwin's DLL naming conventions.

dvklopfenstein:
Great. Thanks. I will give it a try on my Cygwin machine.

Issue #377, opened 2016-05-16
Re-bgzip files according to .gzi


jacrayner:
Hi, I would like to be able to bgzip according to a bgzip index file. 
The issue I have is that different implementations of bgzip segregate the uncompressed data into blocks in different ways. For instance if I convert a BAM to CRAM and back to BAM, will not necessarily  produce a identical BAM due to bgzip compression. Compression level/strategy is easy enough to match but  block segregation is less so.

The index file of bgzip files contains uncompressed offset and compressed offset for each block. The uncompressed offset essentially tells you how much uncompressed data is stored in each block and thus how to segregate the uncompressed data when re-bgzipping to match a previously bgzipped file.  

I have achieved this by adding another bgzip_write function to bgzf.c which just writes the data you give it into a block each time it is called. 

 ssize_t bgzf_block_write(BGZF *fp, const void *data, size_t length)
{
    if ( !fp->is_compressed )
        return hwrite(fp->fp, data, length);
    const uint8_t *input = (const uint8_t*)data;
    assert(fp->is_write);
    uint8_t* buffer = (uint8_t*)fp->uncompressed_block;
    memcpy(buffer, input, length);
    fp->block_offset += length;
    if (lazy_flush(fp) != 0) return -1;
    return length;
}

And then I added corresponding options and logic to bgzip.c to parse an index file and then feed uncompressed data to this function based on the index. 

Could I ask for comments whether something like this could be included in bgzip and any advice on best implementation if you seen any problems in mine? 

Thanks

jkbonfield:
I can see your point, but I fear it's never going to work.  There could be many reasons why BAM to CRAM to BAM can yield different data, not only due to boundaries but potentially also different size for aux elements, perhaps different order in aux tags, headers having M5 entries, etc.  Plus there could be a myriad of different ways the starting BAM was produced - possibly from Picard or maybe from Samtools; did it always start a BGZF block on a new sequence boundary, etc.  Put simply, we cannot guarantee the uncompressed BAM is bit level identical so keeping the index identical is inherently going to be problematic.

What problem precisely are you trying to solve here?  Why does it matter that the index is identical?  Perhaps CRAM -> BAM producing both BAM and BAM.bai simultaneously is a better solution (I'm not sure if it's possible already).

jkbonfield:
Infact I just did a test demonstrating this.

```
[...]

PR #387, opened 2016-06-07
Rebgzip

jacrayner:
Rebgzip a using an index file as file as discussed  in #377. Lets you decompress a BGZF and then re-compress it to obtain the same segregation of blocks. Given that BAM files are by definition bgzipped, it is useful for validation.

I am using this specifically for validating BAM-->CRAM-->BAM conversion, to obtain a bit match of the original BAM but this relies on the cramming process not changing any of the underlying data which may not be the norm depending on your initial BAM and cramming utility.

Possibly need to add a check that the amount of compressed data is as expected by the index, as the compression level and strategy may need to be modified. However this can only be done within bgzf_block_write() when not using multi threading.


bgzf.c
```diff
@@ -950,6 +950,35 @@ ssize_t bgzf_write(BGZF *fp, const void *data, size_t length)
...
+        fp->block_offset += copy_length;
+        input += copy_length; 
+        remaining -= copy_length; 
+        if (fp->block_offset == ublock_size) { 
+	    fprintf(stderr,"block_address: %d\n", fp->block_address);
+            if (lazy_flush(fp) != 0) return -1;
+	    fp->idx->noffs--;  // decrement noffs to track the blocks
+        }
+    }
+    return length - remaining;
```
jkbonfield: With ```while (remaining > 0)```, ```if (copy_length > remaining) copy_length = remaining``` and ```remaining -= copy_length``` remaining should always end up as zero.  There are no break statements in the while loop.

Is the length - remaining a stylistic thing, rather than just ```return length```?

jkbonfield:
The code looks concise and I don't have any issues with the functionality, barring manual testing (I'm about to do that now).   Do you have any automated tests to add to the test harness?  If so it would be good to get them in too, even if it's just a simple check of taking a previous bgzf + index, uncompressing and recompressing again to check identity and exercise the code path some.

bgzf.c
jacrayner: yes just stylistic, ripped out of bgzf_write() which has the same properties.

jacrayner:
This would be the command-line I would use to for decompressing and rebgzipping a file, I have also minimal 2 BGZIP block test files.   
$ bgzip -d -c bgziptest.gz | bgzip -I bgziptest.gz.gzi -c -g

Does this need to be integrated in to test.pl?

jkbonfield:
That would be ideal.  If you have a tiny bgzf file then you can test uncompress, recompress and then md5sum or cmp both to make sure you got identical results.  I don't think it needs anything more complicated than that, so simple is good.

Obviously you'd want a test case where this isn't the case without giving it the .gzi file, but I'm not sure how you get such data.  You obviously know as you need this code!

jacrayner:
Success! Had a bit of hassle because I was doing something stupid in test.pl hence why that last commit is a bit messy. Cheers

jkbonfield:
Merged via 54dc2320f0ce38a66cc73414253844b325dfceca

I did a few minor formatting tweaks; we prefer no trailing white space and (depending on origin of source file) use spaces instead of tabs.  (FWIW I use an emacs mode that autodetects tab vs space and indentation levels, so it generally does the right thing per file!  All very messy I know)

jacrayner:
Great I will take more care in future on formatting and yep new emacs modes are generally of interest to me, I will try out dtrt-indent, Thanks!

jkbonfield:
I use https://www.emacswiki.org/emacs/GuessStyle

I don't know how it compares to the other options.

PR #390, opened 2016-06-15
Add Google Cloud Storage support to htslib.

jakeakopp:


jmarshall:
Please point us at the documentation for the Google Cloud Storage facilities you are using in this code.  In particular:
* Is there precedent for the URL scheme you're suggesting?
* What is the reason for re-URL-encoding the `PATH`?
* Is there precedent for a `GCS_OAUTH_TOKEN` environment variable?

It seems that by adding a **s3_domain** or so key (that would have the value `s3.amazonaws.com` or `storage.googleapis.com`) the existing AWS code could also support GCS via [Simple Migration](https://cloud.google.com/storage/docs/migrating#migration-simple).  What would be the pros and cons for you of doing it this way?

evandbrown:
Just chiming in, hope this helps:

> * Is there precedent for the URL scheme you're suggesting?

The [get](https://cloud.google.com/storage/docs/json_api/v1/objects/get) operation prescribes this scheme.

> * What is the reason for re-URL-encoding the PATH?

The path (object key) [should be URL-encoded](https://cloud.google.com/storage/docs/json_api/#encoding) after it is extracted from the `gs://` scheme (e.g., `gs://BUCKET/some/object/name`  -> `https://www.googleapis.com/storage/v1/b/BUCKET/o/some%2Fobject%2Fname?alt=media`)

> ...the existing AWS code could also support GCS via Simple Migration. What would be the pros and cons for you of doing it this way?

Simple Migration (interop mode) works well, but it requires setting a [default project](https://cloud.google.com/storage/docs/migrating#defaultproj) which restricts accessible buckets to that project. It also requires creating, managing, and rotating the one-off migration credentials. My experience is that interop mode is great for one-time migrations of data, but support for the native GCS API is desirable for ongoing access.

Hope that helps,

Evan

smrgit:
I'll also chime in that one use-case here is for data in GCS that is owned by one project but has been made world-readable (*ie* "allUsers" have "READER" access), and is accessible to anyone, even someone who doesn't have a GCP project at all.

If anyone would like to do some testing, here is an open-access bam file from the CCLE dataset:
```
bash-4.2$ gsutil ls -l gs://isb-cgc-open/ccle/BRCA/DNA-Seq/ | sort -nr | grep "HCC1954.5"
558055062336  2016-03-11T17:42:24Z  gs://isb-cgc-open/ccle/BRCA/DNA-Seq/G15512.HCC1954.5.bam
    10126368  2016-03-11T17:42:24Z  gs://isb-cgc-open/ccle/BRCA/DNA-Seq/G15512.HCC1954.5.bam.bai
```

Sheila

jmarshall:
Thanks all for the additional information.  Please imagine that the htslib maintainers know nothing about GCS APIs and point us at documentation accordingly. :smile:

So the _get_ operation defines the back-end `https://www.googleapis.com/storage/v1/b/BUCKET/o/OBJECT` URL, but what I'm interested in is a description of the `gs:` URL pseudo-scheme.  (Note that this PR is unclear whether it's supporting `gs:` or `gcs:`, but it seems like the latter is just a mistake.)

So far, `gs:` appears to be an illusion provided by the `gsutil` tool.  Can you point me at where in the documentation the `gs://bucket/path/to/foo` conventions are described?  Are `gs:` URLs useable with other tools besides `gsutil`?

@jakeakopp: Did you invent the `GCS_OAUTH_TOKEN` environment variable, or is that used by other existing tools too, thus providing a precedent?  All: Does this environment variable seem like the best way to do authentication for these URLs, or are there configuration files that we could be parsing, similarly to how the htslib AWS code reads _.aws/credentials_ et al?

smrgit:
@jmarshall makes a good point about the suggested ``GCS_OAUTH_TOKEN`` environment variable -- perhaps it would be better to make use of [Google Application Default Credentials](https://developers.google.com/identity/protocols/application-default-credentials)?  (That page also describes usage of an environment variable called ``GOOGLE_APPLICATION_CREDENTIALS``.) 

And, just to clarify, GCS is the commonly used abbreviation for [Google Cloud Storage](https://cloud.google.com/storage/docs/) but the url's used to reference objects in GCS indeed start with ``gs://``.

jakeakopp:
@jmarshall @smrgit unfortunately there's no client library for C, so we decided to go with passing the (application default) token via an environment variable.

smrgit:
@jakeakopp thanks for that clarification -- though it seems like a client library for C would be a pretty desirable thing -- any idea whether such a thing might be in the works?  it seems like calling Python from C is doable ... see [here](http://stackoverflow.com/questions/3286448/calling-a-python-method-from-c-c-and-extracting-its-return-value), [here](https://docs.python.org/3/extending/embedding.html) and [here](https://docs.python.org/3/c-api/)

but if there's no way around the environment variable, would it be better to use ``GOOGLE_APPLICATION_CREDENTIALS`` which is mentioned elsewhere rather than the brand new (as far as I can tell) ``GCS_OAUTH_TOKEN`` ?  

@jmarshall do you have any other comments or questions?  I'm quite keen to get this functionality added to htslib (and thereby eventually to samtools and pysam)

jmarshall:
Keeping htslib portable and buildable by neophytes is difficult enough as it is, so I'm afraid calling out to Python is a non-starter.

I have a number of questions, and with a couple of people apparently from Google interested in this PR, hopefully these questions will find answers.

* `gs:` appears to be an illusion provided by the `gsutil` tool. Can you point me at where in the documentation the `gs://bucket/path/to/foo` conventions are described? Are `gs:` URLs useable with other tools besides `gsutil`?

  It might be useful to extend the authority part of the URL as `gs://[USER@]BUCKET/…`, to allow overriding a user or profile for credentials, as I did with `s3:` URLs.  Is there existing practice for this?  Such documentation would provide hints.

* Why have you chosen the [JSON API](https://cloud.google.com/storage/docs/json_api/) rather than the [XML API](https://cloud.google.com/storage/docs/xml-api/overview)?

* I would prefer not to invent things out of whole cloth for htslib, as the proposed `$GCS_OAUTH_TOKEN` appears to be.  It seems to me to be preferable for htslib to read the **same** environment variables and/or configuration files as the existing tools that people are already using, as they will already have their credentials set up there.  (This surely is how htslib would operate if it were written against a hypothetical S3 or GCS C client library.)

  (For example, for `s3:` URLs, htslib reads the `AWS_*` environment variables and _~/.aws/credentials_ etc config files that already contain the credentials that users have specified for their other AWS tools.)

  It appears that standalone `gsutil config` writes to _~/.boto_.  I propose that htslib should read _~/.boto_ (or as overridden by `$BOTO_CONFIG`) and use `Credentials.gs_oauth2_refresh_token` and/or `Credentials.gs_access_key_id`/`Credentials.gs_secret_access_key` from there.  Presumably the latter can be used with `Authorization: AWS`-style authentication, and I'm guessing that the former contains the same text that your proposed environment variable would?

  Similarly it appears that Cloud SDK-flavoured `gsutil` writes credentials to _~/.config/gcloud_.  So the proposal is that htslib would read credentials from whichever of this file and _~/.boto_ exists.  (Assuming that this file's “non-user-editable” format is parseable by mere mortals!)

  Is this proposal workable and appropriate?  Is there any moral or technical reason for htslib not to use the tokens found in these config files to access `gs:` files?

* At the moment, I do not have a GCS account.  I am a little loath to use a personal credit card to sign up for a time-limited free trial account for work-related tasks.  Is there a mechanism by which Google engineers interested in having GCS functionality in htslib could provide the htslib maintainers with a small GCS credit for test purposes?

smrgit:
@jmarshall -- if you send an email to info@isb-cgc.org with your google identity, we (the [ISB-CGC](www.isb-cgc.org) team) can provide you (and other htslib maintainers who may need it) with a GCP for test purposes 

I have also pinged @jrobinso to ask if he might be able to add his wisdom to this discussion since he has implemented access to bam files stored in GCS inside IGV (though maybe not from C)

jrobinso:
Hi,

IGV supports this by using googles web service for accessing cloud 
storage files by https.  Its simple to do if you already support https.  
Here's the javascript function that converts a gs:// url to the 
equivalent https.

igv.translateGoogleCloudURL =function(gsUrl) {

     var i = gsUrl.indexOf('/',5);
     if (i <0) {
         console.log("Invalid gs url: " + gsUrl);
         return gsUrl;
     }

     var bucket = gsUrl.substring(5,i);
     var object =encodeURIComponent(gsUrl.substring(i +1));

     return "https://www.googleapis.com/storage/v1/b/" +bucket +"/o/" +object +"?alt=media";



}

jrobinso:
BTW,  I have no idea if the comment above is relevant at all to this discussion, probably isn't,  I was just attempting to respond to the request.     The authorization token ("access token") is passed in an Authorization header, as in
Authorization="Bearer " + acToken;

deflaux:
@jmarshall regarding support for testing by htslib maintainers - can you please email google-genomics-contact@googlegroups.com and we will send you the details.  Thanks!

thobrla:
@jmarshall 

The `gs://` URI nomenclature is basically the same as `s3://` nomenclature originally used by `s3cmd`.  It diverges slightly when dealing with object generations, which are not completely compatible with s3's notion of object versions.  The general form is `gs://bucket/object#generation` for referring to a specific object generation, and `gs://bucket/object` for referring to a live object.  I'm not aware of a documented standard for `s3:` or `gs:` URLs, though.

As for the question of XML vs. JSON API: there's no global reason to use one or the other, nor is there an official recommendation.  They're both supported APIs, with slightly differing feature sets.  The XML API is closer in form to s3, but s3-compatibility mode supports only a subset of GCS's features.

jakeakopp:
@jmarshall since htslib is a different client than gsutil, we have been advised to not share credentials between them.

I think between the various replies all your questions have responses now - please let us know what you think!

smrgit:
@jmarshall, I wanted to let you know that I have independently tested this functionality and for the purposes of the [ISB Cancer Genomics Cloud](www.isb-cgc.org) I'm quite keen to get this integrated into htslib so that it can percolate up to samtools, pysam, etc.

I have verified that the ``GCS_OAUTH_TOKEN`` usage works as expected for bam files that require an access token, and that no auth process (or token) is required for bam files that are world-readable in Google Cloud Storage.  Here is an example that you should be able to run pretty much from anywhere (once you've rebuilt htslib and samtools with the modifications in this p/r):
```
./samtools view gs://isb-cgc-open/ccle/LUSC/DNA-Seq/C836.NCI-H1339.2.bam 7:140453130-140453140
```
thanks!

smrgit:
@jmarshall -- are there any specific concerns about accepting this p/r as is?  I am working with a couple of other groups who are very interested in this functionality and I have provided them with the steps necessary to build htslib locally with these changes and then build samtools.  The NCI cloud pilots are hosting over 1 PB of bam files in Google Cloud Storage and it would really benefit the research community to be able to use tools such as samtools and pysam to access these bam files in GCS w/o having to copy them to a VM disk first.  Please let me know if there is anything I can do to move this forward.

lh3:
Just to add that a few broadies here are interested in google cloud storage. On acquiring credentials, `gcloud` writes them to `$HOME/.config/gcloud/credentials` with the standard installation/init procedure. It is a JSON file. It seems overkilling to introduce another library dependency just to parse this file. I don't like custom environment variables, either, but after some searches, it seems the best solution google gives us so far.

smrgit:
@jmarshall -- the [ISB-CGC](www.isb-cgc.org) team would like to be able to provide interested users with a pre-built version of htslib and samtools that includes this functionality.  Do you have a preference and/or recommendations regarding how we go about doing this?

jmarshall:
We have been focussing on other things for a while, but rest assured that we intend for GCS functionality to be in the next release of HTSlib.  There are upcoming changes to allow a form of `hopen()` that will allow extra HTTP headers to be provided, which will allow S3 and GCS support to be separated out from _hfile_libcurl.c_.  So the functionality won't be in the exact form of this PR, but this PR has demonstrated what is needed.

@thobrla wrote
> As for the question of XML vs. JSON API: there's no global reason to use one or the other, nor is there an official recommendation. They're both supported APIs, with slightly differing feature sets. The XML API is closer in form to s3, but s3-compatibility mode supports only a subset of GCS's features.

HTSlib's needs are minimal, so (as of June when I looked at this) are surely in the common subset of feature sets.  Thanks for the information that these APIs are equally supported.  It looks like by using the XML API we won't need to URL-encode the path, so that is simpler and clearer.  (I assume by _s3-compatibility mode_ you're referring to the Simple Migration stuff, which is orthogonal to this and not relevant.)

@jakeakopp wrote
> since htslib is a different client than gsutil, we have been advised to not share credentials between them.

I remain unexcited by the bespoke environment variable, but we can use it as a temporary step towards doing this properly.  @jakeakopp: Perhaps you could expand on this non-answer, and describe what OAUTH-based authentication for a client like samtools/htslib would ideally look like.

----

@smrgit Ideally of course, the recommendation would be to wait until version 1.4 is released with this functionality included.  In the meantime, feel free to provide a patched version to interested users, but please alter the version number (`PACKAGE_VERSION` in _htslib/Makefile_) to make it clear that it's not vanilla HTSlib.

smrgit:
@jmarshall that's great news -- do you have a target date for your next release?

jakeakopp:
@jmarshall I'm a bit confused - are you saying I should abandon this and you will implement support separately? If so that's fine, but I wanted to make sure.

Sorry for my "non-answer" - I'm happy to try and clarify if you still have questions. The ideal situation would be a Google API client library written in C, but sadly this doesn't seem to be coming in the forseeable future (i.e. C would be on this list: https://developers.google.com/identity/protocols/OAuth2#libraries).

smrgit:
@jakeakopp are there any impacts to this from the recently announced/released ["gcloud auth application-default login"](https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login) process?

pgrosu:
@jmarshall, @smrgit It's fairly simple to implement it in C by providing the OAuth2 token (or key) as an argument in the REST link.  For instance, you can do it all via `curl` without the use of `gsutil`.  Below is an example of reading the following file:

```
gs://genomics-public-data/1000-genomes-phase-3/README
```

Reading the object metadata:

```
curl -X GET -H "Content-Type: application/json" https://www.googleapis.com/storage/v1/b/genomics-public-data/o/1000-genomes-phase-3%2FREADME?qk=[YOUR_OAUTH_TOKEN]
```

You'll get something like this:

```
{
 "kind": "storage#object",
 ...
 "selfLink": "https://www.googleapis.com/storage/v1/b/genomics-public-data/o/1000-genomes-phase-3%2FREADME",
 "name": "1000-genomes-phase-3/README",
 "bucket": "genomics-public-data",
 ...
 "mediaLink": "https://www.googleapis.com/download/storage/v1/b/genomics-public-data/o/1000-genomes-phase-3%2FREADME...A_bunch_of_numbers...alt=media",
}
```

Then like @evandbrown mentioned, you can read the file's contents via the `mediaLink` just like this, though you would need to put it in quotes:

````
$ curl "https://www.googleapis.com/download/storage/v1/b/genomics-public-data/o/1000-genomes-phase-3%2FREADME...A_bunch_of_numbers...alt=media"

For more information on the data in this folder see:
https://cloud.google.com/genomics/data/1000-genomes-phase-3

The *full*mirror* of 1,000 Genomes data is also available from Google Cloud Storage underneath this path:

  https://console.developers.google.com/storage/browser/genomics-public-data/ftp-trace.ncbi.nih.gov/1000genomes/ftp
$
```

Hope it helps,
~p

jakeakopp:
@smrgit there were some minor changes to the commands that generate the token. I updated them in commit #4 of this PR.

pgrosu:
@jmarshall, @smrgit By the way you can get tokens in using C as well.  Here's the `curl` version, which is C-based:

**1)** If you look inside your `client_secrets.json` file, you will see something like this which you will need to use for the next sections of getting the token:

```
$ cat client_secrets.json  | sed -e 's/,/,\n/g'
{"installed":{"auth_uri":"https://accounts.google.com/o/oauth2/auth",
"client_secret":"YOUR_SECRET",
"token_uri":"https://accounts.google.com/o/oauth2/token",
"client_email":"",
"redirect_uris":["urn:ietf:wg:oauth:2.0:oob",
"oob"],
"client_x509_cert_url":"",
"client_id":"YOUR_ID",
"auth_provider_x509_cert_url":"https://www.googleapis.com/oauth2/v1/certs"}}
```

**2)** If you put together the following link to get the authorization code for the scope(s) you need, you can then paste in a web browser to get the code - I used the `genomics` scope as an example:

```
https://accounts.google.com/o/oauth2/auth?client_id=YOUR_ID&scope=https://www.googleapis.com/auth/genomics&response_type=code&redirect_uri=urn:ietf:wg:oauth:2.0:oob
```

After clicking the `Allow` button, this will bring you to a webpage with the code you need for the next step of getting the token.

**3)**  To get the token, put together the following `curl` request:

```
  curl \
    -d "code=THE_CODE_RETURNED_ABOVE" \
    -d 'client_id=YOUR_ID' \
    -d 'client_secret=YOUR_SECRET' \
    -d 'redirect_uri=urn:ietf:wg:oauth:2.0:oob' \
    -d 'grant_type=authorization_code' \
    https://accounts.google.com/o/oauth2/token
```

If you run this, you will get something like this:

```
{
  "access_token" : "THE_RETURNED_TOKEN",
  "expires_in" : 3600,
  "refresh_token" : "THE_REFRESH_TOKEN",
  "token_type" : "Bearer"
}
```

Again you can easily put this together in C, and just prompt the user to paste the initial link into a browser to get the authorization code.

Easy peasy :)
~p

jkbonfield:
PR #446 has now been merged to develop.  Please test it and let us know if it is sufficient.

jkbonfield:
Closing this as the reimplementation in PR #446 has been merged and feedback is positive.

PR #395, opened 2016-06-23
Added a ksprintf_g for %g specialisation.

jkbonfield:
VCF spends a lot of time doing ksprintf(s,"%g",val), especially on
VCFs with many samples.  The ksprintf_g is a dedicated function for
this one specific task that runs significantly quicker than the glibc
printf code giving around 4x speed improvement overall.  (Assumed less on single sample VCF.)

On a 1000genomes test user time was previously 50.5s and after this
patch was 12.5s, to convert uncompressed BCF to uncompressed VCF. (But
still -Ou for uncomp BCF is 0.2s, so VCF generation is still slow).

I haven't look at VCF decoding (vs encoding), but likely there is room
there too.  uBCF->uBCF is ~0.2s; uVCF->uBCF is 21.7s.

### DIFF
--- a/htslib/kstring.h
+++ b/htslib/kstring.h
@@ -82,7 +82,7 @@ extern "C" {
 
 	int kvsprintf(kstring_t *s, const char *fmt, va_list ap) KS_ATTR_PRINTF(2,0);
 	int ksprintf(kstring_t *s, const char *fmt, ...) KS_ATTR_PRINTF(2,3);
-    int ksprintf_g(kstring_t *s, double d); // custom %g only handler
+    int kputd(kstring_t *s, double d); // custom %g only handler
 	int ksplit_core(char *s, int delimiter, int *_max, int **_offsets);
 	char *kstrstr(const char *str, const char *pat, int **_prep);
 	char *kstrnstr(const char *str, const char *pat, int n, int **_prep);
--- a/htslib/kstring.h
+++ b/htslib/kstring.h
@@ -82,6 +82,7 @@ extern "C" {
 
 	int kvsprintf(kstring_t *s, const char *fmt, va_list ap) KS_ATTR_PRINTF(2,0);
 	int ksprintf(kstring_t *s, const char *fmt, ...) KS_ATTR_PRINTF(2,3);
+    int ksprintf_g(kstring_t *s, double d); // custom %g only handler
 	int ksplit_core(char *s, int delimiter, int *_max, int **_offsets);
 	char *kstrstr(const char *str, const char *pat, int **_prep);
 	char *kstrnstr(const char *str, const char *pat, int n, int **_prep);
--- a/kstring.c
+++ b/kstring.c
@@ -30,13 +30,124 @@
 #include <ctype.h>
 #include <string.h>
 #include <stdint.h>
+#include <math.h>
 #include "htslib/kstring.h"
+
+int ksprintf_g(kstring_t *s, double d) {
+	int len = 0;
+	char buf[21], *cp = buf+20, *ep;
+	if (d == 0) {
+		if (signbit(d)) {
+			kputsn("-0",2,s);
+			return 2;
+		} else {
+			kputsn("0",1,s);
+			return 1;
+		}
+	}
+
+	if (d < 0) {
+		kputc('-',s);
+		d=-d;
+	}
+	if (d < 0.0001 || d > 999999) {
+		if (ks_resize(s, s->l + 50) < 0)
+			return EOF;
+		// We let stdio handle the exponent cases
+		len = sprintf(s->s + s->l, "%g", d);
+		s->l += len;
+		return len;
+	}
+
+	uint64_t i = d*10000000000;
+	// Correction for rounding - rather ugly
+
+	// Optimised for small numbers.
+	// Better still would be __builtin_clz on hi/lo 32 and get the
+	// starting point very rapidly.
+	if (d<.0001)
+		i+=0;
+	else if (d<0.001)
+		 i+=5;
+	else if (d < 0.01)
+		i+=50;
+	else if (d < 0.1)
+		i+=500;
+	else if (d < 1)
+		i+=5000;
+	else if (d < 10)
+		i+=50000;
+	else if (d < 100)
+		i+=500000;
+	else if (d < 1000)
+		i+=5000000;
+	else if (d < 10000)
+		i+=50000000;
+	else if (d < 100000)
+		i+=500000000;
+	else
+		i+=5000000000;
+
+	do {
+		*--cp = '0' + i%10;
+		i /= 10;
+	} while (i >= 1);
+	buf[20] = 0;
+	int p = buf+20-cp;
+	if (p <= 10) { // d < 1
+		//assert(d/1);
+		cp[6] = 0; ep = cp+5;// 6 precision
+		while (p < 10) {
+			*--cp = '0';
+			p++;
+		}
+		*--cp = '.';
+		*--cp = '0';
+	} else {
+		char *xp = --cp;
+		while (p > 10) {
+			xp[0] = xp[1];
+			p--;
+			xp++;
+		}
+		xp[0] = '.';
+		cp[7] = 0; ep=cp+6;
+		if (cp[6] == '.') cp[6] = 0;
+	}
+
+	// Cull trailing zeros
+	while (*ep == '0' && ep > cp)
+		ep--;
+	char *z = ep+1;
+	while (ep > cp) {
+		if (*ep == '.') {
+			if (z[-1] == '.')
+				z[-1] = 0;
+			else
+				z[0] = 0;
+			break;
+		}
+		ep--;
+	}
+
+	len = strlen(cp);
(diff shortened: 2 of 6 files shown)


pd3:
This is great, +100 for seeing this merged soon.

mcshane:
There is an annoying GATK bug that I have just come across, where GATK chokes on VCF files that have passed through bcftools. See:

http://gatkforums.broadinstitute.org/gatk/discussion/4956/error-message-java-lang-integer-cannot-be-cast-to-java-lang-double

Turns out GATK doesn't like bcftools changing  `val=45.000` to `val=45` via a `ksprintf(s,"%g",val)` and it now thinks it is an integer despite the header saying to treat as a float. Given it is 1.5 years since that posting, doesn't look like GATK will fix soon. @jkbonfield is this something we could do here? For data read in as `45.000`, we would write it out as `45.0` rather than `45` as currently?

+1 for merging, especially if we can incorporate something to do the above.

jkbonfield:
I've no idea how to get %g to handle 45 as 45.0; likely it isn't possible.  However with this patch we're printing numbers using our own algorithm so it ought to be possible to bake this logic into it.

jkbonfield:
I have a version with outputs "45" as "45.".  I could also make it produce "45.0", but this becomes a bit problematic when we have 6 digit values.  With %g 7 digits goes into exponents, so "1234567" is "1.23457e+06", however 6 digits before dot is written to 6 significant places; thus "123456.7" is "123457", now "123457.".

That's a bit misleading, but not as misleading as "123457.0".  Is the addition of the "." on the end without specifying the next digit enough to get GATK to interpret the values as floats?  It's the minimal change if so.

PS. For standalone test program is https://gist.github.com/jkbonfield/32702b1b2aa4de7b7e483d6c1156c51a

I can make the changes to this PR if you're happy it works and is appropriate.  Previously I was trying to keep the output identical.

jmarshall:
What is this algorithm and where does it come from?  Does it take cognisance of the relevant literature, quite nicely [summarised here](http://www.serpentine.com/blog/2011/06/29/here-be-dragons-advances-in-problems-you-didnt-even-know-you-had/)?

jkbonfield:
I'll confess it's my own hack, implemented by turning the limited-range floating point value into a 64-bit integer and then doing successive integer divisions to get the string.  It's definitely MUCH faster than glibc.  For the more out of bounds values (ie all the exponential stuff) I just fall back to printf, as these aren't that common in our outputs so I didn't feel the need to optimise them (plus they're harder).

Given it's a 4 fold speed increase for outputting VCF, this particular algorithm must be much more than 4x faster.  I suspect it therefore beats Grisu (which claims 5x faster than glibc).  I haven't benchmarked it though.

I'll try and do an exhaustive comparison of all values from -999999 to +999999 to validate the result and to test the speed.  If I can also get grisu code from somewhere I'll verify that too.

jkbonfield:
Ok that was harder work than it looked.   I hate overcomplicated C++!  I couldn't find my original exhaustive test code from when I first implemented this so I redid it.  I should probably have left something inside comments or a #ifdef TEST_MAIN wrapper.

@ deskpro107386[jkb/tmp]; time ./a.out 0 > /tmp/_0

This is ksprintf_g, sprintf "%g" and DoubleToStringConverter::ToPrecision from https://github.com/google/double-conversion.

```
real	0m1.470s
user	0m0.786s
sys	0m0.149s
@ deskpro107386[jkb/tmp]; time ./a.out 1 > /tmp/_1

real	0m7.320s
user	0m6.661s
sys	0m0.132s
@ deskpro107386[jkb/tmp]; time ./a.out 2 > /tmp/_2

real	0m7.850s
user	0m7.774s
sys	0m0.080s
```
On a larger test set (10 times smaller granularity in testing) the new code comes out around 8x faster than glibc.

It's possible the additional C++ hoops needed to get their Grisu algorithm called from my test harness meant that all speed gains were rendered null and void by a reimplementation of kstring.  I couldn't find a simple "here's a buffer - write to it!" function. Grr.

There is another implementation of the Dragon4 algorithm here (http://www.ryanjuckett.com/programming/printing-floating-point-numbers/part-4/), but the benchmarks show it isn't a HUGE benefit over the standard library implementation.  Of course this depends what numbers you check.  I only did the range that commonly occur in VCF and just fall through to glibc for the rest.  This keeps it simple.

As for testing, I validated it thus:

```
#define LOOP for (d = 0.0001; d < 1000000; d*=1.000001)

    switch (method) {
    case 0:
        LOOP {
            s.s = buf+idx;
            s.l = 0;
            idx += ksprintf_g(&s, d);
        }
        out = buf;
        break;

    case 1:
        LOOP {
            idx += sprintf(buf+idx, "%g", d);
        }
        out = buf;
        break;

    case 2: {
        StringBuilder b(buf, 1024*1024*1024);
        LOOP {
            dc.ToPrecision(d, 6, &b);
        }
        out = b.Finalize();
        idx = strlen(out);
        break;
    }
    }

    fwrite(out, 1, idx, stdout);
```

The first two methods give identical output, while the grisu3 one differs in a few places.  That's not to say my code is perfect.  Clearly there are some rounding errors in very subtle places as picking specific values from the command line give me rounding differences in the last digit, but frankly I don't care if it rounds up instead of down!  They weren't spotted by the test loop as it didn't generate enough perfect .5 boundary cases.

I don't know what else I can do really.  I'm certainly not going to switch to someone elses mammothly complex algorithm when it's slower, harder to understand and an order of magnitude more code.

Anyway, I did spot an error in the return values so it wasn't wasted effort. :-)

mcshane:
Hi @jkbonfield. Just bumping this as I am getting to need the GATK workaround now.

jkbonfield:
I have a version that prints "10." instead of "10" now, *however* it also means the QUAL column also ends in "." after every quality as these are integer values printed as floats.  Is this correct?  It seems rather strange.  I assume GATK reads those as integers, so this will replace one parsing bug with another.

pd3:
Can we write a small dedicated program for this instead? It could be very fast. In addition to @jkbonfield's comments, it feels awfully wrong to hack this on htslib's side. This really should be fixed in GATK.

jkbonfield:
I take it back - it's just our test data consists solely of integer QUALs.  All other files I look at have a multitude of floating point.  I'll fix the failing tests and push.   However I would advise more aggressive checking as I'm up to my eyeballs in mpeg stuff and haven't got time to give this the required level of testing.

jkbonfield:
I'm in agreement with Petr here - we should file this as a GATK bug IMO and temporarily use a script to add "." or ".0" to all floating point numbers until then.  However the code is there - cherry pick the bits you want and ignore the bits you don't. :-)

NeillGibson:
Hi @jkbonfield @pd3 @mcshane .  Is there any progress on this float formatting issue?

We currently use BCBio for large WGS read alignment and variant calling analysis and one remaining botttleneck is single cpu concat of all analysis block VCF.gz  files to all chromosome VCF.gz file.

Currently BCBio uses `GATK CatVariants` which is twice as slow as `bcftools concat` to keep downstream compatibility with GATK derived tools. 

If this float formatting issue would be resolved then this would save us and other organizations  days of waiting for the VCF files being concatenated, e.g. from a week to 3 days of concatenating.  

GATK by the way says this is not a bug in their software but a functionality / bug in the `Apache Commons Java Expression Language (JEXL)` library that they depend on.  So that would then maybe be the correct place to file the bug. 
http://commons.apache.org/proper/commons-jexl/

Is `bcftools naive concat` of BCF also affected by this issue or would that also be a viable work around? 

Thank you for the very nice and useful bcftools software and I hope that this float formatting / compatibility issue can be resolved.

pd3:
Hi @NeillGibson , the naive concatenation does not change the formatting. If you can assure that the samples come in the same order, then it should work and also should be much much faster.

mcshane:
I think naive concat on the bcftools develop branch supports `vcf.gz` now too. Is that right @pd3?

Internally the vote has been not want to hack this into htslib, but more opinions are welcome.

pd3:
That's correct, vcf.gz is supported as well as bcf.

NeillGibson:
Thank you very much for your answers. If the `naive vcf.gz / bcf.gz concat` works then that would be perfect for avoiding the float formatting / gatk compatibility issue and getting a major performance upgrade for the concat.  

This does require that the variant calling pipes in BCBio produce consistent VCF headers. 
It would be nice if there was the functionality in bcftools to check that the headers are consistent before doing the naive concat.

NeillGibson:
Hi @pd3 . I tested the naive vcf.gz concat  and bcftools stats and picard genotype concordance confirm that the variants and genotypes are identical to a GATK CatVariants concatenated vcf.gz 

Do you have a timeline for when a new bcftools stable version will be released that includes the naive vcf.gz concat  functionality? Thank you.

pd3:
The new release is tied with changes in htslib and the timeline there is unclear, therefore I don't know, sorry! The current bcftools github version is considered stable.

NeillGibson:
Ok thank you for the information.

mcshane:
Merged in #452 without the hack to add `.` like `45` => `45.`

PR #397, opened 2016-06-28
Threading pool

jkbonfield:
# **For consideration and evaluation**

Mammoth BGZF threading overhaul.  See also the samtools and bcftools PRs.

Htslib
------

- Promoted thread_pool.[ch] from CRAM subdirectory to the top-level
  exposed API.  We make use of this instead of the older threading in
  bgzf.c.  TODO: consider type and function naming scheme.

  This code also has a #ifdef TEST_MAIN portion.  This is designed to
  be both debugging to check it works (although it isn't a full test
  suite as it has no knowledge of the expected result; we'd need to
  save the output and compare, but also as a tutorial for different
  styles of threading using the pool.  It has demos of unordered, a
  couple ordered approaches (with and without an async dispatcher) and
  a pipeline example using multiple queues.  That TEST_MAIN code is
  probably the best place to start if you wish to see the API.

- Improved multithreaded BGZF encoding.  It's roughly twice as fast
  now as it no longer takes a boom & bust approach, instead running
  continuously in the background.

- Added multithreaded BGZF decoding.

- New hts_create_threads(), hts_destroy_threads() and
  hts_set_thread_pool() function.  The former two are simply wrappers
  around t_pool_init and t_pool_destroy, indicating that we should
  probably rename all the functions in thread_pool.c instead.  This
  needs some discussion as to what naming scheme is most appropriate.

  hts_set_thread_pool uses an additional structure (htsThreadPool)
  consisting of the pool itself (t_pool) and a queue size (int).  It's
  easier to code this way, but I don't particularly like it.
  Suggestions welcomed.

- Bgzf has a new bgzf_thread_pool() function which is analogous to the
  old bgzf_mt() function, but with a supplied pool instead. (Use
  hts_set_thread_pool in preference.)

- New hts_set_cache_size() function, selected by --input-fmt-option cache_size=<size>.
  This makes little difference unless doing random access.

  It is, however, incompatible with the current multi-threading
  implementation and will be ignored when MT is enabled.

Samtools
-------

- [ERRONEOUS!] Somehow managed to also acquire the travis change to
  use autoconf.  I can remove, but I suspect we want this anyway.
  (Separate PR already exists so if merged first it'll just resolve
  itself.)

- Samtools index now supports threading.

- Samtools sort now multi-threads the decoder too.  It doesn't share a
  thread pool as the code is structured to only do one type of thing
  at any point anyway (all decoding, or all sorting).  Also simplified
  the merge step when coming from sort so it doesn't rewrite the RG
  headers or per-read aux tags.

- Samtools view now multi-threads BAM decoding too.

Bcftools
--------

- Added multi-threaded BCF decoding for annotate, index, merge, norm,
  roh, stats and view.  More tools can be done trivially, but I don't
  know enough on how to use and test them.

Benchmarks
==========

```
alias svi-samtools=/software/svi/bin/samtools-1.3.1
in=/nfs/users/nfs_j/jkb/scratch/data/NA12892.chrom20.ILLUMINA.bwa.CEU.high_coverage.20100611.bam
```

BAM to BAM transcoding
----------------------

```
@ gen1c[samtools.../samtools]; time ./samtools view -b -o /tmp/out.bam $in 20:10000000-20000000
real    1m12.211s
user    1m10.940s
sys     0m1.060s

@ gen1c[samtools.../samtools]; time ./samtools view -@16 -b -o /tmp/out.bam $in 20:10000000-20000\
000
real    0m5.507s
user    1m19.737s
sys     0m2.924s

@ gen1c[samtools.../samtools]; time svi-samtools view -b -o /tmp/out.bam $in 20:10000000-20000000
real    1m10.138s
user    1m8.920s
sys     0m1.024s
@ gen1c[samtools.../samtools]; time svi-samtools view -@16 -b -o /tmp/out.bam $in 20:10000000-200\
00000
real    0m14.220s
user    1m18.337s
sys     0m2.180s
```

=> ~2.6x faster.

```
@ gen1c[tmp/samtools]; time nh13-samtools -n 16 view -b -o /tmp/out.bam $in 20:10000000-20000000
Setting the number of threads per PBGZF file handle to 16
Reading with 16 threads.
Writing with 16 threads.
real    0m8.447s
user    1m36.890s
sys     0m4.680s
```

Indexing
--------

```
@ gen1c[samtools.../samtools]; time ./samtools index $in
real    0m57.156s
user    0m45.639s
sys     0m1.352s

@ gen1c[samtools.../samtools]; time ./samtools index -@16 $in
real    0m6.184s
user    1m4.332s
sys     0m2.484s
```
=> 9x faster
```
@ gen1c[tmp/samtools]; time nh13-samtools -n 16 index $in
Setting the number of threads per PBGZF file handle to 16
Reading with 16 threads.
real    0m19.016s
user    1m22.193s
sys     0m11.101s
```

Sorting
-------

(Memory reduced to force it to still use more than 1 block when multi-threading).

```
@ gen1c[samtools.../samtools]; time ./samtools sort -m 100M -n -T /tmp -o /tmp/out.bam $in
[bam_sort_core] merging from 211 files...
real    13m54.328s
user    16m21.261s
sys     0m28.094s

@ gen1c[samtools.../samtools]; time ./samtools sort -@16 -m 100M -n -T /tmp -o /tmp/out.bam $in
[bam_sort_core] merging from 160 files...
real    2m32.353s
user    15m16.669s
sys     0m32.250s

@ gen1c[samtools.../samtools]; time svi-samtools sort -@16 -m 100M -n -T /tmp -o /tmp/out.bam $in
[bam_sort_core] merging from 160 files...
real    3m32.722s
user    15m42.975s
sys     0m30.542s
```
=> 40% faster.
```
@ gen1c[tmp/samtools]; time nh13-samtools -n 16 sort -m 100M -n $in /tmp/out
Setting the number of threads per PBGZF file handle to 16
(Fails)

@ gen1c[tmp/samtools]; time nh13-samtools -n 16 sort -m 1000M -n $in /tmp/out
Setting the number of threads per PBGZF file handle to 16
Reading with 16 threads.
Writing with 16 threads.
Writing with 16 threads.
...
[bam_sort_core] merging from 16 files...
Reading with 16 threads.
Reading with 16 threads.
...
Writing with 16 threads.
real    5m28.331s
user    18m55.471s
sys     1m14.301s
```

Cache_size
----------
```
ranges() { perl -e 'srand(1); for ($i=10000000;$i<11000000;$i+=int(rand()*1000)) {print "20:$i-$i\
\n"}'; }
ranges2() { perl -e 'srand(1); for ($i=10000000;$i<20000000;$i+=int(rand()*10000)) {print "20:$i-\
$i\n"}'; }
@ gen1c[samtools.../samtools]; time svi-samtools view $in $(ranges) | wc -l
52748
real    0m13.317s
user    0m13.105s
sys     0m0.204s

@ gen1c[samtools.../samtools]; time ./samtools view $in $(ranges) | wc -l
52748
real    0m13.364s
user    0m13.185s
sys     0m0.172s

@ gen1c[samtools.../samtools]; time ./samtools view $in --input-fmt-option cache_size=16M $(range\
s) | wc -l
52748
real    0m11.435s
user    0m11.221s
sys     0m0.204s

@ gen1c[samtools.../samtools]; time ./samtools view $in --input-fmt-option cache_size=64M $(range\
s) | wc -l
52748
real    0m5.414s
user    0m5.276s
sys     0m0.136s

@ gen1c[samtools.../samtools]; time ./samtools view $in --input-fmt-option cache_size=256M $(rang\
es) | wc -l
52748
real    0m2.085s
user    0m1.948s
sys     0m0.136s
```

With less dense ranges the impact is negligible due to fewer cache hits:
```
@ gen1c[samtools.../samtools]; time ./samtools view $in $(ranges2) | wc -l
52429
real   0m13.313s
user   0m12.893s
sys    0m0.192s

@ gen1c[samtools.../samtools]; time ./samtools view --input-fmt-option cache_size=256M $in $(rang\
es2) | wc -l
52429
real    0m11.169s
user    0m10.849s
sys     0m0.312s
```
Multi-threading in this case is detrimental as it spends a lot of time
reading/decoding ahead and then discarding the data:
```
@ gen1c[samtools.../samtools]; time svi-samtools view -@16 $in $(ranges) | wc -l
52748
real    0m13.376s
user    0m13.177s
sys     0m0.188s

@ gen1c[samtools.../samtools]; time ./samtools view -@16 $in $(ranges) | wc -l
52748
real    0m9.233s
user    1m13.641s
sys     0m12.453s

@ gen1c[samtools.../samtools]; time ./samtools view -@16 $in --input-fmt-option cache_size=256M $\
(ranges) | wc -l
52748
real    0m7.971s
user    1m1.328s
sys     0m10.029s
```

Bcftools index
--------------
```
@ gen1c[samtools.../bcftools]; time svi-bcftools index -f /lustre/scratch109/sanger/jkb/all.bcf
real    0m6.655s
user    0m6.480s
sys      0m0.148s

@ gen1c[samtools.../bcftools]; time ./bcftools index -f --threads 16 /lustre/scratch109/sanger/jk\
b/all.bcf
real    0m0.841s
user    0m9.333s
sys     0m0.372s
```
=> ~8x faster

BCF->BCF
--------
```
@ gen1c[samtools.../bcftools]; time svi-bcftools view -Ob -o /tmp/_.bcf /lustre/scratch109/sanger\
/jkb/all.bcf
real    1m25.445s
user    1m24.341s
sys     0m0.872s

@ gen1c[samtools.../bcftools]; time ./bcftools view -Ob -o /tmp/_.bcf /lustre/scratch109/sanger/j\
kb/all.bcf
real    1m25.797s
user    1m24.373s
sys     0m1.216s

@ gen1c[samtools.../bcftools]; time svi-bcftools view --threads 16 -Ob -o /tmp/_.bcf /lustre/scra\
tch109/sanger/jkb/all.bcf
real    0m13.513s
user    1m34.594s
sys     0m1.932s

time ./bcftools view --threads 16 -Ob -o /tmp/_.bcf /lustre/scratch109/sanger/jkb/all.bcf
real    0m6.281s
user    1m34.870s
sys     0m2.548s
```
=> 2.2x faster


htslib.mk
```diff
@@ -68,7 +68,8 @@ HTSLIB_PUBLIC_HEADERS = \
 	$(HTSDIR)/htslib/tbx.h \
 	$(HTSDIR)/htslib/vcf.h \
 	$(HTSDIR)/htslib/vcf_sweep.h \
-	$(HTSDIR)/htslib/vcfutils.h
+	$(HTSDIR)/htslib/vcfutils.h \
+	$(HTSDIR)/htslib/thread_pool.h
```
jmarshall: These are in alphabetical order.

jmarshall:
> Promoted thread_pool.[ch] from CRAM subdirectory to the top-level exposed API.

This is a surprise.  I for one didn't realise it was proposed to expose the guts of _thread_pool.h_ as a public API…?

jmarshall:
There are 44 commits here to be considered.  Around a fifth of them (_Minor improvement to binary searching_ (already on **develop** as 57462fb729fd467cbfa43f4bccf67), 3 × _Merge develop_, _Fixed broken code_, _White space updates_, _Sorry, updates to the previous commits_, perhaps 2 more _Merge develop_) are immaterial and I'd appreciate it if they were omitted or squashed as appropriate.

jkbonfield:
Right now all I've done is add basic BGZF multi-threading, but see the examples in thread_pool.c.  If we wanted to, say, rewrite merge to be using the pool then it would involve adding extra jobs to dispatch to the thread pool etc in much the same way that bgzf.c is.  This means exposing more of the API.

The nature of this API is something I wanted to discuss, hence making this pull request and asking for feedback.  Right now it has a bit of hts_thread_* type API as well as the thread_pool.h, and I'm not sure whether to continue down the road of adding an hts interface to each function, whether to rename the thread_pool.h ones, or whether to cherry pick the few specific bits we need as and when the occasion arises.

Regarding lots of commits; yes, it reflects reality.  Sometimes I even made a cock up and went back and fixed things, but that's just the nature of coding.  I could either make a quick fix and commit it or spend 10 minutes unravelling git history and attempting to construct a nice sanitised one, in the process losing my train of thought on what I'm actually doing (the thread locking stuff is far from easy!).  I expect at some point we'll want to tidy things up, but in the mean time it's beneficial for people to be evaluating the end product rather than finding flaws with the route I took to get there.

jmarshall:
> Sometimes I even made a cock up and went back and fixed things, but that's just the nature of coding. I could either make a quick fix and commit it or spend 10 minutes unravelling git history

You should add `git commit --amend` to your repertoire — it's even quicker than committing a new quick fix commit.

But yes, your choice of updating your PR to omit and/or squash now or later.

jkbonfield:
On Tue, Jun 28, 2016 at 08:33:39AM -0700, John Marshall wrote:
> You should add `git commit --amend` to your repertoire -- it's even quicker than committing a new quick fix commit.

I used that quite a lot actually but obviously it's not visible.
There were also times I needed to amend a commit several commits back.
I've since found out that rebase permits reordering of commits as well
as squashing, so that's probably what I needed to use.

Anyway, for now it's really not worth trying to follow the long list
of commits (including various bug insertions and subsequent deletions
/ rewrites a week later), but the end product and whether the API is
sane.

It's best to start with the thread_pool.c examples and then we can
decide which bits of the API need to be public and whether we want a
naming convention (almost certainly yes, but I didn't do it as I don't
want to rename them twice).  I'm sure there are bits I can make static
or remove from thread_pool.h and into (eg) thread_pool_internal.h.

-- 


-- 

jkbonfield:
Pfft, just noticed a probaln change accidentally saved in the wrong checkout.  I'll tidy it up post-meeting.

jkbonfield:
Tidied up the rogue commit and a couple other niggles.

Still to do - work out how best to deal with htsThreadPool combined pool + qsize structure.  (Likely solution is two hts_opt parameters instead of one combined.)

Possible change of hts_tpool_dispatch to only take the process-queue argument rather than the pool as well.  (See FIXME.)  Comments welcomed.

Final to-do is for someone else to actually use it in anger for something.

nathanweeks:
I'll mention that I've prototyped forks of samtools/htslib that use OpenMP 3.1 to more efficiently parallelize the code path for BAM sorting (leaving the CRAM code untouched). This includes using task parallelism to overlap multi-threading decoding with sorting (where each sort task sorts several million records), and improving concurrency in the encoding. Sorted records are merged in-memory before being written to a single output file. If there's interest, I could clean up the code & make it "pull-ready" for review.

The performance of sorting a ~20 GiB BAM file in-memory was up to 3.92x faster than samtools 1.3 when 32 threads were used (1.55X faster when extra disk I/O for temp files was required when using the default memory). 

From a software engineering standpoint, this would admittedly be a big change from the custom thread pool but implementation---perhaps too big of a paradigm change to consider (though it might be of interest to some in other contexts). At least OpenMP 3.1 is supported by pretty much every major C compiler (http://openmp.org/wp/openmp-compilers/), with the exception that the OS X XCode implementation of Clang/LLVM currently lacks support. The code changes implemented thus far resulted in a net increase of ~200 lines of code for htslib, and ~50 lines for samtools.

As an aside, a performance limitation of the current samtools sort implementation is that if the user specifies a large amount of per-thread memory, and the input BAM (or CRAM) fits within the total memory space (-@ NUMBER_OF_THREADS X -m PER_THREAD_MEMORY), then only a single thread will perform the sort. Another drawback is if the input exceeds the user-specified memory limit, the sort will occur in parallel, but each thread will write its output to a separate file. If there's interest, I could retrofit (sans-OpenMP) a non-task-based version of the in-memory parallel sort/merge code into the existing samtools/bam_sort.c. This should benefit performance when the entire data set fits within memory, otherwise it at least creates far fewer intermediate files to merge (one file for every approximately "-@ x -m" bytes, instead of "-@" files).

jkbonfield:
I found the same memory problems when benchmarking sort, which is why I had to artificially reduce it to stop it going single-threaded on me.  I'm not sure how much faster my code would make sort if it actually still worked in parallel when given enough memory. Sort is rather warty code that most definitely could do with a revamp, so thank you for looking at that.

The thread pool code is here because it already existed (in the cram subdirectory).  I just hiked it one level up and made bgzf.c use it too.  The old multi-threading there was rather simplistic and suffered hugely from a boom & bust model of thread handling.

Openmp is simpler IMO, and **personally** I don't object to it given it's all in pragmas and is just ignored by compilers that don't support it.   However I think a thread task/pool system is more flexible overall as it gives much easier control of sharing parallelism than omp.

What I'm curious to know is how your fixes to the sort algorithm (not omp, just restructuring) work when combined with this PR.  Ie how much of your gain is omp and how much is simply fixed bad code structure.

jkbonfield:
On a 6.1Gb file with lots of temporary merging (50Mb mem size) the old algorithm was taking 6m24 real time (~40m cpu, using 32 threads; 16 cores x2 hyperthread).  The new algorithm was 3m54s real time so around 45% quicker.

As a quick hack, I tested mode "wbx0" instead of "wbx1" for the worker() function when it's writing out the temporary files, and ran it with -T /dev/shm.  This is a poor equivalent to actually sorting in memory, but the code doesn't cope with this without a lot of hacking around which I didn't want to do.

This then took 2m25 sec, so around 2.6x faster.  It's still not your 3.9x faster, but I could see on top that the initial phase was very flip-flop in nature between read/decode and sort as the two don't run in parallel with each other.  This is something that could be achieved by code restructuring I think.  I'd be curious to know what your changes look like.  If we can achieve the same performance using the thread pool then I'd be inclined to go with it as it's more portable than openmp.

I'm sure there are lots of places where openmp could work though and I like the idea of it as (where it works well) it's a really trivial change.

nathanweeks:
The OpenMP version I benchmarked includes code from the memory-optimization pull request I submitted (https://github.com/samtools/samtools/pull/593), and used a BAM file over twice as large---and the larger the BAM file, the bigger the performance difference from https://github.com/samtools/samtools/pull/593, mostly due to reduced overhead in free()ing memory for millions of BAM records at the end. E.g., I profiled samtools 1.3 using hpctoolkit, and on this run ~18% of the total runtime was spent deallocating the BAM array memory:

samtools-1 3

After implementing the previously-described OpenMP multi-threading (but before interleaving the decode/sort tasks), the sort stage (purple in the picture below) was shorter relative to the decode and encode stages, and the "free()" overhead at the end was gone (note that this picture is with 16 threads instead of 32, and I neglected to record the times when I took this image):

samtools-openmp

I don't think it would be too difficult to implement the in-memory sort/merge optimization in the context of the current pthreads (non-thread-pool) implementation (though it might involve some code replication). That could get it into the hands of users sooner. From there, you could try adding it to the proposed thread pool implementation. Just let me know if you'd like to try that approach.

kyleabeauchamp:
Does anyone know if Xcode 8 supports OpenMP?  I know that the OpenMP LLVM implementation is now complete (http://openmp.llvm.org/), so it's presumably only matter of time until it becomes part of XCode?

nathanweeks:
I'm not sure; perhaps someone can download the Xcode 8 beta and find out. In my opinion, the only practical portability issue with OpenMP in htslib/samtools is with clang/LLVM on OS X. If the interleaved decode/sort tasks implementation is excluded (which actually requires the "taskgroup" construct from OpenMP 4.0, released April 22 .2014, not OpenMP 3.1 as I mentioned earlier), and the "atomic read" construct (which requires OpenMP 3.1, supported in gcc since version 4.7, released March 22, 2012) is converted to the somewhat slower "critical" construct---probably only a very minor performance hit given its use---then a compiler supporting OpenMP 3.0 could be used (supported in gcc since 4.4, released April 21, 2009). Also, the GCC OpenMP runtime (libgomp) thread pool implementation has been extensively tested through its use in many applications, and should be fairly robust.

That said, there is already a pull request to use the CRAM thread pool implementation throughout htslib/samtools, whereas to provide that level of consistency, the prototype OpenMP code would have to be implemented in other places as well.

jkbonfield:
FroggerIt!

In doing some more aggressive benchmarking I appear to have found a race condition where all threads are in cond_wait!  This is clearly not intended.  I need to figure out how this was possible.

PR #401, opened 2016-08-01
Protect against sequences starting beyond reference end.

jkbonfield:
This was previously causing the slice reference md5sum to crash.

Fixes samtools/samtools#600, but note this perhaps isn't a perfect
fix.  The test data for that issue is broken, but ideally we should be
able to reproduce the broken input after round-tripping.  Currently we
lose track of MD/NM tags.

However making cram totally lossless even when faced with invalid data
is an issue for another date.

NOTE: the CRAM file produced here cannot be decoded by htsjdk (however it also couldn't turn the test BAM into CRAM either).  Perhaps it's just a case of garbage in, garbage out, but fixing a crash is nonetheless important.


jmarshall:
Might not be a perfect fix, but it's certainly an improvement!

PR #409, opened 2016-09-09
Handle VCF lines with misssing FORMAT=.

pd3:
For example, this is a valid VCF line

1       300 .   C   A   .    PASS   .    .       .       .

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -1618,8 +1618,14 @@ static int vcf_parse_format(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v, char *p
         return -1;
     }
 
-    // get format information from the dictionary
     v->n_fmt = 0;
+    if ( p[0]=='.' && p[1]==0 ) // FORMAT field is empty "."
+    {
+        v->n_sample = bcf_hdr_nsamples(h);
+        return 0;
+    }
+
+    // get format information from the dictionary
     for (j = 0, t = kstrtok(p, ":", &aux1); t; t = kstrtok(0, 0, &aux1), ++j) {
         if (j >= MAX_N_FMT) {
             v->errcode |= BCF_ERR_LIMITS;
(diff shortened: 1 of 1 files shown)


jkbonfield:
The code seems nice and concise and it's backwards compatible so no issues there.

However I think the VCF specification isn't clear though on whether this is a good idea (and I don't know VCF well enough).  It states under the genotype fields section *"First a FORMAT field is given specifying the data types and order (colon-separated alphanumeric String). This is followed by one field
per sample,"*... followed by *"If any of the fields is missing, it is replaced with the missing value."*.

I think it's valid to assume that FORMAT itself may be missing, but it's not explicit in the spec.  Is this interpretation correct?  If so I'm happy for this to be merged.  If however the correct interpretation of *"If genotype information is present,..."* is that there FORMAT field and all subsequent fields should be completely absent, then this PR is the wrong solution.

pd3:
Although this is not specifically mentioned in the specs, that's the same logic all other fields use as well. The alternative would be to trim the line after INFO. That's OK and we do it for VCFs without samples. But if there are samples, the column-oriented logic would be broken. VCF in general is quite flexible and allows, but does not require, to trim multiple empty FORMAT fields.

daviesrob:
This is in the middle of the kstrtok() loop over the format field, so things like `GT:.:DT` would bail out in a slightly odd way in the middle.  Would it be better to hoist this out of the loop and check for a single `.` in the entire field? I.e. before line 1623:

```
if (strcmp(p, ".") == 0) {
    v->n_sample = bcf_hdr_nsamples(h);
    return 0;
}
```

You should also update the specification to make it clear that this usage is allowed - it's not very obvious at the moment.

pd3:
You are right, haven't thought of catching VCF format errors here. How about changing the condition like this:

    if ( t[0]=='.' && t[1]==0 )
    { 
         v->n_sample = bcf_hdr_nsamples(h);
         if ( j!=0 )
         {
             fprintf(stderr, "[E::%s] Invalid FORMAT tag name '.''\n", __func__);
             v->errcode |= BCF_ERR_TAG_INVALID;
         }
         return 0;   
    }

This way no additional code is executed in normal records.

daviesrob:
If you did that, entries like `.:GT` would still be a problem.  It should also return -1 for the error condition.

I think strictly a `.` on its own should be allowed, but a `.` with anything else should be illegal.

I see that `vcf_format` already writes this format for records where `n_fmt == 0` and `n_samples > 0`.  So we do need something like this to allow htslib to read its own output...

mcshane:
Moved to #411

PR #410, opened 2016-09-11
Add bgzf_index_load_fh()

sahlberg:
This function allow reading the index from a pre-existing FILE
instead of using a path.

Signed-off-by: Ronnie Sahlberg ronniesahlberg@gmail.com

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -1239,6 +1239,41 @@ static inline int fread_uint64(uint64_t *xptr, FILE *f)
     return 0;
 }
 
+int bgzf_index_load_fh(BGZF *fp, FILE *idx)
+{
+    fp->idx = (bgzidx_t*) calloc(1,sizeof(bgzidx_t));
+    if (fp->idx == NULL) goto fail;
+    uint64_t x;
+    if (fread_uint64(&x, idx) < 0) goto fail;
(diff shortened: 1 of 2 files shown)


sahlberg:
I have a use-case where I do not have a path available to the index file (https://github.com/sahlberg/fuse-bgzip).
The path to the index file is hidden below an overlay mount that hides both the .gz and the .gzi file. Since the oonly way to load an index file is by path, this means I have to copy the index to a different location first where it can be loaded from.

Things would become much easier if I could pass a FILE* to the hidden file instead.
Using openat() I can create file descriptors to these hidden files and thus avoid having to copy them to a different location.

sahlberg:
ping?

daviesrob:
We're looking to make bgzf_index_load use hFILE rather than FILE so that it works with s3 and other object stores.  This is done in #443 among other changes.

Adding an extra interface that takes an hFILE as input is fairly easy so we should be able to do something like this as well.  A file descriptor can be wrapped up as an hFILE using hdopen().

sahlberg:
@daviesrob Awesome.

sahlberg:
Closing this as a duplicate as this feature will be added in #443

PR #415, opened 2016-09-23
Fixed one of those obvious-in-hindsight bugs in bam_aux_update_str()

jenniferliddle:
<blush>

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1304,17 +1304,16 @@ int bam_aux_update_str(bam1_t *b, const char tag[2], int len, const char *data)
     bam_aux_del(b,s);
     s -= 2;
     int l_aux = bam_get_l_aux(b);
-    uint8_t *aux = bam_get_aux(b);
 
     b->l_data += 3 + len;
     if (b->m_data < b->l_data) {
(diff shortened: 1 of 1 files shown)


sam.c
```diff
@@ -1304,17 +1304,16 @@ int bam_aux_update_str(bam1_t *b, const char tag[2], int len, const char *data)
     bam_aux_del(b,s);
```
mp15: Unrelated to this but perhaps it would be wise to replace the abort() with a return -2; and respect the value of hts_verbose?

jmarshall: Perhaps not; if anything the whole line could be replaced with `assert`.  This is programmer error (either wrong tag used or failure to check tag is type `Z`) rather than runtime failure or user error, so should fail immediately and gracelessly.

jmarshall:
Thanks, merged as 19c189438f852e6e62dbda73f854d465cebb3d9f.

I'm a bit peeved this slipped through…  The `m_data`/`data` stuff would have been caught by the compiler but for excess casts, and the cached `aux` problem should have been caught by test cases.  Can you please check whether the _test/sam.c_ tests are exercising the realloc case and add some more if not.

PR #416, opened 2016-09-27
Reworked knetfile to handle FTP authentication with credentials provi…

Aldaer:
As of now, samtools (and other htslib-based tools) could use only anonymous ftp access, since that was hardcoded in knetfile.c. This change allows parsing username and password from FTP URL (ftp://username:password@ftp.address/filename). If no username and password is provided, new version falls back to anonymous connection. If only username is provided, it attempts to connect with that username and empty password. Tested OK under CentOS 7.

### DIFF
--- a/htslib/knetfile.h
+++ b/htslib/knetfile.h
@@ -60,6 +60,7 @@ typedef struct knetFile_s {
 
 	// the following are for HTTP only
 	char *path, *http_host;
+	char *user_cmd, *password_cmd;
 } knetFile;
 
 #define knet_tell(fp) ((fp)->offset)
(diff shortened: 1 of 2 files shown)


jmarshall:
The libcurl backend already supports this, and deals with unusual response codes etc much better than knet does.  Is there a reason why you can't configure with `--enable-libcurl` and use that instead of knet?

Your PR looks basically sound, but the libcurl backend has superseded knet and we're not terribly keen to actively maintain knet.

Aldaer:
Well, by default samtools still use knet, not libcurl. If would be nice to add support for FTP authentication out of the box, without the need to reconfigure. Not pressing, though.

daviesrob:
Closing as knet's days in htslib are probably numbered.

Issue #420, opened 2016-10-06
PTHREAD error when compiling htslib


jdmontenegro:
$> autoheader
$> autoconf
$> ./configure
$> make
(...)
echo '#define HTS_VERSION "1.3.2-129-gdaae2ea"' > version.h
gcc -g -O2 -I.  -c -o hts.o hts.c
gcc -g -O2 -I.  -c -o md5.o md5.c
gcc -g -O2 -I.  -c -o probaln.o probaln.c
gcc -g -O2 -I.  -c -o realn.o realn.c
gcc -g -O2 -I.  -c -o regidx.o regidx.c
gcc -g -O2 -I.  -c -o sam.o sam.c
gcc -g -O2 -I.  -c -o synced_bcf_reader.o synced_bcf_reader.c
gcc -g -O2 -I.  -c -o vcf_sweep.o vcf_sweep.c
gcc -g -O2 -I.  -c -o tbx.o tbx.c
gcc -g -O2 -I.  -c -o thread_pool.o thread_pool.c
thread_pool.c: In function ‘hts_tpool_init’:
thread_pool.c:658:38: error: ‘PTHREAD_MUTEX_RECURSIVE’ undeclared (first use in this function)
     pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
                                      ^
thread_pool.c:658:38: note: each undeclared identifier is reported only once for each function it appears in
make: **\* [thread_pool.o] Error 1

cannot build the development version of htslib in the server:

$> uname -a

Linux zeus 3.0.101-80-default #1 SMP Fri Jul 15 14:30:41 UTC 2016 (eb2ba81) x86_64 x86_64 x86_64 GNU/Linux

Any ideas how to fix it?

Cheers,

jkbonfield:
Hmm this define has been around since 2008 at least:

http://pubs.opengroup.org/onlinepubs/9699919799/

How old is that linux distro? 3.0 is 2011ish, so I'd think it's new enough so maybe it's some compiler defines we miss.

Does this compile if you define _POSIX_C_SOURCE?  If not, try _POSIX_SOURCE or if desparate _GNU_SOURCE.  We need to know why this function/argument is apparently missing on your OS.

It's possible it can be rewritten to avoid this, but a recursive lock was the least complex solution.

jdmontenegro:
Hi James,

The system administrator managed to get it compiled. I asked him and he
mentioned the need to add an extra flag to gcc during compilation.
[...]

PR #423, opened 2016-10-06
Fix undefined PTHREAD_MUTEX_RECURSIVE compilation error

nathanweeks:
A portable (POSIX) way to address issue https://github.com/samtools/htslib/issues/420 , which I have also observed in the Cray Linux Environment (based on SuSE 11.3) is to define _XOPEN_SOURCE 700 at the top of the affected file.

### DIFF
--- a/thread_pool.c
+++ b/thread_pool.c
@@ -26,6 +26,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <config.h>
 #endif
 
+#define _XOPEN_SOURCE 700 // for PTHREAD_MUTEX_RECURSIVE
 #include <stdlib.h>
 #include <inttypes.h>
 #include <signal.h>


jmarshall:
[Seems legit](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap02.html#tag_02_02_04), although I'm not sure we're using anything XSI so might prefer to define `_POSIX_C_SOURCE` as 200809L (see §2.2.1) — if that works, modulo other bugs on these platforms.

If these platforms use glibc, I suppose the problem is that they predate [this fix](https://sourceware.org/git/?p=glibc.git;a=commit;h=d3c7e68655d65b5e68cfc14c89aefa8dd3f8e3e2).

We should perhaps define this for all source files via _config.h_, or for all source files when it's needed via _configure_, rather than just for one file.  Or is worrying about ODR overly pedantic when it comes to these POSIX mode defines?

nathanweeks:
This system does have a pthread.h that predates that fix, and just defining _POSIX_C_SOURCE as 200809L doesn't work. Addding -D_POSIX_C_SOURCE=200809L to the compile command also uncovered a few XSI-isms (in addition to implicitly-declared functions that have been omitted here):

```
kfunc.c:74:16: error: 'M_SQRT2' undeclared (first use in this function)
  z = fabs(x) * M_SQRT2;
                ^~~~~~~
...
htslib/ksort.h:264:14: warning: implicit declaration of function 'drand48' [-Wimplicit-function-declaration]
    j = (int)(drand48() * i);         \
  ...
errmod.c:88:34: error: 'M_LN10' undeclared (first use in this function)
                 beta[k] = -10. / M_LN10 * logl(sum1 / sum);
                                  ^~~~~~
...
errmod.c:97:45: error: 'M_LN2' undeclared (first use in this function)
             em->lhet[n<<8|k] = lC[n<<8|k] - M_LN2 * n;
```

As a matter of principle, I agree with enforcing some level of standard conformance globally, as it should enhance portability to lesser-used platforms---e.g., has anybody tried compiling htslib on Solaris recently?---and increase the likelihood of reproducibility (10 years from now will someone still be able to compile this version of htslib on, e.g., RHEL 9.x?). OTOH, I don't want to distract developers from issues currently affecting more users.

jmarshall:
Thanks for the reminder about all those maths constants etc being XSI.

Re Solaris, in fact it compiles and passes the test suite just fine with some small build tweaks
(Oracle Solaris 11.3, Developer Studio 12.5; build with e.g. `gmake CC=cc LIBS='-lsocket -lnsl'`; remove `-pthread` from the final link commands, as it seems it doesn't link that when linking).

jmarshall:
Thanks Nathan; merged as afd9b56b2ae3fa7a5245dcb0bd5120b6ca359829.

jkbonfield:
I'm curious whether XOPEN_SOURCE 700 was required, or whether something lower than 500 or 600 worked?  I'm just experimenting at the moment with trying to work out what our actual minimum requirements are for compilation environments. 

This thread seems to imply 500 is necessary:

https://github.com/idris-lang/Idris-dev/issues/1059#issuecomment-66381867

jkbonfield:
Regarding -D_POSIX_C_SOURCE=200809L, we're using ```drand48``` instead of ```rand``` (```random``` is also xopen or bsd) and ```usleep``` (albeit an admittance of failure when it is in use!) so it's not a simple case of ifndef then define for the M_SQRT2 etc defines.

I think I'd prefer to simply state we have a dependency on c99 and _XOPEN_SOURCE=600 or higher, glossing over the slight bit of c11 (anonymous unions) for now until we observe whether it causes issues anywhere.

We'll be testing changes more aggressively on various platforms at some stage, but for now I've done a few code tidyups over here:

https://github.com/jkbonfield/htslib/tree/C99

PR #432, opened 2016-11-02
Suppress index date warning when hts_verbose == 0

dancooke:
Prevents a warning message being written to stderr from calls to  hts_idx_load2 when the target index is younger than the read file.


jmarshall:
Fair enough; I'm surprised you only found one message to suppress!  I guess you have your own htslib-using code that you're calling this from (as samtools and bcftools don't really have a way to set `hts_verbose` down to 0)?

dancooke:
Indeed; this seems to be the only one that's affecting my application right now. I'll suppress more as I come across them.

PR #438, opened 2016-11-22
Cram bug fixes

daviesrob:
Fixes a couple of problems detected during American Fuzzy Lop testing:

* cram_decode_estimate_sizes() would segfault if either the SQ or RN data series codecs were not defined.
* The rANS uncompressors could try to read beyond the end of the input data, for malformed inputs.

### DIFF
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -1031,6 +1031,7 @@ void cram_decode_estimate_sizes(cram_block_compression_hdr *hdr, cram_slice *s,
 
     /* Qual */
     cd = hdr->codecs[DS_QS];
+    if (cd == NULL) return;
     bnum1 = cram_codec_to_id(cd, &bnum2);
     if (bnum1 < 0 && bnum2 >= 0) bnum1 = bnum2;
     if (cram_ds_unique(hdr, cd, bnum1)) {
@@ -1042,6 +1043,7 @@ void cram_decode_estimate_sizes(cram_block_compression_hdr *hdr, cram_slice *s,
 
     /* Name */
     cd = hdr->codecs[DS_RN];
+    if (cd == NULL) return;
     bnum1 = cram_codec_to_id(cd, &bnum2);
     if (bnum1 < 0 && bnum2 >= 0) bnum1 = bnum2;
     if (cram_ds_unique(hdr, cd, bnum1)) {
--- a/cram/rANS_byte.h
+++ b/cram/rANS_byte.h
@@ -333,4 +333,18 @@ static inline void RansDecRenorm(RansState* r, uint8_t** pptr)
     *r = x;
 }
 
+// Renormalize, with extra checks for falling off the end of the input.
+static inline void RansDecRenormSafe(RansState* r, uint8_t** pptr, uint8_t *ptr_end)
+{
+    uint32_t x = *r;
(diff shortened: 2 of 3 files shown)


jmarshall:
Comments, especially on the rANS part, @jkbonfield?

jkbonfield:
I'll need to spend more time checking it, but it may be worth scouring https://github.com/jkbonfield/rans_static/blob/master/rANS_static4x8.c to see what other changes can be made.

I'm a bit concerned over whether it harms the speed too much given this is in a very tight loop.  That will depend on branch prediction perhaps.  I wonder if there is a way it can spot whether it is getting close outside of the loop (eg assume worst case on renorm) and have a tight inner loop with a secondary more careful loop for the final bytes.  But check speed first - it may not matter much.

daviesrob:
The changes pretty much came out of io_lib, so if there is any slowness then that suffers from it too.  I didn't bring in any of the more heroic parts though (e.g. RansDecRenorm2 in assembly) so it's not going to be quite as quick as io_lib's version.

I did change ptr_end to really point to the end of the buffer instead of 8 bytes before it.  That's mainly because I also check for running out of data before the decoder proper starts, and having to add random 8s all over the place would have just made things too confusing.  Hopefully the resulting (ptr_end-8) will be pulled out of the main loop by the optimiser.

The changes from #361 should have made things a bit faster anyway, so we should end up with something quicker than the last release.

I might try running a couple of benchmarks...

daviesrob:
Some benchmark results.  All programs were compiled with `CFLAGS='-g -Wall -O3'`, gcc version 4.6.3 on seq3c.  I tried three different htslibs (50db54b, before the latest cram updates were merged from #361 ; 293a426, current develop; and pr438, this pull request).  All runs were flagstats on NA12892.chrom20.ILLUMINA.bwa.CEU.high_coverage.20100611.cram.  Happily they all gave the same answer.

Test                 | Min       | LQ        | Median    | UQ        | Max
-------------------- | --------- | --------- | --------- | --------- | ---------
/tmp/tm.50db54b:real | 0m16.230s | 0m16.300s | 0m16.324s | 0m16.873s | 0m17.369s
/tmp/tm.293a426:real | 0m13.393s | 0m13.425s | 0m13.438s | 0m13.487s | 0m14.359s
/tmp/tm.pr438:real   | 0m13.361s | 0m13.411s | 0m13.421s | 0m13.577s | 0m14.121s
/tmp/tm.io_lib:real  |  0m9.806s |  0m9.865s |  0m9.874s | 0m10.401s | 0m10.564s

So, this change makes no real difference compared to current develop.  io_lib is still fastest, but I'll leave it to @jkbonfield to implement the other changes from there.  This PR is just to fix a couple of bugs, not to make htslib as fast as io_lib.

jmarshall:
So if there's no real slowdown compared to current develop, I'll merge these bug fixes along with the GA4GH stuff.

jkbonfield:
Looks good. 

Regarding any future benchmarking though, it's best to try scram_flagstat -b (the -b is benchmark mode) and samtools view -f 0xffff.  This is so that neither tool optimises away decoding, as normal flagstat will for most columns, but also neither tool spends time formatting output.

jkbonfield:
Checked speeds using samtools view -f 0xffff instead and it's within 1%.

PR #439, opened 2016-11-23
GA4GH streaming protocol

jmarshall:
This is the GA4GH stuff finally ready to merge.  There was one more gnarly bug at the end there…

One further thing to add to this is a minor _htsfile.c_ reorganisation to facilitate testing this with `htsfile -c`.  Or that can come later on **develop** post-merging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -295,7 +295,7 @@ md5.o md5.pico: md5.c config.h $(htslib_hts_h)
 plugin.o plugin.pico: plugin.c config.h $(hts_internal_h) $(htslib_kstring_h)
 probaln.o probaln.pico: probaln.c config.h $(htslib_hts_h)
 realn.o realn.pico: realn.c config.h $(htslib_hts_h) $(htslib_sam_h)
-textutils.o textutils.pico: textutils.c config.h $(hts_internal_h)
+textutils.o textutils.pico: textutils.c config.h $(htslib_hfile_h) $(htslib_kstring_h) $(hts_internal_h)
 
 cram/cram_codecs.o cram/cram_codecs.pico: cram/cram_codecs.c config.h $(cram_h)
 cram/cram_decode.o cram/cram_decode.pico: cram/cram_decode.c config.h $(cram_h) cram/os.h $(htslib_hts_h)
--- a/Makefile
+++ b/Makefile
@@ -150,6 +150,7 @@ LIBHTS_OBJS = \
 	hfile_net.o \
 	hts.o \
 	md5.o \
+	multipart.o \
 	probaln.o \
 	realn.o \
 	regidx.o \
@@ -292,6 +293,7 @@ vcfutils.o vcfutils.pico: vcfutils.c config.h $(htslib_vcfutils_h) $(htslib_kbit
 kfunc.o kfunc.pico: kfunc.c config.h $(htslib_kfunc_h)
 regidx.o regidx.pico: regidx.c config.h $(htslib_hts_h) $(htslib_kstring_h) $(htslib_kseq_h) $(htslib_khash_str2int_h) $(htslib_regidx_h) $(hts_internal_h)
 md5.o md5.pico: md5.c config.h $(htslib_hts_h)
+multipart.o multipart.pico: multipart.c config.h $(htslib_kstring_h) $(hts_internal_h) $(hfile_internal_h)
 plugin.o plugin.pico: plugin.c config.h $(hts_internal_h) $(htslib_kstring_h)
 probaln.o probaln.pico: probaln.c config.h $(htslib_hts_h)
 realn.o realn.pico: realn.c config.h $(htslib_hts_h) $(htslib_sam_h)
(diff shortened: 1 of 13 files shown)


jmarshall:
Seems to cause a `samtools reheader --in-place foo.cram` samtools test suite failure.  Hmmm…

daviesrob:
Looks like it's related to the hFILE buffering changes.  It writes all of the changes it wants into the buffer, but then fails to write them out.  In `cram_close`, `writebuffer_is_nonempty(fp)` says empty, even though the buffer has been altered.  The implementation of `writebuffer_is_nonempty` is just:

`return fp->begin > fp->end;`

which I guess is OK for write only mode, but not when the file has been opened read-write.

Would `fp->begin > fp->buffer` work? Or maybe an explicit flag to say the buffer needs flushing?

It looks like it might also flush more than it strictly needs to.  If it, say, read 4000 bytes, then overwrote the last one, `flush_buffer` would currently write the entire buffer.  If `hFILE` included a pointer to the first byte that was overwritten (so starting at fp->end and then updated if necessary on each `hwrite`) then `flush_buffer` could skip the part of the buffer that was untouched.  Would need yet another ABI change, though...

daviesrob:
Actually, that last bit isn't quite right.  I see that `hseek` always flushes out if there is data to be written.  So it's not possible for `fp->begin` to move backwards without a flush happening.

Instead `hseek` could store the last position it set `begin` to (say in `write_origin`).  Then `writebuffer_is_nonempty` becomes `fp->begin > fp->write_origin` and `flush_buffer` needs to flush the data between `fp->write_origin` and `fp->begin`.

jmarshall:
> Looks like it's related to the hFILE buffering changes.


Yep. I have a fix...
> return fp->begin > fp->end;
> 
> which I guess is OK for write only mode, but not when the file has been opened read-write.
> 
The idea is to call hflush() or hseek() when switching between reading and writing. Will check whether this code does that.



-- 

daviesrob:
> The idea is to call hflush() or hseek() when switching between reading and writing. Will check whether this code does that.

It effectively does read, seek, write, write, close.  So there is a seek between reading and writing.  The trouble is that the amount read is bigger than the amount written, so `fp->begin` never gets past `fp->end`.

A quick fix might be to make `hwrite` set `fp->end` to where it starts writing (but only if it happens to be bigger).  That would make `writebuffer_is_nonempty()` work again, at the expense of discarding the data after the write (not a major problem, I think).

daviesrob:
Fix was slightly harder than I thought.  See https://github.com/daviesrob/htslib/tree/pr439_fix_inplace

pgrosu:
@jmarshall @daviesrob Since [HTSlib](https://github.com/samtools/htslib) is a core library to the global community, may I ask if you could please add some documentation on the specifics of the GA4GH streaming protocol, including the protocol definition and discussion around it.  If you could help me, by pointing me to a public community discussion at the [GA4GH locations](https://github.com/ga4gh) where this proposal was reviewed before this proposed implementation, as [HTSlib](https://github.com/samtools/htslib) has quite a high standard around it before new additions to its codebase are accepted.  Do you by any chance have any benchmarks showing significant improvement of using this implementation, as compared to current methods based on specific standards?

Thank you,
Paul

jmarshall:
@pgrosu Yes, that is in progress. I guess you haven't been involved with this particular working group.

pgrosu:
@jmarshall  So usually before being involved I like to understand a little bit of what the group has done, and is planning to do.  I have tried to find this group but I cannot find it, as the only thing I found was that [this group exists](http://ga4gh.org/#/streaming-team).  All the other GA4GH groups contribute their progress publicly via Google Groups or through [Github](https://github.com/ga4gh), but unfortunately I cannot seem to find this group's public contributions or discussions.  Maybe you can help me with a place on the Internet where this group posts this?

I am sure there must be a way to be involved, follow the progress of this group, and how to get community input.  Even most code implementations on HTSlib provide supporting references to them, when one performs the following search:

https://github.com/samtools/htslib/search?utf8=%E2%9C%93&q=%22http%3A%2F%2F%22&type=Code

On just about everything I rely on scientific transparency to understand why new approach is explored, and the supporting analytical evidence for it.  It just makes it hard to work though, and evaluate why a piece of code appears in a critical codebase, when there could be other approaches.  

By the way, how would one one go about joining the Directory and Streaming group?

Thank you,
Paul

jmarshall:
So Rob's fix has all writing functions first check for a non-empty read-buffer and discard it (and seek to correct the file position) if there is one.  My fix has `hflush()` discard any read-buffer and do a corrective seek, and requires callers to use `hflush()` when switching between reading and writing _in either direction_.

So mine's cleaner, and Rob's is a bit more failsafe and parallels stdio read/write switching requirements.  Sigh, hmmm…

jmarshall:
…or the simple fix of just disabling the optimisation on files opened for update.

pgrosu:
@jmarshall  Why was this merged?  There was no review for the last code updates by other members, and we still have an open set of questions.  This feels a bit irregular to the normal review and merge process.

PR #442, opened 2016-12-07
Make hts_expand handle realloc failure a bit better.

daviesrob:
The hts_expand() and hts_expand0() macros don't return a value, and will
set (ptr) to NULL if realloc fails.  This could lead to segmentation faults
if callers don't check the value of (ptr).  As few if any do, and some
callers are in external packages, the best solution is to make the
hts_expand macros print an error message and call exit(1).

It's not ideal behaviour for library code, but fixing this in any other
way is just too hard.

This implementation removes the assumption that (m) and (n) are of type
int.  It should work with any integer type no bigger than size_t.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -2182,3 +2182,52 @@ hts_idx_t *hts_idx_load2(const char *fn, const char *fnidx)
 
     return hts_idx_load_local(fnidx);
 }
+
+
+
+/**********************
+ ***     Memory     ***
+ **********************/
+
+/* For use with hts_expand macros *only* */
+size_t hts_realloc_or_die(size_t n, size_t m, size_t m_sz, size_t size,
+                          int clear, void **ptr, const char *func) {
+    /* If new_m and size are both below this limit, multiplying them
+       together can't overflow */
+    const size_t safe = (size_t) 1 << (sizeof(size_t) * 4);
+    void *new_ptr;
+    size_t bytes, new_m;
+
+    new_m = n;
+    kroundup_size_t(new_m);
+
+    bytes = size * new_m;
+
+    /* Check for overflow.  Both ensure that new_m will fit in m (we make the
+       pessimistic assumption that m is signed), and that bytes has not
+       wrapped around. */
+    if (new_m > ((1 << (m_sz * 8 - 1)) - 1)
+        || ((size > safe || new_m > safe)
+            && bytes / new_m != size)) {
+        errno = ENOMEM;
+        goto die;
+    }
+
+    new_ptr = realloc(*ptr, bytes);
+    if (new_ptr == NULL) goto die;
+
+    if (clear) {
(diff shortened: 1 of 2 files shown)


bioinformed:
I **strongly** disagree with this approach.  It is awful for library code and callers should be fixed.  Please-- do not proceed with adding more code that cannot be recovered from.  If it is deemed necessary, then only enable this functionality if it can be disabled with the preprocessor, so libraries like `pysam` don't have to hack the htslib source code.

bioinformed:
I strongly disagree with this approach. It is awful for library code and
callers should be fixed. Please-- do not proceed with adding more code that
cannot be recovered from. If it is deemed necessary, then only enable this
functionality if it can be disabled with the preprocessor, so libraries
like pysam don't have to hack the htslib source code.

jmarshall:
So do we, but elegance and practicality are sometimes a tradeoff.  Before this PR, parts of HTSlib and other callers crashed; now they exit more gracefully.  The circumstances of crashing/exiting are unchanged, so this PR has not changed HTSlib's behaviour wrt recovery.  Moreover we have never received any reports about this crashing, so clearly `hts_expand` failing is a memory exhaustion scenario that is exceedingly rare in practice.

Patches (encompassing `hts_expand` in all of htslib/samtools/bcftools) are welcome.

bioinformed:
I agree that aborting with a message is better than a random crash, but
there are so many other places where ignoring return values or side-effects
will result in crashes -- this is C after all.  The flip side is that since
crashes due to `hts_exand` are so rare and are nearly irreproducible, do we
really expect to see bug reports?

daviesrob:
I can assure you that a great deal of thought went into this problem before opting for this solution.  Any solution that avoids calling exit either leaves things no better than they are (memory leaks, NULL pointer dereferences); makes things worse (NULL pointer dereferences become out-of-bounds array accesses) or involves making changes to the 143 places where hts_expand has been used in bcftools.

Future updates will remove the use of hts_expand from HTSlib and samtools.  bcftools can get its own implementation.  Once that's done we can mark it as deprecated in HTSlib and the whole sorry mess goes away.

It's actually quite easy to trigger hts_expand failures by creative use of the IDX tag in vcf/BCF files.  This was found by [american fuzzy lop](http://lcamtuf.coredump.cx/afl/), which has been looking for bugs in htslib for a while now.  The problem is easy to reproduce, but at least you can't get a core dump by exploiting it now.

bioinformed:
`pysam` makes fairly limited use of `bcftools` via high-level API calls,
so our concern about embedded `abort()`s is mainly in `htslib`.  So I
appreciate the logic better now that I understand the longer term goal of
removing `hts_expand` from htslib.  Thanks for taking the time to explain.
Do you have an estimate on when this future update will be implemented?

PR #443, opened 2016-12-07
Allow fai index to be in a different location to the indexed file.

### LINKED ISSUE #375: add the ability to specify an index file to fai_load
Hi all,

a suggested feature: BAM/CRAM files etc permit the loading of a named index. Could such an argument be added to fai_load?

See here: https://github.com/pysam-developers/pysam/issues/270

Best wishes,
Andreas

daviesrob:
Convert bgzf_index_load, bgzf_index_dump, fai_load, fai_build, fai_read and
fai_save to use hfile instead of stdio.  This allows access to remote
indexes via http, ftp etc. and the plugin infrastructure.

Add new API interfaces fai_build3() and fai_load3() which take separate
names for the fai and gzi index files.  If an index file name is
passed in as NULL, it is derived from the name of the file being indexed
as with fai_build() and fai_load().  As a result, fai_build() and
fai_load() are replaced by simple wrappers that call fai_build3() and
fai_load3() with NULL index file names.

The download_and_open() function which made local copies of remote index
files is removed.  The side effect of creating local files was not
desirable in some cases, and download_and_open() suffered from race
conditions if two processes tried to access the same index simultaneously.
It was also not called for .gzi files.  fai_build3() and fai_load3() can
directly access remote files for both indexed, .fai and .gzi files.

This removes fai_save() as a public symbol in libhts.so, but this
function does not appear in a public header file so is not part of
the official HTSlib API or ABI.

Fixes #375

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -1587,10 +1587,10 @@ int bgzf_index_add_block(BGZF *fp)
     return 0;
 }
 
-static inline int fwrite_uint64(uint64_t x, FILE *f)
+static inline int hwrite_uint64(uint64_t x, hFILE *f)
 {
     if (ed_is_big()) x = ed_swap_8(x);
-    if (fwrite(&x, sizeof x, 1, f) != 1) return -1;
+    if (hwrite(f, &x, sizeof(x)) != sizeof(x)) return -1;
     return 0;
 }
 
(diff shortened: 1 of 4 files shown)


sahlberg:
Please also add a bgzf_index_load() variant that takes a hFILE instead of a path.
See #410

PR #444, opened 2016-12-10
Provide more informative error message when unknown SAM tag type is e…

cinquin:
…ncountered.


jenniferliddle:
Does this need a 
`#undef _parse_err_param`
at the end of the function? For consistency, if nothing else.

cinquin:
@jenniferliddle  Sure; I've added a matching `undef`.

PR #452, opened 2017-01-10
Add a kputd for %g specialisation

mcshane:
Update to PR #395 from @jkbonfield to speed up `ksprintf(s,"%g",val)`. See discussion there.

Squashes those commits, reverts the addition of trailing `.` for whole floating point numbers. Also fixes a small bug for negative exponential values.


jkbonfield:
Agreed on the bug fix.  I didn't double check the s.l field, only return value before.

Looks good to go from my side.

PR #453, opened 2017-01-10
vcfutils: replace exit() with return -1 in bcf_remove_allele_set

mcshane:
follow up to ee4725892ee3f29a813ba906d7c8b07928a3f352

* `bcf_remove_allele_set` to return `int` rather than `void`
* replace `exit(1)` in `bcf_remove_allele_set` with `return -1`
  along with cleanup on error
* wrap error messages in `hts_verbose>1`
* indicate `bcf_remove_alleles` is deprecated with HTS_DEPRECATED

### DIFF
--- a/htslib/vcfutils.h
+++ b/htslib/vcfutils.h
@@ -1,7 +1,7 @@
 /// @file htslib/vcfutils.h
 /// Allele-related utility functions.
 /*
-    Copyright (C) 2012, 2013, 2015 Genome Research Ltd.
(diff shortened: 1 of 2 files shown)


mcshane:
Note: will need to update this line (https://github.com/samtools/bcftools/blob/develop/vcfview.c#L447) in bcftools if merged.

jmarshall:
Thanks for the updates… squashed and merged as 19014511d2f8366ea6404c928f75069913ab3c5a.

PR #457, opened 2017-01-17
Allow CC to be set externally in Makefile

aaronk:
Came to this change through mfumagalli's ngsTools (https://github.com/mfumagalli/ngsTools).  It's a project that uses htslib as a submodule and builds through the Makefile, rather than your autotools setup.  I'm assuming that the Makefile is here for such circumstances.

Just setting CC via ?= so that we can externally control the compiler.


jmarshall:
You can already externally control the compiler via `make CC=blah` or `make -e` without this patch.

See also samtools/samtools#533 et al.

PR #458, opened 2017-01-24
Add +ref modifier to the VCF synced reader

pd3:
When selecting compatible records, it was not possible to match ALT=.
lines with SNPs or indels.

This commit also changes the behavior for MNP records, making them
compatible with SNP records. I think that's the correct thing to do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,25 @@ compiler:
   - clang
   - gcc
 
-script: make -e && make test
+os:
+  - linux
+  - osx
+
+env:
+  - USE_CONFIG=no
+  - USE_CONFIG=yes
+
+# For linux systems
+addons:
+  apt:
+    packages:
+    - liblzma-dev
+    - libbz2-dev
+
+# For MacOSX systems
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update    ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install xz; fi
+
+script:
+  - if test "$USE_CONFIG" = "yes" ; then autoreconf && ./configure ; fi && make -e && make test


daviesrob:
Could you add the line to the NEWS file to summarize the change, please?  Or alternatively leave it as a comment here and we can add it when merging.

We're trying news-as-you go for this version, instead of trying to collect it all together in one go.

jkbonfield:
I initially thought that maybe as +REF is a modifier that it should be its own bit-field (eg COLLAPSE_REF) ORed with COLLAPSE_SNPS and COLLAPSE_INDELS.  However I can see this may cause a lot more code to have to change as all previous code would change behaviour and would need to check that COLLAPSE_REF isn't set.

So having entirely new entries COLLAPSE_REF_SNPS and COLLAPSE_REF_INDELS while somewhat breaking the notion of a true bit-field is perhaps better for backwards compatibiity (and I think it's not a full bit-field already anyway).

Ie.  Looks OK to me, but may need some documentation somewhere.  I know the bulk of this will be in bcftools, but maybe add a brief comment about it here too, as has been done in the #defines for the others.

pd3:
@jkbonfield That was my initial thought too, but then settled on this approach as it is more general: for example one could wish to pair REFs only with SNPs but not with Indels.

jkbonfield:
Thanks Petr.  Looks good.

pd3:
I just realized that the code should be extended to include also a `some+ref` case, however strange it may sound. The motivation is to be able to merge SNP records with REFs, but at the same time also records from the following example:

       FileA:
           POS=10 ALT=C

       FileB
           POS=10 ALT=A
           POS=10 ALT=C

Currently one can either require that some alleles match (`-c some`) and loose the reference records, or include the reference records (`-c snps+ref`) and pair incorrectly ALT=C with ALT=A.

Another option is to make `-c snps` smarter so that when duplicate positions are encountered, it will prioritize the matching alleles. This is probably a better solution.

Issue #450, opened 2017-01-02
Build on ARM64 Fails


octonion:
Make goes well until I get to the cram routines, then I get the following error:

echo '#define HTS_VERSION "1.3.2-180-g6d927df-dirty"' > version.h
gcc -g -O2 -I.  -c -o hts.o hts.c
gcc -g -O2 -I.  -c -o cram/cram_encode.o cram/cram_encode.c
cram/cram_encode.c: In function ‘process_one_read’:
cram/cram_encode.c:3183:1: error: expected declaration or statement at end of input
 }
 ^
Makefile:136: recipe for target 'cram/cram_encode.o' failed
make: *** [cram/cram_encode.o] Error 1

Builds fine on an Intel box, however. Any idea what could be the problem?

daviesrob:
It's a misplaced opening bracket at line 2601 (if should be inside #ifdef ALLOW_UAC).  It will be fixed soon, along with a few other portability problems.

Issue #355, opened 2016-03-16
BCF code makes no attempt to account for endianness


jmarshall:
The _vcf.c_ `bcf_hdr_read()` and `bcf_hdr_write()` functions work only on a little-endian host — `hlen` is read/written assuming that the host and on-disk representations are the same.

(I have not audited the other BCF code, e.g., record reading/writing.)

See also [_bcftools/vcfconcat.c_'s `naive_concat()`](https://github.com/samtools/bcftools/blob/9cf0d82493db16f74e9ff278f0d5573c50694b90/vcfconcat.c#L584-L586) which similarly reads `hlen` without accounting for endianness.

jmarshall:
The particular `hlen` bugs previously identified appear to have been ameliorated by @daviesrob's 37c85e4b9737abc88871daf724552959b612e7cb and 70622cffc711b0e501c431bc221a6abdbdf3a6bd, which presumably were part of PR #459.

Has the BCF code been further audited to check whether BCF I/O is now free from endianness bugs?

jkbonfield:
We did have functioning big-endian variants of htslib back in 2014 (eg see comments in https://github.com/samtools/htslib/pull/99), but I don't know what happened to those PRs.  I know I did a SPARC branch, which was subsequently replaced by Rob's complete rewrite, but I'm not sure either ever made it in as a completed work.

Realistically it all needs retesting to be sure.

jmarshall:
While the HTSlib code is improved (as previously noted: https://github.com/samtools/htslib/issues/355#issuecomment-329719915), note for example over in BCFtools on current **develop** [_bcftools/vcfconcat.c_'s `naive_concat()`](https://github.com/samtools/bcftools/blob/fbe438e26722a89bc980468db60147f899cc983d/vcfconcat.c#L775-L777) still reads `hlen` without accounting for endianness.

jkbonfield:
See https://github.com/daviesrob/htslib/tree/SPARC

I'm not sure why this work was never put in.  I did a lot of work on this, subsequently followed up by Rob, but no PR ever got made.  Odd.

daviesrob:
sam/bam/cram reading should be OK, but there are almost certainly issues with vcf/bcf.

Unfortunately my big-endian testing platform (a netbsd SPARC VM) lost its ability to talk to Github (maybe endian issues in some of the newer encryption code?) so I haven't been able to do much on this recently.

jmarshall:
[...]

Issue #119, opened 2014-08-15
System endiannes and unaligned memory access issues on mips


azlicic:
While trying to fix samtools issue #268 (https://github.com/samtools/samtools/issues/268), I came to a conclusion that htslib is a part of the problem.

I will send you a pull request for htslib as soon as I'm done.

Best Regards
Aleksandar Zlicic

PR #459, opened 2017-01-25
Fix undefined behaviour and improve endian-related behaviour

daviesrob:
Undefined behaviour includes:
    
* Illegal shifts
* Calls to memcpy(ptr, NULL, 0)
* Unaligned access

A new header file htslib/hts_endian.h is added.  This includes functions to read and write little-endian
values, which may be unaligned, in a portable way.  Some endian-specific code is rewritten to be
endian-neutral, but with optional optimisations for little-endian platforms that allow unaligned access.
    
Behaviour on big-endian systems is also changed, notably bam auxiliary data is now stored in little-endian order instead of being byte-swapped.  This means code accessing aux data needs to byte swap integer and float values on big-endian platforms.  To assist this, more bam_aux functions are added.  These include bam_auxB_len(), bam_auxB2i() and bam_auxB2f() for accessing array elements.
    
Justification for this:
    
* The data was previously stored inconsistently depending on if you read a sam, bam or cram file.  This meant big-endian platforms basically didn't work before this change.
* The values are often unaligned.  Some platforms (e.g. sparc, mips, armhf) need special handling for unaligned data, and it's easy to do the byte swapping at the same time.
* No time is wasted byte swapping aux values that aren't going to be accessed (although this is only an advantage for bam).

Undefined behaviour detected by compiling with -fsanitize=undefined and running the test harnesses for htslib, samtools and bcftools.

Endian compatibility tested using netbsd on sparc (emulated with qemu).

Should help with issues  #98, #119, #355, #450, https://github.com/samtools/samtools/issues/268 and https://github.com/samtools/samtools/issues/612.  I'll mark them as fixed once we have more evidence that the problem has actually been solved for the platforms in those issues.  Samtools issues are likely to need some extra work on samtools itself.

### DIFF
--- a/htslib/hts_endian.h
+++ b/htslib/hts_endian.h
@@ -1,0 +1,354 @@
+/// @file hts_endian.h
+/// Byte swapping and unaligned access functions.
+/*
+   Copyright (C) 2017 Genome Research Ltd.
+
+    Author: Rob Davies <rmd@sanger.ac.uk>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.  */
+
+#ifndef HTS_ENDIAN_H
+#define HTS_ENDIAN_H
+
+#include <stdint.h>
+
+/*
+ * Compile-time endianness tests.
+ *
+ * Note that these tests may fail.  They should only be used to enable
+ * faster versions of endian-neutral implementations.  The endian-neutral
+ * version should always be available as a fall-back.
+ *
+ * See https://sourceforge.net/p/predef/wiki/Endianness/
+ */
+
+/* Save typing as both endian and unaligned tests want to know about x86 */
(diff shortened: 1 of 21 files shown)


daviesrob:
Force-pushed to fix the annoying missing semicolons....  Sorry....

htslib/hts_endian.h
```diff
@@ -0,0 +1,354 @@
...
+
+/// Get an int64_t value from an unsigned byte array
+/** @param buf Pointer to source byte array, may be unaligned
+ *  @return A 64 bit signed integer
+ *  The input data is interpreted as 2's complement representation in
+ *  little-endian byte order.
+ */
+static inline int64_t le_to_i64(const uint8_t *buf) {
+    uint64_t v = le_to_u64(buf);
+    return (v < 0x8000000000000000ULL
```
jkbonfield: Being pedantic (which is valid given the sole reason for this PR is due to compiler pedantry), then ```0x8000000000000000ULL``` should be ```__UINT64_C(0x8000000000000000)```.

jkbonfield: Actually maybe not an issue.  ULL is long-long which is defined to be at *least* 64-bits, so maybe larger and inefficient, but always big enough to use in conjunction with a uint64_t expression without overflow.

sam.c
```diff
@@ -1174,36 +1176,36 @@ int sam_format1(const bam_hdr_t *h, const bam1_t *b, kstring_t *str)
...
             if (s+4 <= b->data + b->l_data) {
                 kputsn("i:", 2, str);
-                kputw(*(int32_t*)s, str);
+                kputw(le_to_i32(s), str);
                 s += 4;
             } else return -1;
         } else if (type == 'f') {
             if (s+4 <= b->data + b->l_data) {
-                ksprintf(str, "f:%g", *(float*)s);
+                ksprintf(str, "f:%g", le_to_float(s));
```
jkbonfield: May want this as a kputs f: and a kputd, but it's really minor and no relevance unless we genuingly have B-arrays of floats. (I've yet to see one.)  Perhaps solve it if/when we see it later.

vcf.c
```diff
@@ -451,7 +452,8 @@ int bcf_hdr_register_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec)
 
     // INFO/FILTER/FORMAT
     char *id = NULL;
-    int type = -1, num = -1, var = -1, idx = -1;
+    uint32_t type = -1, var = -1;
```
jkbonfield: Just double checking that the assignment of -1 to an unsigned type is intended.

jkbonfield:
Many interruptions later, I've looked over this and it seems OK barring the couple bits to double check as commented.  I haven't gone through all the vcf.c vint stuff in detail as frankly it's boggling already, but the extra tests give me confidence.

I've dealt with the conflicts in my local checkout.

vcf.c
daviesrob: Yes, in as much as it does what the old code did.  The main change here is that `var` and `type` are masked correctly when packed into `info`.

mp15: Can I suggest we use UINT32_MAX from stdint.h instead? Things like this make me shiver.

daviesrob: Know your C standard!   This conversion is well-defined in section 6.3.1.3:
"Otherwise, if the new type is unsigned, the value is converted by repeatedly adding or
subtracting one more than the maximum value that can be represented in the new type
until the value is in the range of the new type." (with footnote: "The rules describe arithmetic on the mathematical value, not the value of a given type of expression.")

Although it's really down to how this handles invalid input, which is a whole other can o'worms.

PR #461, opened 2017-01-26
BGZF skip empty blocks, do not give up reading prematurely. Fixes #45

### LINKED ISSUE #45: An empty BGZF block elsewhere in the file stream causes premature EOF.
Adding ^_\213^H^D^@^@^@^@^@377^F^@BC^B^@^[^@^C^@^@^@^@^@^@^@^@^@ elsewhere in a BAM file (other than the end) causes samtools view to end prematurely. This will either produce a decoding error, or if it happens to coincide with a genuine end of sequence record then it'll silently just truncate the output. This is visible if we add it as the first block after the SAM header.

Picard continues past this and ignores it as an end of file identifier unless it actually occurs at the end. The proposed amendment to the specification states "should by convention finish with a specific empty BGZF block". Note it does not say anything about the behaviour of empty blocks anywhere else, so we should not be interpreting them as a special indicator for EOF.

pd3:


daviesrob:
Thanks for this.  Merged as 2e0e1246a09ab5e74ddce23b780273ab34f14b67.

We've also added other commits to handle the multi-threaded case not covered here, and to spot missing `EOF` blocks on streamed BGZF data.  Hopefully `EOF` blocks should now work as expected.

PR #462, opened 2017-01-30
More thread pool fixes (see issue #537).

jkbonfield:
These come with some test code for thrashing the bgzf multi-threading,
but the tests are not normal tests and do not get executed by make
check.  Instead they are designed to be run manually to look for
lockups and to be executed under valgrind (memcheck, drd, helgrind
tools) to look for issues.

The bugs fixed include race conditions on both reader and writer
between shutting down the thread pool and starting the reader/writer
itself and races between shutting down and the reader/writer exiting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,11 @@ BUILT_TEST_PROGRAMS = \
 	test/test-regidx \
 	test/test_view \
 	test/test-vcf-api \
-	test/test-vcf-sweep
+	test/test-vcf-sweep \
+	test/thrash_threads1 \
+	test/thrash_threads2 \
+	test/thrash_threads3 \
+	test/thrash_threads4
 
 all: lib-static lib-shared $(BUILT_PROGRAMS) plugins $(BUILT_TEST_PROGRAMS)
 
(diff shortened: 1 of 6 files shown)


daviesrob:
I suspect this may be working around a reference counting problem in the thread pool.  When `hts_tpool_process_init` creates a queue, it sets `q->ref_count   = 1`.  This means the two ends of the queue have to synchronize before calling `hts_tpool_process_destroy` to prevent one or other of them from trying to access the queue after it has been torn down.  If the reference count was set to 2 for ordinary processes (1 for in_only) then the threads holding the ends of the queue could both call `hts_tpool_process_destroy` independently when they are not going to access it any more.  The queue should then be destroyed at the correct time without any need for independent synchronization.

We might also want to distinguish the case of shutting the queue down completely, and only shutting down the input side (so it can finish processing anything left on the queue).  This would be analogous to the idea of half closing a socket using `shutdown()`

jkbonfield:
Note that half of this PR is about startup and not shutdown.  I assume you're happy with that side.

However having got the additional state flags to deal with that case it's not too much extra code to use the same logic on shutdown side.

jkbonfield:
Rebased on top of latest develop bgzf fixes.

jkbonfield:
Rebased again and fixed raw gzip streams; seeking now fails.  Is this valid?
Technically we could support seeking on gzip streams built using SYNC_FULL_FLUSH for each and every block, but I think the only reason gzip support exists at all is because of naive "gzip foo.vcf" usage which won't be doing this.

jkbonfield:
Following the comments of @daviesrob regarding wanting a half-shutdown thread pool model instead, I have made a new PR to replace this (so closing this one).   It more complex and wide-changing than the previous PR, but arguably more useful longer term.

PR #463, opened 2017-01-30
Configure BZ2/LZMA and make htslib.pc more accurate

jmarshall:
This is an alternative to PR #331 that adds a little bit to _configure.ac_ rather than doubling the autoconf LOC by adding a couple of auxiliary m4 scripts.  More importantly it also brings the pkgconfig file up to date, including mentioning `-lbz2` and `-llzma` so that samtools etc have a chance to know how to link statically against libhts.  (I don't think James's PR addresses this, except by way of PR samtools/samtools#524 which just hard-codes them.)

Pkg-config believers can get this information with `pkg-config --variable static_libs htslib`, or more conservative folk can extract it with `sed`:

    sed -n '/^static_libs=/s/[^=]*=//p' htslib-uninstalled.pc

This could be added to _samtools/Makefile_ as above, or it might be possible to do it at configure time in _ax_with_htslib.m4_ operating on the just-created _htslib.pc.tmp_ (though this might have to be done with `AC_CONFIG_COMMANDS` or so, to have it delayed until _config.status_ is run).


jmarshall:
@daviesrob points out that there is no _bzip2.pc_ on e.g. Ubuntu, as it turns out the one I used when testing this was a downstream Fedora patch.  And therefore that maybe we should just add `-lbz2` to `Libs.private` rather than adding a dependency in `Requires.private`.

IMHO this is a demonstration of why [pkg-config is doomed to failure](https://twitter.com/jomarnz/status/826726790458597378) :smile:.  My inclination is to continue to use pkg-config according to its documentation and claim that the [bug is elsewhere](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=853803).

jmarshall:
Hmmm… I already have a similar extension to my existing PR's branch…

daviesrob:
Added some more commits here:

- Fix the bzip2 problem
- Create a small Makefile fragment for samtools/bcftools to use when static linking against the source tree
- Add some missing -l flags so that the list in the .pc file is exactly what is needed by htslib.

Trying to get _ax_with_htslib.m4_ to read the .pc file didn't go too well.  The problem is that if configure is configuring subdirectories as well (via AC_CONFIG_SUBDIRS) then the subdirectories are configured _after_ the root.  This means the htslib.pc.tmp gets written far too late in the process.  The easiest solution was to use a Makefile fragment where we have much better control over the order in which things happen.

I'll put in a couple of companion pull requests that make samtools and bcftools use this.  I've tested for various combinations of ./configure or not, and plugins or not, for both htslib, samtools and bcftools on Linux and MacOS.  It seems to do the right thing on all of them now.

daviesrob:
@jmarshall Hopefully this means I'm going along the right lines here...

jmarshall:
Indeed.  You know I've already done all this for htslib and samtools, right?

daviesrob:
Where?

jkbonfield:
Merged, but please also consider the change in the original bzip2 PR#331 of hard-coding bzip2 and lzma into the (non-autoconf) Makefile.

The CRAM spec lists bzip2 and lzma in it so we should be supporting it.  We don't want a build system that by default can detect they're absent and simply work around it offering reduced functionality without the user explicitly stating that as their aim.  That's already possible in the configure script, so we should direct people down that route if they want to do anything clever.

PR #468, opened 2017-02-08
Allow fai index to be in a different location to the indexed file; write bgzf index to hFILE

### LINKED ISSUE #375: add the ability to specify an index file to fai_load

daviesrob:
Update to #443.  Has been re-based, and extra functions are added to allow bgzf indexes to be written to an existing hFILE (see #410 ).

Summary of changes:
* bgzf_index_load, bgzf_index_dump, fai_load, fai_build, fai_read and fai_save to use hfile instead of stdio
* Add new API interfaces fai_build3() and fai_load3() which take separate names for the fai and gzi index files
* The download_and_open() function which made local copies of remote index files is removed.  .fai indexes are no longer cached locally; .gzi indexes were never cached
* fai_save() is removed as a public symbol in libhts.so - it never appeared in a public header file so was not part of the official API
* Add new API interfaces bgzf_index_load_hfile and bgzf_index_dump_hfile to allow bgzf indexes to be read from / written to an existing hFILE.
* Add test/test_bgzf.c with various unit tests for bgzf-related functions

Fixes #375

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ lib*.so.*
 /test/hfile
 /test/hts_endian
 /test/sam
+/test/test_bgzf
 /test/test-regidx
 /test/test-vcf-api
 /test/test-vcf-sweep
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,7 @@ BUILT_TEST_PROGRAMS = \
 	test/fieldarith \
 	test/hfile \
 	test/sam \
+	test/test_bgzf \
 	test/test-regidx \
 	test/test_view \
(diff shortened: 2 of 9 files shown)


daviesrob:
Merged as 84a89da6cf0eb828d39f9a85b82411d58c840e16, if anyone want to find it.

PR #479, opened 2017-03-04
Portability and docs

anderskaplan:
Fixed some minor issues: pointer offsetting with void pointers (which is not supported in standard C) and missing include directives. Also added some documentation for the thread pool test program.

### DIFF
--- a/test/test_view.c
+++ b/test/test_view.c
@@ -29,6 +29,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <unistd.h>
 #include <stdlib.h>
 #include <string.h>
+#include <getopt.h>
 
 #include "cram/cram.h"
 
--- a/cram/cram_structs.h
+++ b/cram/cram_structs.h
@@ -48,6 +48,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <pthread.h>
 #include <stdint.h>
+#include <sys/types.h>
 
 #include "htslib/thread_pool.h"
 #include "cram/string_alloc.h"
--- a/regidx.c
+++ b/regidx.c
@@ -147,7 +147,7 @@ int regidx_insert(regidx_t *idx, char *line)
     if ( idx->payload_size )
     {
         if ( m_prev < list->mregs ) list->payload = realloc(list->payload,idx->payload_size*list->mregs);
-        memcpy(list->payload + idx->payload_size*(list->nregs-1), idx->payload, idx->payload_size);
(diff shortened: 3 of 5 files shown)


test/test_view.c
```diff
@@ -29,6 +29,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <unistd.h>
 #include <stdlib.h>
 #include <string.h>
+#include <getopt.h>
```
jkbonfield: What was getopt.h required for? I think this is meant to be for the GNU getopt_long extension, which POSIX getopt coming from unistd.h.

Which OS were you compiling on and with which compiler?

jkbonfield:
Docs for thread pool tests look fine.  It's not really in a state for a proper regression test yet, but documenting it is a nice idea so I'm happy with that.

I was curious about the ```(char *)``` casts until I noticed the reason is to avoid arithmetic on ```void *```.  I'm kind of suprised most compilers let that one past.

I see there are many more portability complaints if we look harder. Eg gcc6 -std=gnu99 is fine, but -std=c99 has a *long* list of issues (constants undefined like M_LN10, use of alloca, etc).  I'm guessing we'd find lots of portabilities nasties if we try MSVC too.  Still this is all a movement in the right direction. :)

jkbonfield:
Rebased and merged, although I'm still curious to know which compiler/os combination doesn't have getopt in unistd.h.

test/test_view.c
anderskaplan: I'm working on a simple Windows (MSVC, VS2015) build, so not a regular posix system. I like the VS IDE and I'm curious if I can get it to work without any big changes to the code.

Including getopt.h is a bit of a kludge, but as I saw it done in another c file, I figured it would be an acceptable thing to do. It's not strictly needed though. I can find other ways to work around it.

jkbonfield: No problem.  Including extra header files isn't a major problem and as long as it helps one platform out there I have no issue with it at all.  If I had then I wouldn't have merged it as was.  It was simply my curiosity and to consider whether there are better ways for us to be doing automated testing on a wider range of platforms.

lh3: As I remember, an ancient Sanger machine did not have getopt.h that time. I could be wrong, though. That was like over 10 years ago when I was even not sitting in an office at Sanger. I would be cautious of having "getopt.h" included.

jkbonfield: Getopt.h is primarily for GNU long options.  We already use this in samtools and htslib applications (eg tabix), so this isn't an additional dependency.

Potentially we could add something like:

```
#ifdef __MSC_VER
#  include <getopt.h>
#endif
```

However it's not really worth the hassle.

PR #480, opened 2017-03-09
Reworked synced VCF/BCF reading

pd3:
The original version of allele matching was too simplistic and handled
records with duplicate positions by removing ("collapsing") them by
type - if requested so via the bcf_srs_t.collapse option. This behavior
is changed by this commit, records will never be discarded by the
reader. Instead, compatible records will be paired intelligently
based on user criteria. If duplicate positions are not desired, they
must be removed by the caller or by an external tool, such as `bcftools norm`.

Notes:

- the usage of the old "collapse" is discouraged, but backward-compatible
  at the API level, with behavior changed as described above

- bcf_sr_set_opt() call was introduced to avoid the need to access reader's
  internal structures directly. Ideally at some point it should became
  an opaque struct, to minimize the need for API/ABI breaking changes in
  future

- the allele pairing is still not perfect, it does not try to resolve
  ambiguous representations of identical alleles, such as multiallelic
  indels AA>A vs AAA>AA,AAA>A

- the new implementation slows the reader by 40-20%. This was measured
  with `bcftools isec`, which has no significant overhead: the original
  version took 113 seconds to process, the new reader took 157 seconds
  (two sites-only BCFs with 35M sites). In real life this never is a
  bottleneck, for example the difference in speed was not measurable when
  176 gBCFs with 64M sites are merged.

### DIFF
--- a/bcf_sr_sort.c
+++ b/bcf_sr_sort.c
@@ -1,0 +1,645 @@
+/* 
+    Copyright (C) 2017 Genome Research Ltd.
+
+    Author: Petr Danecek <pd3@sanger.ac.uk>
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
+    in the Software without restriction, including without limitation the rights
+    to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+    copies of the Software, and to permit persons to whom the Software is
+    furnished to do so, subject to the following conditions:
+    
+    The above copyright notice and this permission notice shall be included in
+    all copies or substantial portions of the Software.
+    
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+    THE SOFTWARE.
+*/
+
+#include "bcf_sr_sort.h"
+#include "htslib/khash_str2int.h"
+
+#define SR_REF   1
+#define SR_SNP   2
+#define SR_INDEL 4
+#define SR_OTHER 8
+#define SR_SCORE(srt,a,b) (srt)->score[((a)<<4)|(b)]
+
+// Resize a bit set.
+static inline kbitset_t *kbs_resize(kbitset_t *bs, size_t ni)
+{
+    if ( !bs ) return kbs_init(ni);
+    size_t n = (ni + KBS_ELTBITS-1) / KBS_ELTBITS;
+    if ( n==bs->n ) return bs;
+
+    bs = (kbitset_t *) realloc(bs, sizeof(kbitset_t) + n * sizeof(unsigned long));
+    if ( bs==NULL ) return NULL;
+    if ( n > bs->n )
+        memset(bs->b + bs->n, 0, (n - bs->n) * sizeof (unsigned long));
+    bs->n = n;
+    bs->b[n] = ~0UL;
+    return bs;
+}
+
+// Logical AND
+static inline int kbs_logical_and(kbitset_t *bs1, kbitset_t *bs2)
+{
+    // General case, bitsets of unequal size:
+    //  int i, n = bs1->n < bs2->n ? bs1->n : bs2->n;
+    int i, n = bs1->n;
+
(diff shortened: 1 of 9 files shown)


bcf_sr_sort.c
```diff
@@ -0,0 +1,645 @@
...
+            if ( ivar->type & jvar->type && multi_is_subset(ivar,jvar) ) return UINT32_MAX; // one of the alleles is identical
+
+            uint32_t score = SR_SCORE(srt,ivar->type,jvar->type);
+            if ( !score ) return 0;     // some of the varsets in the two groups are not compatible, will not pair
+            if ( min>score ) min = score;
+        }
+    }
+    if ( srt->pair & BCF_SR_PAIR_EXACT ) return 0;
+
+    assert( min!=UINT32_MAX );
```
daviesrob: Under what circumstances could this assert fire?  Would it ever be possible to get through the whole loop without setting `min`?

bcf_sr_sort.c
```diff
@@ -0,0 +1,645 @@
...
+    if ( !srt->str.l ) return strdup("");
+    int i;
+    hts_expand(char*,srt->noff,srt->mcharp,srt->charp);
+    for (i=0; i<srt->noff; i++)
+    {
+        srt->charp[i] = srt->str.s + srt->off[i];
+        if ( i>0 ) srt->charp[i][-1] = 0;
+    }
+    qsort(srt->charp, srt->noff, sizeof(*srt->charp), cmpstringp);
+    char *ret = (char*) malloc(srt->str.l + 1), *ptr = ret;
```
daviesrob: Unchecked `malloc`!
I would be very happy if everything that could allocate memory checked to see if the memory was actually allocated.  This will need to be fixed at some point.

bcf_sr_sort.c
pd3: This is a left-over from debugging. The loop changed in the meanwhile so it cannot happen any more.

bcf_sr_sort.c
pd3: Sure, I can add checks for that.

bcf_sr_sort.c
```diff
@@ -24,21 +24,30 @@
...
 #define SR_INDEL 4
 #define SR_OTHER 8
 #define SR_SCORE(srt,a,b) (srt)->score[((a)<<4)|(b)]
 
+static inline void *die_if_null(void *ptr)
+{
+    if ( ptr ) return ptr;
+    errno = ENOMEM;
+    fprintf(stderr, "[E::%s] %s\n", __func__,strerror(errno));
+    exit(1);
```
jkbonfield: I suspect this isn't really what Rob expected.

It's OK for an application to do this, but when using a library I'd never expect it to simply call ```exit``` on me.  Instead the functions calling malloc should return an error indicator (eg the traditional 0 vs -1 or pointer vs NULL) if they fail so the calling application can do whatever is appropriate.

Normally that will also be exit, but perhaps after shutting things down, tidying up temporary files and similar.

pd3: Well, I was following what's in `hts.c:hts_realloc_or_die`

daviesrob: Please don't copy that.  It's there because `hts_expand()` is basically otherwise unfixable (which also means we really shouldn't be using `hts_expand()`, but that's another issue).  See the comments on #442 to see what our users think of this sort of behaviour.

As @jkbonfield says, good libraries should return error codes, rather than calling exit or dumping core.

daviesrob:
Would you mind if I did some squashing here, mainly to tidy up the travis-related hiccup?

daviesrob:
Merging this leads to four test failures in bcftools' `make test`.  Is this expected?

```
test_vcf_isec:
        ./bcftools isec -n =2 -c both /tmp/SJT087M4u0/isec.a.vcf.gz/tmp/SJT087M4u0/isec.b.vcf.gz 2>/dev/null

        The outputs differ:
                ./test/isec.ab.both.out
                ./test/isec.ab.both.out.new
.. failed ...

test_vcf_isec:
        ./bcftools isec -Ob -n =2 -c both /tmp/SJT087M4u0/isec.a.vcf.gz/tmp/SJT087M4u0/isec.b.vcf.gz 2>/dev/null

        The outputs differ:
                ./test/isec.ab.both.out
                ./test/isec.ab.both.out.new
.. failed ...
test_vcf_isec:
        ./bcftools isec -n =2 -c any /tmp/SJT087M4u0/isec.a.vcf.gz/tmp/SJT087M4u0/isec.b.vcf.gz 2>/dev/null

        The outputs differ:
                ./test/isec.ab.any.out
                ./test/isec.ab.any.out.new
.. failed ...

test_vcf_isec:
        ./bcftools isec -Ob -n =2 -c any /tmp/SJT087M4u0/isec.a.vcf.gz/tmp/SJT087M4u0/isec.b.vcf.gz 2>/dev/null

        The outputs differ:
                ./test/isec.ab.any.out
                ./test/isec.ab.any.out.new
.. failed ...

diff ./test/isec.ab.both.out ./test/isec.ab.both.out.new
3a4
> 1	3106154	C	T,CT	11

diff ./test/isec.ab.any.out ./test/isec.ab.any.out.new
1a2
> 1	3062915	G	T	11
2a4
> 1	3106154	C	T,CT	11

```

pd3:
Yes, of course, squashing is OK, thank you. 

I also have a fix for the failing bcftools tests, it waits until the pull request is merged.

daviesrob:
First three commits squashed and merged as 6a44f49
Memory allocation checks will be added later.

PR #481, opened 2017-03-09
Make liblzma and libbz2 default when not using configure

daviesrob:
Change Makefile defaults so using `make` without `configure` turns on the liblzma and libbz2 options for CRAM.  NEWS is updated to report this change.

Change `configure` default for libcurl to `--enable-libcurl=check` as we want to prefer libcurl.   The check for libcrypto is changed slightly so that it's not fatal if it's missing and the default option `--enable-s3=check` was used.  This is so that running `./configure` doesn't unexpectedly fail if libcurl is present but libcrypto is not.

Add some information about dependencies to INSTALL, as we now need more libraries to be present by default.


jkbonfield:
Merged, although your PR had incorrect authorship of https://github.com/samtools/htslib/pull/331/commits/02d159cef1b8aab2b3b34391fe08e6ede272817b commit.

PR #484, opened 2017-03-12
Line endings

anderskaplan:
This pull request fixes two issues with line endings on Windows: the first one being that the test script compare_sam.pl doesn't handle them properly (despite what the perl documentation says). The second is that git converts line endings to CRLF by default on Windows machines on all text files -- including the reference fasta files -- which causes the fasta file and the fasta index file to become out of sync with each other. 

Also added some minor improvements to the test suite.

### DIFF
--- a/test/.gitattributes
+++ b/test/.gitattributes
@@ -1,0 +1,4 @@
+# Remove the text attribute from reference files, so that git doesn't convert
+# line separators on Windows machines. It causes the index files to become out
+# of sync with the fasta files.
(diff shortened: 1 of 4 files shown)


anderskaplan:
Never mind; I'll submit a more complete PR later.

jkbonfield:
Sorry we hadn't got around to replying yet.  I haven't had time to investigate this, but we are starting to look at how to support Windows better so your efforts are welcomed.

Thanks.

Issue #487, opened 2017-03-13
Changes required to compile on illumos (née OpenSolaris)


jblachly:
[illumos](https://www.illumos.org/) is a high performance server oriented operating system descended from Solaris. `htslib` currently does not build out of the box, but does work well with a few minor changes. In addition, at least item 1 (possibly also needing no. 2) below should allow building on Oracle Solaris.

1. Add check for requirement to link with libsocket (needed for both illumos and Solaris)
2. #undefine _XOPEN_SOURCE 700 in thread_pool.c (need for illumos, and probably Solaris if using gcc rather than SunSTudio?). The only reason this was added was to support broken linux distributions, which is mostly an older problem (but understandably some HPC clusters have old installations that cannot be upgraded by end users).  My fix allows both platforms to compile. 
3. Silence warnings about implicit definition of alloca() when not using a GNU libc (this is optional)

PR incoming. This is my first time really modifying GNU autoconf scripts, so feedback is welcomed.

daviesrob:
As I understand it, Illumos is the kernel and there are various distributions built on top of it.  Could you let me know which distribution you were using, please?

jblachly:
Almost -- illumos is the combination of the kernel plus the libc. I think of illumos as something in between linux (kernel only) and BSD (kernel, libc, and userland). The problems I am reporting have to do with the libc and so should be independent of distribution. However, you are right that distribution is important information; I did this work on Smartos, but confirmed that the issues exist, and are resolved by the PR, also on OpenIndiana. I have no reason to believe OmniOS does not behave precisely the same.

daviesrob:
Thanks.  I have an OpenIndiana VM running now.  I'll try playing with this once I've got it to talk to the internet.

daviesrob:
Fixed using part of #488, plus #485 and #498.  Please give the latest develop a try and let us know how you get on.

PR #488, opened 2017-03-13
Illumos

jblachly:
To fix issue #487 (build on illumos)
I broke the three fixes into three separate commits.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -182,6 +182,11 @@ is installed.
 FAILED.  This error must be resolved in order to build HTSlib successfully.])
 fi
 
+dnl connect() etc. fns are in libc on linux, but libsocket on illumos/Solaris
+libsocket=unneeded
+AC_SEARCH_LIBS(connect, socket, [libsocket=needed], [])
+
+
 if test "$enable_bz2" != no; then
   bz2_devel=ok
   AC_CHECK_HEADER([bzlib.h], [], [bz2_devel=missing], [;])
--- a/configure.ac
+++ b/configure.ac
@@ -186,6 +186,16 @@ dnl connect() etc. fns are in libc on linux, but libsocket on illumos/Solaris
 libsocket=unneeded
 AC_SEARCH_LIBS(connect, socket, [libsocket=needed], [])
 
+AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
+#include <pthreads.h>
+#ifndef PTHREAD_MUTEX_RECURSIVE
+# error macro not defined, probably a broken Linux, see https://github.com/samtools/htslib/pull/423
+#endif
+]])], [add_xopen="yes"], [add_xopen="no"])
+if test $add_xopen == yes; then
+  AC_DEFINE([ADD_XOPEN_700], 1,
+    [Define _XOPEN_SOURCE 700 for pthread on broken systems])
+fi
 
 if test "$enable_bz2" != no; then
   bz2_devel=ok
--- a/configure.ac
+++ b/configure.ac
@@ -134,6 +134,9 @@ esac
 AC_MSG_RESULT([$host_result])
 AC_SUBST([PLATFORM])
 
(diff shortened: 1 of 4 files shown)


jkbonfield:
The XOPEN_SOURCE thing makes me wonder which C version we really are targetting.  We've been fairly lax in stating this, but for sure there are a lot of use of c99 code that a strict c89 compiler would choke on.  Indeed adding -std=c89 to the compiler options reveals that not a single .o file is built (eg due to // comments, variables declared not at the start of a block, and so on).

Your comment indicates that _XOPEN_SOURCE=700 requires c99.  Maybe we should just define c99 as our base.  I did some tests with gcc -std=c99 -D_XOPEN_SOURCE=700 and the only hard error is alloca() usage (which IMO we shouldn't use for portability reasons anyway - it's always been a suspect function), and various non-standard things like strncasecmp.  I think it needs more investigations.

jkbonfield:
As per samtools I took a look at alloca usage and see it in a few places.

1. Hts region parsing to take a temporary copy of the chromosome name.  We could just use char reg[1024] and fail if someone attempts to give a chromosome name that long (seriously you'd deserve to get no support from the community if you attempt to be that daft!).  If we're kind we could malloc, but I don't see the point.

2. Vcf_parse.  This code is a total monster with up to 4 nested loops.  I can see why it has alloca therefore, but it's easier solved by keeping track of maximum size seen so far and a realloc.  A tad more work, but portable. 

3. bcf_unpack.  This is called from lots of places, possibly in loops?  I suspect it therefore needs to be efficient.  There doesn't look to be a place we can cache preallocated arrays although we could have a fixed size and malloc/free if and only if we need a larger one so it's fast for the usual case.  However looking at the code I wonder if it can be rewritten to avoid a temporary buffer by writing directly into ```d->allele[i]``` in one pass.

4. bcf_subset.  This could be a static sized buffer of MAX_N_FMT already defined to be 255.

Thanks for bringing this to our attention.  If you wish I can look at removing the alloca usage here and in samtools.

jblachly:
@jkbonfield Thanks for taking a look at these!  Replying here only instead of on both PRs.

I definitely think that newer standard compilation is the right solution given the codebase.  Regarding C99, there are some POSIX, but non ISO C99 usage in the codebase (for example, `M_SQRT2`). Thus compilation with `-std=c99` will fail (the reason it worked for you is that you had also defined _XOPEN_SOURCE >= 500 ; see math.h IIRC).  Compilation with `-std=gnu99` will work fine, OTOH. However, this may prevent using clang as a  compiler, since it does not recognize gnu99 standard. So, if you do update to -std=c99 you need to #define (globally) some other symbol like _XOPEN_SOURCE or one of the POSIX macros.  However, as you mentioned, with C99 standard (even with the POSIX macros defined) will throw warnings about implicit definitions of string functions.

Regarding alloca(), it compiles  on my system even without including `alloca.h` although the implicit glibc defn is slightly different (return type IIRC) than my libc. Certainly, you are welcomed to try to replace the usage of `alloca`, but unless we know there are platforms that absolutely can't use it, I don't know if it is worth your time? FWIW I included `alloca.h` on an Ubuntu16 system and compilation proceeded fine with -std=gnu99.

daviesrob:
Thanks for sending this.  I have now merged the first commit, to fix the libsocket issue.

We decided to implement the fixes in the other two commits in slightly different ways, both in #498:

- `_XOPEN_SOURCE` has been removed.  The Linux distributions that need this are old enough to have dropped off support, as far as I can tell.  If it still causes a problem for anyone then we can advise them on how to work around it.
- @jkbonfield removed all of the calls to `alloca()`

It was also necessary to merge #485 to fix a problem with BCF's use of NaNs.

The resulting combination now builds and passes all of its tests on a machine running the OpenIndiana flavour of Illumos, using the SunStudio C compiler (I think you were possibly using gcc?)

We will try to get some automated testing running on Illumos in the near future, so we can spot regressions....

PR #496, opened 2017-03-24
Windows portability

anderskaplan:
The purpose of this pull request is to make libhts and the test suite work better with some Windows quirks: the CRLF line endings and the distinction between binary and text files. 

Plus a few other minor improvements -- see the individual commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,9 @@
 *.o
 *.pico
+*.obj
 *.dSYM
 *.exe
 *.dll
-*.obj
 *.pc.tmp
 *-uninstalled.pc
 /version.h
@@ -49,6 +49,5 @@ lib*.so.*
 /test/thrash_threads[1-6]
 /test/*.tmp
 /test/*.tmp.*
-/test/*.new
 
 /TAGS
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 *.pico
 *.dSYM
 *.exe
+*.dll
+*.obj
 *.pc.tmp
 *-uninstalled.pc
 /version.h
@@ -47,5 +49,6 @@ lib*.so.*
 /test/thrash_threads[1-6]
 /test/*.tmp
 /test/*.tmp.*
+/test/*.new
 
 /TAGS
--- a/hfile.c
+++ b/hfile.c
@@ -573,6 +573,9 @@ error:
 
 hFILE *hdopen(int fd, const char *mode)
 {
+#if defined HAVE_SETMODE && defined O_BINARY
+    if (setmode(fd, O_BINARY) < 0) return NULL;
+#endif
     hFILE_fd *fp = (hFILE_fd*) hfile_init(sizeof (hFILE_fd), mode, blksize(fd));
     if (fp == NULL) return NULL;
 
@@ -594,9 +597,6 @@ static hFILE *hopen_fd_fileuri(const char *url, const char *mode)
 static hFILE *hopen_fd_stdinout(const char *mode)
 {
     int fd = (strchr(mode, 'r') != NULL)? STDIN_FILENO : STDOUT_FILENO;
-#if defined HAVE_SETMODE && defined O_BINARY
-    if (setmode(fd, O_BINARY) < 0) return NULL;
-#endif
     return hdopen(fd, mode);
 }
 
--- a/hfile.c
+++ b/hfile.c
@@ -573,9 +573,6 @@ error:
 
 hFILE *hdopen(int fd, const char *mode)
 {
-#if defined HAVE_SETMODE && defined O_BINARY
-    if (setmode(fd, O_BINARY) < 0) return NULL;
-#endif
     hFILE_fd *fp = (hFILE_fd*) hfile_init(sizeof (hFILE_fd), mode, blksize(fd));
     if (fp == NULL) return NULL;
 
@@ -597,6 +594,9 @@ static hFILE *hopen_fd_fileuri(const char *url, const char *mode)
 static hFILE *hopen_fd_stdinout(const char *mode)
 {
     int fd = (strchr(mode, 'r') != NULL)? STDIN_FILENO : STDOUT_FILENO;
+#if defined HAVE_SETMODE && defined O_BINARY
+    if (setmode(fd, O_BINARY) < 0) return NULL;
+#endif
     return hdopen(fd, mode);
 }
 
--- a/test/.gitattributes
+++ b/test/.gitattributes
@@ -1,0 +1,4 @@
+# Remove the text attribute from reference files, so that git doesn't convert
+# line separators on Windows machines. It causes the index files to become out
+# of sync with the fasta files.
+*.fa* -text
--- a/test/.gitattributes
+++ b/test/.gitattributes
@@ -1,4 +1,0 @@
-# Remove the text attribute from reference files, so that git doesn't convert
-# line separators on Windows machines. It causes the index files to become out
-# of sync with the fasta files.
-*.fa* -text
--- a/test/hfile.c
+++ b/test/hfile.c
@@ -61,16 +61,17 @@ char *slurp(const char *filename)
 {
     char *text;
     struct stat sbuf;
-    size_t filesize;
-    FILE *f = fopen(filename, "r");
-    if (f == NULL) fail("fopen(\"%s\", \"r\")", filename);
+    size_t filesize, readsize;
+    FILE *f = fopen(filename, "rb");
+    if (f == NULL) fail("fopen(\"%s\", \"rb\")", filename);
     if (fstat(fileno(f), &sbuf) != 0) fail("fstat(\"%s\")", filename);
     filesize = sbuf.st_size;
 
     text = (char *) malloc(filesize + 1);
     if (text == NULL) fail("malloc(text)");
 
-    if (fread(text, 1, filesize, f) != filesize) fail("fread");
+    readsize = fread(text, 1, filesize, f);
+    if (readsize != filesize) fail("fread");
     fclose(f);
 
     text[filesize] = '\0';
--- a/test/hfile.c
+++ b/test/hfile.c
@@ -70,8 +70,7 @@ char *slurp(const char *filename)
     text = (char *) malloc(filesize + 1);
     if (text == NULL) fail("malloc(text)");
 
-    readsize = fread(text, 1, filesize, f);
-    if (readsize != filesize) fail("fread");
+    if (fread(text, 1, filesize, f) != filesize) fail("fread");
     fclose(f);
 
     text[filesize] = '\0';
--- a/test/test_view.c
+++ b/test/test_view.c
@@ -35,6 +35,13 @@ DEALINGS IN THE SOFTWARE.  */
 
 #include "htslib/sam.h"
 
+enum test_op {
+    READ_COMPRESSED  = 1,
+    WRITE_COMPRESSED = 2,
+    READ_CRAM        = 4,
+    WRITE_CRAM       = 8
+};
+
 int main(int argc, char *argv[])
 {
     samFile *in;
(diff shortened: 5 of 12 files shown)


test/hfile.c
```diff
@@ -61,16 +61,17 @@ char *slurp(const char *filename)
...
+    FILE *f = fopen(filename, "rb");
+    if (f == NULL) fail("fopen(\"%s\", \"rb\")", filename);
     if (fstat(fileno(f), &sbuf) != 0) fail("fstat(\"%s\")", filename);
     filesize = sbuf.st_size;
 
     text = (char *) malloc(filesize + 1);
     if (text == NULL) fail("malloc(text)");
 
-    if (fread(text, 1, filesize, f) != filesize) fail("fread");
+    readsize = fread(text, 1, filesize, f);
```
jkbonfield: While I've no major objections to this style (use of a variable over inline checking), I don't particularly see the need for changing it and such non-changes make it harder to evaluate the PR.

test/test_view.c
```diff
@@ -54,23 +54,39 @@ int main(int argc, char *argv[])
 
     while ((c = getopt(argc, argv, "IbDCSl:t:i:o:N:BZ:@:")) >= 0) {
         switch (c) {
-        case 'S': flag |= 1; break;
+        case 'I': ignore_sam_err = 1; break;
```
jkbonfield: Adding usage is most welcomed, thanks, but as for prior comment, why have the options changed order?  It doesn't look to be for consistency with the usage statement.

hfile.c
```diff
@@ -573,6 +573,9 @@ static hFILE *hopen_fd(const char *filename, const char *mode)
 
 hFILE *hdopen(int fd, const char *mode)
 {
+#if defined HAVE_SETMODE && defined O_BINARY
+    if (setmode(fd, O_BINARY) < 0) return NULL;
+#endif
```
jmarshall: This is incorrect.  See commentary in #283.

.gitignore
```diff
@@ -2,6 +2,8 @@
 *.pico
 *.dSYM
 *.exe
+*.dll
+*.obj
```
jmarshall: _.gitignore_ already has appropriate patterns for `.dll`, no?  To date, Windows support has been aiming at Cygwin and GCC so `*.o` has sufficed, but it would be good to put `.obj` alongside `.o`/`.pico`.

.gitignore
```diff
@@ -47,5 +49,6 @@ lib*.so.*
 /test/thrash_threads[1-6]
 /test/*.tmp
 /test/*.tmp.*
+/test/*.new
```
jmarshall: `*.new` files are output files representing failed test cases and are not cleaned by `make clean`.  They need to be visible in `git status` so that you can see what tests have failed and can see them to remove them by hand.  Hence they are intentionally not in _.gitignore_.  (If that policy were to change, further changes would be necessary — perhaps just adding them to `make clean` too.)

test/.gitattributes
```diff
@@ -0,0 +1,4 @@
+# Remove the text attribute from reference files, so that git doesn't convert
+# line separators on Windows machines. It causes the index files to become out
+# of sync with the fasta files.
+*.fa* -text
```
jmarshall: Put this in the top-level _.gitattributes_ file.  But wouldn't it be better to regenerate the _.fai_ files instead?

jkbonfield: I think the problem here is that git clone itself changes these files.  We can't really regenerate the .fai files on a git clone, unless we have the Makefile *always* regenerate .fai from .fai.  Even this is problematic as there is no htslib command-line interface to fai_build: our only calls to it are over in ```samtools faidx``` instead.

jmarshall: Clearly by _regenerate_ I'm referring to having _test.pl_ generate the derived _.fai_ files (and removing them from source control), just as samtools's _test.pl_'s `test_index()` generates some _.bai_ indices used by later tests.  Additions to _test/test_view.c_ or a new _test/test_fai.c_ would indeed be needed to exercise `fai_build()`, but this would increase HTSlib's standalone test coverage so would obviously be a good thing…

daviesrob: The problem with regenerating them ourselves is that then we are testing our own output, which can lead to bugs cancelling each other out.  On the whole, I'd prefer to keep the .fai files in version control for now.  I agree that the change would be best in the top-level file.

test/test_view.c
anderskaplan: I changed the order to be consistent with the string passed to getopt, right above the switch statement. And for the usage instructions I aimed for what would make most sense for the user. Ideally they should all be consistent, but I didn't touch the getopt string because I'm not familiar with how it's used.

hfile.c
anderskaplan: Without this patch, the test cases in test_bgzf.c which use the function try_bgzf_dopen fail on Windows, because the files are opened in text mode, and then the file descriptors are passed into bgzf_dopen and hdopen.

I considered changing the test cases instead, but couldn't really see why it would be a bad thing to always set binary mode on the file descriptors. Are there any cases where it would be a problem?

test/hfile.c
anderskaplan: The reason for splitting the line in the first case was that I wanted to disable the check on Windows. But that wasn't necessary as long as the file is opened in binary mode. Reverting this one.

.gitignore
anderskaplan: There are "hfile_*.dll", "cyg*.dll", and "lib*.dll" patterns but nothing to catch pthreadVC2.dll, the pthreads library for Windows. I don't see why one would ever want to check in dll's in git, so I suggest that we keep the more generic *.dll pattern instead of the more specific ones.

.gitignore
anderskaplan: Ok, removing

test/.gitattributes
anderskaplan: Moving to top-level .gitattributes.

hfile.c
jkbonfield: @jmarshall, can you please clarify the intention here for binary files.  You state "The general policy in htslib is to open everything in binary mode", but that is precisely what this change does.  Do you mean that all opens should be "rb", "wb" or similar and specifying it at that point?  (Mostly we do that already of course.)  That only covers fopen style APIs though and not the Unix ```open``` function used in the test case.   The ```bgzf_dopen``` function should perhaps add "b" to the mode string before calling ```hdopen```, but this doesn't really seem to fit the "everything is binary" philosophy if we're only doing it for bgzf and not for SAM.

Stdin/stdout as you note are the exceptions which clearly need a ```setmode``` call to change, but again we're turning stdin/stdout to an hfile via the dopen call once more.

.gitignore
jmarshall: What is this _pthreadVC2.dll_ of which you speak? Why would it appear in the htslib source directory rather than some compiler directory? (If there's some larger pull request you're working towards, showing the draft end product would be illuminating…)

hfile.c
daviesrob: Relevant part from #283 (quoting @jmarshall):
>I would prefer not to call `setmode()` in `hdopen()` because:
>
 >  1. `hdopen()` can be called with file descriptors that are not files. In particular, on Windows sockets will not respond to `setmode()` particularly well;
>   2. file descriptors given to `hdopen()` may already have had I/O done on them, so it may by then be too late to call `setmode()`.


So we should fix test_bgzf.c, and note in the documentation for `hdopen()` that the file descriptor passed in must already be in binary mode.

jmarshall: [TL;DR: what Rob said.]

As noted in the [#283 commentary](https://github.com/samtools/htslib/pull/283#issuecomment-145835688) that I already pointed you both to, there are at least two problematic cases: `hdopen()` can be too late to set binary mode, and it is invalid to call `setmode()` on some types of file descriptor given to `hdopen()`.

If you are going to call `hdopen()` or `bgzf_dopen()` then it is your responsibility to have opened the file descriptor in binary mode.

The bug is in _test_bgzf.c_'s `try_bgzf_dopen()` which should be setting `O_BINARY` as `hfile_oflags()` does, and probably (being part of htslib so kosher to include _hfile_internal.h_) it should just call `hfile_oflags()` instead of duplicating that code.

jkbonfield: Good point, "open" vs "dopen" in binary mode; agreed fixing the original open is correct.

We should probably document this somewhere too in the doxygen comments before the hdopen function as it's currently intuitive to believe mode should contain "b" and handle it there.

anderskaplan:
Regarding your question above: no, I haven't gotten to samtools yet. The Windows build was a bit of a detour, as I planned to work on some bugs/feature request for a start, but it turned out to be an interesting one. At this point I have a working MSVC build of htslib which passes the test suite in my fork of the repo, it's the "msvc-build-2" branch if you want to take a look. But for some reason the compiled code is much slower than on linux, so still some investigation to do before I'm comfortable sharing it.

jkbonfield:
I've started looking into this, specifically as a starting point for Mingw64 support.  So I'll probably include your commits as part of a larger pull request.

Thanks.

anderskaplan:
ok, thanks for the update!

jkbonfield:
It's ongoing work, but see https://github.com/jkbonfield/htslib/tree/jkb_win, which includes a rebased and squashed version of your commits.

This also has an appveyor config, which is the Travis-like for windows.  An example build is seen here:

https://ci.appveyor.com/project/jkbonfield/htslib

I'm going to move over to samtools next.  Who knows what horrors that will bring! :-)
Once I'm happy with both I'll create PRs.

jkbonfield:
Both htslib and samtools are now passing their test harnesses on Windows with mingw/msys.  Some caveats:

1. You need ```make check TEST_OPTS="-t c:/msys64/tmp/_"``` as the default msys /tmp/... dir ends up in files of filenames and hence not being automatically modified by the msys command line hackery.  Ideally we'd use a raw mingw build, but we need the rest of msys style environment for tests.

2. The isatty checks in some samtools commands (eg samtools dict) to report usage when run on a terminal otherwise read from stdin if via a pipe or file, fail on msys.  This is a terminal bug I think.  It works when the executable is run under DOS and PowerShell.  It's a known issue with known fixes (https://fossies.org/linux/vim/src/iscygpty.c) but I am unsure if it's worth it.  I disabled those tests under Windows.

3. There are a lot of checks for msys "OS" in the perl test harness.  I don't know which of these are generic windows vs mingw vs msys, but for now it's enough to support one windows environment.  Similarly I'm not sure all the ```#ifdef _WIN32``` checks in the source are actually win32 issues or just mingw/msys.  Do they happen on other compilers?  Again - it's sufficient for now.

4. I had all sorts of nasty issues with curl and linking, but this is an ongoing investigation.

5. I'm not convinced adding drand48 (and co) to htslib is the correct solution, but it's a starting point.  Ideally we'd just stop depending on them!

Anyway, progress has been made and I'll investigate more on Tuesday when I get back from the long weekend. :-)

anderskaplan:
I worked around problem #1 above by wrapping the tempdir function as follows in the test scripts:

```
sub safe_tempdir
{
    my $dir = tempdir(CLEANUP=>1);
    if ($^O eq 'msys')
    {
        $dir = `cygpath -m $dir`;
        $dir =~ s/\015?\012//;
    }
    return $dir;
}
```

anderskaplan:
The super annoying #defines ERROR and MF_APPEND in Windows.h can be avoided by defining NOUSER
and NOGDI before including Windows.h.

anderskaplan:
For #3, #ifdef _WIN32 tests for platform and not compiler. It is defined also with MSVC.

jkbonfield:
Thanks for the suggestions.  Redefining tempdir is a nice idea.   Agreed $^O is a bit messy, but I don't know the right solution here.

The point about _WIN32 is I know it spots the platform rather than the compiler, but I am using it at least in part for compiler issues as well as platform ones.  It can be hard to tell which is which short of trying different compilers.   Mostly it's OK I think.

jkbonfield:
I forgot to add a fixes #commit comment to my new PR, but please see #531 which incorporates this PR (rebased) plus my own changes.

Thanks.

anderskaplan:
Replaced by another pull request, so closing this one.

Issue #403, opened 2016-08-08
Disabling stdout error messages


dancooke:
Would it be possible to make htslib's error reporting optional, perhaps in the form of a preprocessor macro? I'm writing an application which makes heavy use of htslib, and would like to maintain a consistent error reporting style, but this is not possible because htslib keeps polluting my own error reporting, e.g.:

    htsFile* file = sam_open(some_path, "r");
    [E::hts_open_format] fail to open file ....

This error message simply unnecessary; the caller is quite capable of checking the return value.

In cases where there is more than one failure condition (e.g. in `bam_hdr_read`), I think it's preferable to have an explicit method for diagnosing the problem:

    bam_hdr_t* header = sam_hdr_read(file);
    if (!header) {
        char* problem = diagnose_sam_hdr_error(file);
    }

These two approaches would not affect existing code that relies on htslib error reporting, but also afford new library users greater flexibility.

daviesrob:
Most (but sadly not all yet) of htslib's error reporting checks the global variable `hts_verbose` to see if the message should be printed.  For example, in `hts_open_format` we have:

```
error:
    if (hts_verbose >= 2)
        fprintf(stderr, "[E::%s] fail to open file '%s'\n", __func__, fn);
```

So setting `hts_verbose` to 1 or less should make this message go away.

jmarshall:
Those two in `hts_open()` et al in particular drive me mad.  You end up with messages like

    $ samtools view bork.bam
    [E::hts_open_format] fail to open file 'bork.bam'
    samtools view: failed to open "bork.bam" for reading: No such file or directory

and the htslib message is the **less** informative one.  I've had a change stashed for a while to change these to `if (hts_verbose >= 4)` so they don't appear with the default `hts_verbose` (3), but actually I'd rather verify that all `hts_open()` et al invocations in samtools/bcftools/htslib print their own messages and then nuke these two fprintfs within `hts_open()` et al.

anderskaplan:
Would you be interested in a contrib (pull request) that solves this issue? I'm looking for a good open source project to get involved in, and this could be a good starting point.

daviesrob:
@anderskaplan You're welcome to try, but I should warn you that it might not be the ideal thing to start on.  You would end up touching most of the files in htslib.  That tends to result in big pull requests that need lots of revision and may end up not being merged.  It would likely be best done in stages.

What it needs is:

[...]

PR #499, opened 2017-04-04
Logging mechanism

anderskaplan:
A first draft at a logging mechanism as described in https://github.com/samtools/htslib/issues/403.

- Converted the messages in "bgzf.c" as a pilot. Found two highly suspicious bracket placements when doing so; I took the liberty to correct them.
- Opted for a run-time setting rather than a compile-time setting because libhts can be installed as a system library.
- Added functions to get and set the log level, because it's not so clear with an "extern int" in a public API.
- Added a test script to check that the number of placeholders in the format string matches the number of parameters, because it's so easy to make mistakes when fiddling with the log messages.


daviesrob:
Thanks for this, it looks good.  I have a few suggestions:

- Variadic macros are C99 so we can use them, but `##__VA_ARGS__` is an extension, so not completely portable.  For example, the SunStudio compiler complains about:
```
"bgzf.c", line 399: warning: argument mismatch
```
for all the calls with no parameter.  It does still produce working code, though.  Absorbing `format` into `__VA_ARGS__` would fix this, at the expense of making the macro operation a bit more mysterious (possibly fixable via doxygen).

- It would be good to add at least `hts_log()`, and possibly `log_error()` etc. (with `hts_` prefix), to the external API.  This would allow tools that use htslib (notably samtools) to access the same API.

- We might want to make `HTS_LOG_INFO` greater than 3 (the default hts_verbose), so INFO messages do not normally appear.  Generally we only want to write if something bad happened, and otherwise stay quiet.  It might also be handy to have `HTS_LOG_TRACE` (= 8?) to catch the really low-level stuff (for example as in `hfile_gcs.c`).  This might result in `get_severity_tag()` becoming an if..else chain to deal with gaps in the list of enumerated values.

- (Noting the above) I suspect the `log_info()` calls in `bgzf.c` should really be `log_error()` as most of them seem to result in an exit failure.  Happily, this API makes it easier to spot such anomalies.

- If you add `HTS_FORMAT()` from `htslib/hts_defs.h` to the `hts_log()` declaration, it will add an attribute to make gcc and clang [check the format string at compile time](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-format-function-attribute).  You would want to do something like this:

```
void hts_log(enum htsLogLevel severity, const char *context, const char *format, ...) HTS_FORMAT(printf, 3, 4);
```

Then if a format string is wrong, you see a warning like this:
```
warning: format ‘%s’ expects a matching ‘char *’ argument [-Wformat]
```
It can do a more thorough job than your test-logging perl script as the compiler can check for correct data types as well as number of arguments.  If you don't have access to `gcc` to test this, then we could put it in later for you.

anderskaplan:
Thanks for the thorough review! I have fixed most of the issues, but there are a few things around the log levels that I'd like to bring up for discussion.

First, regarding this comment: "We might want to make HTS_LOG_INFO greater than 3 (the default hts_verbose), so INFO messages do not normally appear." I agree that the expected outcome is a good one, it's just that the way of doing it feels a bit backwards to me. I'd rather change the default value of hts_verbose to HTS_LOG_WARNING. Are there any good reasons for not doing that instead?

Second, regarding "It might also be handy to have HTS_LOG_TRACE (= 8?) to catch the really low-level stuff (for example as in hfile_gcs.c)." Sure, it's absolutely possible to add more levels, but I didn't see the benefit of adding more levels unless there is a clear need for them. It just gets more difficult (and arbitrary!) to assign the proper level for an event. Four levels is already quite a lot, and HTS_LOG_DEBUG is supposed to be "pulling out all the stops". I checked the cases in the code base where values of hts_verbose >= 4 are being used. They all look like debugging messages to me. (Except for the level 4 event in hfile.c, "initialization failed for plugin", which I'd rank as a level 2 warning instead.)

Finally, for "I suspect the log_info() calls in bgzf.c should really be log_error() as most of them seem to result in an exit failure." I agree, and I'll make that change in an update to follow.

daviesrob:
For better or worse, `hts_verbose` is a part of htslib's API.  Tools that use htslib, for example [samtools](https://github.com/samtools/samtools) already set it to specific values.   It's also exposed by wrappers like [pysam](https://github.com/pysam-developers/pysam).  As we don't want to unexpectedly change the behaviour of code that uses htslib, I think it would be best to keep the default where it is.

HTS_LOG_TRACE would mainly be for the Amazon S3 and Google cloud support, which will spew out lots of low-level protocol dumps if you make hts_verbose big enough.  This is only useful when debugging the protocol itself; for anything else it's just unnecessary noise.  As a comparison, [syslog](https://en.wikipedia.org/wiki/Syslog) has seven levels, and [log4j](https://en.wikipedia.org/wiki/Log4j) has six, so I don't think five would be too excessive.

As an aside, the pysam people will be very happy when we've managed to get rid of all the places that output to stderr without first checking `hts_verbose`.

bioinformed:
Very much so!

anderskaplan:
The default value of hts_verbose is now back to 3, but interpreted as HTS_LOG_WARNING rather than HTS_LOG_INFO. Also added a HTS_LOG_TRACE log level for the most detailed debugging events.

On closer inspection it turned out that the hts_log_info calls in bgzf.c always followed another hts_log_error event, so I changed them to the debug log level. I think that's closer to the intent.

I hope this first batch is ready to go in with those changes!

daviesrob:
Just a quick update.  We're trying to fix a load of bugs at the moment, but I promise this will be merged just as soon as that's done...

anderskaplan:
thanks for the update!

daviesrob:
I've added some commits to tidy up a few loose ends:

- Make dependency tracking work for htslib/hts_log.h
- Tweak a few log levels in bgzf.h
- Make hts_log() preserve errno, in case fprintf() decides to change it.

I'll have a go at merging it all tomorrow unless anyone objects.

daviesrob:
Now merged.  Thanks very much for this, it will be **very** useful.

anderskaplan:
great news! I'll get started on the next batch of logging calls then.

Issue #436, opened 2016-11-16
support for non-amazonaws endpoint urls (e.g. s3.<company>.com)


blajoie:
We are utilizing EMC on-premise s3-compatible storage.

Htslib does not allow non (.s3.amazonaws.com) endpoint urls (hard-coded for the time being)

We can successfully modify, re-compile and use htslib/samtools if we swap in our own endpoint url as below:

line 846 of hfile_libcurl.c
```
// Use virtual hosted-style access if possible, otherwise path-style.
    if (is_dns_compliant(bucket, path)) {
        kputsn(bucket, path - bucket, &url);
        kputs("<custom_endpoint_url>", &url);
    }
    else {
        kputs("<custom_endpoint_url>/", &url);
        kputsn(bucket, path - bucket, &url);
    }
    kputs(path, &url);
```

Can htslib be modified to accept custom endpoint urls from all potential config locations.

e.g.
1. embeded within link - s3://id:secret:endpoint@bucket/
2. env variable - AWS_ENDPOINT_URL=
3. within ~/.aws/config - aws_endpoint_url=

jmarshall:
Yes, we'll soon support `host_base` in [_.s3cfg_](http://s3tools.org/kb/item14.htm) and some equivalent in _.aws/config_.  I'm not really keen to invent new environment variables — are there other tools using such an environment variable?

At the moment you can select a profile to be used with `s3://profile@bucket/…`.  What we actually need is some documentation of how the S3 URLs are parsed and what configuration files are used…

blajoie:
No - that environment variable was a fictional creation on my part.  Sticking to a well known standard is best.  Or perhaps parsing additional information from the supplied s3 (default or named) profile?

I will play a bit with s3://profile@bucket/... syntax - thanks!   Happy to provide additional feedback and or some basic documentation regarding htslib / s3 interaction.

sb10:
I can see in samtools 1.4 there is a hfile_s3.c that seems to have all the necessary code for getting the correct domain from ~/.s3cfg, but is it actually supposed to be working now?

I tried installing by:

```
sudo apt-get install gcc make zlib1g-dev libbz2-dev liblzma-dev libcurl4-openssl-dev libssl-dev libncurses5-dev -y
wget "https://github.com/samtools/samtools/releases/download/1.4/samtools-1.4.tar.bz2"
tar -xvjf samtools-1.4.tar.bz2
rm samtools-1.4.tar.bz2
[...]

PR #506, opened 2017-04-08
non amazonaws endpoint support (+ force PATH_STYLE)

blajoie:
Hi,

Using the new v1.4 (current master)

It looks like when both ~/.s3cfg and ~/.aws credentials exist, then ~/.aws credentials will be used and ~/.s3cfg will be ignored.

We (@nhjohnson, @egor-dolzhenko) have a need to use the new host_base value extracted from the ~/.s3cfg (for non amazonaws endpoint), but we also do not want to force users to remove/disable ~/.aws credentials (this looks to be the only way to force htslib to always read (host_base) from ~/.s3cfg?)

This PR makes a few changes to hfile_s3.c.

always read from both ~/.s3cfg and ~/.aws/credentials, placing priority on ~/.aws for id/secret/token. This allows host_base to always be extracted and used. We are using a local S3 and specifying a non-amazonaws endpoint is required.

Following the logic outlined here (s3tools/s3cmd#416), we also have the need to always force PATH_STYLE over VIRTUAL_STYLE urls. Our s3 resource does not support VIRTUAL_STYLE urls. To achieve this, a user can set host_base and host_bucket to be equal, which then forces PATH_STYLE urls within htslib.

Any feedback is welcomed!

Thanks

### DIFF
--- a/hfile_s3.c
+++ b/hfile_s3.c
@@ -227,10 +227,15 @@ static hFILE * s3_rewrite(const char *s3url, const char *mode, va_list *argsp)
     kstring_t id = { 0, 0, NULL };
     kstring_t secret = { 0, 0, NULL };
     kstring_t host_base = { 0, 0, NULL };
+    kstring_t host_bucket = { 0, 0, NULL };
     kstring_t token = { 0, 0, NULL };
     kstring_t token_hdr = { 0, 0, NULL };
     kstring_t auth_hdr = { 0, 0, NULL };
-
+    kstring_t id_secondary = { 0, 0, NULL };
+    kstring_t secret_secondary = { 0, 0, NULL };
+    kstring_t token_secondary = { 0, 0, NULL };
+    int force_path_style = 0; // assume boolean
+    
     time_t now = time(NULL);
 #ifdef HAVE_GMTIME_R
     struct tm tm_buffer;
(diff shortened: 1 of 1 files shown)


jmarshall:
This is the wrong approach.  Mixing configuration settings from the two config files will rapidly become incomprehensible.  The correct approach is to give up waiting for aws-cli upstream and define an equivalent of _host_base_ in _~/.aws/credentials_ ourselves — this is #436.

OTOH a way to force path-style v. virtual-style is a good idea.

blajoie:
Agreed completely - Glad we done waiting for the upstream awscli fix :)

I would much rather define the equivalent of a host_base in the ~/.aws/credentials, perhaps even a flag within ~/.aws/credentials to force_path_style, url_mode=path/virtual?

Ideally host_base/url_mode would be set per profile, not global as it appears to be within ~/.s3cfg.

Thanks!

jmarshall:
Likewise agreed.  We've tried hard to use existing keys in _~/.aws/credentials_ and _~/.s3cfg_ (perhaps too hard in the case of endpoint_url!), so will look at what existing facilities to (ab)use for this path/virtual forcing.

HTSlib supports profiles in both these config files.  Just put the settings in a `[fooprofile]` section instead of `[default]` and specify the profile to use in one of the usual ways.  See recent comment on #436 for more details.

blajoie:
Closing this as the better solution is to instead add a new param to ~/.aws/credentials
https://github.com/samtools/htslib/issues/436

PR #514, opened 2017-04-11
Rans hardening

jkbonfield:
Fixes the identified RANS bugs plus some issues with embedded references.

### DIFF
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -2279,6 +2279,10 @@ int cram_decode_slice(cram_fd *fd, cram_container *c, cram_slice *s,
 	    s->ref = (char *)BLOCK_DATA(b);
 	    s->ref_start = s->hdr->ref_seq_start;
 	    s->ref_end   = s->hdr->ref_seq_start + s->hdr->ref_seq_span-1;
+	    if (s->ref_end - s->ref_start > b->uncomp_size) {
+		fprintf(stderr, "Embedded reference is too small.\n");
+		return -1;
+	    }
 	} else if (!fd->no_ref) {
 	    //// Avoid Java cramtools bug by loading entire reference seq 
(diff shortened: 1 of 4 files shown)


pauldwolfe:
Just encountered the check for embedded references being too small for slice, but struggling a bit to figure out how to deal with it elegantly. Under what circumstances can it occur? Is the workaround decode with the full reference fasta?

jkbonfield:
I'm not sure what you're trying to do.

Commit b1b2816 is to cope with fuzz-generated data where the input data may be invalid.  This particular scenario is for data aligned against an embedded reference but the supplied reference is too small.  Our fix is simply to reject this as an error, but this seems appropriate as the input is malformed.

I am not aware of any genuine circumstances where this situation can occur and hence I don't think there is any need for a more elegant solution.  Do you have a specific situation in mind?

pauldwolfe:
Thanks for quick reply. We have a cohort containing about 10k (human) BAMs and we're compressing them all to CRAM. Our process has a post-compression validation step which decompresses back to BAM then compares fidelity with the original BAM to assert the compression was lossless.

Our compression uses the HG37 ref genome, and embeds it:
`samtools view -h -O cram,embed_ref=1 -T [ref genome]
-o [output file] -@ $(nproc) [input file]`

Our decompression attempts to use the embedded genome (no supplied reference):
`samtools view -O bam -o [output BAM] -@ $(nproc) [input CRAM]`

We've seen this failure in about 1% of our CRAMs so far. I've been digging a bit more today and actually it seems to be undeterministic (ie I've seen the same BAM fail and pass the validation)

Could definitely be this is a problem in the compression operation itself, and this error is exposing the corruption. Just trying to figure out the root cause.

Thanks for any additional insight.

jkbonfield:
This is a new issue to the one above, so I've migrated it to its own issue here:

https://github.com/samtools/htslib/issues/1038

Issue #200, opened 2015-04-28
htsfile predicts "SAM version 1" for all text files


tseemann:
I'm not sure what peeking "htsfile" does but on a random Ruby file it says:

```
% htsfile baps.rb
baps.rb:        SAM version 1 sequence text

% head baps.rb
Class Baps < Formula
  homepage "http://www.helsinki.fi/bsg/software/BAPS/"
  # doi ""
 <snip>
```

jmarshall:
At present, `hts_detect_format()` reports pretty much [any text file as SAM](https://github.com/samtools/htslib/blob/develop/hts.c#L271-276).

Thanks for pointing out that in addition to the possibilities in that comment, it should say `unknown` for text that doesn't look tab-delimited.  (It needs to be conservative in saying `unknown` as this is used by `hts_open(filename, "r")` so it must not fail to recognise anything that can in fact be parsed.)

jkbonfield:
On Mon, Apr 27, 2015 at 07:27:18PM -0700, Torsten Seemann wrote:
> I'm not sure what peeking "htsfile" does but on a random Ruby file it says:
> 
> ```
> % htsfile baps.rb
> baps.rb:        SAM version 1 sequence text
> 
> % head baps.rb
> Class Baps < Formula

The unfortunately situation is that SAM has no magic number.  BAM and
CRAM can be identified by their headers.  SAM *may* start with
"@"-something, but it's not necessarily required. Indeed even the
totally blank file (zero length) is a valid SAM file!

I guess what it needs is something cleverer to tell apart SAM from
other arbitrary files.  Ie starting with @..\t or a tab delimited file
of at least 11 columns.

-- 
James Bonfield (jkb@sanger.ac.uk) | Hora aderat briligi. Nunc et Slythia Tova
                                  | Plurima gyrabant gymbolitare vabo;
  A Staden Package developer:     | Et Borogovorum mimzebant undique formae,
https://sf.net/projects/staden/   | Momiferique omnes exgrabure Rathi. 


-- 
 The Wellcome Trust Sanger Institute is operated by Genome Research 
 Limited, a charity registered in England with number 1021457 and a 
 company registered in England with number 2742969, whose registered 
 office is 215 Euston Road, London, NW1 2BE.

tseemann:
I guess there are multiple levels of evidence that point to being a SAM file
* .sam extension
* possible SAM header with ```@```
* possible alignment lines, tab sep 11 columns
[...]

PR #519, opened 2017-04-21
Prevent bgzf assert on reading empty VCF/BCF files

pd3:
Empty files are not recognised as VCF by hts_open and therefore bcf_hdr_read()
attempts to call bgzf_read() and the program fails with

    bcftools: bgzf.c:879: bgzf_read: Assertion `fp->is_write == 0' failed.

This probably a good place for the fix, fp->bgzf should not be accessed before
fp->is_bin is checked.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -645,6 +645,35 @@ int bcf_hdr_parse(bcf_hdr_t *hdr, char *htxt)
         needs_sync += bcf_hdr_add_hrec(hdr, hrec);
         p += len;
     }
+
+    // Detect incorrect header lines, print a warning, and skip them.
+    // Alternatively, the program could refuse to work with such VCF. However,
+    // many VCF operations do not really care about a few malformed header
+    // lines.
+    if ( strncmp("#CHROM\tPOS",p,10) )
+    {
+        char *ptr = p, tmp; 
+        while ( *ptr && *ptr!='\n' ) ptr++;
+        tmp  = *ptr;
+        *ptr = 0;
+        if (hts_verbose >= 2) fprintf(stderr, "[W::%s] Error parsing the header starting with: %s\n", __func__, p);
+        *ptr = tmp;
+
+        ptr = strstr(p, "#CHROM\tPOS");
+        if ( !ptr )
+        {
+            if (hts_verbose >= 1) fprintf(stderr,"[E::%s] Could not parse the header, sample line not found\n", __func__);
+            return -1;
+        }
+        while ( p < ptr )
+        {
+            hrec = bcf_hdr_parse_line(hdr,p,&len);
+            if ( !len ) break;
+            if ( hrec ) needs_sync += bcf_hdr_add_hrec(hdr, hrec);
+            p += len;
+        }
+        p = ptr;
+    }
     int ret = bcf_hdr_parse_sample_line(hdr,p);
     bcf_hdr_sync(hdr);
     bcf_hdr_check_sanity(hdr);
@@ -876,7 +905,7 @@ bcf_hdr_t *bcf_hdr_read(htsFile *hfp)
     if (!htxt) goto fail;
     if (bgzf_read(fp, htxt, hlen) != hlen) goto fail;
     htxt[hlen] = '\0'; // Ensure htxt is terminated
-    bcf_hdr_parse(h, htxt);  // FIXME: Does this return anything meaningful?
+    if ( bcf_hdr_parse(h, htxt) < 0 ) goto fail;
     free(htxt);
     return h;
  fail:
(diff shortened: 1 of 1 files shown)


pd3:
Apologies for putting two unrelated commits in the same pull request. I'd need to do some git magic to separate the two commits that were already pushed in two different branches, hope this is OK

daviesrob:
We can probably use cherry-pick to pull these apart if we need to.

On the first one, how do you get it to go wrong?  If I try to read empty files, I get:
```
Failed to open /dev/null: unknown file type
```

Also happens for empty (b)gzipped files.

pd3:
@daviesrob Try this:

    echo '' | bcftools +check-sparsity

jmarshall:
`echo '' | …` gives a text file with one empty line; either that or a truly-empty file are really empty text VCF files rather than BCF files.  VCF/BCF gets unlucky [here](https://github.com/samtools/htslib/blob/develop/hts.c#L299), and so `format.format` is not good for the current `vcf_hdr_read()` (this is samtools/htslib#200, for which I have a branch to dust off).

I would suggest you fix this with just

```diff
 bcf_hdr_t *bcf_hdr_read(htsFile *hfp)
 {
-    if (hfp->format.format == vcf)
+    if (hfp->format.format != bcf)
         return vcf_hdr_read(hfp);
```

as `vcf_hdr_read()` already emits a similar error message.

(I didn't look at the second unrelated commit.)

daviesrob:
Ah yes, I see.  The assertion is a bit of a red herring; compiling without optimisation gives this from Valgrind:

```
==43114== Invalid read of size 8
==43114==    at 0x4CB167: bgzf_read_block (bgzf.c:637)
==43114==    by 0x4CC6D4: bgzf_read (bgzf.c:884)
==43114==    by 0x4FFE70: bcf_hdr_read (vcf.c:848)
==43114==    by 0x635A502: init_data (check-sparsity.c:78)
==43114==    by 0x635B210: run (check-sparsity.c:264)
==43114==    by 0x49466F: main_plugin (vcfplugin.c:520)
==43114==    by 0x40F9BD: main (main.c:267)
```

So the problem is that it's trying to read a non-bgzf file as bgzf.

@jmarshall I tried your solution.  It works for the single newline case, but I got into an endless loop by doing this (unpatched bcftools dies in the same manner as the newline case):
```
tail -n +2 test/ex2.vcf | BCFTOOLS_PLUGINS=`pwd`/plugins ./bcftools +check-sparsity -
```
Which appears to be due to https://github.com/samtools/bcftools/blob/a468a83/plugins/check-sparsity.c#L185 not including an else clause.

Possibly we should do something like this?
```
     if (hfp->format.format == vcf)
         return vcf_hdr_read(hfp);
+    if (hfp->format.format != bcf) {
+        fprintf(stderr, "[E::%s] Input is not in bcf or vcf format\n", __func__);
+        return NULL;
+    }
 
     BGZF *fp = hfp->fp.bgzf;
     uint8_t magic[5];
```
The requirement that the file is identified as either vcf or bcf already exists if it's opened using `synced_bcf_reader`, so I don't think we'd be making anything unreadable by insisting on this.

jmarshall:
> It works for the single newline case, but I got into an endless loop by doing this (unpatched bcftools dies in the same manner as the newline case):
>
>     tail -n +2 test/ex2.vcf | BCFTOOLS_PLUGINS=`pwd`/plugins ./bcftools +check-sparsity -

IMHO that's mostly because `vcf_hdr_read()` accepts something that's not a valid VCF header as an acceptable header.  If “_[W::bcf_hdr_parse] The first line should be ##fileformat; is the VCF/BCF header broken?_” produced an error and a NULL return, this would have failed as expected.

In these cases, `format.format` is not `vcf` because `hts_detect_format` did not see `##fileformat=VCF` at the start of the file.  It's sort of assumed that the subsequent header readers will be at least as stringent as `hts_detect_format`…

daviesrob:
True, but I've tried to fix bcf error reporting before and it opened a **huge** can of worms.  In the interest of getting a fix in a finite length of time, I think it's easier to check the result we already have on what sort of file we're dealing with.

daviesrob:
Part one fixed via ae4f3df. On to part two...

daviesrob:
OK, I've now got a cherry-picked copy of part two, along with a couple of extra fixes to `bcf_hdr_parse_line()` that ensure it stops in the right place on incorrectly formatted lines.  I think this does more-or-less what you intended.  It's currently lurking in a branch at https://github.com/daviesrob/htslib/tree/vcf_hdr.

I'm not entirely convinced with the wisdom of trying to carry on when errors are detected, though.  In my experience, this just leads to inexplicable problems later on in the pipeline.  It's much better to fail early and get a responsible adult to fix the input than to try to guess what was intended and get it wrong...

How many files have to seen that exhibit the sort of trouble that this is trying to solve?

pd3:
In my experience we encounter broken third-party VCFs quite frequently. Usually there is a practical problem we need to solve, and we need a tool that proceeds anyway and gets the information we need despite some non-important aspects violating the specification. 
From my perspective it is better when the program is able to recover from errors rather than exiting immediately. Unless there is some switch or environment variable that controls the strictness.

pd3:
@daviesrob I checked your code and indeed like it much better than my original
+1

daviesrob:
@pd3 Revised version now in #526, if you'd like to take a quick look.

pd3:
+1 Looks great, thank you

daviesrob:
All now included between ae4f3df and #526.

Issue #400, opened 2016-07-18
compiler optimization: O3 causes segmentation fault, O2 works


tobiasrausch:
Hi,

I have a simple program that uses htslib which works fine for compiler optimization level O2 but fails with a segmentation fault for O3 using gcc version 5.4.0. It works fine using both optimization levels for an older gcc version 4.1.2. 

Is it true that O3 is not supported by htslib or is this a potential bug (valgrind error trace is below)?  I am happy to share the test bam and a tiny test program that reproduces the error if needed.

Thanks, Tobias

Valgrind output:
==13313==
==13313== Process terminating with default action of signal 11 (SIGSEGV)
==13313==  General Protection Fault
==13313==    at 0x4010F4: main (cov.cpp:26)
==13313==
==13313== FILE DESCRIPTORS: 4 open at exit.
==13313== Open file descriptor 3: a.bam
==13313==    at 0x574B790: __open_nocancel (syscall-template.S:84)
==13313==    by 0x4E4DE0A: hopen (in /usr/lib/x86_64-linux-gnu/libhts.so.1.2.1-2ubuntu1)
==13313==    by 0x4E5209E: hts_open (in /usr/lib/x86_64-linux-gnu/libhts.so.1.2.1-2ubuntu1)
==13313==    by 0x400E95: main (cov.cpp:10)
==13313==
==13313== Open file descriptor 2: /dev/pts/17
==13313==    <inherited from parent>
==13313==
==13313== Open file descriptor 1: /dev/pts/17
==13313==    <inherited from parent>
==13313==
==13313== Open file descriptor 0: /dev/pts/17
==13313==    <inherited from parent>
==13313==
==13313==
==13313== HEAP SUMMARY:
==13313==     in use at exit: 243,518 bytes in 94 blocks
==13313==   total heap usage: 135 allocs, 41 frees, 438,394 bytes allocated
==13313==
==13313== 304 (80 direct, 224 indirect) bytes in 1 blocks are definitely lost in loss record 12 of 22
==13313==    at 0x4C2FB55: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==13313==    by 0x4E54BA2: hts_itr_query (in /usr/lib/x86_64-linux-gnu/libhts.so.1.2.1-2ubuntu1)
==13313==    by 0x400F15: main (cov.cpp:17)
==13313==
[...]

PR #523, opened 2017-04-24
Fix l_qname overflow

### LINKED ISSUE #520: l_qname overflows for QNAMEs of 252 or more characters
5d114ebd8e9b80622769b8e575c5a9359cd51273 added extra NULs in memory to ensure that CIGAR data is 4-byte aligned.  To avoid breaking bad code that does not use the accessor macros, these extra bytes were included in in-memory `l_qname`.  For QNAMES of length ≥252 characters, this means that `l_qname` needs to have the value 256, but it is a `uint8_t`.

There are various ways of fixing this, though it's less straightforward to do so while minimising binary compatibility implications.

This is the cause of pysam-developers/pysam#447.

jmarshall:
Fixes #520, while maintaining binary compatibility as best one can.  Code compiled against htslib 1.4 will continue to work when running against a libhts.so containing this fix, but will still fail for ≥252 QNAMEs.  Such code would need to be recompiled to benefit from the fix.

### DIFF
--- a/cram/cram_samtools.c
+++ b/cram/cram_samtools.c
@@ -94,7 +94,7 @@ int bam_construct_seq(bam_seq_t **bp, size_t extra_len,
     b->core.bin     = bam_reg2bin(pos-1, end);
     b->core.qual    = mapq;
     b->core.l_qname = qname_len+qname_nuls;
-    b->core.l_extranul = qname_nuls-1;
+    b->core.l_qname_old = (uint8_t) b->core.l_qname;
     b->core.flag    = flag;
     b->core.n_cigar = ncigar;
     b->core.l_qseq  = len;
@@ -105,8 +105,9 @@ int bam_construct_seq(bam_seq_t **bp, size_t extra_len,
     cp = b->data;
 
     strncpy((char *)cp, qname, qname_len);
-    for (i = 0; i < qname_nuls; i++)
+    for (i = 0; i < qname_nuls-1; i++)
 	cp[qname_len+i] = '\0';
+    cp[qname_len+qname_nuls-1] = qname_nuls-1;
     cp += qname_len+qname_nuls;
     if (ncigar > 0) memcpy(cp, cigar, ncigar*4);
     cp += ncigar*4;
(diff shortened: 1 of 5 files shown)


daviesrob:
That's an impressive fix.

One possible problem is that it removes `core.l_extranul`.  It looks like pysam may fiddle with this field, see https://github.com/pysam-developers/pysam/blob/3c627ff/pysam/libcalignedsegment.pyx#L893.  Do we know of anything else that might?

Would it be easier to just say we don't support names > 252 bytes long, until we get the chance to fix it properly?

AndreasHeger:
Many thanks! I added the l_extranul code so that a record that is created programmatically is binary identical to a record that has been read through the htslib API. More than happy to remove it. 
There is a length check in pysam already, so this could simply be reduced to 252.

daviesrob:
I've been looking into this a bit more, and concluded there's no easy fix.

Long read names seem to come from [Mitty](https://github.com/sbg/Mitty), which assumes it can use all 254 available bytes if it needs to.

1. This patch works, but isn't completely backwards-compatible for code that uses htslib:

- `l_extranul` disappears, so external code that uses it will need to be altered
- As the value stored in `l_extranul` is no longer present, code compiled against the old version will fail when trying to access it.  For example, `bam_name_len()` will not work until recompiled.

I've also tried a couple of other solutions:

2. Reject reads with names > 251 bytes long (see https://github.com/daviesrob/htslib/tree/l_qname).
This may well break Mitty, but most other data should still work, and at least it tells you that there's a problem.

3. Don't add the extra NULs if the read is too long (see https://github.com/daviesrob/htslib/tree/l_qname2).
This mostly works, but #400 comes back albeit only for data that has very long read names.  It's otherwise fine as long as you're on intel and haven't used `gcc -O3`.

So we just have to decide what level of breakage we can put up with.  Option 2 is the most conservative, but Mitty stops working.  We could tell Mitty users to go for version 1.3.1 for now, but as that has completely unaligned access, it might be better to go for option 3 instead where unaligned CIGARs are at least confined to the long name case.

Opinions...?

daviesrob:
I've pushed 079c7b81 as a temporary fix to catch the names we can't currently handle.  I'll keep this open for when we fix it properly.

AndreasHeger:
Thanks, I will follow your lead and disallow query names longer than 252 characters as well.

PR #526, opened 2017-05-04
Improved VCF/BCF header parsing

daviesrob:
Second commit from #519, rebased with extra commits to fix a couple of issues.

- Make `bcf_hdr_parse()` try to include more header records when it encounters a read error.
- Add checks on the return value of `bcf_hdr_parse()`.
- Ensure data structures are cleaned up if `bcf_hdr_parse()` fails.
- Stop `bcf_hdr_parse_line()` from going on to the next line when it doesn't find '='.
- Ensure `bcf_hdr_parse_line()` sets `*len` to the correct number of bytes to get to the next line.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -296,7 +296,7 @@ bcf_hrec_t *bcf_hdr_parse_line(const bcf_hdr_t *h, const char *line, int *len)
     p += 2;
 
     const char *q = p;
-    while ( *q && *q!='=' ) q++;
+    while ( *q && *q!='=' && *q != '\n' ) q++;
     int n = q-p;
     if ( *q!='=' || !n ) { *len = q-line+1; return NULL; } // wrong format
 
@@ -337,10 +337,8 @@ bcf_hrec_t *bcf_hdr_parse_line(const bcf_hdr_t *h, const char *line, int *len)
         {
             // wrong format
             while ( *q && *q!='\n' ) q++;
-            kstring_t tmp = {0,0,0};
-            kputsn(line,q-line,&tmp);
-            fprintf(stderr,"Could not parse the header line: \"%s\"\n", tmp.s);
-            free(tmp.s);
+            fprintf(stderr,"Could not parse the header line: \"%.*s\"\n",
+                    (int) (q - line), line);
             *len = q - line + (*q ? 1 : 0);
             bcf_hrec_destroy(hrec);
             return NULL;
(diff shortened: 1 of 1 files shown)


vcf.c
```diff
@@ -2774,7 +2808,7 @@ bcf_hdr_t *bcf_hdr_merge(bcf_hdr_t *dst, const bcf_hdr_t *src)
         dst = bcf_hdr_init("r");
         kstring_t htxt = {0,0,0};
         bcf_hdr_format(src, 0, &htxt);
-        bcf_hdr_parse(dst, htxt.s);
+        if ( bcf_hdr_parse(dst, htxt.s) < 0 ) dst = NULL;
```
jkbonfield: This leaks memory.  It should have a ```bcf_hdr_destroy``` call in there before ```dst = NULL```.

vcf.c
```diff
@@ -645,6 +650,35 @@ int bcf_hdr_parse(bcf_hdr_t *hdr, char *htxt)
         needs_sync += bcf_hdr_add_hrec(hdr, hrec);
         p += len;
     }
+
+    // Detect incorrect header lines, print a warning, and skip them.
+    // Alternatively, the program could refuse to work with such VCF. However,
+    // many VCF operations do not really care about a few malformed header
+    // lines.
+    if ( strncmp("#CHROM\tPOS",p,10) )
```
jkbonfield: This and the one below need to check for ```\n#CHROM\tPOS```.  By adding a line starting with ##CHROM...  I managed to core dump the parser:

```
#0  0x00007f41df84cc37 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007f41df850028 in __GI_abort () at abort.c:89
#2  0x000000000041dfa0 in bcf_hdr_add_sample (h=0x2441130, s=0x2441a70 "") at vcf.c:98
#3  0x000000000041e151 in bcf_hdr_parse_sample_line (h=h@entry=0x2441130, 
    str=str@entry=0x24413da "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT\tX\ty\t\n##FORMAT=<ID=S,Number=1,Type=String,Description=\"Text\">\n#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT\tS1\tS2\tS3\n")
    at vcf.c:136
#4  0x0000000000427b38 in bcf_hdr_parse (hdr=hdr@entry=0x2441130, htxt=<optimized out>) at vcf.c:682
#5  0x000000000042820a in vcf_hdr_read (fp=0x24410a0) at vcf.c:1591
#6  0x0000000000428685 in bcf_hdr_read (hfp=<optimized out>) at vcf.c:872
#7  0x0000000000403e95 in view_vcf (in=0x24410a0, filename=0x7fff91c7c2ef "_.vcf") at htsfile.c:118
#8  0x0000000000404437 in main (argc=3, argv=0x7fff91c7a6d8) at htsfile.c:246
```

daviesrob: This is actually due to  `bcf_hdr_add_sample()` punishing invalid input by calling `abort()`.  Your `##CHROM` line included a trailing tab, which caused `bcf_hdr_add_sample()` to print:

```
[E::bcf_hdr_add_sample] Empty sample name: trailing spaces/tabs in the header line?
```
and then abort().

I think it should be fixable to return non-zero instead without getting in to too deep a hole.

daviesrob: ... Or maybe not.  bcftools calls `bcf_hdr_add_sample()` in loads of places without checking the return value.  I'll leave this one for later.

jkbonfield:
Changes look OK to me.  @pd3 can you please check and merge the PR if you're happy with it too?

PR #531, opened 2017-05-08
Windows compilation.

jkbonfield:
Implemented building on Windows with Mingw64 & msys.

Also contains an appveyor configuration for Windows CI. See https://ci.appveyor.com/project/jkbonfield/htslib/build/vers.31 for an example of this.


jkbonfield:
Based on investigations to https://github.com/samtools/bcftools/issues/610 I've realised I accidentally set the wrong define in the configure.ac - it should be _XOPEN_SOURCE=600.  Brain fade!

I'll postpone pushing that update until I get more checking done on other OSes, specifically an ancient linux (if I can get any of them to work at all!).

cram/cram_structs.h
```diff
@@ -192,7 +192,7 @@ typedef struct cram_file_def {
 struct cram_slice;
 
 enum cram_block_method {
-    ERROR    = -1,
```
anderskaplan: This change and its parallel MF_APPEND are caused by a clash with Windows.h. The workaround used here certainly works, but I'd recommend a less intrusive one instead: define NOUSER and NOGDI when compiling libhts. That shouldn't interfere with anything on any other platforms.

anderskaplan: I just noted that the affected header files aren't part of the public API. So perhaps less of an issue than I originally thought.

jkbonfield: Indeed.  I figured it's probably a good move anyway to minimally name-space that ERROR enum as the potential for clashes on other systems will be reduced.

htslib/hts_defs.h
```diff
@@ -69,6 +69,18 @@ DEALINGS IN THE SOFTWARE.  */
 #define HTS_DEPRECATED(message)
 #endif
 
+// On mingw the "printf" format type doesn't work.  It needs "gnu_printf"
+// in order to check %lld and %z, otherwise it defaults to checking against
```
anderskaplan: Please note that %lld should be avoided anyway, use the standard & portable PRIxx constants instead.

jkbonfield: Agreed, where we have data types that are int32_t or int64_t we should (and do) use PRId64 etc.  The HTS_FORMAT macro though is generic and we don't have any way to enforce programs that use htslib stick to only int64_t and not "long long", so the change is simply to make it as robust as it can be.

We probably ought to fix such cases in our own code though - eg samtools/bam_stat.c has ```long long``` all over the place and hence (correct, given the type) ```%lld``` too.  It should use uint64_t for most of those.

Makefile
```diff
@@ -138,6 +138,7 @@ print-version:
 
 
 LIBHTS_OBJS = \
+	hts_os.o\
```
jmarshall: Other entries here are clearly sorted by klib, then alphabetically, then subdirectories.

Makefile
```diff
@@ -210,6 +211,8 @@ config.h:
 	echo '/* Default config.h generated by Makefile */' > $@
 	echo '#define HAVE_LIBBZ2 1' >> $@
 	echo '#define HAVE_LIBLZMA 1' >> $@
+	echo '#define HAVE_FSEEKO 1' >> $@
+	echo '#define HAVE_DRAND48 1' >> $@
```
jmarshall: Rather than adding infrastructure for these functions, wouldn't it be better to see where they are used: fseeko (it isn't); drand48 (just in ks_shuffle, which could use native alternatives)?

bgzip.c
```diff
@@ -198,6 +203,9 @@ int main(int argc, char **argv)
 
         if ( index ) bgzf_index_build_init(fp);
         buffer = malloc(WINDOW_SIZE);
+#ifdef _WIN32
+        _setmode(f_src, O_BINARY);
+#endif
```
jmarshall: If the decision here is to have bgzip open everything as binary (cf [second paragraph of this comment](https://github.com/samtools/htslib/pull/283#issuecomment-164023559)), wouldn't it be better to just use `hopen()` rather than sprinkling _WIN32 platform crap all over the code?

hfile.c
```diff
@@ -849,7 +866,8 @@ static const struct hFILE_scheme_handler *find_scheme_handler(const char *s)
         else if (s[i] == ':') break;
         else return NULL;
 
-    if (i == 0 || i >= sizeof scheme) return NULL;
+    // 1 byte schemes are likely windows C:/foo pathnames
```
jmarshall: s/byte/character/

hfile.c
```diff
@@ -506,12 +518,13 @@ static off_t fd_seek(hFILE *fpv, off_t offset, int whence)
...
-    int ret;
+    int ret = 0;
     do {
 #ifdef HAVE_FDATASYNC
+        hFILE_fd *fp = (hFILE_fd *) fpv;
         ret = fdatasync(fp->fd);
-#else
+#elif defined(HAVE_FSYNC)
+        hFILE_fd *fp = (hFILE_fd *) fpv;
         ret = fsync(fp->fd);
```
jmarshall: Less invasive to add a `#elif defined HAVE_FLUSHFILEBUFFERS`-based version (or add a `FlushFileBuffers()`-based `fsync` function elsewhere for the existing `fd_flush` code to use).

hfile.c
```diff
@@ -585,7 +598,11 @@ hFILE *hdopen(int fd, const char *mode)
 static hFILE *hopen_fd_fileuri(const char *url, const char *mode)
 {
     if (strncmp(url, "file://localhost/", 17) == 0) url += 16;
+#ifdef _WIN32
+    else if (strncmp(url, "file:///", 8) == 0) url += 8;
+#else
```
jmarshall: Surely a little obscure, and doesn't handle `file://localhost/C:/foo/bar.txt`.  Clearer alternative would be adding something like this below:

    #ifdef _WIN32
    if (url[0] == '/' && url[2] == ':' && url[3] == '/') url += 1;
    #endif

Makefile
jkbonfield: fseeko *is* used and drand48 is used by more than ks_shuffle.  Possibly we could rewrite the code that uses both, but as these are POSIX functions the intent is simply to put OS-specific deficiencies in one single place so we can maintain our position of desiring a POSIX compatible compilation.  It also makes the PR much simpler, and safer, as it doesn't contain a lot of rewriting of code just to make it run on Windows.

bgzip.c
jkbonfield: I thought we decided that hopen musn't do the ```O_BINARY``` setmode thing itself, indeed I recall you squashing a PR on that very reason.  Plus again we'd be changing bgzip on Unix in order to make it run on Windows, which comes with potential bugs and risks.  I'm happier with a windows-only change personally.

The change was made for a reason (it fixes bugs).  If your view is that bgzip must NOT open everything as binary then we need to change the tool further still, adding a command line argument to tell it whether or not to treat stdin as text or binary.  Yet more complications.

hfile.c
jkbonfield: It's char*, not wchar*, so byte vs character is splitting hairs here!

Makefile
jkbonfield: Agreed.

hfile.c
jkbonfield: I don't get what you mean.  If we define ```HAVE_FLUSHFILEBUFFERS``` then we need our own custom autoconf code to generate that define, instead of using the easy existing macros for ```HAVE_```func style checking, which sounds more invasive rather than less.

Could you give an example of what you mean plase?

hfile.c
jkbonfield: Agreed.

Makefile
jmarshall: The only place that needs to use `fseeko()` is `zfseeko()`, which is itself unused.  Moreover for _zfio.c_ in general: if someone is in a tidying mood, see #552.

jmarshall: The one in `zfopen()` could be changed as is to `fseek` or `rewind`, hence I wrote “that **needs** to use”.  James, there's no need to find something to disagree with in everything I say! 😛

hfile.c
jmarshall: I mean code like jmarshall/htslib@39146e27e146eba21d9243e8151682d0c33ce91d to ensure `fsync()` is always available (but probably moved to _hts_os.h_ or so).

Alternatively that code as a function in _hfile.c_ or just inline in `fd_flush()`, likely controlled by an explicit platform `HAVE_` macro like [these ones](https://github.com/samtools/htslib/blob/dc45f188e5c0bd296b9087ad6119722ebc48aca4/hfile.c#L457-L465).

hfile.c
jmarshall: It's a string provided by the user.  A readable string, not binary data.

bgzip.c
jmarshall: Reread my comment.  It's talking about changing _bgzip.c_ to use `hopen` instead of `open`.  It does not propose altering the implementation of `hopen` itself.

I changed bgzip's BGZF files from `open` to (effectively) `hopen` in b8204c12e1a474cf4b3ed845e0323f71a12c0cda during the PR you mention, which is linked in my previous comment.  I would have rather liked to change its other files too as it simplifies and regularises the code, but I had not analysed the effects that would have on Windows, as I noted at the time.

> The change was made for a reason (it fixes bugs).

Apparently you have now analysed the effects, though that wasn't apparent from the commit messages.  Grand; and I guess it gives `bgzip -i` a chance to have the right offsets.

So the maintenance question is about the risks of a change that affects all platforms while regularising and simplifying the code, versus one that adds the maintenance burden of `#ifdef` spaghetti.

Makefile
jkbonfield: I'm not, but please say what you mean then.  You started by saying it's unused, then by saying it's used by code which isn't used, and now by saying it's used in code that could be rewritten.  The conversation would be shorter if you just said to start with that it could be avoided by replacing zfio with calls to bgzf - which I fully agree with.

However this just wasted all our time instead.

jmarshall:
Incidentally, from the commit message:

> \- Fixed various shared library types in the Makefile.  The existing cygwin code there is unmodified, but also untested.

The Cygwin code was of course tested when it was added (in 2cddfb3c75887ca02d58708e5519c6d62992634c).

hfile.c
jkbonfield: A windows implementation of fsync makes sense, but it probably ought to be surrounded by a ```#ifndef HAVE_FSYNC``` check and appropriate autoconf magic.  Indeed hts_os.h sounds like the right spot for it too.

I'm not such a fan of ```#define HAVE_STRUCT_STAT_ST_BLKSIZE``` appearing mid-way through a file as the use of it can lead people to wrong conclusions (that it's in a header file somewhere and can be used elsewhere), given the uses of these are several pages below their defines.

jkbonfield:
Ok point taken, but untested by me then since these changes.  I could easily have broken cygwin in this process, but haven't tested that.

Makefile
jmarshall: IMHO I was perfectly clear for people willing to read between the lines at all; i.e., approaching the conversation with a view of ”what is the reason behind that particular statement” rather than “that statement is wrong”.  Anyway: PR #553 fixes a minor bug too. Win/win.

configure.ac
```diff
@@ -125,6 +126,15 @@ case $host_alias in
...
     PLUGIN_EXT=.bundle
     ;;
+  *-msys* | *-MSYS* | *-mingw* | *-MINGW*)
+    host_result="MSYS dll"
+    PLATFORM=MSYS
+    PLUGIN_EXT=.dll
+    # This also sets __USE_MINGW_ANSI_STDIO which in turn makes PRId64,
+    # %lld and %z printf formats work.  It also enforces the snprintf to
+    # be C99 compliant so it returns the correct values (in kstring.c).
+    CPPFLAGS="$CPPCFLAGS -D_POSIX_C_SOURCE=600"
```
valeriuo: Please change `-D_POSIX_C_SOURCE=600` to `-D_XOPEN_SOURCE=600`.

jkbonfield:
In lieu of having a local working windows box of my own at the moment, I prodded AppVeyor on this latest commit and sadly it fails everywhere: https://ci.appveyor.com/project/jkbonfield/htslib/build/vers.44

I'm not sure what caused it, whether it was something that has moved on in develop since the PR was initially made or whether it was a requirement on HAVE_FSEEKO being defined (there are lots of related warnings) I am unsure.

It may be best to just use the original commit (https://github.com/samtools/htslib/pull/531/commits/ace007f9d1bf11243085c6363f5ce3be24183b04) and apply the changes one by one to see what broke it.   It's hard to tell without being in front of an actual windows setup.

jkbonfield:
Squashed two most recent commits, rebased on fresh develop and merged.

PR #543, opened 2017-05-20
Logging improvements 2

anderskaplan:
Replaced all the fprintf(stderr, ...) calls that already made use of the hts_verbose setting into hts_log_... calls.

This is the second pull request in a series that aims to solve https://github.com/samtools/htslib/issues/403 .


daviesrob:
Thanks for this.  We will have to see how we get on with enforcing capital letters at the start of the messages, but I guess it's easy enough to remove the test if it proves too annoying.

Issue #540, opened 2017-05-19
Out-of-bounds read in bcf_fmt_array


cryptoad:
There is another out-of-bounds reachable via `vcf_format`, when importing (and displaying) the attached file via `view_vcf`:
```
==227334==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6030000017d0 at pc 0x55ac4df16a62 bp 0x7ffec6eb1b00 sp 0x7ffec6eb1af8
READ of size 1 at 0x6030000017d0 thread T0
    #0 0x55ac4df16a61 in le_to_i8 third_party/htslib/htslib_1_4_1/htslib/hts_endian.h:223:12
    #1 0x55ac4df162b3 in bcf_fmt_array third_party/htslib/htslib_1_4_1/vcf.c:1842:32
    #2 0x55ac4df2860c in vcf_format third_party/htslib/htslib_1_4_1/vcf.c:2551:18
    #3 0x55ac4df13129 in vcf_write third_party/htslib/htslib_1_4_1/vcf.c:2615:5
    #4 0x55ac4df126ef in bcf_write third_party/htslib/htslib_1_4_1/vcf.c:1514:16
    #5 0x55ac4de94c8b in view_vcf(hFILE*, char const*) experimental/users/kostyak/htsfuzz/htsreader.cc:55:5
    #6 0x55ac4de945db in main experimental/users/kostyak/htsfuzz/htsreader.cc:85:11

0x6030000017d0 is located 0 bytes to the right of 32-byte region [0x6030000017b0,0x6030000017d0)
allocated by thread T0 here:
    #0 0x55ac4de80ff5 in __interceptor_realloc third_party/llvm/llvm/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:79:3
    #1 0x55ac4df15a5a in ks_resize third_party/htslib/htslib_1_4_1/htslib/kstring.h:113:16
    #2 0x55ac4df13486 in kputsn third_party/htslib/htslib_1_4_1/htslib/kstring.h:146:6
    #3 0x55ac4df15fa8 in bcf_enc_vchar third_party/htslib/htslib_1_4_1/vcf.c:1809:5
    #4 0x55ac4df183fa in vcf_parse third_party/htslib/htslib_1_4_1/vcf.c:2221:25
    #5 0x55ac4df0b860 in vcf_read third_party/htslib/htslib_1_4_1/vcf.c:2395:12
[...]

Issue #539, opened 2017-05-19
Out-of-bounds read in bcf_unpack_info_core1


cryptoad:
When processed under ASan, the attached file triggers an out-of-bounds read via `view_vcf`:
```
==369612==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6020000024e3 at pc 0x5647659dc310 bp 0x7ffe35258fc0 sp 0x7ffe35258fb8
READ of size 1 at 0x6020000024e3 thread T0
    #0 0x5647659dc30f in bcf_dec_size third_party/htslib/htslib_1_4_1/htslib/vcf.h:993:13
    #1 0x5647659eb65a in bcf_unpack_info_core1 third_party/htslib/htslib_1_4_1/vcf.c:2416:17
    #2 0x5647659eaa34 in bcf_unpack third_party/htslib/htslib_1_4_1/vcf.c:2487:19
    #3 0x5647659ec1ab in vcf_format third_party/htslib/htslib_1_4_1/vcf.c:2504:5
    #4 0x5647659d8129 in vcf_write third_party/htslib/htslib_1_4_1/vcf.c:2615:5
    #5 0x5647659d76ef in bcf_write third_party/htslib/htslib_1_4_1/vcf.c:1514:16
    #6 0x564765959c8b in view_vcf(hFILE*, char const*) experimental/users/kostyak/htsfuzz/htsreader.cc:55:5
    #7 0x5647659595db in main experimental/users/kostyak/htsfuzz/htsreader.cc:85:11

0x6020000024e3 is located 3 bytes to the right of 16-byte region [0x6020000024d0,0x6020000024e0)
allocated by thread T0 here:
    #0 0x564765945ff5 in __interceptor_realloc third_party/llvm/llvm/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:79:3
    #1 0x5647659daa5a in ks_resize third_party/htslib/htslib_1_4_1/htslib/kstring.h:113:16
    #2 0x5647659cd0a2 in kputc third_party/htslib/htslib_1_4_1/htslib/kstring.h:161:6
    #3 0x5647659da399 in bcf_enc_size third_party/htslib/htslib_1_4_1/htslib/vcf.h:940:12
    #4 0x5647659da5f0 in bcf_enc_int1 third_party/htslib/htslib_1_4_1/htslib/vcf.h:959:9
[...]

Issue #246, opened 2015-07-21
vcf/bcf reading has no concept of errors!


jkbonfield:
zlib can spot truncated files, as can bgzf, but the code for bcf_read glosses over all such things and treats any error as EOF. 

Note this isn't the same issues as https://github.com/samtools/samtools/issues/101 which is describing the samtools application as ignoring errors returned by htslib.  This is htslib not being able to return the error in the first place.

It can be demonstrated in bcftools or htsfile, but the error comes from bcf_read returning 0 in many cases.  

```
$ samtools mpileup -g ../samtools/test/mpileup/mpileup.1.bam > /tmp/_.bcf
$ head -50000c /tmp/_.bcf | ./htsfile -c - > /dev/null;echo $?
ret=0
0
$ head -50000c /tmp/_.bcf | zcat > /dev/null;echo $?

gzip: stdin: unexpected end of file
1
```

The ret=0 is debugging I put into htsfile to show the return value from the last bcf_read() call.

jmarshall:
Need to check whether we can have an error return value and whether that involves breaking the ABI.

jkbonfield:
Perhaps we can set ```fp->fp.bgzf->errcode``` (or maybe it's already set?) as a way of distinguishing EOF from errors.  This is a side-channel that would leave the ABI the same.

jmarshall:
That sounds like a good idea.

Re ugliness, IMHO anything that involves non-htslib code directly looking inside a `htsFile *` is a bug.  See comments on the [`htsFile` declaration in _hts.h_](https://github.com/samtools/htslib/blob/develop/htslib/hts.h#L112-L118).  So yes, we'd want to add some kind of accessor function/macro.

jkbonfield:
Do these comments in hts.h have any relation to wanting ABI changes for 1.4?  I suspect not as deprecating the old fields used <= 1.1 won't really help us unless we can remove every field which I don't believe we can.

Regarding this specific issue, while we could add checks to in->fp.bgzf->errcode with an appropriate accessor function at the htsFile level, I wonder if it is sufficient or whether we should take the opportunity to add an errno field to the htsFile structure now too (even if it's not used in the initial implementation).

Obviously bgzf has errcode, cram_fd has an err field, hFILE has an has_errno field - is that sufficient to avoid needing a generic hts errno field?

jkbonfield:
I'm having a look at this now, as we want to get as many ABI breaking things done before next release.

[...]

PR #544, opened 2017-05-22
Fix buffer overrun in vcf_format.

### LINKED ISSUE #537: SIGSEGV in vcf_format
Hey,

I am going to file a bunch of new issues found through fuzzing. The previous one where in the *_read paths, those ones are usually in a *_write path, feeding an fuzzed file to `view_sam` and others of `htsfile.c`. They might not be as dire as something directly in *_read, but are probably worth a fix.

The first one appears to be an over-indexing issue of `h->id` in `vcf_format`, when processing the attached file (compiled with ASan):
```
Program received signal SIGSEGV, Segmentation fault.
0x0000555555889aa2 in vcf_format (h=0x60f000000220, v=0x611000002840, s=0x608000000cb0) at third_party/htslib/htslib_1_4_1/vcf.c:2536
2536                kputs(h->id[BCF_DT_ID][z->key].key, s);
(gdb) x/i $rip
=> 0x555555889aa2 <vcf_format+2738>:    mov    (%rax),%rdi
(gdb) p/x $rax
$1 = 0x61900002bcf0
(gdb) p/x &h->id[0][z->key]
$2 = 0x61900002bcf0
(gdb) p z->key
$3 = 10327
(gdb) p h->n[0]
$4 = 55
```

[poc.zip](https://github.com/samtools/htslib/files/1014842/poc.zip)

jkbonfield:
Fixes #537

Also "fixed" the parser to prevent this from happening in the first
place, but poorly as there is no way to return an error from
bcf_hdr_register_hrec. (see also #246)

The following minimal VCF triggered this crash. Note "dype".

```
##fileformat=VCFv4.1
##INFO=<ID=S,Number=1,dype=String,Description="blah">
##INFO=<ID=I,Number=1,Type=Integer,Description="blah">
##contig=<ID=chr1,length=249250621,assembly=b37>
#CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO
chr1    10327   .       T       C       .       .       S=foo;I=12345
chr1    10327   .       T       C       .       .       S=foo;I=12345
```

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -500,6 +500,11 @@ int bcf_hdr_register_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec)
             if (var != BCF_VL_FIXED) num = 0xfffff;
         }
     }
+    if (strcmp(hrec->key, "INFO") == 0 && type == -1) {
+        if (hts_verbose >= 2)
+            fprintf(stderr, "[E::%s] An INFO field has no Type defined. Assuming String\n", __func__);
+        type = BCF_HT_STR;
+    }
     uint32_t info = ((((uint32_t)num) & 0xfffff)<<12 |
                      (var & 0xf) << 8 |
                      (type & 0xf) << 4 |
@@ -2533,6 +2538,11 @@ int vcf_format(const bcf_hdr_t *h, const bcf1_t *v, kstring_t *s)
             if ( !z->vptr ) continue;
             if ( !first ) kputc(';', s);
             first = 0;
+            if (z->key >= h->n[BCF_DT_ID]) {
+                fprintf(stderr, "[E::%s] invalid BCF, the INFO index is too large.\n", __func__);
+                errno = EINVAL;
+                return -1;
+            }
             kputs(h->id[BCF_DT_ID][z->key].key, s);
             if (z->len <= 0) continue;
             kputc('=', s);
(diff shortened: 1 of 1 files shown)


jkbonfield:
Actually delay this for a while.  I'll do the other vcf/bcf problems in the same PR as separate commits so it can be reviewed and merged together.

jkbonfield:
Also fixes #539 and fixes #540 by dint of protecting against broken headers during reading.  If we constructed a broken BCF record via another route (not reading via htslib), or reading and then editing the headers, then we could still trigger those other two bugs.

I'm unsure if that's an issue though as it actively requires a breakage elsewhere first.

daviesrob:
@pd3 This fixes INFO headers, but there is a related problem for FORMAT.  Currently not putting `Type=` in a format header causes `vcf_parse_format` to print:
```
[E::vcf_parse_format] the format type 15 currently not supported
```
and then call `abort()` - which at least alerts the user to the problem, although not in a very friendly way.

As far as I can tell, the `Number=` field is not used anywhere.

The specification is not very clear on whether `Type=` and `Number=` should be compulsory in FORMAT fields.  What should we do if they are missing?

pd3:
I think it's just an omission in the specification, they should be compulsory. 

What to do if they are missing - the specification says that if the header lines are present, they must be completely well-formed.  Because the tool already throws an error in this situation, I'd keep that. Maybe I was wrong before and the parser should always throw an error on malformed header lines, rather then attempting error recovery. Reheadering is a relatively cheap operation for VCF and should not be needed for BCF.

jkbonfield:
My patch currently has:

```
if (strcmp(hrec->key, "INFO") == 0 && type == -1) {
```

Are you saying that adding FORMAT in there too is a valid restriction to place?  If so we need the specification updated too to reflect that.  Ie:

```
if ((strcmp(hrec->key, "INFO") == 0 || strcmp(hrec->key, "FORMAT") == 0) && type == -1) {
```

I'm happy to amend it.  Should we also be checking FORMAT for the Number key too?  Again, the specification is unclear in this regard.

pd3:
Yes, I would think so

jkbonfield:
Updated to check FORMAT now too.

PR #551, opened 2017-06-04
Logging improvements 3

anderskaplan:
Replaced another batch of fprintf(stderr, ...) calls with hts_log_... calls, this time aiming to cover everything except the CRAM code.

Also fixed some missing header file dependencies in the Makefile that I happened to find.

This is the third pull request in a series that aims to solve #403 .


jmarshall:
> Also fixed some missing header file dependencies in the Makefile that I happened to find.

They're not missing — they come from _htslib_vars.mk_.

daviesrob:
Thanks for the update.  I've fixed the dependency to use the variable from htslib_vars.mk and will merge this now.

PR #553, opened 2017-06-06
Use BGZF* in cram_index_build() instead of zfp*

jmarshall:
Also fixes unchecked-write-error bugs in index writing.

The routines in _zfio.c_ are now unused so can be removed.  As a happy byproduct, HTSlib no longer uses `fseeko()`, so no heroics are needed for platforms where that function is not available.


jkbonfield:
Thanks John.

jmarshall:
If I'd realised there would be as little to change as that, I'd have got around to it years ago :smile:

jkbonfield:
Quite.  "Getting around to it" is always a tricky thing.  I decided to keep going with it noticing vlen.h next to the removed zfio.h so culled the reason for that existing too. :-)

The io_lib legacy lives on still in places, but it's all an incremental improvement.

jmarshall:
Yep… nice.

My plan for that corner was to replace _cram_io.c_'s `open_path_mfile()` and `find_path()` calls with HTSlib's wider facilities that by now have (almost?) expanded to do what's actually needed for reference path handling.  At that point, _mFILE.[ch]_, _open_trace_file.[ch]_, and _vlen.[ch]_ all evaporate.

jkbonfield:
There is still https://github.com/samtools/htslib/pull/417 which is on my radar to check out.  This may well provide most of the functionality needed as it's been written with the same need in mind - faking up file IO from a memory buffer.

Although to be frank the reference handling code was rather "organic" during development and is a rats nest in need of a complete rewrite.  It would be useful outside of CRAM too, for things like mpileup and tview.  That's a bit of a longer project needing a rainy month.

jmarshall:
hFILE fixed buffers, introduced for `data:` URLs but over the top just for those, and var-args `hopen` were designed with that memory buffer PR in mind.  I've written up my past plans for such functionality over on the PR.

PR #555, opened 2017-06-12
Add kbs_resize() function

pd3:
This adds a new m field to avoid unnecessary free()/malloc() calls.


daviesrob:
This changes the ABI by adding an extra field to `kbitset_t`, so isn't quite as easy to merge as the (possibly) less efficient version currently in bcf_sr_sort.c.  If the number of resizes is likely to be low then just calling realloc every time might not be that bad in practice.

Also, I think a better interface to `kbs_resize()` might be:
```C
int kbs_resize(kbitset_t **bs, size_t ni);
```

It makes checking for failure easier, and callers don't have to worry about losing the pointer and therefore leaking memory.  I see you avoided that trap with `realloc()` here, but then fell into it when calling `kbs_resize()` at https://github.com/samtools/bcftools/blob/db521c9ae2732722c30ec357a53d05da87d858de/plugins/fill-tags.c#L405

pd3:
Thank you for taking a look @daviesrob.

I am aware this breaks ABI. However, although the number of resizes will be low with the current usage, it cannot be generalized for future uses, this being a library function. A hacky solution would be to use the last bucket for storing `m`, instead of the current `~0UL`. As far as I understand the code, the only place this value depends on is `kbs_next()` and it would continue to work. If you'd prefer this solution, I wouldn't object.

The return type of pointer to `kbitset_t` cannot be avoided, given its definition. 

You are right that the bcftools code does not check the return value of memory allocations (here and many other places). In this particular case it is OK, because we know that the allocated memory is going to be very small. Admittedly not clean, but the code is astronomically more likely to fail at a different place.

pd3:
Sorry, my mistake about the return type. I did not notice the double pointer `kbitset_t **bs`, which is a better solution.

pd3:
Can this be reviewed and merged so that the issues https://github.com/samtools/bcftools/issues/621 and https://github.com/samtools/bcftools/issues/797 can be closed please?

jmarshall:
The bcftools PR samtools/bcftools#627 which fixes those two issues could in fact be very easily recoded to use the existing `kbitset_t` API and not need `kbs_resize()` at all. See my comment on that PR.

That said, `kbitset_t` was designed with simple cases where you know the size of the collection in advance in mind, and it would be good to extend this so that you could realise that the set needed to be made bigger later on too.  But a useful sensible `kbs_resize()` API function would preserve the current value of the set rather than clearing it.

This could be done without adding an `m` member and breaking ABI; or it would be beneficial (for `kbs_next()` in particular) to add such a member if the HTSlib ABI were being bumped for other reasons anyway. (The last bucket is an `unsigned long` rather than `size_t` so isn't really suitable for abusing to store `m`.)

pd3:
I agree, `kbs_resize()` should not clear, only resize. The pull requested is now updates and squashed.

jmarshall:
Rather than the _kbitset.h_ changes in this PR, I would recommend #710, which initialises the new memory after the `realloc()`, fixes an existing bug, and follows the local formatting conventions.

As htslib appears to be in an ABI-breaking phase, that PR also just adds a max-length member.

daviesrob:
#710, which replaces this is now merged.

PR #560, opened 2017-07-06
Work-around for CIGARs with >64k operations in BAM files

lh3:
This PR implements the idea discussed in samtools/hts-specs#40.

EDIT: test data are [available here](http://lh3lh3.users.sourceforge.net/data/cigar-64k/).

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -355,17 +355,25 @@ int32_t bam_endpos(const bam1_t *b)
 static int bam_tag2cigar(bam1_t *b)
 {
     bam1_core_t *c = &b->core;
-    uint32_t cigar_st, n_cigar4, CG_st, CG_en, ori_len = b->l_data;
+    uint32_t cigar_st, n_cigar4, CG_st, CG_en, ori_len = b->l_data, *cigar0, CG_len;
     uint8_t *CG;
-    if (c->n_cigar > 0 || !(c->flag & BAM_FUNMAP) || c->tid < 0 || c->pos < 0) return 0;
-    if ((CG = bam_aux_get(b, "CG")) == 0) return 0;
-    if (CG[0] != 'B' || CG[1] != 'I') return 0;
-    cigar_st = (uint8_t*)bam_get_cigar(b) - b->data;
-    c->n_cigar = le_to_u32(CG + 2);
+
+    // test where there is a real CIGAR in the CG tag to move
+    if (c->n_cigar != 1 || c->tid < 0 || c->pos < 0) return 0;
+    cigar0 = bam_get_cigar(b);
+    if (bam_cigar_op(cigar0[0]) != BAM_CSOFT_CLIP || bam_cigar_oplen(cigar0[0]) != c->l_qseq) return 0;
+    if ((CG = bam_aux_get(b, "CG")) == 0) return 0; // no CG tag
+    if (CG[0] != 'B' || CG[1] != 'I') return 0; // not of type B,I
+    CG_len = le_to_u32(CG + 2);
+    if (CG_len == 0) return 0; // nothing to move
+
+    // move from the CG tag to the right position
+    cigar_st = (uint8_t*)cigar0 - b->data;
+    c->n_cigar = CG_len;
     n_cigar4 = c->n_cigar * 4;
     CG_st = CG - b->data - 2;
     CG_en = CG_st + 8 + n_cigar4;
-    b->l_data += n_cigar4;
+    b->l_data += n_cigar4 - 4; // we need (c->n_cigar-1)*4 bytes to swap CIGAR to the right place
     if (b->m_data < b->l_data) {
         uint8_t *new_data;
         uint32_t new_max = b->l_data;
@@ -374,12 +382,11 @@ static int bam_tag2cigar(bam1_t *b)
         if (new_data == 0) return -1;
         b->m_data = new_max, b->data = new_data;
     }
-    memmove(b->data + cigar_st + n_cigar4, b->data + cigar_st, ori_len - cigar_st); // make room for the real CIGAR
-    memcpy(b->data + cigar_st, b->data + n_cigar4 + CG_st + 8, n_cigar4); // copy the real CIGAR to the right place
-    if (ori_len > CG_en)
-        memmove(b->data + CG_st + n_cigar4, b->data + CG_en + n_cigar4, ori_len - CG_en);
-    b->l_data -= n_cigar4 + 8;
-    c->flag &= ~BAM_FUNMAP;
+    memmove(b->data + cigar_st + n_cigar4, b->data + cigar_st + 4, ori_len - (cigar_st + 4)); // insert 4*(c->n_cigar-1) empty space to make room
+    memcpy(b->data + cigar_st, b->data + (n_cigar4 - 4) + CG_st + 8, n_cigar4); // copy the real CIGAR to the right place; -4 for the fake CIGAR
+    if (ori_len > CG_en) // move data after the CG tag
+      memmove(b->data + CG_st + n_cigar4 - 4, b->data + CG_en + n_cigar4 - 4, ori_len - CG_en);
+    b->l_data -= n_cigar4 + 8; // 8: CGBI (4 bytes) and CGBI length (4)
     return 1;
 }
 
@@ -461,11 +468,11 @@ int bam_write1(BGZF *fp, const bam1_t *b)
     const bam1_core_t *c = &b->core;
     uint32_t x[8], block_len = b->l_data - c->l_extranul + 32, y;
     int i, ok;
-    if (c->n_cigar > 0xffff) block_len += 8; // "8" for "CGBI" plus 4-byte tag length
+    if (c->n_cigar > 0xffff) block_len += 12; // "12" for "CGBI", 4-byte tag length and 4-byte fake CIGAR
     x[0] = c->tid;
     x[1] = c->pos;
     x[2] = (uint32_t)c->bin<<16 | c->qual<<8 | (c->l_qname - c->l_extranul);
-    if (c->n_cigar > 0xffff) x[3] = (uint32_t)(c->flag | BAM_FUNMAP) << 16;
+    if (c->n_cigar > 0xffff) x[3] = (uint32_t)c->flag << 16 | 1;
     else x[3] = (uint32_t)c->flag << 16 | (c->n_cigar & 0xffff);
     x[4] = c->l_qseq;
     x[5] = c->mtid;
@@ -482,20 +489,20 @@ int bam_write1(BGZF *fp, const bam1_t *b)
     }
     if (ok) ok = (bgzf_write(fp, x, 32) >= 0);
     if (ok) ok = (bgzf_write(fp, b->data, c->l_qname - c->l_extranul) >= 0);
-    if (c->n_cigar <= 0xffff) {
+    if (c->n_cigar <= 0xffff) { // no long CIGAR; write normally
         if (ok) ok = (bgzf_write(fp, b->data + c->l_qname, b->l_data - c->l_qname) >= 0);
-    } else {
-        uint32_t cigar_st, cigar_en;
+    } else { // with long CIGAR, insert a fake CIGAR record and move the real CIGAR to the CG:B,I tag
+        uint8_t buf[4];
+        uint32_t cigar_st, cigar_en, cigar1;
         cigar_st = (uint8_t*)bam_get_cigar(b) - b->data;
         cigar_en = cigar_st + c->n_cigar * 4;
+        cigar1 = (uint32_t)c->l_qseq << 4 | BAM_CSOFT_CLIP;
+        u32_to_le(cigar1, buf);
+        if (ok) ok = (bgzf_write(fp, buf, 4) >= 0); // write cigar: <read_length>S
         if (ok) ok = (bgzf_write(fp, &b->data[cigar_en], b->l_data - cigar_en) >= 0); // write data after CIGAR
         if (ok) ok = (bgzf_write(fp, "CGBI", 4) >= 0); // write CG:B,I
-        if (fp->is_be) {
-            y = c->n_cigar;
-            if (ok) ok = (bgzf_write(fp, ed_swap_4p(&y), 4) >= 0);
-        } else {
-            if (ok) ok = (bgzf_write(fp, &c->n_cigar, 4) >= 0);
-        }
+        u32_to_le(c->n_cigar, buf);
+        if (ok) ok = (bgzf_write(fp, buf, 4) >= 0); // write the true CIGAR length
         if (ok) ok = (bgzf_write(fp, &b->data[cigar_st], c->n_cigar * 4) >= 0); // write the real CIGAR
     }
     if (fp->is_be) swap_data(c, b->l_data, b->data, 0);
--- a/sam.c
+++ b/sam.c
@@ -485,9 +485,13 @@ int bam_write1(BGZF *fp, const bam1_t *b)
         cigar_st = (uint8_t*)bam_get_cigar(b) - b->data;
         cigar_en = cigar_st + c->n_cigar * 4;
         if (ok) ok = (bgzf_write(fp, &b->data[cigar_en], b->l_data - cigar_en) >= 0); // write data after CIGAR
-        memcpy(&x[0], "CGBI", 4);
-        x[1] = c->n_cigar;
-        if (ok) ok = (bgzf_write(fp, x, 8) >= 0); // write CG:B:I and length
+		if (ok) ok = (bgzf_write(fp, "CGBI", 4) >= 0); // write CG:B,I
+		if (fp->is_be) {
+			y = c->n_cigar;
+			if (ok) ok = (bgzf_write(fp, ed_swap_4p(&y), 4) >= 0);
+		} else {
+			if (ok) ok = (bgzf_write(fp, &c->n_cigar, 4) >= 0);
+		}
         if (ok) ok = (bgzf_write(fp, &b->data[cigar_st], c->n_cigar * 4) >= 0); // write the real CIGAR
     }
     if (fp->is_be) swap_data(c, b->l_data, b->data, 0);
--- a/sam.c
+++ b/sam.c
@@ -485,13 +485,13 @@ int bam_write1(BGZF *fp, const bam1_t *b)
         cigar_st = (uint8_t*)bam_get_cigar(b) - b->data;
         cigar_en = cigar_st + c->n_cigar * 4;
         if (ok) ok = (bgzf_write(fp, &b->data[cigar_en], b->l_data - cigar_en) >= 0); // write data after CIGAR
-		if (ok) ok = (bgzf_write(fp, "CGBI", 4) >= 0); // write CG:B,I
-		if (fp->is_be) {
-			y = c->n_cigar;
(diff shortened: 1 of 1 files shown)


sam.c
```diff
@@ -352,6 +352,33 @@ int32_t bam_endpos(const bam1_t *b)
...
+static void bam_tag2cigar(bam1_t *b)
+{
+    bam1_core_t *c = &b->core;
+    uint32_t cigar_st, n_cigar4, CG_st, CG_en, ori_len = b->l_data;
+    uint8_t *CG;
+    if (c->n_cigar > 0 || !(c->flag & BAM_FUNMAP) || c->tid < 0 || c->pos < 0) return;
+    if ((CG = bam_aux_get(b, "CG")) == 0) return;
+    if (CG[0] != 'B' || CG[1] != 'I') return;
+    cigar_st = (uint8_t*)bam_get_cigar(b) - b->data;
+    c->n_cigar = *(uint32_t*)(CG + 2);
```
jkbonfield: I think this could do with some endianness protection via ```ed_swap_4p(&c->n_cigar)```.

sam.c
```diff
@@ -352,6 +352,33 @@ int32_t bam_endpos(const bam1_t *b)
...
+    n_cigar4 = c->n_cigar * 4;
+    CG_st = CG - b->data - 2;
+    CG_en = CG_st + 8 + n_cigar4;
+    b->l_data += n_cigar4;
+    if (b->m_data < b->l_data) {
+        b->m_data = b->l_data;
+        kroundup32(b->m_data);
+        b->data = (uint8_t*)realloc(b->data, b->m_data);
+    }
```
jkbonfield: What aren't I following here.  We migrate an 4-byte B array from aux tag to cigar array, but the total number of 4-byte values is the same.  The only difference should be we have one less aux tag header, so we drop the ```CGBI<size>``` bytes.  Memory allocation should be sufficient as it becomes smaller after migration.  Are you reallocing purely for the purpose of making the shuffle easier?  If so I think allocating a temporary buffer may be easier to understand, although it requires two copies.

If you want to grow and do inline, please check the return value from malloc.

Also I can't see anywhere that checks the aux data is large enough.  We could have a truncated aux array where the size is larger than the aux tag permits.  We ought to error rather than crash in that scenario, or worse it's a buffer overrun and potential exploit due to the memmove writing beyond the end of the buffer.

sam.c
```diff
@@ -421,6 +448,7 @@ int bam_read1(BGZF *fp, bam1_t *b)
         bgzf_read(fp, b->data + c->l_qname, b->l_data - c->l_qname) != b->l_data - c->l_qname)
         return -4;
     if (fp->is_be) swap_data(c, b->l_data, b->data, 0);
+    bam_tag2cigar(b);
```
jkbonfield: Given the memory allocation, please make bam_tag2cigar return int and check the value here.

sam.c
```diff
@@ -453,7 +478,18 @@ int bam_write1(BGZF *fp, const bam1_t *b)
...
-    if (ok) ok = (bgzf_write(fp, b->data + c->l_qname, b->l_data - c->l_qname) >= 0);
+    if (c->n_cigar <= 0xffff) {
+        if (ok) ok = (bgzf_write(fp, b->data + c->l_qname, b->l_data - c->l_qname) >= 0);
+    } else {
+        uint32_t cigar_st, cigar_en;
+        cigar_st = (uint8_t*)bam_get_cigar(b) - b->data;
+        cigar_en = cigar_st + c->n_cigar * 4;
+        if (ok) ok = (bgzf_write(fp, &b->data[cigar_en], b->l_data - cigar_en) >= 0); // write data after CIGAR
+        memcpy(&x[0], "CGBI", 4);
+        x[1] = c->n_cigar;
```
jkbonfield: Also needs byte swapping code.

sam.c
```diff
@@ -1143,6 +1179,7 @@ int sam_parse1(kstring_t *s, bam_hdr_t *h, bam1_t *b)
         } else _parse_err_param(1, "unrecognized type %c", type);
     }
     b->data = (uint8_t*)str.s; b->l_data = str.l; b->m_data = str.m;
+    bam_tag2cigar(b);
```
jkbonfield: Please add check for return value.

jkbonfield:
I see you've already noticed the endianness, great!  Niggles aside, the basic implementation looks like the right approach to me and the sensible place to do it.  I like the idea of calling ```bam_tag2cigar``` when parsing SAM too incase it's been processed by an another BAM reader or older code.  I wonder if we need this logic applying when decoding CRAM too for similar reasons (BAM->CRAM with outdated software).

Also, we need an ancillary fix for the spec too to reserve the new tag type (even if it's just BAM only, we have to prevent us from reusing that tag).

sam.c
lh3: Which function will create a truncated aux array? This would be a bug in htslib. In addition, `bam_aux_append()` doesn't check aux data, either.

`realloc` is to reduce future heap allocations. If there were many long CIGARs, you would need to allocate a temporary buffer for swapping every time.

lh3:
> Also, we need an ancillary fix for the spec too to reserve the new tag type (even if it's just BAM only, we have to prevent us from reusing that tag).

Yes, hts-specs needs to be changed first. I am posting this PR just to show how the idea in samtools/hts-specs#40 works in practice. It does not require much code change.

sam.c
jkbonfield: Good point on realloc.  If we're in the usual pipeline of reusing the same bam then it'll not realloc very often so thumbs up on that idea.

Regarding truncated array, my worry is about malicious data where a malformed BAM file has been loaded.  My understanding is that the BAM reader just reads block_size bytes, but the contents of the aux data may not be consistent with the decoded block_size value.

It may not actually be a buffer write overrun though as by reallocing we're growing the array first before doing the memcpy, but it may lead to exposing uninitialised data.  Consider a broken aux field of CGBI<0xff 0x00 0x00 0x00>foo <end of stream>.  It claims to be 255 ints long, so nearly 1Kb, but only contains "foo" bytes.  The BAM record size itself could be correct, ending at the "foo" point, but the actual decoding of the BAM record would expose an error.  This code however will grow the BAM record by ~1Kb to make way for 255 cigar elements and then memcpy ~1Kb from ```CGBI``` onwards to that cigar array, which possibly includes nearly 1Kb of random contents from a previous malloc/free on the heap.

If bam_aux_append has the same issue then it needs fixing too, but on a quick glance it seems fine to me.

jkbonfield: I take it all back. :-)

The ```bam_aux_get``` call already checks (since 95b1034e93b345fef7aa4980750ba1dedde2c284 by Rob I think) the data being queried doesn't overflow the buffer and returns NULL if so, making the entire ```bam_tag2cigar``` function bail out at this point.  Alarm averted! :-)

sjackman:
Commented at https://github.com/samtools/hts-specs/issues/40#issuecomment-313950483

lh3:
Updated to write fake cigar like `<readLength>S<refLength>N` if there are >65535 CIGAR operators. The reading code is more general. It works as long as the first CIGAR operation is `<readLength>S`.

sam.c
```diff
@@ -352,6 +352,46 @@ int32_t bam_endpos(const bam1_t *b)
...
+        new_data = (uint8_t*)realloc(b->data, new_max);
+        if (new_data == 0) return -1;
+        b->m_data = new_max, b->data = new_data;
+    }
+    memmove(b->data + cigar_st + n_cigar4, b->data + cigar_st + fake_bytes, ori_len - (cigar_st + fake_bytes)); // insert c->n_cigar-fake_bytes empty space to make room
+    memcpy(b->data + cigar_st, b->data + (n_cigar4 - fake_bytes) + CG_st + 8, n_cigar4); // copy the real CIGAR to the right place; -fake_bytes for the fake CIGAR
+    if (ori_len > CG_en) // move data after the CG tag
+      memmove(b->data + CG_st + n_cigar4 - fake_bytes, b->data + CG_en + n_cigar4 - fake_bytes, ori_len - CG_en);
+    b->l_data -= n_cigar4 + 8; // 8: CGBI (4 bytes) and CGBI length (4)
+    b->core.bin = hts_reg2bin(b->core.pos, b->core.pos + bam_cigar2rlen(b->core.n_cigar, bam_get_cigar(b)), 14, 5);
```
jkbonfield: Given the main decoding loop has a bam_cigar2qlen check always performed, it makes sense to combine these into a (hypothetical) ```bam_cigar2rqlens(b->core.n_cigar, bam_get_cigar(b), &qlen, &rlen)``` call that only scans the cigar string once.  Summing both lengths is minimal difference in CPU time than just summing one, so we can exploit this to do some better error checking.

This can be done in the main loop, rather than ```bam_tag2cigar```, and we may as well validate (or just set if ncigar large) reg2bin in all cases as computing bin is a cheap operation compared to scanning cigars.

lh3:
One remaining question: it would be good to call `bam_tag2cigar()` immediately after reading a CRAM record. Where should we add this call? I am thinking to add to `cram_get_bam_seq()` in `cram_decode.c`. Is that a good place?

jkbonfield:
I don't think it should be anywhere in SAM or CRAM reading code as this is purely a BAM workaround.

lh3:
When older tools converting long-cigar BAM to SAM or CRAM, they will leave the `CG` tag in SAM/CRAM unintentionally. Calling `bam_tag2cigar()` allows the latest samtools to work with such SAM/CRAM correctly.

jkbonfield:
Fair enough, but when doing so we should emit a warning (once only ideally), as it is an explicit indication that there is an out-dated part of your pipeline which you need to be aware of.

As for where the call should go, I'd say just after the call to cram_get_bam_seq as then the function can be static and doesn't leak out of sam.c.

lh3:
It is easy to print a warning for every offending read. Does htslib have a mechanism to output one type of warning only once? Or I will just call the `hts_log_warning()` macro?

lh3:
As htslib does not have a mechanism to print a type of warning only once, I will not invent one with this PR. The latest change gives a warning for each offending record. There are few such reads, so the repeated message to stderr won't be that bad. When long-cigar reads become common and this message becomes annoying, the users should really update their other tools anyway.

Test data can be found [here](http://lh3lh3.users.sourceforge.net/data/cigar-64k/). There is also a precompiled samtools-1.6 plus a backport of this PR.

Any more concerns?

sjackman:
May I suggest…
```c
if (give_warning) {
	static int warned;
	if (!warned)
		hts_log_error("%s encodes a CIGAR with %d operators at the CG tag", bam_get_qname(b), c->n_cigar);
	warned = 1;
}
```

sjackman:
Checking and setting `warned` ought to use an atomic primitive if available. The result of the race condition is simply to print the error message more than once, so not so terrible.

lh3:
I thought about that. In theory, it is not the right thing to do with a library, but if @jkbonfield is ok with this solution, I can live with that.

sjackman:
The biggest downside to the above approach is that it will be printed only once per run of the program, rather than once per SAM file, which would be more intuitive.

sjackman:
For that reason, perhaps one warning per offending read is better.

lh3:
If there are no more concerns, could someone review this PR and get it merged? Note that Travis reported an error, but it seems unrelated to my changes. The error is:
```
Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core failed!
...
The command "if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update    ; fi" failed and exited with 1 during .
```
It is a homebrew error.

jmarshall:
It would be preferable for the merging of implementation PRs to hold off until the specification PR samtools/hts-specs#227 is merged or closer to merging. While there is agreement on the general approach, IMHO as a spec maintainer there are details still remaining to be worked out.

lh3:
It seems to me there are no pending issues with samtools/hts-specs#227 so far – CG:Y has been pushed back, which is not my preference anyway. Could you comment on the other PR about specific concerns?

jkbonfield:
Agreed on Travis error - it's just one of the many MacOS X glitches it has.  

Regarding merging,  I took a quick look at the code and commented a bit but haven't yet stress tested it.

jkbonfield:
As I feared, it does indeed have potential security issues when given deliberately broken data. My nasty test has a longer (actual) CIGAR string than CG:B,I tag.  The expectation is that we're replacing a short bit of memory by a large bit of memory and the memmoves are written accordingly, but if we reverse this assumption it starts to shift outside its memory buffers.

There are also issues with signed vs unsigned types.  m_data is unsigned while l_data is signed.  I'm not sure why this is so, but the check for reallocating is wrong.

```
(gdb) n
389	    b->l_data += n_cigar4 - fake_bytes; // we need c->n_cigar-fake_bytes bytes to swap CIGAR to the right place
(gdb) p n_cigar4
$4 = 4
(gdb) p fake_bytes
$5 = 408
(gdb) n
390	    if (b->m_data < b->l_data) {
(gdb) ptype b->m_data
type = unsigned int
(gdb) ptype b->l_data
type = int
```

My example file was produced from this one liner:

```
perl -e 'print "\@SQ\tSN:a\tLN:100\nfoo\t16\ta\t1\t1\t1S1N", "1N" x 100, "\t*\t0\t0\tN\t*\tCG:B,I,0\n"' > a.sam
```

I think we need to fuzz test this basically, with both BAM and SAM inputs so we can check binary and textual parsing of cigar and CG tags.  I don't mind if it simply gives nonsensical output from nonsensical input (eg the CG:B,I tag makes no sense and uses illegal cigar opcodes), but it should never be permitted to crash and definitely not to overrun any buffers.

The basic principle of how it works however is fine.  We just need a few more input validation checks and detection of integer overflows.

lh3:
We can change
```c
if (CG_len == 0) return 0;
```
to
```c
if (CG_len < c->n_cigar) return 0;
```
to prevent this crash. If you are ok with this, I can push another commit.

jkbonfield:
It cuts out one of the issues.  I'm still concerned over the realloc and unsigned vs signed variables though.

I think maybe if we really push the boat out we can get a 2Gb CG tag, wrap around b->l_data, make the ```if (b->m_data < b->l_data)``` check succeed and no need a realloc, and then do a memmove outside of the allocated memory.

It's extreme, but I think it still exists as a possibility unless there is something else that forbids a 2Gb aux tag.  It could perhaps be guarded against simply by adding a maximum length of CG too to prevent this scenario being possible. Eg:

```
if (CG_len < c->n_cigar || CG_len >= (1u<<31)) return 0;
```

jkbonfield:
Actually that should have been CG_len*4, or more robust CG_len >= (1u<<29) I guess.

We also wondered about the impact of faked up CG binary data, eg where the CG tag claims to be 1Mb long, but actually isn't.  There are no checks here about going outside the end of the array, however I think this cannot happen as bam_aux_get itself specifically checks for this case and returns NULL so it simply pretends that the truncated aux tag doesn't exist.

In short, you've probably nailed it, but I'll try and remember how to run afl again.

sam.c
```diff
@@ -423,13 +478,18 @@ int bam_read1(BGZF *fp, bam1_t *b)
...
         return -4;
     if (fp->is_be) swap_data(c, b->l_data, b->data, 0);
-
-    // Sanity check for broken CIGAR alignments
-    if (c->n_cigar > 0 && c->l_qseq > 0 && !(c->flag & BAM_FUNMAP)
-        && bam_cigar2qlen(c->n_cigar, bam_get_cigar(b)) != c->l_qseq) {
-        hts_log_error("CIGAR and query sequence lengths differ for %s",
-                      bam_get_qname(b));
-        return -4;
+    bam_tag2cigar(b, 0, 0);
```
jkbonfield: bam_tag2cigar returns an error (0 or -1), but every call to it doesn't bother to check the return value.  Please fix this.

jkbonfield:
Thanks for the updates Heng.  I can confirm that AFL was finding the problems before and no longer is.  I'll leave it running over the weekend and assuming it's fine merge Monday hopefully

jkbonfield:
Rebased, squashed and merged as aea349acc900eb2718ab459a5896c123d97d6df3.

lh3:
Thanks a lot, @jkbonfield!

sam.c
```diff
@@ -352,6 +364,49 @@ int32_t bam_endpos(const bam1_t *b)
         return b->core.pos + 1;
 }
 
+static int bam_tag2cigar(bam1_t *b, int recal_bin, int give_warning) // return 0 if CIGAR is untouched; 1 if CIGAR is updated with CG
+{
+    bam1_core_t *c = &b->core;
+    uint32_t cigar_st, n_cigar4, CG_st, CG_en, ori_len = b->l_data, *cigar0, CG_len, fake_bytes;
+    uint8_t *CG;
+
+    // test where there is a real CIGAR in the CG tag to move
```
BenJWard: Hi @lh3, I'm adding support for cigars in CG tags into [BioJulia/BioAlignments.jl](https://github.com/BioJulia/BioAlignments.jl), and am currently implementing the test for the fake cigar, which is defined as `[k<<4|4,m<<4|3]` in the spec, where k is l_seq and m is the length of the reference in the alignment. I don't see a field I can say obviously corresponds to m in BAM records.

In your C of bam_tag2cigar, if I understand it correctly, I see you check whether the first op is a softclip, and if it's length == l_seq `if (bam_cigar_op(cigar0[0]) != BAM_CSOFT_CLIP || bam_cigar_oplen(cigar0[0]) != c->l_qseq) return 0;`
But I don't see in the C where the second cigar op of "mN" is checked, is it therefore something that is a given if kS is present then mN will be present?

Thanks,
Ben.

jkbonfield: There isn't a field directly in the BAM structure to define 'm' as it is the number of *reference* bases spanned by the sequence and not *query* (so not l_qseq).  The corresponding code for writing it is also in sam.c:

```
        cigar[0] = (uint32_t)c->l_qseq << 4 | BAM_CSOFT_CLIP;
        cigar[1] = (uint32_t)bam_cigar2rlen(c->n_cigar, bam_get_cigar(b)) << 4 | BAM_CREF_SKIP;
```

You'll see the bam_cigar2rlen function is the appropriate code for working out 'm'.

BenJWard: @jkbonfield Ah I see, so if I'm properly understanding, since `m` is computed with `bam_cigar2rlen`, using the actual long cigar, then when reading in a BAM record, in bam_tag2cigar, only the first cigar op is checked for `k` and `BAM_CSOFT_CLIP` here:

```
// test where there is a real CIGAR in the CG tag to move
    if (c->n_cigar == 0 || c->tid < 0 || c->pos < 0) return 0;
    cigar0 = bam_get_cigar(b);
    if (bam_cigar_op(cigar0[0]) != BAM_CSOFT_CLIP || bam_cigar_oplen(cigar0[0]) != c->l_qseq) return 0;
    fake_bytes = c->n_cigar * 4;
```

Preciely because you can't yet compute `m` as the real cigar has not been read yet in order to do so?

jkbonfield: That would be the generous interpretation, but it probably boils down to not bothering to check. :-)  Note it's not solely entire-soft-clip, but also the presence of the CG tag.  The two combined are pretty specific, although it's not a bad idea to also validate that CG and cigar are consistent with each other.

Note the main reason for the existence of the m part of kSmN was so that tools that wanted to do coverage graphs could still use the minimal cigar string to derive the alignment end coordinate.  The code started with just kS and so the htslib implementation probably shows some of that history in how it is structured.

PR #563, opened 2017-07-12
Logging improvements 4

anderskaplan:
Replaced fprintf(stderr, ...) calls with hts_log_... calls in the CRAM code. This pull request is part of a series that aims to solve #403 .

Most of the changes are search-and-replace, but there is one change that I'd like to bring up for discussion: I've removed the CRAM-specific verbosity flag. No use with a separate flag for that part of the code, right? And it's not part of the public API. (In case there are actually good reasons to keep it, those changes are isolated in a separate commit and so can be easily reversed.)

I've also put #ifdefs around debugging functions in a few cases. The reason is that I don't think the implementations should be there when the invocations are commented out. After the change, both the implementation and invocation are enabled or disabled by a single define.

### DIFF
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -558,7 +558,7 @@ cram_codec *cram_beta_decode_init(char *data, int size,
     else if (option == E_BYTE_ARRAY || option == E_BYTE)
 	c->decode = cram_beta_decode_char;
     else {
-	fprintf(stderr, "BYTE_ARRAYs not supported by this codec\n");
+	hts_log_error("BYTE_ARRAYs not supported by this codec");
 	return NULL;
     }
     c->free   = cram_beta_decode_free;
@@ -748,7 +748,7 @@ cram_codec *cram_subexp_decode_init(char *data, int size,
     char *cp = data;
 
     if (option != E_INT) {
-	fprintf(stderr, "This codec only supports INT encodings\n");
+	hts_log_error("This codec only supports INT encodings");
 	return NULL;
     }
 
@@ -764,7 +764,7 @@ cram_codec *cram_subexp_decode_init(char *data, int size,
     cp += safe_itf8_get(cp, data + size, &c->subexp.k);
 
     if (cp - data != size || c->subexp.k < 0) {
-	fprintf(stderr, "Malformed subexp header stream\n");
+	hts_log_error("Malformed subexp header stream");
 	free(c);
 	return NULL;
     }
@@ -814,7 +814,7 @@ cram_codec *cram_gamma_decode_init(char *data, int size,
     char *cp = data;
 
     if (option != E_INT) {
-	fprintf(stderr, "This codec only supports INT encodings\n");
+	hts_log_error("This codec only supports INT encodings");
 	return NULL;
     }
 
@@ -985,13 +985,13 @@ cram_codec *cram_huffman_decode_init(char *data, int size,
     int l;
 
     if (option == E_BYTE_ARRAY_BLOCK) {
-	fprintf(stderr, "BYTE_ARRAYs not supported by this codec\n");
+	hts_log_error("BYTE_ARRAYs not supported by this codec");
 	return NULL;
     }
 
     cp += safe_itf8_get(cp, data_end, &ncodes);
     if (ncodes < 0) {
-        fprintf(stderr, "Invalid number of symbols in huffman stream\n");
+	hts_log_error("Invalid number of symbols in huffman stream");
         return NULL;
     }
     if (ncodes >= SIZE_MAX / sizeof(*codes)) {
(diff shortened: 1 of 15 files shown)


cram/cram_decode.c
```diff
@@ -2368,10 +2365,10 @@ int cram_decode_slice(cram_fd *fd, cram_container *c, cram_slice *s,
 	if ((!s->ref && s->hdr->ref_base_id < 0)
 	    || memcmp(digest, s->hdr->md5, 16) != 0) {
 	    char M[33];
-	    fprintf(stderr, "ERROR: md5sum reference mismatch for ref "
-		    "%d pos %d..%d\n", ref_id, s->ref_start, s->ref_end);
-	    fprintf(stderr, "CRAM: %s\n", md5_print(s->hdr->md5, M));
-	    fprintf(stderr, "Ref : %s\n", md5_print(digest, M));
+	    hts_log_error("MD5 checksum reference mismatch for ref %d pos %d..%d",
+		ref_id, s->ref_start, s->ref_end);
+	    hts_log_error("CRAM: %s", md5_print(s->hdr->md5, M));
```
valeriuo: Please surround char arrays with simple quotes (`'%s'`) when logging. It makes it easy to spot the type and be sure no space has sneaked in.

cram/cram_codecs.c
```diff
@@ -558,7 +558,7 @@ cram_codec *cram_beta_decode_init(char *data, int size,
     else if (option == E_BYTE_ARRAY || option == E_BYTE)
 	c->decode = cram_beta_decode_char;
     else {
-	fprintf(stderr, "BYTE_ARRAYs not supported by this codec\n");
+	hts_log_error("BYTE_ARRAYs not supported by this codec");
```
valeriuo: Please replace tabs with spaces (4) in all your changes. I know the file is littered with them, but let's not introduce any more.

anderskaplan: I'd rather not change the formatting as part of a pull request targeting something else, so I've tried to stick to whatever was in the file already.

But I totally agree that the mix of tabs and spaces is horrible. I will gladly make a pull request covering ONLY tabs-to-spaces reformatting if people agree it's a good idea.

jmarshall: See also the last paragraph of e770a1b390bc72ba467831a67ffb2d4e1c2091f4, though the need for aligning _cram/*_ formatting with that of the same code in io_lib may have since changed.

jkbonfield: I'd be happy with rationalising the spacing throughout, besides there are usually white space options for diffing files.

Frankly the only reason I had a mix of tabs and spaces is simply because it's the emacs default and it's invisible (plus tab = 8 space is standard on Unix) so I never really cared!  Things have moved on a bit though and it's probably time to coax the code into pure space mode.

I agree it's best done in its own pull request though, and keeping the same formatting as current code is good for now.

cram/cram_decode.c
jkbonfield: I had a look at htslib's sam.c and see it has a mix of ```%s```, ```"%s"``` and ```'%s'```, depending on message.

These error messages were written by a different author, initially in a different project, to the htslib ones so there is probably an inconsistency of style.   I have no strong preferences on the style though and feel free to improve wording if you spot anything phrased poorly too.  Thank you @anderskaplan for your work on logging.

cram/cram_decode.c
```diff
@@ -2741,11 +2741,11 @@ int cram_decode_slice(cram_fd *fd, cram_container *c, cram_slice *s,
     // As we grow blocks we overallocate by up to 50%. So shrink
     // back to their final sizes here.
     //
-//    fprintf(stderr, "%d %d // %d %d // %d %d // %d %d\n",
-//    	    (int)s->seqs_blk->byte, (int)s->seqs_blk->alloc, 
-//    	    (int)s->qual_blk->byte, (int)s->qual_blk->alloc, 
-//    	    (int)s->name_blk->byte, (int)s->name_blk->alloc, 
-//    	    (int)s->aux_blk->byte,  (int)s->aux_blk->alloc);
+    //fprintf(stderr, "%d %d // %d %d // %d %d // %d %d\n",
```
jkbonfield: No complaints on the change, but the reason why I often end up commenting out this way is ironically due to the tabbing issues mentioned before.

I use rectangular edits, which works well at the start of a line, but less well part way along a line if there are mixes of tabs and spaces.  So another good reason to quell the tabs.

cram/cram_structs.h
```diff
@@ -706,7 +706,6 @@ typedef struct cram_fd {
 
     // options
     int decode_md; // Whether to export MD and NM tags
-    int verbose;
```
jkbonfield: This changes the size of cram_fd struct and moves the elements down a few bytes.  The ```cram_fd``` definition is internal and the external code that accesses it does so via accessor functions `cram_fd_get_fp`, `cram_fd_get_header`, `cram_minor_vers` and `cram_major_vers` in cram_external.c.

So I *think* this is fine for compatibility and doesn't break ABI at all despite the structure size change.

cram/cram_decode.c
anderskaplan: Right. The reason for the change here is that I've searched the code for "fprintf(stderr" several times when scanning for printouts that should be changed to hts_log. This formatting makes it more clear what is commented out and what is not.

cram/cram_codecs.c
anderskaplan: For the record: I'm happy with ANY standard, as long as there is only one of them. :)

cram/cram_decode.c
anderskaplan: I'll add that to my to-do list, but I don't think this PR is the right place to do it.

PR #568, opened 2017-07-22
Fix bcf_translate to skip deleted INFO and FORMAT fields to avoid segfaults

bioinformed:
## Background

`bcf_translate` works correctly on newly created or loaded BCF records, but it does not properly account for records which have been modified by having one or more INFO or FORMAT fields deleted.  This bug has resulted in segmentation faults in real usage.

## Nature of fix

Prior to attempt to update the info value (either in place or after resizing), update the entry's key id (to keep all ids valid) and then check whether `line->d.info[i].vptr` is `NULL`.   This indicates that the key and value have been deleted.  Otherwise, `vptr` is incorrectly assumed to point to a valid value and a segmentation fault results.

The same basic flaw appears to be possible for deleted FORMAT values when `line->d.fmt[i].p` is `NULL` and a similar fix was applied.

The resize cases were also unable to handle when when `p_free` is set.  This was also corrected.

## Other issues not addressed in this PR

`bcf_translate` uses a data structure stored within the source BCF header data structure.  This is a poor design decision, because it presumes that any header will be used to translate to no more than one destination header.  Worse, there is no check to verify that the same destination header is used or that the source and destination have not changed since the translation table was initialized.  Both requirements are unchecked and can cause crashes or write corrupted data.  There is no good reason that the translation data structure is part of the header, except that it would require breaking ABI compatibility.  This suggests adding a `bcf_translate2` function that takes an explicit translation table as a parameter with clear documentation on the assumptions and lifetime of the translation struct.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -2929,11 +2929,12 @@ int bcf_translate(const bcf_hdr_t *dst_hdr, bcf_hdr_t *src_hdr, bcf1_t *line)
         int src_id = line->d.info[i].key;
         int dst_id = src_hdr->transl[BCF_DT_ID][src_id];
         if ( dst_id<0 ) continue;
+        line->d.info[i].key = dst_id;
+        if ( !line->d.info[i].vptr ) continue;  // skip deleted
         int src_size = src_id>>7 ? ( src_id>>15 ? BCF_BT_INT32 : BCF_BT_INT16) : BCF_BT_INT8;
         int dst_size = dst_id>>7 ? ( dst_id>>15 ? BCF_BT_INT32 : BCF_BT_INT16) : BCF_BT_INT8;
         if ( src_size==dst_size )   // can overwrite
         {
-            line->d.info[i].key = dst_id;
             uint8_t *vptr = line->d.info[i].vptr - line->d.info[i].vptr_off;
             if ( dst_size==BCF_BT_INT8 ) { vptr[1] = (uint8_t)dst_id; }
             else if ( dst_size==BCF_BT_INT16 ) { *(uint16_t*)vptr = (uint16_t)dst_id; }
@@ -2950,7 +2951,6 @@ int bcf_translate(const bcf_hdr_t *dst_hdr, bcf_hdr_t *src_hdr, bcf1_t *line)
             kputsn((char*)info->vptr, info->vptr_len, &str);
             info->vptr = (uint8_t*)str.s + info->vptr_off;
             info->vptr_free = 1;
-            info->key = dst_id;
             line->d.shared_dirty |= BCF1_DIRTY_INF;
         }
     }
--- a/vcf.c
+++ b/vcf.c
@@ -2961,11 +2961,12 @@ int bcf_translate(const bcf_hdr_t *dst_hdr, bcf_hdr_t *src_hdr, bcf1_t *line)
         int src_id = line->d.fmt[i].id;
         int dst_id = src_hdr->transl[BCF_DT_ID][src_id];
         if ( dst_id<0 ) continue;
+        line->d.fmt[i].id = dst_id;
+        if( !line->d.fmt[i].p ) continue;  // skip deleted
         int src_size = src_id>>7 ? ( src_id>>15 ? BCF_BT_INT32 : BCF_BT_INT16) : BCF_BT_INT8;
         int dst_size = dst_id>>7 ? ( dst_id>>15 ? BCF_BT_INT32 : BCF_BT_INT16) : BCF_BT_INT8;
         if ( src_size==dst_size )   // can overwrite
         {
-            line->d.fmt[i].id = dst_id;
             uint8_t *p = line->d.fmt[i].p - line->d.fmt[i].p_off;    // pointer to the vector size (4bits) and BT type (4bits)
             if ( dst_size==BCF_BT_INT8 ) { p[1] = dst_id; }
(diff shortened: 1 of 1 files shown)


daviesrob:
Could this actually be due to a missing call to `bcf1_sync()`, which should tidy up the NULL `vptr` values along with a number of other possible inconsistencies?  @pd3 may want to comment on this, I'm having a bit of trouble getting my head around how it's supposed to work...

bioinformed:
@daviesrob: Maybe?  However it isn't at all clear from the code or header docs that `bcf_translate()` must be called on a synced `bcf1_t` instance.

pd3:
@bioinformed is correct, the API is not general enough and assumes it is run on newly loaded records. Is the pull request motivated by a real life necessity or is it a theoretical concern that could be solved by improving the documentation? 
Can you make sure that the same is not a problem for FORMAT? I do not remember the details.

bioinformed:
@pd3: We've seen segfaults while attempting to validate our production pipeline due to this bug and have no other workaround except, perhaps, adding a call to `bcf1_sync()` earlier within `bcf_translate()`. 

We have tested `bcf_translate()` on records after FORMAT elements are deleted and have not seen segfaults.  However, they may still be possible, because our test cases are all single sample and we delete both FORMAT tag and values.  In fact, this code in `bcf_update_format()` suggests that the same kind of fault should happen, since `fmt.p = NULL` and `bcf1_sync` is not called:

```c
    if ( !n )   // n == 0 when deleting a tag
    {
        if ( fmt )
        {
            // Mark the tag for removal, free existing memory if necessary
            if ( fmt->p_free )
            {
                free(fmt->p - fmt->p_off);
                fmt->p_free = 0;
            }
            line->d.indiv_dirty = 1;
            fmt->p = NULL;   // <--------------
        }
        return 0;
    }
```

I'll update the PR to account for this case as well, though I won't have time to fully explore why we're *not* seeing these failures in our code.  

(It is very difficult to understand what is going on with all of these internal pointers without any documentation.  I've 80% figured it out at various times, but never to the point where I felt like I really know what all of that machinery was doing.)

bioinformed:
@pd3: I've updated the PR as described above.  `bcf_translate()` still contains the following assertions:
```c
            assert( !info->vptr_free );  // INFO case
```
and
```c
            assert( !fmt->p_free );  // FORMAT case
```

Is it now safe to remove these assertions or is more work needed when these values are set?

pd3:
The asserts are there to prevent cases where the record has been modified (but not deleted) and the new value could not be stored in place. Therefore the asserts still could be triggered. 

The original intention was to allow `bcf_translate` on a newly loaded record only, before its fields are modified in any way. If this is not general enough, then my preference would be to change the API as described in the "other issues not addressed in this PR" above. 

Otherwise the pull request could get a tentative +0.5 for handling one new usage case.

bioinformed:
@pd3: Would a better partial fix be to add a call to `bcf1_sync()` early in `bcf_translate`?  For our use-cases, we either call `bcf_translate()` just after reading or just prior to writing, so this approach would have negligible performance impact.

I am willing to work on a separate PR to implement the approach mentioned in "Other issues not addressed", but that will solve the problem of managing the lifetime and use translation tables.  This PR will still be needed and it doesn't make sense to require a single PR for both.

And mostly for my own edification, would the correct fix for the realloc case be:
```diff
--- a/htslib/vcf.c
+++ b/htslib/vcf.c
@@ -2975,12 +2975,12 @@ int bcf_translate(const bcf_hdr_t *dst_hdr, bcf_hdr_t *src_hdr, bcf1_t *line)
         else    // must realloc
         {
             bcf_fmt_t *fmt = &line->d.fmt[i];
-            assert( !fmt->p_free );
             kstring_t str = {0,0,0};
             bcf_enc_int1(&str, dst_id);
             bcf_enc_size(&str, fmt->n, fmt->type);
             fmt->p_off = str.l;
             kputsn((char*)fmt->p, fmt->p_len, &str);
+            if( fmt->p_free ) free(fmt.p - fmt.p_off);
             fmt->p = (uint8_t*)str.s + fmt->p_off;
             fmt->p_free = 1;
             line->d.indiv_dirty = 1;
```

pd3:
Hi, sorry about the delay! I think the fix should work (except for the `fmt.` vs `fmt->` typos). The syncing should work as well, although it would be good to test how much this affects the performance when applied to VCFs with thousands of samples.

Out of curiosity, why cannot the translate be called on a fresh record, what is the use case for calling the translate only after the record has been modified?

bioinformed:
@pd3: Thanks for looking at the proposed fix.  I'll correct the `->` vs. `.` and add it to the patch.  (Too much time spent in Python/Cython).

There are two reasons why `bcf_translate` can't always be called on fresh records:

1. The pysam object API tries to hide as many implementation details as possible from the htslib C implementation as possible.  Restricting when a method can be called is not ideal and would push us to make header translation part of the fetch API.  While not awful, we'd rather not complicate a seemingly unrelated API when there is a relatively simple fix for what is actually broken.

2. The actual use case that triggered this failure involves writing records after merging several files, local phasing, performing allele clumping, and demultiplexed to one or more files.  These records originate from multiple files, are highly processed, and are written to multiple files with minimal header records.  Header translation just prior to writing is the most obvious solution, since the destination file isn't known until much later in processing.

bioinformed:
@pd3: Can you please re-review?  This is a necessary fix for being able to call `bcf_translate` more generally and I believe I have covered all cases.  

I'm also working on a separate and orthogonal PR to address the location and consistency of translation tables.

pd3:
Great thank you. I just reviewed, squashed the commits, added a test and will merge.

PR #569, opened 2017-07-24
make hfile_set_blksize declaration consistent

tt6746690:


daviesrob:
Strictly, the definition in `hfile.c` is:
```
int hfile_set_blksize(hFILE *fp, size_t bufsiz)
```
So `bufsiz` rather than `bufsize`.

I'll fix this to match.

PR #574, opened 2017-08-07
Add return value documentation to sam_read1()

ctsa:
For `sam_read1`, it appears that a return value of -1 indicates a non-erroneous end-of-file/stream, and other negative values indicate various errors, but there are no docs on this return value in `htslib/sam.h`.

If this function is part of the stable api, could this interface be documented? I submitted the following line as a minimum starting point.


mp15:
We really should document all the return values, this is a start though. Thank you

Issue #467, opened 2017-02-07
Improve CRC in htslib


jkbonfield:
The CRC32 supplied by zlib, as used in BAM and CRAM is ancient and slow.  (Why don't linux distributions replace the archaic zlib with one of the many newer and faster versions? Grrr!)

The remains of PR #361 after merging all the aux tag code is a faster CRC, but as mentioned in that PR it should be its own thing.  Since then I've found a better source of fast CRCs:  https://github.com/ebiggers/libdeflate

```
b21c634a	"../enwik9bwt1024"	830 ms	1203 MB/s     Zlib
b21c634a	"../enwik9bwt1024"	541 ms	1845 MB/s     Libdeflate -O3
b21c634a	"../enwik9bwt1024"	290 ms	3441 MB/s     Crc32 16-way (as PR361)
b21c634a	"../enwik9bwt1024"	76 ms	13139 MB/s    Libdeflate -O3 -march=native
```

The march=native build uses PCMULQDQ SIMD on x86-64 machines meaning it's around .62 cycles per byte.  This means we can speed up CRC further still, but not on default builds.  Libdeflate's code is a compatible license, so we'd need to decide whether to lift in the algorithm (just 1 or 2 files) or to add the option to externally link.

Probably externally link is a good choice as the library has a whole replacement of zlib and it's around 2.4x faster at both compression and decompression.

kyleabeauchamp:
I think this issue can be closed now, right?

Also, it I've got a working bioconda build ready to be merged with libdeflate (https://github.com/bioconda/bioconda-recipes/pull/8521).  The speed improvements (with conservative optimization settings) are still pretty substantial!

jkbonfield:
Yes well spotted and thanks.

PR #579, opened 2017-08-14
Added BGZF CRC32 checking.  Fixes #402.

### LINKED ISSUE #402: CRCs in BAM aren't validated
The presence of CRCs sounds like a nice prospect to have, to check for data corruptions, but looking at the code in bgzf.c more (and io_lib's equivalent alas too), it simply doesn't work that way.

Demonstration:

```
@ seq3a[samtools.../htslib]; samtools view -o _old.bam -b test/xx#pair.sam 
@ seq3a[samtools.../htslib]; (dd if=_old.bam bs=1 count=149;echo z;dd if=_old.bam bs=1 skip=151) > _new.bam
149+0 records in
149+0 records out
149 bytes (149 B) copied, 0.000744582 s, 200 kB/s
33+0 records in
33+0 records out
33 bytes (33 B) copied, 0.000584354 s, 56.5 kB/s
@ seq3a[samtools.../htslib]; samtools view -h _old.bam |md5sum
2349b1bbbb2e5af5e81a786e3a1c72b1  -
@ seq3a[samtools.../htslib]; samtools view -h _new.bam |md5sum
2349b1bbbb2e5af5e81a786e3a1c72b1  -

@ seq3a[samtools.../htslib]; zcat _old.bam | md5sum
40c962b3fec07ec96a59ff31c4872601  -
@ seq3a[samtools.../htslib]; zcat _new.bam | md5sum
gzip: _new.bam: invalid compressed data--crc error
40c962b3fec07ec96a59ff31c4872601  -
@ seq3a[samtools.../htslib]; 
```

So zcat can spot the invalid crc, but samtools glosses over it completely.  I believe the problem is the way it is implemented as manually stripping off the gzip wrapper and then asking zlib to process the raw deflate stream:

```
// Inflate the block in fp->compressed_block into fp->uncompressed_block
static int inflate_block(BGZF* fp, int block_length)
{
    z_stream zs;
    zs.zalloc = NULL;
    zs.zfree = NULL;
    zs.msg = NULL;
    zs.next_in = (Bytef*)fp->compressed_block + 18;
    zs.avail_in = block_length - 16;
    zs.next_out = (Bytef*)fp->uncompressed_block;
    zs.avail_out = BGZF_MAX_BLOCK_SIZE;

    int ret = inflateInit2(&zs, -15);
```

So it is given compressed_block + 18 (18 being the size of the gzip header + BC extra bytes) instead of compressed_block itself decoded using `inflateInit2(&zs, 31)`.  Oddly it also claims the modified size is `block_length -16` and not `block_length -18`, but I suspect that's simply a harmless bug.

I haven't investigated the performance impact of adding back CRC checking, nor whether it is as simple as it appears to be, but it's rather sobering given I'd always assumed the checksums meant that BAM would identify data corruptions.  (I suspect the main reason we spot corruptions is simply that it is very likely to break decompression.)

PS. I checked the exit status too, so it's not simply that this was the last block to be modified.  I can get the same error by corrupting the CRC in the header block too and samtools still continues to end of file and yields the same output.

jkbonfield:
Note this normally is an insignificant overhead, but if decoding
uncompressed BAMs it becomes dominant.  Eg perf on samtools flagstat
shows:

 70.47%  samtools.tmp  libz.so.1.2.8       [.] crc32
  8.91%  samtools.tmp  [kernel.kallsyms]   [k] 0xffffffff8104f45a
  8.40%  samtools.tmp  libc-2.19.so        [.] __memcpy_sse2_unaligned
  4.80%  samtools.tmp  samtools.tmp        [.] bgzf_read
  3.95%  samtools.tmp  samtools.tmp        [.] bam_read1

One solution is either to have an option to not check (I don't like
this), or to replace the default zlib crc algorithm with a much more
performant alternative.  See #467 for discussion.  Libdeflate's SSE
crc is around 11x faster, meaning this overhead would be around 20%
instead.

Perhaps it's time we had a build option linking against libdeflate
anyway as it's almost certainly better for all of the work instead of
just this bit.
https://encode.ru/threads/2482-libdeflate-a-new-optimized-DEFLATE-implementation

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -39,6 +39,7 @@
 #include "htslib/bgzf.h"
 #include "htslib/hfile.h"
 #include "htslib/thread_pool.h"
+#include "htslib/hts_endian.h"
 #include "cram/pooled_alloc.h"
 
 #define BGZF_CACHE
@@ -460,6 +461,16 @@ static int inflate_block(BGZF* fp, int block_length)
                               (Bytef*)fp->compressed_block + 18, block_length - 18);
     if (ret < 0) {
         fp->errcode |= BGZF_ERR_ZLIB;
+        return -1;
+    }
+
+    // Check CRC of uncompressed block matches the gzip header.
+    // NB: we may wish to switch out the zlib crc32 for something more performant.
+    // See PR#361 and issue#467
+    uint32_t c1 = crc32(0L, (unsigned char *)fp->uncompressed_block, dlen);
+    uint32_t c2 = le_to_u32((uint8_t *)fp->compressed_block + block_length-8);
+    if (c1 != c2) {
+        fp->errcode |= BGZF_ERR_CRC;
         return -1;
     }
 
--- a/htslib/bgzf.h
+++ b/htslib/bgzf.h
@@ -49,6 +49,7 @@ extern "C" {
 #define BGZF_ERR_IO     4
 #define BGZF_ERR_MISUSE 8
 #define BGZF_ERR_MT     16 // stream cannot be multi-threaded
+#define BGZF_ERR_CRC    32
 
 struct hFILE;
 struct hts_tpool;


kyleabeauchamp:
Q: the libdeflate improvements would not improve CRAMv3 performance because that has custom compression code, right?

jkbonfield:
Not quite.

Firstly, CRAM also does CRC checking and zlib's CRC is 10x slower than the libdeflate implementation.  Hence faster CRC benefits both formats.

Secondly CRAM can still (and often does) make use of the Deflate algorithm, eg for read names, although this is a much reduced percentage compared to the work load of BAM, BCF etc (which is 100% deflate).  It also has the option to use bzip2 or lzma, but most people aren't yet doing so I believe.

jkbonfield:
As a follow up to that, `samtools view -f 0xffff foo.cram` to benchmark reading (with no writing - all filtered out) vs io_lib's equivalent `scram_flagstat -b foo.cram` gives around 29% faster speed to Scramble.  This is, in part, down to the use of a better CRC implementation, but there are also other speedups (rans codec) I think which weren't migrated over for some reason.  (I'd still recommend htslib as the canonical and most stable CRAM support.)

So in CRAM around 5.5% of the samtools view was CRC and 6% inflate.    Compare this to BAM where 75% is inflate and 17% is CRC.

PR #582, opened 2017-08-18
Update multipart to work with revised htsget specification

daviesrob:
PR samtools/hts-specs#230 updates the htsget specification so that all htsget JSON starts with a protocol identifier.  This is implemented by wrapping the payload inside a JSON object with the single key "htsget".  Update multipart.c and hts_detect_format() in hts.c to understand the revised format.

As the protocol now has an official name, change some mentions of 'json' to 'htsget'.  One of these is in the htsExactFormat enumerated type, which is in htslib/hts.h.  This is left in place for API compatibility, but marked as deprecated (if using a compiler that supports this).

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -109,7 +109,7 @@ static enum htsFormatCategory format_category(enum htsExactFormat fmt)
     case bed:
         return region_list;
 
-    case json:
+    case htsget:
         return unknown_category;
 
     case unknown_format:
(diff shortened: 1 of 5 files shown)


jmarshall:
> One of these is in the htsExactFormat enumerated type, which is in htslib/hts.h

Did you consider just adding a new `htsget_json` or so enumerator alongside, leaving the existing `json` as a currently temporarily unused placeholder?  There are other such placeholders in `htsExactFormat` already. Deprecating `json` (and inventing machinery to do so!) seems unnecessary.

daviesrob:
I was attempting to leave the numeric value unchanged, for ABI compatibility.  This meant `json` was in the way.  Although I doubt anything outside htslib actually uses it at the moment, so adding a new value might not be too harmful.

Issue #580, opened 2017-08-14
hts.c test_and_fetch ignores fwrite errors


depristo:
We recently discovered an bug in the test_and_fetch routine in hts.c. We were running on a filesystem that doesn't allow large files and were directly processing a BAM file stored in a GCS bucket. The automatic detection of the index worked as expected, but the run unexpected crashed.

The root cause is the unchecked call to fwrite in:

https://github.com/samtools/htslib/blob/cc55632993313cf4c23be82f8b13f7113d61e675/hts.c#L2285

That call should be updated to check that l bytes have been written and raise an appropriate error if a write failure occurred.

daviesrob:
Yes, it should be checking.  Pull request incoming...

PR #583, opened 2017-08-22
Make hts_test_and_fetch check for fwrite() errors

daviesrob:
Plus a few other changes to make it and hts_idx_getfn() work a bit more sensibly.  For example, the latter will now notice if the input file name doesn't have a '.' in it when trying to remove the suffix.

Fixes issue #580

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -2258,8 +2258,8 @@ int hts_itr_next(BGZF *fp, hts_itr_t *iter, void *r, void *data)
 //         -2 on other errors
 static int test_and_fetch(const char *fn, const char **local_fn)
 {
-    hFILE *fp_remote;
-    FILE *fp = NULL;
+    hFILE *remote_hfp;
+    FILE *local_fp = NULL;
     uint8_t *buf = NULL;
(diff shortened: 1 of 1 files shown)


depristo:
Thanks for fixing this issue so quickly Rob.

hts.c
```diff
@@ -2254,14 +2254,17 @@ int hts_itr_next(BGZF *fp, hts_itr_t *iter, void *r, void *data)
 /**********************
  *** Retrieve index ***
  **********************/
-
-static char *test_and_fetch(const char *fn)
+// Returns -1 if index couldn't be opened.
+//         -2 on other errors
+static int test_and_fetch(const char *fn, const char **local_fn)
 {
+    hFILE *fp_remote;
```
valeriuo: The naming of the variables is a bit inconsistent. Could we have either `fp_remote` and `fn_local` or `remote_fp` and `local_fn`?

daviesrob: Agreed.  I've pushed a new commit to make them more consistent.

PR #589, opened 2017-09-08
Abstract ref_to_m5 from cram_io

ThomasHickman:
This pull request extracts the function `m5_to_ref` from `cram_io` and adds tests. `m5_to_ref` and helper functions for it are added in a new file called `ref.c`.

I've also added a new `hFILE` extension called hFILE_ref, internal to ref.c as a wrapper hFILE to carry information required by `cram_populate_ref` (such as file name, mFILE handle for MMapped files etc.). and also implemented the proposed in-memory I/O proposed in #417, so that this `m5_to_ref` returns a hFILE. Sorry about the number of commits, it took a bit of trying to get this working - I could squash the commits down into one if necessary


ThomasHickman:
I've squashed the commits into two commits and made a pull request for the in-memory I/O at #590. The original history is now at https://github.com/wtsi-hgi/htslib/tree/ref-old

ThomasHickman:
I've had a look at the double buffering problem - it's hard to stop buffering happening in a hFILE, so I think it's best to have a implement a "wrapper hFILE", which contains extra information, but redirects all hFILE calls to the wrapper object. I've implemented this above and made `hFILE_ref`s a hFILE wrapper

jkbonfield:
Now the mem-io branch has been merged, is it possible to untangle this PR by removing those merges and rebasing on develop?  I gave it a try, but it's a bit tricky and I ended up with something that didn't compile!  You understand this bit better than me.

hfile.c
```diff
@@ -459,6 +488,14 @@ off_t hseek(hFILE *fp, off_t offset, int whence)
 
 int hclose(hFILE *fp)
 {
+    if(fp->is_hFILE_wrapper){
+        fp->backend->close(fp);
```
jkbonfield: This generates a warning:

```
hfile.c: In function 'hclose':
hfile.c:493:27: warning: ignoring return value of function declared with attribute warn_unused_result [-Wunused-result]
         fp->backend->close(fp);
```

ref.c
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    mkdir_prefix(path, mode);
+    mkdir(path, mode);
+    chmod(path, mode);
+    *cp = '/';
+}
+
+int ref_close(hFILE* hf){
+    int error_code = 0;
```
jkbonfield: Unused variable

ref.c
```diff
@@ -0,0 +1,401 @@
...
+    hFILE_ref *hfref = (hFILE_ref*)hf;
+    free(hfref->file_name);
+
+    if(hfref->mf)
+        return mfclose(hfref->mf);
+}
```
jkbonfield: Undefined return value here.  Please add an else clause.

ThomasHickman:
I've rebased the current `develop` branch onto this, and cleaned up what's in the `Abstract ref_to_m5 from cram_io` and `Add hFile wrappers` commits, so that whitespace isn't changed and commits change what they say.
I've also reverted the changes in `open_trace_file.c`, as getting the filename that this returns isn't necessary, now I've implemented hFILE wrappers, fixed some other issues that I found while looking though the code and fixed the warnings reported by @jkbonfield

ThomasHickman:
For ease of reviewing the code, this is what I've done in this pull request:

1. Moved get_cache_basedir, expand_cache_path, get_int_threadid, mkdir_prefix to ref.c, all of which were helpers functions of parts of cram_populate_ref which dealt with ref fetching.
2. Created `m5_to_ref` from `cram_populate_ref`, moving over the ref fetching code.
3. Made `cram_populate_ref` compatiable with results returned by `m5_to_ref`
3. Created hFILE wrappers, which are hFILEs which don't contain a buffer or a file pointer but contains extra information fields and a pointer to another hFILE. All public calls are redirected to be in terms or the inner hFILE, except `hclose` which calls both the wrapper and the inner hFILE's close function. Backends that are hFILE_wrappers are created with only a close method in the hFILE_backend structure and the open method calls `hfile_init_wrapper` instead of `hfile_init`.
4. Created a hFILE_ref, which is a hFILE wrapper to store information needed by `cram_populate_ref` and to store information to correctly close mFILEs on hclose.

If it would be easier to review, I could also move the changes that implement hFILE_wrappers into a new pull request.

valeriuo:
Please make sure that all tests pass, on both HTSlib and samtools. I get 13 failures when building samtools (develop) against this HTSlib branch.

ThomasHickman:
In the above commits:
- I've fixed the test failures in samtools, https://github.com/wtsi-hgi/htslib/blob/51a3095d02ee1acb9c19f20314b32d863d0f75de/cram/cram_io.c#L1929 was getting  executed when the M5 field wasn't present, leading to a segfault.
- I've created an option: `disable_http_tests`, which disables http fetching tests. This option is implicitly enabled when libcurl is not detected (or `./configure` has not been executed), so the CI tests should work.  This option is enabled in all tests by default, but I don't think this is going to be much of a problem - as this is only fetching https://www.ebi.ac.uk/ena/cram/md5/bbf4de6d8497a119dda6e074521643dc 2 times.
- I've changed the interface of `m5_to_ref` to return the reference file and return NULL when there's an error. I'll update https://github.com/samtools/samtools/pull/748 with the new implementation

configure.ac
```diff
@@ -121,6 +121,11 @@ AC_ARG_ENABLE([s3],
                   [support Amazon AWS S3 URLs])],
   [], [enable_s3=check])
 
+AC_ARG_ENABLE([http-tests],
+  [AS_HELP_STRING([--disable-http-tests],
+                  [disables tests that require querying to a http server])],
+  [disable_http_tests=yes], [disable_http_tests=no])
```
valeriuo: Also, it was decided that HTTP should be disabled by default. So the option should be renamed `--enable-http-tests` and the associated variable should only be set to true when this option is selected.

ref.c
```diff
@@ -0,0 +1,406 @@
...
+    char* seq;
+    seq = mfsteal(mf, &sz);
+
+    if(!seq){
+        seq = mf->data;
+    }
+    else{
+        mf = NULL;
+    }
+
```
jkbonfield: I'm unsure about the above.

`mfsteal` returns the buffer and size from the mFILE and then destroys (frees) the mFILE.  It's designed to turn an mFILE into a C buffer basically.  Thus after the `mfsteal` call `mf` is now undefined and we cannot dereference it and nor should we pass it to `open_hfile_ref`.

ref.c
```diff
@@ -0,0 +1,406 @@
...
+            thrid++; // Ensure filename changes even if time/clock haven't
+
+            sprintf(path_tmp, "%s.tmp_%d_%u_%u", path, pid, thrid, t);
+            fp = hopen(path_tmp, "wx");
+        } while (fp == NULL && errno == EEXIST);
+        if (!fp) {
+            perror(path_tmp);
+
+            // Not fatal - we have the data already so keep going.
+        }
```
jkbonfield: We lost a `return 0` (NULL now) in the `if (!fp)` clause from the original cram_io.c copy.  The intent was if we fail to create a local temporary file name to cache the reference, then so be it - we do still have our reference and the code can continue to work, but for whatever reason (eg permission denied) being able to write to the cache is a non-fatal condition.

However continuing means we end up calling hwrite, hfclose etc on NULL.

test/test-ref.c
```diff
@@ -0,0 +1,60 @@
...
+        return EXIT_FAILURE;
+    }
+
+    if(hclose(ref) != 0){
+        fprintf(stderr, "Cannot close hfile\n");
+        return EXIT_FAILURE;
+    }
+
+    return error_code;
+}
```
jkbonfield: Please add a newline at the end of file.

ref.c
jkbonfield: Hmm looking further I see how it works.  It only uses mf if mfsteal failed, otherwise it sets mf to NULL.  mfsteal can only fail if mfdetach fails, which in turn only fails if we're using mmap.

I think the logic in this function is therefore correct (sorry), but I'm wondering if it's a symptom of my mfsteal being too complex in error handling.

At the very least we probably should do something like this to get to a consistent point whether or not the steal fails:

```
if (!seq) {
    seq = mf->data;
    mf->data = NULL;
    mfdestroy(mf);
}
```

and remove the mf parameter from `open_file_href`.  This all ties up with the ref_close comments though too, as that's where `href->mf` is used.

jkbonfield: Sorry for spamming.  I've finally figured out what's happening here and I feel guilty for commenting on what turns out to be entirely my own demons!

So, just incase anyone else who is reading this cares...

The original code in cram_populate_ref was:

```
    /* Otherwise search full REF_PATH; slower as loads entire file */
    if ((mf = open_path_mfile(tag->str+3, ref_path, NULL))) {
        size_t sz;
        r->seq = mfsteal(mf, &sz);
        if (r->seq) {
            r->mf = NULL;
        } else {
            // keep mf around as we couldn't detach                                                              
            r->seq = mf->data;
            r->mf = mf;
        }
        r->length = sz;
        r->is_md5 = 1;
    } else {
```

(The comment would probably have been useful to keep, albeit with explanation that mfdetach won't free the data if we used mmap.)

Basically if the `open_path_mfile` used mmap on an md5 file then we need to keep the file descriptor open as we need it for munmap later on.  Otherwise once it's open we can just steal the pointer can close it.  This would have been the original way it was done, with mmap appearing later on.  With hindsight it probably should have kept mf valid in all scenarios, for simplicity, and never attempted the mfsteal part at all.   We probably needed a "mf_close_if_able" function or similar that closed the file descriptor if not mmaped so it's no longer consuming resources, but keep the rest of the mFILE struct present.

Hence hFILE has now taken the role as replacement for `ref_entry *r` and `r->mf`.  It's the same code as my original, just with different structure names.  Sorry Thomas. :)

I think what all of this points towards is that mFILE is overkill for what we need here.  It was designed as a transparent drop in for stdio FILE I/O so it supports read/write, seeking, update, etc.  We really just need a load-or-mmap function, which is what the hfile mem layer provides (minus the missing mmap bit).

PS. My head hurts!

ref.c
jkbonfield: I have validated that this is an issue and that adding `return 0` after the comment fixes it.
Try setting REF_CACHE to a directory that you don't have write access to and running the test-ref tool.  It'll segmentation fault.

cram/cram_io.c
```diff
@@ -1901,326 +1903,109 @@ int cram_set_header(cram_fd *fd, SAM_hdr *hdr) {
...
-	hts_md5_hex(md5_buf2, md5_buf1);
-
-	if (strncmp(tag->str+3, md5_buf2, 32) != 0) {
-	    hts_log_error("Mismatching md5sum for downloaded reference");
-	    hclose_abruptly(fp);
-	    unlink(path_tmp);
-	    return -1;
-	}
+            r->seq = seq;
+            r->mf = hfref->mf;
```
jkbonfield: As we have `hfref=ref` above on line 1981 and `free(ref)` at line 2001, this is accessing freed memory.  The free statement needs moving down a few lines.

jkbonfield: Furthermore, the `free(ref)` isn't sufficient.  If ref is an hFILE_wrapper then we also need to free the innerhf too. Eg:

```
            if (ref->is_hFILE_wrapper)
                free(((hFILE_wrapper *)ref)->innerhf); // hack!                                                  
            free(ref); // Can't hclose as this would free the buffer                                             
```

This is all feeling like the wrong solution and it's going to fail if we ever have other forms of wrappers, but without the above we leak memory.

I'm trying to recall the reasoning for the wrapper and why it can't just be an `hFILE_mem` hfile class without any wrapping up.  I think the wrapper is there because we have another type (`hFILE_ref`), but perhaps we just want hFILE_mem expanded to be able to do the same as the ref one.

The ref version adds mf and file_name fields.  mf is only needed due to mmap and the requirement to unmap at the end.  File_name I can't see a use for in this code.   If we have requirements for these then we may be better off adding them both to an extended hFILE_mem structure instead of introducing an entirely new wrapper layer.

ThomasHickman: @jkbonfield The use of hfile_wrappers comes from the need to use the existing code in cram_populate_ref and functions that call it, without much modification - I didn't want to rewrite the logic in `cram_populate_ref` and `cram_get_ref` (and maybe other places). This required a structure that could hold both in-memory structures to be contained as well holding arbitrary hFiles generated from https://github.com/wtsi-hgi/htslib/blob/1a6a7e242d75c9b746653ee548d2bce430cb905b/ref.c#L304. The file name, length and mFile (if used) also needed to be stored, where the file name and length are used here: https://github.com/wtsi-hgi/htslib/blob/1a6a7e242d75c9b746653ee548d2bce430cb905b/cram/cram_io.c#L1989 and here: https://github.com/wtsi-hgi/htslib/blob/1a6a7e242d75c9b746653ee548d2bce430cb905b/cram/cram_io.c#L1983. 

I could rewrite relevant code such that `m5_to_ref` would use an extended in-memory hFile pointing either memory or mmapped data, where this extended structure would contain a flag indicating whether this is an mmapped structure and it's length. However, it may be better to work to merge this PR as it is with the inelegant implementation, then refractor the code at a later time.

PR #598, opened 2017-09-26
Fix cram integer overflows

### LINKED ISSUE #593: Various CRAM related issues
Here are some additional fuzzing found issues in htslib.

Settings have changed a bit, notably int overflows are caught at the time of the overflow, so the actual impact of the overflow has not been determined. There might be security consequences, or it might be benign, unfortunately I don't have time to go past a simple explanation per bug (and you have a better understand than I anyway).

As before, I am using view_sam/view_vcf on fuzzed inputs. We do not have the very latest version of the code so apologies in advance if things have been fixed already.

daviesrob:
Fixes #593

The second case in the issue above shows a bug in how huffman table entries are sorted which could be triggered by valid data.  If two integer values are more than 2^32 apart (say, 2000000000 and -20000000000) then they would compare and sort incorrectly.  As the same comparison is used when reading and writing, it's possible that HTSlib could create data that was invalid, but it could read itself.  Hopefully we haven't written anything like that as we tend to avoid using the Huffman codec.  I haven't yet been able to work out if htsjdk can produce the same incorrect output (it uses the same subtraction trick when making the comparison).

In the third case, a RAW block had different compressed and uncompressed sizes.  This enforces the rule (in the specification) that they should be the same.

### DIFF
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -1063,7 +1063,7 @@ cram_codec *cram_huffman_decode_init(char *data, int size,
     qsort(codes, ncodes, sizeof(*codes), code_sort);
 
     /* Assign canonical codes */
-    val = -1, last_len = 0, max_val = 1;
+    val = -1, last_len = 0, max_val = 0;
     for (i = 0; i < ncodes; i++) {
 	val++;
         if (val > max_val)
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -855,7 +855,7 @@ static int code_sort(const void *vp1, const void *vp2) {
     if (c1->len != c2->len)
 	return c1->len - c2->len;
     else
-	return c1->symbol - c2->symbol;
+	return c1->symbol < c2->symbol ? -1 : (c1->symbol > c2->symbol ? 1 : 0);
 }
 
 void cram_huffman_decode_free(cram_codec *c) {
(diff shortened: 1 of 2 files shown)


jkbonfield:
I assume you mean more than 2^31 apart and not 2^32?  It looks initially like code lengths have the same issues, but the way they're generated means they can never be large.   Theoretically the maximum code length of a genuine huffman code in a block of length L is the Lth (or Lth-1?) fibonacci number.  Thus is say 100Mb of data it could be possible to generate long codes, albeit extremely unlikely.  I don't know if we block that case from happening.   I don't think the spec has an upper limit on code length, but to be honest HUFFMAN should be retired anyway so I'm not inclined to spend too much time on that part of the spec.

cram/cram_codecs.c
```diff
@@ -1050,24 +1047,32 @@ cram_codec *cram_huffman_decode_init(char *data, int size,
...
+        free(codes);
+        return NULL;
     }
 
     /* Sort by bit length and then by symbol value */
     qsort(codes, ncodes, sizeof(*codes), code_sort);
 
     /* Assign canonical codes */
-    val = -1, last_len = 0;
+    val = -1, last_len = 0, max_val = 1;
```
jkbonfield: About the only place huffman tables are now generated in htslib is for zero length codes as we have no CRAM codec capable of stating something is a constant.  A huffman tree of one symbol (zero length) is the closest we get.

What if we have two symbols claiming zero length?  At this point we have two loops through here with val 0, val 1, but the second one doesn't trigger "val > max_val" error.  Should max_val be 0 to start with?

PS.  The rest of the PR looks fine.

daviesrob:
Yes, that should be 2^31.  I've put in a check for hitting the implementation limit on the code length (there is no limit in the specification).  It's unlikely to be hit in practice, unless someone tries to make an epic Huffman table covering all possible 32 bit integers.  If they do, at least they'll know we can't read it now.

I agree, max_val should start at zero to disallow more than one zero-length code.  I have pushed a commit to do that.

PR #599, opened 2017-10-04
Allow overriding the cURL CA bundle path

gkelly:
Allow overriding the cURL CA bundle via the CURL_CA_BUNDLE environment
variable. This allows users to specify a new CA root certificate to use
when communicating with a host (such as an htsget server using a
self-signed certificate).


gkelly:
@kemp-ggh

daviesrob:
Thanks.  I'll merge this for you.

PR #600, opened 2017-10-04
Add Bearer token support to hfile_libcurl (for htsget)

daviesrob:
Allows you to set `HTS_AUTH_LOCATION` to point to a file or URL where HTSlib can find an OAUTH2 Bearer token (see [RFC 6750](https://tools.ietf.org/html/rfc6750).  `%h` in `HTS_AUTH_LOCATION` will be substituted with the host (and currently port) part of the URL, so that tokens can be sent to specific sites.

The file can either contain just the token (e.g. `mF_9.B5f-4.1JqM`) or an OAUTH2 access token response:
```JSON
{
       "access_token":"mF_9.B5f-4.1JqM",
       "token_type":"Bearer",
       "expires_in":3600,
       "refresh_token":"tGzv3JOkF0XG5Qx2TlKWIA"
}
```

`HTS_AUTH_LOCATION` can be a URL as well as file path.  For URLs, care is taken to avoid trying to read an access token for them (otherwise infinite recursion could result).  Allowing URLs enables a way of implementing authentication agents that handle issues like token refresh.

### DIFF
--- a/hfile_libcurl.c
+++ b/hfile_libcurl.c
@@ -259,6 +259,7 @@ static void libcurl_exit()
         for (i = kh_begin(curl.auth_map); i != kh_end(curl.auth_map); ++i) {
             if (kh_exist(curl.auth_map, i)) {
                 free_auth(kh_value(curl.auth_map, i));
+                kh_key(curl.auth_map, i) = NULL;
                 kh_value(curl.auth_map, i) = NULL;
             }
         }
@@ -511,7 +512,7 @@ static int add_auth_header(hFILE_libcurl *fp) {
     if (renew_auth_token(fp->headers.auth, &changed) < 0)
         goto unlock_fail;
 
-    if (!changed) {
+    if (!changed && fp->headers.auth_hdr_num < 0) {
         pthread_mutex_unlock(&fp->headers.auth->lock);
         return 0;
     }
@@ -616,6 +617,7 @@ static int get_auth_token(hFILE_libcurl *fp, const char *url) {
                 free_auth(tok);
                 tok = NULL;
             }
+            kh_value(curl.auth_map, idx) = tok;
         }
     }
     pthread_mutex_unlock(&curl.auth_lock);
@@ -965,7 +967,6 @@ static int libcurl_close(hFILE *fpv)
         fp->headers.callback(fp->headers.callback_data, NULL);
     free_headers(&fp->headers.fixed, 1);
     free_headers(&fp->headers.extra, 1);
-    free(fp->headers.auth);
 
     if (save_errno) { errno = save_errno; return -1; }
     else return 0;
--- a/hfile_libcurl.c
+++ b/hfile_libcurl.c
@@ -365,6 +365,12 @@ static int add_callback_headers(hFILE_libcurl *fp) {
     return -1;
 }
 
+/*
+ * Read an OAUTH2-style Bearer access token (see
+ * https://tools.ietf.org/html/rfc6750#section-4).
+ * Returns 'v' for valid; 'i' for invalid (token missing or wrong sort);
+ * '?' for a JSON parse error; 'm' if it runs out of memory.
+ */
 static int read_auth_json(auth_token *tok, hFILE *auth_fp) {
     hts_json_token t;
     kstring_t str = {0, 0, NULL};
--- a/hfile_libcurl.c
+++ b/hfile_libcurl.c
@@ -1103,8 +1103,6 @@ error:
     return NULL;
 
 early_error:
-    save = errno;
-    errno = save;
     return NULL;
 }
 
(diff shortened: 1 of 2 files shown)


hfile_libcurl.c
```diff
@@ -312,6 +365,265 @@ static int add_callback_headers(hFILE_libcurl *fp) {
...
+        } else if (strcmp(t.str, "expires_in") == 0) {
+            if ((ret = hts_json_fnext(auth_fp, &t, &str)) != 'n') goto error;
+            expiry = ks_release(&str);
+        } else if (hts_json_fskip_value(auth_fp, '\0') != 'v') {
+            ret = '?';
+            goto error;
+        }
+    }
+
+    if (!token || (type && strcmp(type, "Bearer") != 0)) {
```
jkbonfield: https://tools.ietf.org/html/rfc7230#section-3.2 states that http headers are case insensitive.  This should probably be another strcasecmp.  Similarly elsewhere.

hfile_libcurl.c
```diff
@@ -312,6 +365,265 @@ static int add_callback_headers(hFILE_libcurl *fp) {
     return -1;
 }
 
+static int read_auth_json(auth_token *tok, hFILE *auth_fp) {
```
jkbonfield: This could do with a comment explaining the cryptic return values of ?, i, v and m.  The only one checked for is v apparently.

hfile_libcurl.c
daviesrob: 2. Is fixed.
1. Is beyond the scope of this PR.

daviesrob: These aren't HTTP headers, they're JSON values.  You need to be looking at [RFC 6749](https://tools.ietf.org/html/rfc6749) and [RFC 6750](https://tools.ietf.org/html/rfc6750).
I can't find anything to say they are supposed to be case-insensitive, although I suppose it wouldn't do any harm to make them.

jkbonfield:
I've finished the review, but it could do with documentation adding somewhere.  Possibly (albeit not ideally) to samtools.1 man page as we don't yet have any htslib manuals.  Specifically the environment variables, the use of %h and how it all hangs together.  I'm happy for that to be a separate PR though, and indeed it'd need to be if it goes to samtools.1 anyway.

dkj:
If no `Authorization`  header is provided in the htsget JSON then this will (from my efforts at testing, not looking at code) fall back to using a token specified in the `HTS_AUTH_LOCATION`. I think that `Authorization`  header for the second stage of htsget should only come from the htsget JSON and not from `HTS_AUTH_LOCATION`.

dkj:
@daviesrob, I'm anticipating (but not sure of) scenario where end user sets and populates a plain (no `%h`) `HTS_AUTH_LOCATION` and htsget  server returns JSON with urls to some other third party storage/bucket server using signatures in the headers rather than an `Authorization` header so that the token will be leaked to that 3rd party storage system. I think this is low risk, and can be remediated by the user with the use of `%h` in token locations, so probably not be a blocker, but recommend it become a new issue.

juhtornr:
It seems that authorisation headers are sent only to the first URL provided. For example when trying to access two different htsget URLs using samtools:

[cloud-user@processing-demo ~]$ echo -e "HTTP/1.1 200 OK\n\n" | nc -l 3333 &
[1] 29155
[cloud-user@processing-demo ~]$ echo -e "HTTP/1.1 200 OK\n\n" | nc -l 3334 &
[2] 29157
[cloud-user@processing-demo ~]$ samtools mpileup 'http://localhost:3333/1/' 'http://localhost:3334/2/'
GET /1/ HTTP/1.1
User-Agent: htslib/0.2.0-rc6-1119-g23e7c3e libcurl/7.29.0
Host: localhost:3333
Accept: */*
Authorization: Bearer xyz

GET /2/ HTTP/1.1
User-Agent: htslib/0.2.0-rc6-1119-g23e7c3e libcurl/7.29.0
Host: localhost:3334
Accept: */*

[mpileup] 2 samples in 2 input files
<mpileup> Set max per-file depth to 4000
[1]-  Done                    echo -e "HTTP/1.1 200 OK\n\n" | nc -l 3333
[2]+  Done                    echo -e "HTTP/1.1 200 OK\n\n" | nc -l 3334

[cloud-user@processing-demo ~]$ cat ~/tokens/default 
xyz
[cloud-user@processing-demo ~]$ export | grep HTS
declare -x HTS_ALLOW_UNENCRYPTED_AUTHORIZATION_HEADER="I understand the risks"
declare -x HTS_AUTH_LOCATION="/home/cloud-user/tokens/default"

daviesrob:
@jkbonfield , @juhtornr   Problems with token caching should now be fixed.
@dkj The HTS_AUTH_LOCATION token should be suppressed now for htsget second-stage fetches.

PR #601, opened 2017-10-06
Avoid memory double-free on vhopen_libcurl failure

ramyala:
libcurl_open cleans up headers completely on failure (see hfile_libcurl.c:761).
So, its not required to call free_headers on failure in vhopen_libcurl.


ramyala:
This was addressed in the develop branch.

PR #602, opened 2017-10-06
Add support for TCGA GDC URL schema handler

ramyala:
Support performing samtools read-only operations directly on
TCGA Genemic Data Commons repository. URLs need to be of format:

tcga://UUID

Range operations are also supported by querying and downloading
the corresponding index file from TCGA.


ramyala:
bump?

daviesrob:
I'm not sure if we want to include this in the main htslib repository.  We really don't want to encourage having non-standard URL schemas for every single sequence archive in the world.  Ideally they should all converge on something like [htsget](http://samtools.github.io/hts-specs/htsget.html) as a standard way of managing sequence retrieval and access control.

Meanwhile, it might be better to build this as a stand-alone plug-in. You can see a couple of examples of this in the [htslib-plugins](https://github.com/samtools/htslib-plugins) repository.

lh3:
I am with @daviesrob. htsget will be the way forward.

PR #603, opened 2017-10-12
Fix annotations: hash [map] => hash [set]

16m:
- KHASH_SET_INIT_INT annotations is correct, while KHASH_SET_INIT_INT64 is not. KHASH_SET_INIT_INT64 should be hash set, not map, just as KHASH_SET_INIT_INT is.
```
 /*! @abstract     Instantiate a **hash set** containing integer keys */
 #define KHASH_SET_INIT_INT(name)
 
 /*! @abstract     Instantiate a **hash map** containing integer keys*/
 #define KHASH_MAP_INIT_INT(name, khval_t)
 
 /*! @abstract     Instantiate a **hash map** containing 64-bit integer keys*/
 #define KHASH_SET_INIT_INT64(name)
```

### DIFF
--- a/htslib/khash.h
+++ b/htslib/khash.h
@@ -594,7 +594,7 @@ static kh_inline khint_t __ac_Wang_hash(khint_t key)
 	KHASH_INIT(name, khint32_t, khval_t, 1, kh_int_hash_func, kh_int_hash_equal)
 
 /*! @function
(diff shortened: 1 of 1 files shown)


daviesrob:
Thanks.  You are correct.  They are sets.

Issue #604, opened 2017-10-13
htslib http seek() doesn't fail on errors


ramyala:
The following python code would call into htslib seek() and hts_iter_next() for reads. When libcurl_seek fails the following code still passes and leads to erronous output. I've noticed this happen when connections fail with ssl issues or 503 Service Unavailable errors. Ideal behavior should be for libcurl_seek to return a failed code instead of success which would allow an exception to be triggered on pysam (so user can either re-establish connection or handle the issue gracefully).

```
num_reads = 0
for read in pysam.fetch(chrom, start, end):
   process(read)
   num_reads += 1
print "my_process:Total reads: ", num_reads
```

```
[I::hseek] fp: 0xc6d37e8520, mode: SET, seek to: 5895168049 
* Found bundle for host gdc-api.nci.nih.gov: 0xc6d3816310 [can pipeline] 
* Hostname gdc-api.nci.nih.gov was found in DNS cache 
*   Trying 192.170.230.228... 
* TCP_NODELAY set 
* Connected to gdc-api.nci.nih.gov (192.170.230.228) port 443 (#71) 
* found 664 certificates in /etc/ssl/certs 
* ALPN, offering http/1.1 
* gnutls_handshake() failed: Error in the pull function. 
* Curl_http_done: called premature == 1 
* stopped the pause stream! 
* Closing connection 71 
INFO:my_process:Total reads: 0 ```

Error 2:

```
[I::hts_itr_query] idx: 0x6525a40da0, tid: 6, beg: 51396407, end: 51396425 
[I::hts_itr_next] GRAB MORE: bgzf_fp: 0x65249c7a10, iter: finished: 0, read_rest: 0, curr: (0:0-0@0x0), inst: (6:51396407-51396425@1), i: -1, n_off: 1, offset: 0x65249642d0 
[I::hseek] fp: 0x65239704d0, mode: SET, seek to: 4518809043 
* Found bundle for host gdc-api.nci.nih.gov: 0x652399b310 [can pipeline] 
* Hostname gdc-api.nci.nih.gov was found in DNS cache 
*   Trying 192.170.230.228... 
* TCP_NODELAY set 
* Connected to gdc-api.nci.nih.gov (192.170.230.228) port 443 (#49) 
* found 664 certificates in /etc/ssl/certs 
* ALPN, offering http/1.1 
* SSL connection using TLS1.2 / ECDHE_RSA_AES_256_GCM_SHA384 
* 	 server certificate verification OK 
[...]

PR #606, opened 2017-10-16
Error on seek failure when iterating over reads

ramyala:
Reference #604 

libcurl_seek can fail for several reasons and frequently a retry is required (eg. the remote server is under high load).

Currently, errno is set to ESPIPE and any further seeks are disabled indiscriminately. Instead, ESPIPE should be set on errors where Range requests are unsupported (and perhaps other curated scenarios). This would allow application logic to handle the error appropriately.

### DIFF
--- a/hfile.c
+++ b/hfile.c
@@ -452,6 +452,7 @@ off_t hseek(hFILE *fp, off_t offset, int whence)
     // Seeking succeeded, so discard any non-empty read buffer
     fp->begin = fp->end = fp->buffer;
     fp->at_eof = 0;
+    fp->has_errno = 0;
 
     fp->offset = pos;
     return pos;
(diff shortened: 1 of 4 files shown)


ramyala:
bump

daviesrob:
I don't think this is the right way to solve the problem.  If we want to implement retry logic then it would best be done in `hfile_libcurl` and not at a higher level.

`hFILE` is supposed to be (as much as possible) transparent to the method used to fetch the data.  If we push retry logic to higher levels then callers need to know if the data came from an HTTP request, in which case retry may make sense, or some other means where trying again is pointless.

Putting retry logic in callers also means you need to add it to a lot of places.  For example, this PR only propagates the need to try again in BAM iterators.  It doesn't deal with CRAM iterators, or cases where `hopen()` fails.  If instead we implement retry in `hfile_libcurl` we can make it work for everything that tries to open or seek on an HTTP(S) stream, while not affecting behaviour on regular files.

ramyala:
Agreed. I took a easier solution to apply retry logic at a higher level in pysam. However, this PR only addresses error propagation in htslib which should differentiate among a failure and a legitimate EOF.

PR #607, opened 2017-10-16
Fix data: urls in multipart requests

daviesrob:
Fixes googlegenomics/htsget#16.  Commit 4c6b93cb835 broke data: urls in htsget requests.

Not all URLs accessed by multipart.c go through hfile_libcurl.
As we want to pass an hfile_libcurl-specific hopen() parameter
and there is currently no way to find if a parameter will be
accepted, we check to see if the URL will be handled by
hfile_libcurl before calling hopen().

find_scheme_handler() is made non-static, but not made part of
the public API.

### DIFF
--- a/hfile.c
+++ b/hfile.c
@@ -880,7 +880,7 @@ static hFILE *hopen_unknown_scheme(const char *fname, const char *mode)
 }
 
 /* Returns the appropriate handler, or NULL if the string isn't an URL.  */
-static const struct hFILE_scheme_handler *find_scheme_handler(const char *s)
+const struct hFILE_scheme_handler *find_scheme_handler(const char *s)
 {
     static const struct hFILE_scheme_handler unknown_scheme =
         { hopen_unknown_scheme, hfile_always_local, "built-in", 0 };
--- a/hfile.c
+++ b/hfile.c
@@ -880,7 +880,7 @@ static hFILE *hopen_unknown_scheme(const char *fname, const char *mode)
 }
 
 /* Returns the appropriate handler, or NULL if the string isn't an URL.  */
-const struct hFILE_scheme_handler *find_scheme_handler(const char *s)
+const struct hFILE_scheme_handler *hfile_find_scheme_handler(const char *s)
 {
     static const struct hFILE_scheme_handler unknown_scheme =
         { hopen_unknown_scheme, hfile_always_local, "built-in", 0 };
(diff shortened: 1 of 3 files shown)


jmarshall:
> find_scheme_handler() is made non-static, but not made part of the public API.

Then you need to change its name to `hfile_find_scheme_handler()` or so, to avoid polluting the namespace.

But instead of this hackery, did you consider adding a `vopen` method to the `data` `hFILE_scheme_handler` so that it could just ignore this libcurl-specific parameter?

daviesrob:
Good point about changing the name.  I considered adding `vopen`, but it isn't enough.  The handler  would also need to recognise and ignore the `auth_token_enabled` parameter, and this would have to be done for all handlers, not just `data`.

The basic problem is that it's not possible to find out which extra parameters (if any) are accepted at the moment without trying them in an `hopen()` call.  We need to add some proper introspection, but I don't think that's possible without changing the ABI to add an extra call.

daviesrob:
Of course, another solution would be to make `vopen()` ignore any parameters it does not understand instead of reporting an error.  This wouldn't be too difficult as they all follow the same format at the moment.  Opinions welcome on if this would be a good idea.

jmarshall:
There's a grand total of four extant handlers that don't just forward their parameters to hfile_libcurl, and surely only one of them is likely to appear in htsget requests.

When I was designing this, I was imagining that a vocabulary of well-known parameter keys would naturally propagate to all the handlers as appropriate. So my opinion FWIW is that other handlers accepting well-known keys or ignoring unknown ones (but this is what constrained you to using `char*` here rather than the natural `int`, which is not ideal) would be the better approach. For simple handlers like `data`, “ignoring unknown keys” is basically just another way saying the handler understands the parameter key but it has no effect on the handler's behaviour anyway.

daviesrob:
So, would the best solution then be to get rid of [hfile.c:922](https://github.com/samtools/htslib/blob/7c5083b/hfile.c#L922) and switch the logic around in [lines 913 and 914](https://github.com/samtools/htslib/blob/7c5083b/hfile.c#L913) so that `hopen()` just calls `handler->open()` when `vopen` is unavailable?  Or should we write a function to actually go through the parameters and check they are the ones we expect?

You are correct, `char *` is there to keep consistency with other parameters.  I guess we could add another sigil to indicate that the following value is an integer (`:i`?), or even not present at all.

daviesrob:
Replaced by #609

PR #608, opened 2017-10-17
Multiple region iterator for BAM and CRAM

valeriuo:
1. Creation of a single iterator (with ordered file offsets) for BAM and CRAM files, starting from a specific format index and a region list.
2. Logic for stepping through the iterator and fetching reads which overlap regions of interest.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -851,7 +851,6 @@ htsFile *hts_hopen(hFILE *hfile, const char *fn, const char *mode)
         if (!fp->is_write)
             cram_set_option(fp->fp.cram, CRAM_OPT_DECODE_MD, 1);
         fp->is_cram = 1;
-        fp->fp.cram->new_seek = 0;
         break;
 
     case text_format:
@@ -2315,7 +2314,7 @@ hts_itr_multi_t *hts_itr_multi_cram(const hts_idx_t *idx, hts_itr_multi_t *iter)
             if (!off)
                 return NULL;
 
-            for(j=0; j < curr_reg->count; j++) {
+            for (j=0; j < curr_reg->count; j++) {
                 curr_intv = &curr_reg->intervals[j];
                 if (curr_intv->beg < 0) curr_intv->beg = 0;
                 if (curr_intv->end < curr_intv->beg)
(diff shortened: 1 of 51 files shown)


hts.c
```diff
@@ -2575,6 +2574,8 @@ hts_itr_multi_t *hts_itr_regions(const hts_idx_t *idx, hts_reglist_t *reglist, i
             }
 
             itr->reg_list[i].tid = getid(hdr, reglist[i].reg);
+            if (itr->reg_list[i].tid < 0) 
+                fprintf(stderr,"[hts_itr_regions]: region '%s' specifies an unknown reference name. Continue anyway.\n", reglist[i].reg);
```
jkbonfield: This should be using ```hts_log_error``` or ```hts_log_warning``` instead of fprintf, which then gives control over whether it is displayed.

jkbonfield:
Squashed, rebased and merged as b36740b6f82a74d8ccec022f0f641a966339fa70

jmarshall:
For future reference, it is good if squash commits like b36740b6f82a74d8ccec022f0f641a966339fa70 mention the PR number in their commit message, so readers don't have to hunt for the original PRs.

I haven't looked at the code, only the resulting compiler output:

```
hts.c:1989:81: warning: suggest parentheses around arithmetic in operand of ‘|’
                             iter->off[iter->n_off].max = (uint64_t)tid<<32 | end+1;

hts.c:2016:9: warning: unused variable 'res' [-Wunused-variable]
    int res = 0;
        ^
hts.c:2213:40: warning: comparison of unsigned expression < 0 is always false
      [-Wtautological-compare]
                    if (curr_intv->beg < 0) curr_intv->beg = 0;
                        ~~~~~~~~~~~~~~ ^ ~
hts.c:2174:34: warning: unused variable 't_fin' [-Wunused-variable]
    int i, j, l, n_off = 0, bin, t_fin;
                                 ^
hts.c:2325:36: warning: comparison of unsigned expression < 0 is always false
      [-Wtautological-compare]
                if (curr_intv->beg < 0) curr_intv->beg = 0;
                    ~~~~~~~~~~~~~~ ^ ~
hts.c:2546:12: warning: unused variable 'j' [-Wunused-variable]
    int i, j;
           ^
hts.c:2547:14: warning: unused variable 'off' [-Wunused-variable]
    uint64_t off;

```

jkbonfield:
Unfortunately I only just spotted this myself too.  It's a symptom of something we've needed to fix for ages - "make" shows these errors but "./configure; make" does not, as the configure script removes -Wall from the build command line. :/

jmarshall:
`./configure CFLAGS='-Wall -O2 -g'; make` is an easy habit. Trouble with adding it for a plain invocation is that _configure.ac_ ought to probe for the existence of the non-standard option…

jkbonfield:
@daviesrob added this to io_lib's configure script, with amendments: http://ac-archive.sourceforge.net/ac-archive/vl_prog_cc_warnings.html

It's a bit off topic though and we need to get these warnings fixed in develop.

valeriuo:
I'll make a fix today.

PR #612, opened 2017-10-23
Changed threat_pool.c examples to use hts_tpool_result_data(r) over r->data

clwgg:
This is to avoid compiler errors when using the library as described in the examples.

Also fixed some typos in the comments of thread_pool.c and htslib/thread_pool.h


jkbonfield:
Thank you.  Hiding the data pointer was added in 3bfd2811.  My mistake for forgetting to update the examples.

I can't see anything that could make the tests break given the location of changes, but  will follow procedure and wait for Travis to finish before merge.

PR #615, opened 2017-11-02
Fixes compiler warnings.

valeriuo:
Warnings caused by #608.


jkbonfield:
Thanks. (And sorry for putting that thanks in wrong place, duh!)

PR #617, opened 2017-11-07
Calculate sum in logarithmic form

ginggs:
using the expansion:
log(a + c) = log(a) + log(1 + c/a)
fixes calculation of PLs on ARM and POWER
see: https://github.com/samtools/bcftools/issues/702

### DIFF
--- a/errmod.c
+++ b/errmod.c
@@ -64,7 +64,7 @@ static double* logbinomial_table( const int n_size )
 static void cal_coef(errmod_t *em, double depcorr, double eta)
 {
     int k, n, q;
-    long double sum, sum1;
+    double sum, sum1;
     double *lC;
 
     // initialize ->fk
@@ -84,10 +84,11 @@ static void cal_coef(errmod_t *em, double depcorr, double eta)
         double le1 = log(1.0 - e);
         for (n = 1; n <= 255; ++n) {
             double *beta = em->beta + (q<<16|n<<8);
-            sum1 = sum = 0.0;
-            for (k = n; k >= 0; --k, sum1 = sum) {
-                sum = sum1 + expl(lC[n<<8|k] + k*le + (n-k)*le1);
-                beta[k] = -10. / M_LN10 * logl(sum1 / sum);
+            sum1 = lC[n<<8|n] + n*le;
+            beta[n] = HUGE_VAL;
+            for (k = n - 1; k >= 0; --k, sum1 = sum) {
+                sum = sum1 + log1p(exp(lC[n<<8|k] + k*le + (n-k)*le1 - sum1));
+                beta[k] = -10. / M_LN10 * (sum1 - sum);
             }
         }
     }


pd3:
Thank you for this!

daviesrob:
Can we get away with making `sum` and `sum1` double, and using the double versions of the functions, i.e. `log1p` and `exp`?  Comparing the values stored in `beta`, the absolute difference is very small (about 2e-12 in the worst case).  Both the htslib and bcftools tests still pass.

I suspect that ARM and POWER are treating `long double` as `double` so changing them should make them the same size on all platforms we support.

ginggs:
@daviesrob Having looked at this, I would say yes, you can make ```sum``` and ```sum1``` doubles now.

For interest's sake, on ARMv7 ```long double``` is indeed ```double```, on ARMv8 it is 128-bit quadruple precision, and on POWER it is double-double (the exact sum of two doubles) ~106-bit precision.

jmarshall:
ISTR `INFINITY`can cause portability problems. It's from C99 and probably POSIX, and isn't present in MS VS 11 for example. Not sure what Visual Studio version the Windows aficionados are targeting these days…

Is it practical to recode to avoid using this constant macro?

mp15:
I doubt anyone is writing any commercial apps using MS VS v11 given that they would have a hard time targeting much past Windows 8 with it. For other development they can just download the community edition for free. C99 is not unreasonable, especially given how much floating point we do, C11 is still probably a bridge too far though.

jkbonfield:
I don't think C99 is an issue given we basically took the decision it's the minimum C standard we support (along with XOPEN_SOURCE=600 IIRC).

On windows this means we use one of the mingw variants, but we may one day get around to investigating Microsoft compiler.  It's a much bigger task though due to a totally different command line syntax (-Fo:file instead of -o file, etc).

ginggs:
@jmarshall 
> Is it practical to recode to avoid using this constant macro?

We could simply do:
```beta[n] = 1.0 / 0.0;```

lh3:
Division by 0 will trigger floating point failures if a program asks CPU to do so. A library can't control downstream programs and thus should not use `1.0/0.0`.

ginggs:
@lh3 OK, well the original code was:
```beta[k] = -10. / M_LN10 * logl(sum1 / sum);```
with ```sum = 0.0``` when ```k = n```

but this seems to work, too:
```beta[n] = - log(0);```

ginggs:
err no, my mistake, ```sum1 = 0.0``` when ```k=n``` so there was no division by zero in the original code.

daviesrob:
In that case, `beta[n]` would be zero, wouldn't it?

[The code that uses this table](https://github.com/samtools/htslib/blob/6e0f0982/errmod.c#L153) looks like it never accesses `beta[n]` as `aux.c[base]` can't get to more that `n - 1` by line 163.  So I suspect the exact value doesn't matter too much.

daviesrob:
Or actually, `log(0)`, which is `-HUGE_VAL` and a possible floating exception...

`HUGE_VAL` appears to be [C89](http://www.schweikhardt.net/identifiers.htm), so it should be widely available.

ginggs:
@daviesrob:
> Or actually, ```log(0)```, which is ```-HUGE_VAL``` and a possible floating exception...

and multiplied by ```-10 / log(10)``` gave me ```+inf``` on the systems I tested on, which is why I originally went for ```INFINITY```.

daviesrob:
`HUGE_VAL` may be infinity on platforms that have it (see, for example, [here](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/math.h.html)).  As we require IEEE floating point for other reasons, it's pretty sure all platforms we support do have infinity (if not `INFINITY`).

I think `HUGE_VAL` should be fine here.  Anyone object?

pd3:
`HUGE_VAL` is already used throughout bcftools, nobody ever complained.

ginggs:
Changes pushed.  Please let me know if you'd like the commits squashed.

daviesrob:
Squashing would be useful, I think.  Or I can do it for you, if you prefer?

ginggs:
@daviesrob Squashed.

daviesrob:
Sorry, could you rebase it onto develop as well, please?  It appears another commit managed to sneak on to it this morning.

ginggs:
@daviesrob No problem, rebased.

daviesrob:
Hooray! Got it in.

Thanks for fixing this for us.  Much appreciated.

ginggs:
You are welcome!

PR #620, opened 2017-11-09
On cache full, remove the oldest object

ramyala:
Currently cache eviction is random. This causes a significant performance hit for network based protocols (libcurl) when reads are sequential (most BED files are sorted so this is a fair assumption).

This patch implements oldest block eviction on cache full.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -69,11 +69,23 @@ static const uint8_t g_magic[19] = "\037\213\010\4\0\0\0\0\0\377\6\0\102\103\2\0
 typedef struct {
     int size;
     uint8_t *block;
-    int64_t end_offset;
+    uint64_t end_offset;
+    uint64_t next_cached_block_address;
+    unsigned has_next_cached_block:1;
 } cache_t;
+
 #include "htslib/khash.h"
 KHASH_MAP_INIT_INT64(cache, cache_t)
-#endif
+
+typedef struct {
+    khash_t(cache)* hash;
+    uint64_t oldest_block_address;
+    uint64_t latest_block_address;
+    unsigned in_use:1; // set if there is atleast one object in cache
+} bgzf_cache_t;
+
+#define bgzf_cache(x) ((bgzf_cache_t*) x->cache)
+#endif // BGZF_CACHE
 
 #ifdef BGZF_MT
 
@@ -215,7 +227,10 @@ static BGZF *bgzf_read_init(hFILE *hfpr)
     fp->is_compressed = (n==18 && magic[0]==0x1f && magic[1]==0x8b);
     fp->is_gzip = ( !fp->is_compressed || ((magic[3]&4) && memcmp(&magic[12], "BC\2\0",4)==0) ) ? 0 : 1;
 #ifdef BGZF_CACHE
-    fp->cache = kh_init(cache);
+    bgzf_cache_t *bc = calloc(1, sizeof(bgzf_cache_t));
+    if (bc == NULL) { free(fp->uncompressed_block); free(fp); return NULL; }
+    bc->hash = kh_init(cache);
+    fp->cache = bc;
 #endif
     return fp;
 }
@@ -524,11 +539,12 @@ static int check_header(const uint8_t *header)
 static void free_cache(BGZF *fp)
 {
     khint_t k;
-    khash_t(cache) *h = (khash_t(cache)*)fp->cache;
     if (fp->is_write) return;
+    khash_t(cache) *h = bgzf_cache(fp)->hash;
     for (k = kh_begin(h); k < kh_end(h); ++k)
         if (kh_exist(h, k)) free(kh_val(h, k).block);
     kh_destroy(cache, h);
+    free(bgzf_cache(fp));
 }
 
 static int load_block_from_cache(BGZF *fp, int64_t block_address)
@@ -536,7 +552,8 @@ static int load_block_from_cache(BGZF *fp, int64_t block_address)
     khint_t k;
     cache_t *p;
 
-    khash_t(cache) *h = (khash_t(cache)*)fp->cache;
+    assert(fp->is_write == 0); // cache should never be used in write mode
+    khash_t(cache) *h = bgzf_cache(fp)->hash;
     k = kh_get(cache, h, block_address);
     if (k == kh_end(h)) return 0;
     p = &kh_val(h, k);
@@ -559,19 +576,27 @@ static void cache_block(BGZF *fp, int size)
     int ret;
     khint_t k;
     cache_t *p;
-    //fprintf(stderr, "Cache block at %llx\n", (int)fp->block_address);
-    khash_t(cache) *h = (khash_t(cache)*)fp->cache;
+
+    assert(fp->is_write == 0); // cache should never be used in write mode
+    bgzf_cache_t *bc = bgzf_cache(fp);
+    assert(bc != NULL);
+    khash_t(cache) *h = bc->hash;
     if (BGZF_MAX_BLOCK_SIZE >= fp->cache_size) return;
     if ((kh_size(h) + 1) * BGZF_MAX_BLOCK_SIZE > (uint32_t)fp->cache_size) {
-        /* A better way would be to remove the oldest block in the
-         * cache, but here we remove a random one for simplicity. This
-         * should not have a big impact on performance. */
-        for (k = kh_begin(h); k < kh_end(h); ++k)
-            if (kh_exist(h, k)) break;
-        if (k < kh_end(h)) {
-            free(kh_val(h, k).block);
-            kh_del(cache, h, k);
+        // Remove the oldest cached block
+        assert(bc->in_use == 1);
+        k = kh_get(cache, h, bc->oldest_block_address);
+        assert(k != kh_end(h));
+        p = &kh_val(h, k);
+        if (p->has_next_cached_block) {
+            // The block being freed has a next block pointer. It is now the oldest block.
+            bc->oldest_block_address = p->next_cached_block_address;
+        } else {
+            // Last object in cache was evicted. Mark state appropriately.
+            bc->in_use = 0;
         }
+        free(kh_val(h, k).block);
+        kh_del(cache, h, k);
     }
     k = kh_put(cache, h, fp->block_address, &ret);
     if (ret == 0) return; // if this happens, a bug!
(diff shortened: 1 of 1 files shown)


jkbonfield:
Thanks for the PR.

Do you have any feeling for what the performance changes are with this?  I'm wary of bgzf - it's a total minefield so I'm cautious about changes that don't have a demonstrable benefit.

My concern also is a random cache removal is sometimes inappropriate, but doesn't have a bad worst case performance.  A least recently used cache removal does have a worst case performance of never caching anything.  Eg consider a cache size of 100 blocks, with code that reads 101 blocks, seeks back, and re-reads 101 blocks.  This scenario always purges the cached block that we're going to need next and has a 0% cache hit rate.  The random removal has a hit rate of 99%.

ramyala:
I got a dramatic 3-5x improvement because connection setup on HTTP is very costly in our environment (2-3 seconds to re-establish). Our workloads are defined as reading from regions defined in a sorted BED file, with about 100mb coverage across the human genome.

In this case you can imagine a 100MB cache having say 3200 block of cache capacity. Once the cache fills up, the random eviction hurts performance dramatically because during sorted reads, you are most likely to seek back to the most recently read blocks. (BAI index defining a large byte region for the corresponding genome region). In practice i've seen a large part of these "random" evictions actually be the most recent block being evicted, which hurt performance dramatically once cache filled up.

Irrespective, I think a cache design based on oldest block eviction is an important design feature to consider given a large set of tools assume sorted BED files, so we have an indication of what to expect in a majority of the use cases.

jkbonfield:
Ok that sounds very significant.  It's curious it has this impact and the comment on it not being very random may be true.  I took a quick look at the code and it uses the int64 khash algorithm as:

```
#define kh_int64_hash_func(key) (khint32_t)((key)>>33^(key)^(key)<<11)
```

The keys though are compressed offsets (normally), so generally the low bits are pretty random, but I think you're correct that this is somehow failing to randomly discard a block and instead favouring the previous one.

jkbonfield:
I tried tracking how often the cache block removed is the last block inserted when streaming through a small 300Mb BAM file and it was a staggering 75% of the time!  I don't yet know why this is so.  However picking a random start point and looping through the hash until we get to the same point, instead of always starting at the start and removing the first found, solves this.

I suspect it's something to do with the nature of the hashing algorithm and the closed-hash approach where you hash and then increment through buckets to find the first non-empty one.  This may clump things together.

Anyway, does this simpler patch solve things for you too?  If so I'm inclined to go with it over a much more complex alternative.  Sorry it's not a PR, but it's trivial enough to just manually hack in and you're clearly a coder. 

```
@@ -557,18 +566,28 @@ static int load_block_from_cache(BGZF *fp, int64_t block_address)
 static void cache_block(BGZF *fp, int size)
 {
     int ret;
-    khint_t k;
+    khint_t k, k_orig;
     cache_t *p;
     //fprintf(stderr, "Cache block at %llx\n", (int)fp->block_address);
     khash_t(cache) *h = (khash_t(cache)*)fp->cache;
     if (BGZF_MAX_BLOCK_SIZE >= fp->cache_size) return;
     if ((kh_size(h) + 1) * BGZF_MAX_BLOCK_SIZE > (uint32_t)fp->cache_size) {
         /* A better way would be to remove the oldest block in the
          * cache, but here we remove a random one for simplicity. This
          * should not have a big impact on performance. */
-        for (k = kh_begin(h); k < kh_end(h); ++k)
-            if (kh_exist(h, k)) break;
-        if (k < kh_end(h)) {
+        k = k_orig = random() % (kh_end(h)-kh_begin(h)) + kh_begin(h);
+        do {
+            if (kh_exist(h, k))
+                break;
+            if (++k == kh_end(h))
+                k = kh_begin(h);
+        } while (k != k_orig);
+            
+        if (k != k_orig) {
             free(kh_val(h, k).block);
             kh_del(cache, h, k);
         }
```

lh3:
An alternative is to use a better hash function. You can replace the macro in khash.h with:
```c
static inline uint32_t kh_int64_hash_func(uint64_t key)
{
	key = (~key + (key << 21));
	key = key ^ key >> 24;
	key = ((key + (key << 3)) + (key << 8));
	key = key ^ key >> 14;
	key = ((key + (key << 2)) + (key << 4));
	key = key ^ key >> 28;
	key = (key + (key << 31));
	return (uint32_t)key;
}
```
or you can define this function in bgzf.c and ask khash to use this hash function.

ramyala:
Thank you for the suggestions. I don't think these comments address the key design concerns I brought up for the cache. 

A cache should be optimized for the most common use cases, which is sequential and stream processing. Doing any random eviction will substantially affect performance even if its 10-15% of the time on large files (100+GB). I know personally I can't accept that performance hit and this PR solves the issue elegantly with minimal changes to bgzf's core code. And, it comes with ample asserts for good measure :)

lh3:
As you are only improving caching, you should not touch the BGZF struct. A modification to BGZF is a significant change. If I had to implement precise FIFO, I would isolate caching related variables like:
```c
typedef struct {
  khash_t(cache) *hash;
  uint64_t first, last; // don't use int64_t
} bgzf_cache_t;
```
Then `void *cache` in `BGZF` is a pointer to the new `bgzf_cache_t` struct.

ramyala:
Good point. I'll update the PR.

first and last have to be able to distinguish between being used and not, and I'm using -1 as an unused state. uint64_t forces another bit to see validity since block_address = 0 is a valid value.

We already have this be the case with off_t being assigned a negative value when unused so I thought that would be ok.

ramyala:
I updated the branch with requested changes. Please review.

jkbonfield:
Ok so this is complex, but I think I understand the issues.

1.  The problem with non-randomness before is nothing to do with poor hash functions.  The hash function isn't great, but the input data (start addresses of gzipped blocks) is already pretty random in the low bits so the hash function is fine.  The problem is that we have formed clumps.  The hash table is a closed hash table, meaning when it is full we keep iteratively probing until we find the first empty bucket.  By always removing the first used bucket over time we're preserving a large consecutive clump at the end, gradually turning the hash table into a linked list of which the last key is in the empty portion at the start of the hash table and corresponding the one we delete.  This is true of ANY hash function due to random key replacement + selective key removal.

2. The LRU (actually LR inserted) hash code has the problem I expected; it works fine for small queries, but doesn't gracefully decay when given large re-reads.  It'll either be 100% successful or 0% successful, so hits a cliff quite hard.  This is not true of random removal which is a gradual removal.

3. A round-robin removal, instead of random() to preserve anyone foolish enough to assume srand/random are entirely their own private functions, works better even than random removal.

I'll do some benchmarks on a round-robin removal vs LRI removal (this PR).  For these I iterate on chr1 from 1MB to 10Mb in 100k chunks, querying 1 byte (so 1:1200000-1200000) and with larger queries up to ~100Kb (so 1:1200000-1300000).  Easily done via a quick awk hack to produce the regions list:

```
awk 'END {for (i=1000000; i < 10000000; i+=1000) {printf("1:%d-%d\n", i, i+100000)}}' /dev/null > _reg
```

Ongoing...

lh3:
@jkbonfield yes, you are right about hash function.

In general, which cache scheduling algorithm to choose highly depends on the access pattern. It would be good to know how others are using BGZF caching. Caching was introduced to speed up tview over network. I rarely use it to be honest.

ramyala:
That matches what I've seen empirically. Regarding @lh3's comment on usage, I strongly feel sequential access should be the top priority in caching strategy due to:
1. A large number of pipelines depend on unix pipes which essentially is sequential access
2. BAI index format specifies byte ranges across regions, and many times regions which are close enough will seek back to the same byte range. This leads to repeated querying of most recently read bytes.

Any randomness introduced into cache eviction strategy is a tax on these base use cases.

jkbonfield:
What size cache are you typically using?  How large are the regions you are querying?  How far apart typically?

Agreed the highest priority is to sequential processing of regions, although note, you may wish to look at the recent multi-region iterator which does a far better job of handling this.

jkbonfield:
I've got a minimal fix for the existing code - random removal but without the selective pressure to preserve offsets that hash to high values and remove offsets that hash to low values: https://github.com/jkbonfield/htslib/tree/bgzf_cache_fix

I'm interested to know if this acceptable.  In my testing it seems OK and works sufficiently well on sorted streams, OK on mostly sorted queries, and about as expected on unsorted ones (ie it depends on the size of the cache).

I can see that an LRU cache could work better in some scenarios (and poorer in others), but this PR isn't LRU - it's FIFO (or least-recently inserted).  FIFO caches can be better than random too in some cases, but if the cache size is large then the differences become very slight.  I'm still a bit wary of a caching strategy that has a hard cutoff of working perfectly to working never rather than one with gradual decay as cache size becomes too small, but whether this is a real concern entirely depends on the actual usage in cache size set, depth of data and the size and localities of queries.

Anyone else have thoughts?

daviesrob:
After testing this against #629 and a true least-recently used implementation, I decided to go with #629 in the end.  It's quite a lot simpler, it works better when the cache is small and makes very little difference when the cache is big.

While testing this, I noticed that there is an `hseek()` call in `load_block_from_cache()`.  This has the effect of causing many more seeks to happen when the cache was turned on than when it was off, which was rather undesirable when getting data via http.  This has now been mitigated by #648, which makes hfile_libcurl delay seek operations until the next read.

Here are some counts for the number of re-connections when stepping through 100Kb of NA12892 chr. 20 in 100 base increments:

Cache | Without #648 | With #648
------- | ------- | -------
None | 988 | 988
Original (1M) | 2632 | 985
PR #620 (1M) | 4498 | 706
PR #629 (1M) | 7163 | 717
Original (2M) | 5202 | 931
PR #620 (2M) | 13921 | 416
PR #629 (2M) | 19504 | 441
Original (10M) | 20215 | 481
PR #620 (10M) | 22071 | 248
PR #629 (10M) | 22058 | 248

And some timings, in seconds:

Cache | Without #648 | With #648
------- | ------- | -------
None | 14 | 14
Original (1M) | 15 | 14
PR #620 (1M) | 15 | 10
PR #629 (1M) | 16 | 9
Original (2M) | 16 | 12
PR #620 (2M) | 18 | 7
PR #629 (2M) | 24 | 4
Original (10M) | 20 | 4
PR #620 (10M) | 28 | 2
PR #629 (10M) | 19 | 2

PR #627, opened 2017-11-20
Cram iter 625

### LINKED ISSUE #625: HTS_IDX_START support in cram?
Hi,

I use the HTS_IDX_START [macro](https://github.com/samtools/htslib/blob/develop/htslib/hts.h#L494) with `hts_itr_query` to iterate over entire bam files. When running on a cram file, I get this error:

```
[cram_itr_query] tid=-3 not implemented for CRAM files -- Exiting Aborted
```

Is there a plan to implement this feature? Alternatively, is there an easy way to iterate over an entire cram file?

Thanks,
Jared

jkbonfield:
Fixes #625

Modifies cram_index.c to cope with HTS_IDX_* special "tid" values, plus minimal changes to sam.c (normal iterator) and hts.c (multi-region iterator) to permit use of these on CRAM.

The second commit adds a -M option to test/test_view.c to permit command line testing of the new multi-region iterator.  This may be considered separate to this PR if desired.  It also exposes some weaknesses in the new API in that it lacks any way of constructing an hts_reglist_t struct.  Ideally we'd use something like:

```
hts_reglist_t *reglist = NULL;
for (i = optind + 1; i < argc; ++i) {
    if (hts_reglist_add(&reglist, argv[i]) < 0)
        return -1;
}
```

Perhaps with another form for adding BED lines?  However that is a separate issue.

### DIFF
--- a/test/test_view.c
+++ b/test/test_view.c
@@ -30,6 +30,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <stdlib.h>
 #include <string.h>
 #include <getopt.h>
+#include <stdint.h>
 
 #include "cram/cram.h"
 
@@ -58,8 +59,9 @@ int main(int argc, char *argv[])
     int extra_hdr_nuls = 0;
     int benchmark = 0;
     int nthreads = 0; // shared pool
+    int multi_reg = 0;
 
-    while ((c = getopt(argc, argv, "DSIt:i:bCl:o:N:BZ:@:")) >= 0) {
+    while ((c = getopt(argc, argv, "DSIt:i:bCl:o:N:BZ:@:M")) >= 0) {
         switch (c) {
         case 'D': flag |= READ_CRAM; break;
         case 'S': flag |= READ_COMPRESSED; break;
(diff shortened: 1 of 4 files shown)


test/test_view.c
```diff
@@ -185,13 +189,72 @@ int main(int argc, char *argv[])
...
+                if (cp) *cp = 0;
+
+                // Does any of this need to be sorted?
+                // For now we assume any such limits are passed onto whoever
+                // is running the command line.
+                for (j = 0; j < reg_count; j++)
+                    if (strcmp(reg_list[j].reg, argv[i]) == 0)
+                        break;
+                if (j == reg_count) {
+                    reg_list[j=reg_count++].reg = argv[i];
```
valeriuo: `j=reg_count` part is a bit redundant here, since `j` was already equal to `reg_count`.

valeriuo:
Merged in 6fee01859b88c019ccb2d91a72f6f74f10db4468 and ff955652a29e9f71f15e1f4bd39758d11f8a5fb6

PR #628, opened 2017-11-20
Use macOS's included liblzma (thus avoiding reliance on Brew etc)

jmarshall:
A vanilla macOS installation in fact includes _/usr/lib/liblzma.dylib_ but sadly does not ship the corresponding header files. This PR adds a fake _lzma_stub.h_ header that provides just enough declarations so that _cram/cram_io.c_ can be compiled, for use when _configure_ detects a usable _liblzma_ but no _lzma.h_.

Also some incidentals:

* Rename _win/_ to _os/_ for genericness, and also in sympathy with the sentiments in the [GNU Coding Standards §2.3](https://www.gnu.org/prep/standards/html_node/Trademarks.html).

* **Fix hts_drand48() et al**, which as shipped in v1.6 **always** has the implementations in _os/rand.c_ activated. This is obviously a concern on FreeBSD (which comes with an implementation that uses the same system-reserved identifiers, i.e. `_rand48_seed` etc)… you may wish to further highlight the NEWS item added in this PR.


daviesrob:
I don't understand your problem with `hts_drand48()` etc.

They are only ever exported with the `hts_` prefix.  Does FreeBSD have functions with this prefix?
Likewise _rand48_seed etc. are declared static, so should not interfere with any system-defined variables with the same name.

Although you are correct that it should have included `config.h`...

jmarshall:
I don't have a problem with `hts_drand48()` etc, only with non-`hts`-prefixed items.

All I know is what `nm` tells me. Probably things are safe-ish as `_rand48_seed` etc are static (which they weren't originally, but fortunately you caught that), though technically all bets are off once HTSlib starts using file-scope identifiers with leading underscores…

daviesrob:
nm says...

```
$ nm -a libhts.so | grep rand
0000000000030f08 t _dorand48
00000000002c5338 d _rand48_add
00000000002c5332 d _rand48_mult
00000000002c532c d _rand48_seed
00000000000310e4 T hts_drand48
0000000000031062 T hts_erand48
00000000000310f6 T hts_lrand48
0000000000031011 T hts_srand48
```

So they're local, and not really causing a problem.

We could always change them to `hts_rand48_seed` etc. if you're really worried, but I don't think there's any need for the text you added to the NEWS file, is there?

jmarshall:
I was somewhat concerned that these internal symbols appear in the symbol table of the fully-linked _libhts.so_ at all. Anyway, that NEWS text was really just a suggestion to get the idea of mentioning this in NEWS on the table; the maintainers should do with it what they will.

Meanwhile, any reaction whatsoever to any other aspect of this PR would be appreciated. :smile:

jkbonfield:
If we're worried about these symbols then we should be worried about every static in the library, including eg:

```hfile.c:static const struct hFILE_backend fd_backend = ...```

I think you're just misinterpreting how nm and symbols work.

The other points are interesting and have made us wonder if it should apply elsewhere.  Hacking our own header files to avoid using system ones feels really icky and the wrong solution, but pragmatically may simply be the best way of avoiding lots of queries.  If we go down this road though, as Rob pointed out in the office, we should probably add a fake local curl.h too as that's the other common culprit.

Personally though I dislike this fix.  I think we're tackling the problem the wrong way.  What is it that people really want?  I think our users fall into two groups.

1) Developers who wish to use the library in their own tools.  They ought to know enough about dependencies, installing dev packages, etc.

2) End users who just want a program to run.  We're failing this community by making them all build from source, when we could just provide prebuilt binaries instead - perhaps built automatically as part of travis and appveyor tests.

jmarshall:
> If we're worried about these symbols then we should be worried about every static in the library

Okay, I withdraw. As it is static, there is probably no mechanism for this `_rand48_seed` to cause trouble in reality. Nonetheless, there is no excuse for using identifiers starting with underscores in a non-system library like HTSlib.

> If we go down this road though, as Rob pointed out in the office, we should probably add a fake local curl.h too as that's the other common culprit.

The difference is that there is no major platform that provides _libcurl.a/.so_ but not the corresponding header files. For _liblzma_, there is. (I suppose Rob is thinking of Linux with a libcurl runtime package installed but no dev package. To support that you'd need a fake header and at least a fake _libcurl.so_ symlink, which would be rather more work and more fragile.)

jkbonfield:
It took a while to parse that comment, but yes I see the issue now (libcurl.so vs libcurl.so.3).  I've no idea if there are systems with the base .so that don't have headers.  Assuming not then.

Looking at the lzma stub it looks quite large, but yes it's necessary due to the structure init call and use of internal fields, so I think this is indeed the minimal hack we can get away with.  It still feels ugly and wrong, but I'd go with the flow if others feel it's an appropriate hack to use.

PS.  I hate macs! :P

daviesrob:
Rebased (minus the change to NEWS) and merged.  Mac users can now avoid using homebrew, at least for HTSlib.

atks:
@daviesrob 

I'm encountering this missing lzma.h issue again with OSX 10.13.3 with  htslib 1.7-23-g9f9631d from the development branch. 

Can you have a look into this?

daviesrob:
@atks What errors are you seeing when you try to build?

PR #629, opened 2017-11-21
Bug fix to random bgzf cache removal.

jkbonfield:
An alternative to https://github.com/samtools/htslib/pull/620 for consideration.

Difference here is random removal (via round-robin approach) vs removal of oldest object.  Pros and cons of each.  (Thanks @ramyala for spotting it was bugged and raising the other PR implementation.)

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -76,9 +76,9 @@ typedef struct {
 KHASH_MAP_INIT_INT64(cache, cache_t)
 #endif
 
-struct __bgzf_cache_t {
+struct bgzf_cache_t {
     khash_t(cache) *h;
-    int last_pos;
+    khint_t last_pos;
 };
 
 #ifdef BGZF_MT
@@ -225,7 +225,11 @@ static BGZF *bgzf_read_init(hFILE *hfpr)
         free(fp);
         return NULL;
     }
-    fp->cache->h = kh_init(cache);
+    if (!(fp->cache->h = kh_init(cache))) {
+        free(fp->cache);
+        free(fp);
+        return NULL;
+    }
     fp->cache->last_pos = 0;
 #endif
     return fp;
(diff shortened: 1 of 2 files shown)


ramyala:
I ran some tests with your patch and saw massive degradation (4x) in performance across our pipelines vs. the LRI cache in #620. The difference is because we rely either on sorted BED files or pipe as an input for all our pipelines. 

The other issue I saw is the performance got worse when the ratio of BAM file to cache size is larger. (1 GB BAM with 100MB cache took 10 seconds, vs. 10 GB w/100MB cache at 7minutes).

I understand the hesitation on making the cache unusable in an unsorted BED use-case, but I think the cost imposed is great for what I feel is a majority use case (which is what a cache should try to help?)

jkbonfield:
I'm struggling to understand what is going on here.  Is there something else changing the speed?

You previously benchmarked the original (and buggy) random caching at 3-5x slower.  This turned out to be due to a bug in the way the cache worked, which degraded the size down to a single item cached.  The above code fixes this.

However you still see it as 4x slower.  Are you certain you tested the right version?  In my own benchmarks it's sometimes slightly faster than yours and sometimes slightly slower, but very close in most regards.  I tested both sorted and unsorted.

Do you have an example BAM and bed file that we can experiment with?

If we really conclude that the random approach isn't sufficient, then the next best would be least recently used discard, rather than oldest inserted discard, as this is better when it comes to the blocks corresponding to the larger bins in a bai file.

valeriuo:
@ramyala, can you provide us with your test files (BAM and BED), so that we benchmark our solutions against the same data? We acknowledged there is an issue with the caching algorithm, that we should fix soon, but there seems to be a disagreement regarding the best approach.

jkbonfield:
Happy with new commits.  Good for merge.

PR #632, opened 2017-11-29
Add configure options to enable warnings and optionally Werror

daviesrob:
The m4 file here is an evolution of the one from the Staden io_lib.  In tests it works better that the equivalent one in autoconf_archive, which seems to be broken for some compilers.  This one first tries to identify the compiler and then attempts to run it with the option, so should be reasonably safe.

Also included is an option to enable -Werror or equivalent.  Currently it only works on GCC compatible compilers and Sun Studio, as they're the ones I've been able to test so far.  Setting -Werror can break some tests (notably AC_SEARCH_LIBS), so this option is only applied to CFLAGS towards the end of the configure script.

Default settings are to enable warnings, but not -Werror.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -29,6 +29,8 @@ AC_PREREQ(2.63)  dnl This version introduced 4-argument AC_CHECK_HEADER
 AC_CONFIG_SRCDIR(hts.c)
 AC_CONFIG_HEADERS(config.h)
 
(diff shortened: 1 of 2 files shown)


jkbonfield:
Works like a charm.  Are you going to add this to Samtools and Bcftools too?

PR #638, opened 2018-01-04
Fixes the finished flag usage.

valeriuo:
In a previous implementation of `kstrtok`, the LSB of `tab[0]` was used as a flag to indicate that the tokenization process had ended. However, a new dedicated flag (`finished`) was later added to the `aux` structure, but the change to the new flag hasn't been completely carried out.

### DIFF
--- a/kstring.c
+++ b/kstring.c
@@ -174,28 +174,29 @@ int ksprintf(kstring_t *s, const char *fmt, ...)
 	return l;
 }
 
-char *kstrtok(const char *str, const char *sep, ks_tokaux_t *aux)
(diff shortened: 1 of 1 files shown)


jkbonfield:
Does this ```aux->finished``` change originate from https://github.com/attractivechaos/klib/commit/d8d0c5fbcb61e56a9cc607e16a65a290978fc47c ?

Either way, we should probably prod upstream (@lh3) so the master can be fixed too.

James

valeriuo:
I tracked it down to https://github.com/lh3/samtools/commit/d211e652d93791d2e112d334added243ffe5fc3e, but the commits look similar.

PR #641, opened 2018-01-08
Fixed return value of cram_get_bam_seq and fixed a bug in required_fields option.

### LINKED ISSUE #640: Alignments returned by sam_itr_query are affected by required_fields in CRAM files
For some reason, whether a given alignment is returned by `sam_itr_queryi()` in a CRAM file can be affected by whether and which `required_fields` have been set when opening the CRAM file. Let's take as an example [this test file from deepTools](https://github.com/deeptools/deepTools/blob/feature/implement619_CRAM/deeptools/test/test_data/test1.cram) (you can find the index and the fasta file for decoding it in the same directory). I'll use the following (generally poorly written) program as a test example:

```
#include <stdio.h>
#include <stdlib.h>
#include "htslib/sam.h"
#include "htslib/hts.h"

int main(int argc, char *argv[]) {
    samFile *cram = sam_open("test1.cram", "rc");
    hts_idx_t *idx = sam_index_load(cram, "test1.cram");
    char foo[1024];
    hts_opt *opts = NULL;
    hts_itr_t *iter = NULL;
    int cnt = 0, ret;
    bam1_t *b = bam_init1();

    //Set some options
    sprintf(foo, "required_fields=%i", argv[1]);
    hts_opt_add(&opts, foo);
    hts_opt_apply(cram, opts);
    hts_opt_free(opts);

    //Iterate over a range, counting the number of alignments
    iter = sam_itr_queryi(idx, 0, 99, 1100);
    while((ret=sam_itr_next(cram, iter, b)) > 0) cnt++;
    hts_itr_destroy(iter);

    sam_close(cram);
    bam_destroy1(b);
    free(idx);
    printf("There were %i entries found\n", cnt);
    return 0;
}

```

If I compile that to `./foo` then I can easily play with the results:

```
$ ./foo 0xF
There were 0 entries found
$ ./foo 0xFF
There were 0 entries found
$ ./foo 0xFFF
There were 143 entries found
$ ./foo 0x8FF
There were 140 entries found
```

The same sort of thing happens with `samtools view -c --input-fmt-option required_fields=0x8FF test1.cram 3R:99-1100` in the 1.6 release and going back to at least 1.3, so it's not purely a coding error on my part. I assume that this is not intended behavior.

jkbonfield:
1. bam_construct_seq returned 0 instead of number of bytes written to
the bam_seq_t struct.  This in turn bubbled up to cram_get_seq
returning only the number of auxiliary bytes written to instead of the
total record size.

Under specific circumstances (no RG tag, no NM/MD requested, and no
other aux fields), this meant cram_get_bam_seq returned 0 on a
successful read.  This in turn was interpreted as EOF by the iterator.
Given bam_read1 returns the size of the bam structure it feels
appropriate for the cram version to mirror it, as originally intended.

2. If we use the required_fields option to exclude sequence data, then
we also exclude loading of the reference (which is a useful
optimisation).  However this then also meant the cram_decode_seq
function wasn't correctly modifying the alignment end, leaving it set
to alignment start.

The consequence is range queries combined with required_fields usage
could miss reads which start before the range and end within it.

Fixes #640

### DIFF
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -1798,6 +1798,9 @@ static int cram_decode_seq(cram_fd *fd, cram_container *c, cram_slice *s,
 		}
 		ref_pos += cr->len - seq_pos + 1;
 	    }
+	} else if (cr->ref_id >= 0) {
+	    // So alignment end can be computed even when not decoding sequence
+	    ref_pos += cr->len - seq_pos + 1;
 	}
 
 	if (ncigar+1 >= cigar_alloc) {
@@ -1950,7 +1953,7 @@ static int cram_decode_aux_1_0(cram_container *c, cram_slice *s,
 
 	cr->aux_size += out_sz + 3;
     }
-    
+
     return r;
 }
 
@@ -1961,7 +1964,7 @@ static int cram_decode_aux(cram_container *c, cram_slice *s,
     int32_t TL = 0;
     unsigned char *TN;
     uint32_t ds = c->comp_hdr->data_series;
-	    
+
     if (!(ds & (CRAM_TL|CRAM_aux))) {
 	cr->aux = 0;
 	cr->aux_size = 0;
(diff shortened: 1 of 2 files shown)


dpryan79:
Out of curiosity, are there currently any plans for a 1.6.1 release with this included?

daviesrob:
We're planning to make a new release soon, which will include this.

dpryan79:
Great, thanks @daviesrob

PR #642, opened 2018-01-09
Significant speed up of the BCF synced reader (in some situations).

pd3:
Skip the expensive variant matching algorithm when it is unnecessary and only one file is active.
Alas, this does not solve the general case of multiple active readers, something to look further into.

### DIFF
--- a/bcf_sr_sort.c
+++ b/bcf_sr_sort.c
@@ -327,6 +327,18 @@ char *grp_create_key(sr_sort_t *srt)
     }
     return ret;
 }
+void bcf_sr_sort_set_active(sr_sort_t *srt, int idx)
+{
+    hts_expand(int,idx+1,srt->mactive,srt->active);
+    srt->nactive = 1;
+    srt->active[srt->nactive - 1] = idx;
+}
+void bcf_sr_sort_add_active(sr_sort_t *srt, int idx)
+{
+    hts_expand(int,idx+1,srt->mactive,srt->active);
+    srt->nactive++;
+    srt->active[srt->nactive - 1] = idx;
+}
 static void bcf_sr_sort_set(bcf_srs_t *readers, sr_sort_t *srt, const char *chr, int min_pos)
 {
     if ( !srt->grp_str2int )
@@ -357,11 +369,11 @@ static void bcf_sr_sort_set(bcf_srs_t *readers, sr_sort_t *srt, const char *chr,
     memset(&grp,0,sizeof(grp_t));
 
     // group VCFs into groups, each with a unique combination of variants in the duplicate lines
-    int ireader,ivar,irec,igrp,ivset;
-    for (ireader=0; ireader<readers->nreaders; ireader++)
+    int ireader,ivar,irec,igrp,ivset,iact;
+    for (ireader=0; ireader<readers->nreaders; ireader++) srt->vcf_buf[ireader].nrec = 0;
+    for (iact=0; iact<srt->nactive; iact++)
     {
-        srt->vcf_buf[ireader].nrec = 0;
-
+        ireader = srt->active[iact];
         bcf_sr_t *reader = &readers->readers[ireader];
         int rid   = bcf_hdr_name2id(reader->header, chr);
         grp.nvar  = 0;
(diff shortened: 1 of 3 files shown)


bcf_sr_sort.c
```diff
@@ -546,23 +558,8 @@ static void bcf_sr_sort_set(bcf_srs_t *readers, sr_sort_t *srt, const char *chr,
...
-        bcf_sr_t *reader = &readers->readers[0];
-        assert( reader->buffer[1]->pos==min_pos );
-        bcf1_t *tmp = reader->buffer[0];
-        for (j=1; j<=reader->nbuffer; j++) reader->buffer[j-1] = reader->buffer[j];
-        reader->buffer[ reader->nbuffer ] = tmp;
-        reader->nbuffer--;
-        readers->has_line[0] = 1;
-        return 1;
-    }
+    assert( srt->nactive>0 );
```
jkbonfield: Is this assert something that is there for protection against coding errors? Ie absolutely regardless of user input, well formed or otherwise, ```src->nactive``` will always be positive.  Or is it something that is protection for malformed data files?

If the former then it is fine, but if the latter then it should be replaced by an appropriate return code to signify error and handled by the calling application properly.

I don't know enough about this and the other asserts to be sure.

bcf_sr_sort.c
```diff
@@ -575,6 +572,19 @@ int bcf_sr_sort_next(bcf_srs_t *readers, sr_sort_t *srt, const char *chr, int mi
...
         srt->chr = NULL;
     }
+    if ( srt->nactive == 1 )
+    {
+        if ( readers->nreaders>1 )
+            memset(readers->has_line, 0, readers->nreaders*sizeof(*readers->has_line));
+        bcf_sr_t *reader = &readers->readers[srt->active[0]];
+        assert( reader->buffer[1]->pos==min_pos );
+        bcf1_t *tmp = reader->buffer[0];
+        for (j=1; j<=reader->nbuffer; j++) reader->buffer[j-1] = reader->buffer[j];
```
jkbonfield: How large is nbuffer expected to be?  This is probably more efficient, as well as clearer, written as

```
memmove(&reader->buffer[j-1], &reader->buffer[j], reader->nbuffer);
```

I haven't profiled it, but if it's called many times over it's more efficient still using a cyclic ring-buffer of some sort so there is no data-copying required.  It may be over-optimisation though so probably not worth it and I expect you've profiled this already to spot the need for this PR.

bcf_sr_sort.c
pd3: This is only a protection against coding errors.

bcf_sr_sort.c
pd3: The difference with `memmove` will not be measurable, the size of these buffers is very small, typically 1-2, rarely bigger than 10.  The bottleneck is `bcf_sr_sort_set`, I may need to rethink the strategy and the use of hashes there.

Thank you for looking at this.

bcf_sr_sort.c
```diff
@@ -327,6 +327,18 @@ char *grp_create_key(sr_sort_t *srt)
     }
     return ret;
 }
+void bcf_sr_sort_set_active(sr_sort_t *srt, int idx)
```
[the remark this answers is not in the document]
pd3: OK

bcf_sr_sort.c
jkbonfield: Thanks.  No issue with that then.

bcf_sr_sort.c
jkbonfield: Fair enough.  Thanks.

jkbonfield:
It works on the tests I've done, but I can't see any difference in performance at all.  What tests are you doing?

I tried something which I think is relatively fast - bcftools norm.  Are there better commands where the main bcftools command consumes less time and more is spent in the reader?  Even so, my tests run *very* quick for both.  Eg chr1 uncompressed BCF file with around 300,000 VCF lines is taking ~1.4sec both before and after this fix, so it feels like almost no time is spent in the synced reader.

pd3:
The case of `bcftools norm` already takes a shortcut and avoids the expensive `bcf_sr_sort_set`. A good test case where this matters is `bcftools concat -a` run on hundreds of disjoint files. A future test case to address is the same or similar, say `bcftools isec`, on hundreds of overlapping files.

jkbonfield:
Thanks for the tips on testing.  I saw a 7x speed up in my 250-way concat test.

Rebased & merged as ff823afd1a8f71a18e737e45acb34c14e7b24b5d

I realised my comment about return type int vs void was irrelevant as the function is purely internal.   I was thinking about ensuring we don't have to change prototypes if we later fix hts_expand to not die and instead return an error code.  However it's a useful placeholder even when not-exposed.

PR #648, opened 2018-01-11
hfile_libcurl seek improvements

daviesrob:
The BGZF cache calls `hseek()` whenever `load_block_from_cache()` [returns a block from the cache](https://github.com/samtools/htslib/blob/da27bfd89a/bgzf.c#L548).  This can lead to a lot of seeks happening without any actual reads between them.  On real files this doesn't matter very much, but with hfile_libcurl it results in lots of pointless HTTP requests for data that is never used.  Even worse, the problem gets bigger if the cache size in increased as a seek happens on every cache hit.

This patch fixes the problem by making `libcurl_seek()` perform the full process the first time it is called on each file handle (to test that seeking actually works), then for all following calls simply record the location where the next read should start and delay the actual work until `libcurl_read()` is called.  Testing on the first call means non-seekable streams should cause hseek() to return an error as would normally be expected.

The second commit here makes hfile_libcurl avoid small forward seeks by reading and discarding the data from the input stream.  The assumption is that this will be quicker than issuing a new request as it avoids several round trips to the server, and in fact libcurl may already have downloaded the data anyway.

Testing has shown that when the BGZF cache is working properly, this can reduce the number of reconnections by more than an order of magnitude under some conditions.

### DIFF
--- a/hfile_libcurl.c
+++ b/hfile_libcurl.c
@@ -49,6 +49,12 @@ DEALINGS IN THE SOFTWARE.  */
 // and slow servers
 #define AUTH_REFRESH_EARLY_SECS 60
 
+// Minimum number of bytes to skip when seeking forward.  Seeks less than
+// this will just read the data and throw it away.  The optimal value
+// depends on how long it takes to make a new connection compared
+// to how fast the data arrives.
+#define MIN_SEEK_FORWARD 1000000
+
 typedef struct {
     char *path;
     char *token;
@@ -101,7 +107,8 @@ typedef struct {
     unsigned tried_seek : 1; // At least one seek has been attempted
     int nrunning;
     http_headers headers;
-    off_t delayed_seek;     // Location to seek to before reading
+    off_t delayed_seek;      // Location to seek to before reading
+    off_t last_offset;       // Location we're seeking from
 } hFILE_libcurl;
 
 static off_t libcurl_seek(hFILE *fpv, off_t offset, int whence);
@@ -720,6 +727,8 @@ static ssize_t libcurl_read(hFILE *fpv, void *bufferv, size_t nbytes)
 {
     hFILE_libcurl *fp = (hFILE_libcurl *) fpv;
     char *buffer = (char *) bufferv;
+    off_t to_skip = -1;
+    ssize_t got = 0;
     CURLcode err;
 
     if (fp->delayed_seek >= 0) {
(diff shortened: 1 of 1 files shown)


hfile_libcurl.c
```diff
@@ -715,18 +727,54 @@ static ssize_t libcurl_read(hFILE *fpv, void *bufferv, size_t nbytes)
...
+                to_skip -= got;
+            } else {
+                got -= to_skip;
+                if (got > 0) {  // If enough was skipped, return the rest
+                    memmove(buffer, buffer + to_skip, got);
+                    to_skip = -1;
+                }
+            }
+        }
+    } while (to_skip >= 0 && ! fp->finished);
```
jkbonfield: Why `to_skip >= 0` and not `to_skip > 0`.  Skipping zero bytes sounds like it should require no data to be read?

Also I'm unsure if this is efficient when asking to read a few bytes.  Eg if nbytes is 10 but we're skipping 10,000 then we have 1000 loops as we can only consume 10 bytes at a time.

Would it make more sense to have something like:

```
// Discard any data if we're doing a small seek ahead
while (to_skip > 0 && !fp->finished) {
    char discard[MIN_SEEK_FORWARD]; // maybe a static global to avoid large stack size
   ... curl perform to consume data ...
   to_skip -= got;
}

// Read the requested data (code as before)
```

Maybe this access pattern never happens though.  Do we ever do small reads in hfile, rather than entire blocks?

I also don't understand the "if enough was skipped, return the rest" bit.  How does this happen?  The caller asked for nbytes, and we set buffer.len to nbytes, but how do we then ever have enough?  I think it's a NOP at the moment, but if it wasn't it'd be bugged anyway.  Eg if we want 10 bytes and skip 5, then we read 10, discard 5, and return the other 5 (not 10).

Is it working because we then return 5 and require the caller to spot the short read and call us in a loop until it got as much as they initially wanted?  That's probably necessary already, but exploiting it is yet more round trips which can probably be avoided by a discard to HERE and read to THERE two-stage approach.

daviesrob: It's `to_skip >= 0` to ensure that the buffer always ends up with some of the data that we wanted to read in it (by forcing another loop round when skip has reached zero).  If we didn't do that, a skip of an exact multiple of `nbytes` would end up with nothing in the buffer which would be misinterpreted as end of file.

The rest of this is down to how libcurl works.  It expects to be able to pass at most `CURL_MAX_WRITE_SIZE` bytes to a callback function (in this case `recv_callback()`).  The callback must either use up all of this data or return `CURL_WRITEFUNC_PAUSE`, otherwise libcurl will stop and report an error.  This means `nbytes` must be at least `CURL_MAX_WRITE_SIZE` for `libcurl_read()` to work at all.

Given this, it seems easier to borrow the output buffer for the discard operation as we know that it's big enough already, and it saves having to allocate more storage.  It's also important to never call `wait_perform()` with an undersized buffer as it will never finish, so we try to read as much as possible into the buffer on each loop round.  When we've skipped enough, we return the remaining data.  As you've noted, this will generally not be as much as requested but that isn't a problem as `hfile` will ask for more as soon as it's used up the bit that it's been given.

PR #649, opened 2018-01-16
Accomodate use of some internal APIs in plug-ins

### LINKED ISSUE #623: hfile_libcurl plugin uses private htslib APIs
Since PR #600, hfile_libcurl uses the JSON tokeniser, which is private HTSlib functionality declared in _hts_internal.h_. This is a problem when hfile_libcurl is built as a plugin.

Plugins should not use private HTSlib functions (apart from those in _hfile_internal.h_ for hfile plugins, obviously) as they are not tightly bound to particular htslib versions — this is part of the point of being a plugin!

As alluded to in the 39ca08917a1ed4b94ed6c2aae4a2038f26edb193 commit message, the JSON tokeniser should be finalised as a public API before it is used in plugins.

daviesrob:
Use of functions declared in hts_internal.h has crept into two
plug-ins: hfile_s3 uses isalnum_c(), etc. and hfile_libcurl uses
the JSON parsing API.  This makes these APIs more fixed than is
suggested by their being defined in the internal header files
and not the public ones.

Move the function declarations to a new file (textutils_internal.h)
which includes a note that the functions may be used in plug-ins.
This file is included from both hts_internal.h and hfile_internal.h.
Alter hfile_s3.c and hfile_libcurl.c to no longer include
hts_internal.h.

The definition of struct hts_json_token is left in hts_internal.h.
New functions to allocate this struct and access the contents are
added with declarations in textutils_internal.h.  This is to
enable changes to the structure in the future without breaking
existing plug-ins that use it.

Fixes #623.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,8 +175,9 @@ cram_sam_header_h = cram/sam_header.h cram/string_alloc.h cram/pooled_alloc.h $(
 cram_samtools_h = cram/cram_samtools.h $(htslib_sam_h) $(cram_sam_header_h)
 cram_structs_h = cram/cram_structs.h $(htslib_thread_pool_h) cram/string_alloc.h $(htslib_khash_h)
 cram_open_trace_file_h = cram/open_trace_file.h cram/mFILE.h
-hfile_internal_h = hfile_internal.h $(htslib_hfile_h)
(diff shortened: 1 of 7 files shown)


jkbonfield:
Squashed and merged as e539b32

PR #650, opened 2018-01-16
CUD operations on HD line

valeriuo:
Adds create/update/delete operations to tag:value pairs of the HD line from a SAM header.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1,6 +1,6 @@
 /*  sam.c -- SAM and BAM file I/O and manipulation.
 
-    Copyright (C) 2008-2010, 2012-2017 Genome Research Ltd.
+    Copyright (C) 2008-2010, 2012-2018 Genome Research Ltd.
     Copyright (C) 2010, 2012, 2013 Broad Institute.
 
     Author: Heng Li <lh3@sanger.ac.uk>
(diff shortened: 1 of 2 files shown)


sam.c
```diff
@@ -1088,6 +1088,73 @@ int sam_hdr_write(htsFile *fp, const bam_hdr_t *h)
...
+    char *p, *q, *beg = NULL, *end = NULL, *newtext;
+    if (!h || !key)
+        return -1;
+
+    if (h->l_text > 3) {
+        if (strncmp(h->text, "@HD", 3) == 0) { //@HD line exists
+            if ((p = strchr(h->text, '\n')) == 0) return -1;
+            *p = '\0'; // for strstr call
+
+            char tmp[5];
```
daviesrob: Could maybe avoid the str* functions below by assigning:
```C
char tmp[5] = { '\t', key[0], key[0] ? key[1] : '\0', ':', '\0' };
```
Could also be simplified if we ensure that key is two chars long first.

sam.c
```diff
@@ -1088,6 +1088,73 @@ int sam_hdr_write(htsFile *fp, const bam_hdr_t *h)
...
+                beg = end = p;
+                *p = '\n';
+            }
+        }
+    }
+    if (beg == NULL) { // no @HD
+        if (!sam_version)
+            return -1;
+
+        h->l_text += strlen(sam_version) + 12;
```
daviesrob: Might want to check for overflow.  h->l_text is unsigned 32 bit.

yfarjoun:
the tests have been running for the past 9 hours....do you know if it's due to your changes or something in Travis itself?

yfarjoun:
ah, it's on the travis website...there's a problem there....nm me!

PR #653, opened 2018-01-25
Fix end-of-range check & setting fd->eof in cram_next_slice()

### LINKED ISSUE #651: htslib 1.6 breaking CRAM fetch: "truncated file or corrupt BAM index file"
From https://github.com/pysam-developers/pysam/issues/567

It appears that certain CRAM files that work fine under htslib < 1.6 (tested with 1.5 and 1.3.1, using either samtools or pysam <0.13) throw errors when queried using tools built against htslib 1.6. Reproducing examples are given in the linked issue.

cc @kyleabeauchamp

daviesrob:
In the loop that skips over containers in range queries, check that the range and container references are the same before comparing the start and end positions. Also check that range.refid is not -1 (unmapped), as in that case the start and end positions may not be valid.

Ensure that fd->eof is set when making range requests and the container found does not have the desired reference id.

Fixes #651

Also avoid accessing the bam record if cram_get_bam_seq() fails.  It's not likely to contain valid data. Spotted when looking into the problem above.  Instances fixed in cram_readrec(), sam_bam_cram_readrec() and sam_read1().

### DIFF
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -2941,7 +2941,8 @@ static cram_slice *cram_next_slice(cram_fd *fd, cram_container **cp) {
 	if (fd->range.refid != -2) {
 	    while (c->ref_seq_id != -2 &&
 		   (c->ref_seq_id < fd->range.refid ||
-		    c->ref_seq_start + c->ref_seq_span-1 < fd->range.start)) {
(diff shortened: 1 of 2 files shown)


jkbonfield:
Looks good and merged.  Thanks for the fixes.

PR #655, opened 2018-01-30
Mixed Cram bug fixes

### LINKED ISSUE #654: Repeated attempt to iterate over any portion of the same 'multi_seq' CRAM slice results in an erroneous failure to populate the reference.
I've recently looked into several reported cram errors from the manta sv caller (which uses htslib), and have been able to reproduce the problem in `samtools view`.

The error seems to occur for an hts iterator reading a CRAM file when:
1. A request is made for a region from the CRAM which is treated using the multi_seq protocol (ie. `ref_id == -2`)
2. A second consecutive request is made for the same region or any region from another chromosome falling in the range of chroms covered by the multi_seq region.

When the error occurs, the following conditions are observed:
1.  The error message "[E::cram_get_ref] Failed to populate reference for id X" appears for all X covered in the multi_seq range except the last 2.
2. Unlike other scenarios where a 'failed to populate reference' is triggered, there is no error report of the form "[E::cram_decode_slice] Unable to fetch reference", and nothing programatically detectable via the htslib API (to my knowledge).
3. Reads returned from the second (or higher) request to the multi_seq region are returned with the ANY/'=' basecall substituted in for all match positions in the read, even though these are not being used in the source CRAM.

I have not been able to reproduce the condition with data I can share, but here is an example of how I am triggering the problem and the corresponding error messages:
```
$ samtools view example.cram -T /illumina/scratch/BeaconVariantCalling/Resources/References/GRCh38_1kg/GRCh38_full_analysis_set_plus_decoy_hla.fa chrUn_JTFH01000020v1_decoy chrUn_JTFH01000020v1_decoy >|  reads
[E::cram_get_ref] Failed to populate reference for id 855
[E::cram_get_ref] Failed to populate reference for id 856
[E::cram_get_ref] Failed to populate reference for id 857
[E::cram_get_ref] Failed to populate reference for id 858
[E::cram_get_ref] Failed to populate reference for id 859
[E::cram_get_ref] Failed to populate reference for id 860
[E::cram_get_ref] Failed to populate reference for id 862
```

Note the id of chrUn_JTFH01000020v1_decoy  is 862.  In the above example reads from the first request for chromosome 'chrUn_JTFH01000020v1_decoy ' appear with full read sequences, in the second pass through 'chrUn_JTFH01000020v1_decoy' all match segments in the read are replaced with the ANY/'=' symbol.

The same error occurs for any combination of chromosomes in this multi-seq region, for instance, using samtools 1.7:
```
$ samtools view example.cram -T /illumina/scratch/BeaconVariantCalling/Resources/References/GRCh38_1kg/GRCh38_full_analysis_set_plus_decoy_hla.fa chrUn_JTFH01000019v1_decoy chrUn_JTFH01000018v1_decoy >|  reads
[E::cram_get_ref] Failed to populate reference for id 855
[E::cram_get_ref] Failed to populate reference for id 856
[E::cram_get_ref] Failed to populate reference for id 857
[E::cram_get_ref] Failed to populate reference for id 858
[E::cram_get_ref] Failed to populate reference for id 859
[E::cram_get_ref] Failed to populate reference for id 860
[E::cram_get_ref] Failed to populate reference for id 862
```

I have not been able to recreate this with simulated or public data, in large part because I do not understand the conditions that trigger multi_seq slices vs. regular single-chromosome cram slices. It is also possible that I'm dealing with a cram file with undetected corruption, but the variety of manta users using different cram pipelines but reporting similar problems (https://github.com/Illumina/manta/issues/109) makes me suspect that this is either unlikely or else, if there were such pervasive undetected corruption this itself would be the concern.

What I know of the mechanism of the problem is as follows:
1. In the first call to `cram_decode_slice()`, the sub-functions call to `cram_get_ref()` runs as expected.
2. After this call, `cram_ref_decr_locked` is first called on the target id (say, 857 in the above example), and then on target+1 (858, in which case last_id is 857). This leads to `fd->refs->ref_id[857].length` being set to 0, here:

https://github.com/samtools/htslib/blob/78dbcce8616228c5b4a5ac8a141386995e0747ea/cram/cram_io.c#L2260

3. On the second call to `cram_decode_slice()`, the sub-functions call to `cram_get_ref()` detects the length == 0 condition and interprets this as a failure to load the sequence from the fasta file provided on the command-line -- the method then falls beck to some reference cache and ftp/md5 reference recovery methods. On my test machine at least, these recovery methods do not work, so that I observe the full error pattern described above. For a user where these steps might work, I would contend it is still a bug to supply the correct fasta and the commandline and having htslib download the same sequence over ftp instead.

Happy to put in some more effort to create a reduced cram test case if I can get more pointers on the above described phenomena.

jkbonfield:
A mish-mash of CRAM bug fixes.  Fixes #654


ctsa:
In additional to #654, I confirm this closes the parent issue from manta https://github.com/Illumina/manta/issues/109.

jkbonfield:
Thanks for the testing @ctsa.  I see I need to tidy my PR a little bit with some stray comments left in after debugging, but hoping we can review and merge this soon.

cram/cram_io.c
```diff
@@ -4403,8 +4410,11 @@ int cram_set_voption(cram_fd *fd, enum hts_fmt_option opt, va_list args) {
 	break;
 
     case CRAM_OPT_RANGE:
+	pthread_mutex_lock(&fd->range_lock);
 	fd->range = *va_arg(args, cram_range *);
-	return cram_seek_to_refpos(fd, &fd->range);
+	pthread_mutex_unlock(&fd->range_lock);
+	int ret = cram_seek_to_refpos(fd, &fd->range);
```
daviesrob: I think there could be a race here, where `cram_decode_slice()` in a thread tries to use the new `fd->range` before `cram_seek()` is able to drain the job from the queue.  I guess it might not matter as the results are thrown away, but it could be avoided by passing a `cram_range *` to somewhere other than `&fd->range` to `cram_seek_to_refpos`, and then getting that to fill in `fd->range` after it's done the seek.  Insisting that the second argument to `cram_seek_to_refpos()` is not `&fd->range` would allow some of the locks in that function to be removed as well.

cram/cram_decode.c
```diff
@@ -2442,14 +2442,34 @@ int cram_decode_slice(cram_fd *fd, cram_container *c, cram_slice *s,
...
-							    1, 0);
-			s->ref = refs[cr->ref_id];
+			pthread_mutex_lock(&fd->range_lock);
+			if (fd->range.refid < 0 || cr->ref_id == fd->range.refid) {
+			    pthread_mutex_unlock(&fd->range_lock);
+			    if (!refs[cr->ref_id])
+				refs[cr->ref_id] = cram_get_ref(fd, cr->ref_id,
+								1, 0);
+			    if (!(s->ref = refs[cr->ref_id])) {
+				pthread_mutex_unlock(&fd->range_lock);
```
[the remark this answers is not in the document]
jkbonfield: Thanks.  Not sure what changes I was doing to leave it like this.

cram/cram_decode.c
```diff
@@ -2442,14 +2442,34 @@ int cram_decode_slice(cram_fd *fd, cram_container *c, cram_slice *s,
...
+				return -1;
+			    }
+			} else {
+			    pthread_mutex_unlock(&fd->range_lock);
+			    // For multi-ref containers, we don't need to fetch all
+			    // refs if we're only querying one.
+			    s->ref = NULL;
+			}
 
+			pthread_mutex_lock(&fd->range_lock);
```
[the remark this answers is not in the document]
jkbonfield: Some functions like cram_get_ref can be slow and also contain various mutexes.  I want to keep the locking as small as possible, but it can be refactored to be cleaner.

cram/cram_decode.c
```diff
@@ -2442,14 +2442,34 @@ int cram_decode_slice(cram_fd *fd, cram_container *c, cram_slice *s,
...
-			    refs[last_ref_id] = NULL;
+			    if (fd->range.refid != -2 && last_ref_id == fd->range.refid) {
+				pthread_mutex_unlock(&fd->range_lock);
+				cram_ref_decr(fd->refs, last_ref_id);
+				refs[last_ref_id] = NULL;
+			    } else {
+				pthread_mutex_unlock(&fd->range_lock);
+			    }
+			} else {
+			    pthread_mutex_unlock(&fd->range_lock);
```
jkbonfield: ... especially this tangled mess!

cram/cram_io.c
jkbonfield: I don't think the race matters as we're immediately tearing down the threads anyway.  If they erroneously start reading a block that they didn't need because the range increased then it'll be short lived.  However I agree it can be simplified and will do so.

PR #657, opened 2018-02-02
support for multiple AWS regions

mwchristiansen:
The tool currently only works on S3 buckets in us-east-1.  This fix will allow people to read from data in buckets in any S3 region.  I have tested this works in us-west-2.  

Thank you for reviewing and merging into the master.


mwchristiansen:
please refer to PR 661.  I think that one is better

PR #661, opened 2018-02-21
add s3 region support

mwchristiansen:
Reads the environment variable AWS_DEFAULT_REGION or parses an ini file for "region" to read S3 buckets from the non-default region (us-east-1)

This fix was made possible thanks to contributions from @jlrgraham and @opsroller

### DIFF
--- a/hfile_s3.c
+++ b/hfile_s3.c
@@ -41,6 +41,7 @@ typedef struct {
     kstring_t id;
     kstring_t token;
     kstring_t secret;
+    kstring_t region;
     char *bucket;
     kstring_t auth_hdr;
     time_t auth_time;
@@ -247,6 +248,7 @@ static void free_auth_data(s3_auth_data *ad) {
     free(ad->id.s);
     free(ad->token.s);
     free(ad->secret.s);
+    free(ad->region.s);
     free(ad->bucket);
     free(ad->auth_hdr.s);
     free(ad);
@@ -358,6 +360,7 @@ static hFILE * s3_rewrite(const char *s3url, const char *mode, va_list *argsp)
         if ((v = getenv("AWS_ACCESS_KEY_ID")) != NULL) kputs(v, &ad->id);
         if ((v = getenv("AWS_SECRET_ACCESS_KEY")) != NULL) kputs(v, &ad->secret);
         if ((v = getenv("AWS_SESSION_TOKEN")) != NULL) kputs(v, &ad->token);
+        if ((v = getenv("AWS_DEFAULT_REGION")) != NULL) kputs(v, &ad->region);
 
         if ((v = getenv("AWS_DEFAULT_PROFILE")) != NULL) kputs(v, &profile);
         else if ((v = getenv("AWS_PROFILE")) != NULL) kputs(v, &profile);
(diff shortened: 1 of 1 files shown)


daviesrob:
Hi, I've spotted a few issues with this that would need to be fixed before it can be merged.  Most notably it breaks if you don't set a region, or do set `host_base` in `.s3cfg` due to a couple of bugs around [line 386](https://github.com/mwchristiansen/htslib/blob/8b0092b/hfile_s3.c#L386).  The three `kputs` lines need to be in a block, otherwise only the first one is guarded by the `if (host_base.l == 0)`.  If `host_base` already contains something then it tries to append `<region>.amazonaws.com` whether you wanted it or not, resulting in an unusable host name.  It also doesn't check if `region` has been set - you get a segmentation fault if it wasn't.

Looking at the [S3 regions list](https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region), it turns out that the `s3-<region>.amazonaws.com` doesn't exist for some endpoints, including `us-east-1`.  `s3.<region>.amazonaws.com` does work for all of the endpoints outside China, so that would be a better choice.  Putting all this together, a better solution for setting `host_base` might be:
```c
    if (host_base.l == 0) {
        ksprintf(&host_base, "%s%s.amazonaws.com",
                 ad->region.l > 0 ? "s3." : "s3",
                 ad->region.l > 0 ? ad->region.s : "");
    }
```

In the configuration files, `region` appears to be valid in `.aws/credentials`, but in `.s3cfg` it looks like `bucket_location` actually performs the same function.  There are some complications though as the default you get when you set up `s3cmd` seems to be `US`, which doesn't actually correspond to any known endpoint.  This doesn't harm `s3cmd` because it redirects to the correct place.  Presumably if we used this parameter we would need to do the [same mapping as s3cmd](https://github.com/s3tools/s3cmd/blob/aa18fa41/S3/S3.py#L432), i.e. convert `US` to `us-east-1` (or possibly nothing) and `EU` to `eu-west-1`.

The other problem with `.s3cfg` is that the default file sets `host_base` to `s3.amazonaws.com`, which would override the region setting.  We would either need to tell `.s3cfg` users to change `host_base` in order to use a different endpoint (which they could do without this pull request), or ignore the default setting so that it can be reset based on `bucket_location`.

I've tried some experiments with reading a bucket in `eu-west-1` from various endpoints.  Obviously it works correctly from `s3.eu-west-1.amazonaws.com`.  `s3.amazonaws.com` also works, presumably through some behind-the-scenes magic.  Any other endpoint, including `s3.us-east-1.amazonaws.com`, returns `301 Moved Permanently` which libcurl fails to follow -- presumably there is no Location header for it to use.  Trying the same with `s3cmd` in debug mode shows that it reads the error message to get the correct endpoint and tries again.  Ideally we should do something similar, although it would be quite a bit of work to implement.  We also need to get around to supporting version 4 signatures, as we're currently quite limited in the number of regions HTSlib can access directly.

Finally, could you base pull requests on the `develop` branch rather than `master`, please?  For better or worse we try to follow the [git-flow](http://nvie.com/posts/a-successful-git-branching-model/) model where all new work is done on `develop` and only merged to and from `master` at release time.

PR #666, opened 2018-03-09
Cram multi-slice container fixes.

jkbonfield:
All bar one of these are related to multi-threading.

The first one though is for range requests with multi-slice containers.

Sorry it's complex, but the multi-threading and multi-slice handling is just that - complex (and likely fragile!).
It's been hammered in valgrind with memcheck (--leak-check=full), drd and helgrind.   The entire test harness got put through memcheck, which now only shows memory leaks in the perl parts.


jkbonfield:
Also passes AppVeyor: https://ci.appveyor.com/project/jkbonfield/htslib

I'm not sure why it isn't filling out this page though.

jkbonfield:
The latest code has been tested on some more randomised CRAM files, with the number of reads per slice varying on each slice and similarly slices per container, as well as having random choice of HUFFMAN vs EXTERNAL codecs in each container so no two containers use the same compression header struct.

(Sanger locals see ~jkb/scratch/data/*.rand[123].cram)

It looks to pass everything I've thrown at it so far, unlike the current develop branch, so would appear to be a big step in the right direction.

liuxf09:
I hope `Picard` can merge this change very soon.

jkbonfield:
Picard is a different project, and the (original) Java implementation of CRAM.  I don't know how much multi-threading it has, but I'm not aware of the same bug there.  If there is, please raise it on their github project.

As for this one, it's been merged already.

PR #667, opened 2018-03-13
Minor bug fix

kgururaj:


daviesrob:
It turns out @valeriuo's seg fault was down to `bcf_sr_remove_reader` shifting the array of readers to fill in the hole, so you have to take the renumbering into account.

I've converted some tabs into spaces (we avoid tabs due to disagreements about how long they should be) and will merge the result.

PR #673, opened 2018-03-23
fqidx - a modification of faidx for FASTQ

mp15:
Bit of code duplication but this might be useful to other folks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,7 @@ LIBHTS_OBJS = \
 	bgzf.o \
 	errmod.o \
 	faidx.o \
+	fqidx.o \
 	hfile.o \
 	hfile_net.o \
 	hts.o \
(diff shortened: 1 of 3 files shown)


daviesrob:
There is rather a lot of duplication here.  Would it be possible to combine the fasta and fastq versions?  It might make faidx a little more complicated, but I think it would be better than having to look after twice as much code.

The documentation on the index in the header file isn't correct.  It looks like it was copied from faidx.h but not updated.  On the index format itself, it might be better to put the `qual_offset` value as the last item on the index line as that would be more compatible with the existing `faidx` format.  `fai_read()` currently ignores anything after the numbers it expects to find, so putting the extra value at the end would allow `faidx` to return the desired sequence even when given a fastq index.

PR #675, opened 2018-03-26
Add option to set compression level in bgzip

nathanweeks:
It would be nice if bgzip supported compression levels other than the default (like some of the samtools subcommands do). This pull request adds `-l INT` short and `--compress-level INT` long options to bgzip.

### DIFF
--- a/bgzip.c
+++ b/bgzip.c
@@ -73,24 +73,25 @@ static int bgzip_main_usage(void)
     fprintf(stderr, "Version: %s\n", hts_version());
     fprintf(stderr, "Usage:   bgzip [OPTIONS] [FILE] ...\n");
     fprintf(stderr, "Options:\n");
-    fprintf(stderr, "   -b, --offset INT        decompress at virtual file pointer (0-based uncompressed offset)\n");
-    fprintf(stderr, "   -c, --stdout            write on standard output, keep original files unchanged\n");
-    fprintf(stderr, "   -d, --decompress        decompress\n");
-    fprintf(stderr, "   -f, --force             overwrite files without asking\n");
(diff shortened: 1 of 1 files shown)


daviesrob:
Thanks for this.  I've rebased it onto develop and merged as effcb1e1.

In a separate commit (2d5fa5dfe) I've also added a manual page for bgzip, as I noticed that the old documentation (in the tabix man page) was a bit sparse.  I've also included documentation of the .gzi index format in the new page as there's nothing on it elsewhere.

jmarshall:
Thanks for merging this. Did you consider using the API (`bgzf_open()`'s mode parameter) to set the level rather than fiddling with the internals afterwards — which probably works for now, but no guarantees `bgzf_write_init()` won't do special set up in the future…

nathanweeks:
@daviesrob : thanks, it's great to have an updated man page for bgzip.

@jmarshall : I had looked at the mode parameter for `bgzf_open()`, but didn't use it (in the interest of laziness---it looked like fewer changed lines of codeto not use it). The possibility that future changes to the BGZF struct might break this hadn't occurred to me, but I guess it makes sense to treat BGZF as an opaque data type. It probably would be a good idea to set this in a more future-proof way.

daviesrob:
@jmarshall: Good spot.  I've added commit 6b85e52b06 to make it use the approved method of setting the compression level.

jmarshall:
Thanks @daviesrob :smile: The other triviality I noticed was the capitalisation in the Usage display. I wouldn't mind capitalising all of them (and if someone wanted to rephrase/reformat three entries for 80 columns that'd be great too!).

PR #678, opened 2018-03-29
Memory leak when iterating over new queries

rpetrovski:
The following code consumes memory indefinitely. Memory leak is gone once the change is applied.
Steps:
1. build htslib
2. compile test.c with:
gcc -O0  -ggdb -I htslib/install/include test.c -L htslib/install/lib/ -l:libhts.a -lz -lpthread -llzma -lbz2
3. run ./a.out some.cram chr1
4. watch virtual memory going up in top
5. apply patch, rebuild test.c, notice virtual memory does not change

test.c:
```c++
#include <stdlib.h>
#include <string.h>
#include <stdio.h>
#include <unistd.h>
#include <math.h>
#include <inttypes.h>
#include <stdbool.h>
#include <assert.h>
#include "htslib/sam.h"
#include "htslib/faidx.h"
#include "htslib/kstring.h"
#include "htslib/khash.h"
int main(int argc,char** argv)
{
        hts_itr_t *iter=NULL;
        hts_idx_t *idx=NULL;
        samFile *in = NULL;
        bam1_t *b= NULL;
        bam_hdr_t *header = NULL;
        if(argc!=3) return -1;
        in = sam_open(argv[1], "r");

        if(in==NULL) return -1;
        if ((header = sam_hdr_read(in)) == 0) return -1;

        idx = sam_index_load(in,  argv[1]);
        if(idx==NULL) return -1;

        b = bam_init1();
        fputs("reading\n",stdout);
        do
        {
                if (iter) hts_itr_destroy(iter);
                iter = sam_itr_querys(idx, header, argv[2]);
                if(!iter) return -1;
//              fputs("DO STUFF\n",stdout);
        }
        while (sam_itr_next(in, iter, b) >= 0);

        fputs("done reading\n",stdout);

        hts_itr_destroy(iter);
        bam_destroy1(b);
        hts_idx_destroy(idx);
        bam_hdr_destroy(header);
        sam_close(in);
        return 0;
}
```

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -2753,6 +2753,11 @@ void cram_free_container(cram_container *c) {
 	free(c->slices);
     }
 
+    if (c->slice)
+    {
+        cram_free_slice(c->slice);
+    }
+
     for (id = DS_RN; id < DS_TN; id++)
 	if (c->stats[id]) cram_stats_free(c->stats[id]);
 


daviesrob:
Your test destroys and recreates the iterator on each loop.  It will just return the same read forever -- is that what you want?

Anyway, I think we may have fixed this already somewhere in 6b6c8a3fdd.  I can't reproduce the problem even with the original version of the test program.  Have you tried running it against the latest `devel` branch?

daviesrob:
I'll take it back.... I've now managed to get a very rapid growth.

Helps if you test with a cram file....

rpetrovski:
My test exemplifies the issue. This is obviously not what I normally need.
But it bears the essence. I did see the problem on latest devel. And
valgrind pointed me to the source. I will double check. Are you using cram?
The issue does not show with bam.

rpetrovski:
b.t.w. This pull request is broken. I've made another one https://github.com/samtools/htslib/pull/679 which passes automated checks.

Let's continue there.

PR #680, opened 2018-04-03
S3 multi region support

mwchristiansen:
Reads the environment variable AWS_DEFAULT_REGION or parses an ini file for "region" to read S3 buckets from multiple regions (us-west-2) while preserving functionality to access the default region (us-east-1)

This fix was made possible thanks to contributions from @jlrgraham and @opsroller

@daviesrob , i hope this address the changes you discussed in PR #661 

Thanks for your feedback

### DIFF
--- a/hfile_s3.c
+++ b/hfile_s3.c
@@ -41,7 +41,7 @@ typedef struct {
     kstring_t id;
     kstring_t token;
     kstring_t secret;
-    char *region;
+    kstring_t region;
     char *bucket;
     kstring_t auth_hdr;
     time_t auth_time;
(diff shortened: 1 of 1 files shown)


daviesrob:
Thanks for the update and sorry not to get back to you sooner.

Would it be possible to change `region` in the `.s3cfg` section to `bucket_location`?  That appears to be the key that `s3cmd` uses for the purpose of setting the region (as far as I can tell - I need to do more research on this).

mwchristiansen:
can we merge these changes?  i am not familiar with how people use s3cmd.  perhaps you have done more research on the matter?

daviesrob:
As we're using s3cmd's configuration file, we should try to behave as much as possible like it does.  In this case, it seems to use `bucket_location` so set the region that it tries to download from / upload to rather than `region` (which I don't think it understands).

I think this would be OK with a rebase to current `develop` and that one change.

daviesrob:
#839 has added this functionality, along with the ability to follow both 301 redirections (for path-style buckets) and extract the correct region out for 400 responses (for version 4 signatures with virtual-hosted buckets).  Hopefully that should have solved the problem for you.

PR #699, opened 2018-05-15
Cigar P support

jkbonfield:
Resolves #157 
Resolves #226 

This is ABI breaking as it adds an extra field to the bam_pileup1_t structure and thus changes the size. (This causes issues for anything using an array of these structs).

The PR now also includes a bug fix to handling of N cigar op.

Note I also have a rewritten and far cleaner implementation of bam_resolve_cigar2 if people prefer, although it needs a little more work still.  I reimplemented it to try and better understand what's happening.


jkbonfield:
Oops, missing something.  I'll revise it. :-)

Makefile
```diff
@@ -447,6 +452,14 @@ test/thrash_threads7: test/thrash_threads7.o libhts.a
 	$(CC) $(LDFLAGS) -o $@ test/thrash_threads7.o libhts.a -lz $(LIBS) -lpthread
 test_thrash: $(BUILT_THRASH_PROGRAMS)
 
+test/fieldarith.o: test/fieldarith.c $(htslib_sam_h)
```
jmarshall: These dependencies are already listed about 30 lines above…

daviesrob: Thanks for the spot.  I'm suspecting a rebase hiccup somewhere in the long history of this PR.  Should be gone soon.

PR #700, opened 2018-05-15
Fix warnings found when compiling with gcc 8.1.0

daviesrob:
Possible integer overflow in kputs.

Possible buffer overflow in cram_populate_ref().

Not caught by gcc, but seen at the same time: Possible overflow in expand_cache_path().

Possible use of uninitialised values in bcf_sr_regions_next().  @pd3 probably needs to check the fix for this one.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -227,11 +227,9 @@ bam_hdr_t *bam_hdr_read(BGZF *fp)
 
 int bam_hdr_write(BGZF *fp, const bam_hdr_t *h)
 {
-    char buf[4];
     int32_t i, name_len, x;
     // write "BAM1"
-    strncpy(buf, "BAM\1", 4);
-    if (bgzf_write(fp, buf, 4) < 0) return -1;
+    if (bgzf_write(fp, "BAM\1", 4) < 0) return -1;
     // write plain text and the number of reference sequences
     if (fp->is_be) {
         x = ed_swap_4(h->l_text);
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -2886,6 +2886,7 @@ static int process_one_read(cram_fd *fd, cram_container *c,
                 kh_val(s->pair[sec], k) = rnum;
         } else {
             new = 1;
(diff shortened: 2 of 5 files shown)


jkbonfield:
CRAM changes look fine.

daviesrob:
@jmarshall I avoided doing that for ABI preservation purposes, although I see it's a `static inline` so maybe we can get away with it.

One of the things I intend to do for the next ABI bump is find as many things that are badly sized as I can, and fix them.

jkbonfield:
+1 for using size_t instead.  Also there are more warnings found if you don't use optimisation during compilation.

daviesrob:
Rebased, squashed and a couple of extra fixes added for warnings that popped up at different optimisation levels.

The kputs() warning is now fixed by making kputsn() and kputsn_() take size_t instead of int.  [abi-compliance-checker](https://github.com/lvc/abi-compliance-checker) says this is fine.

sam.c
```diff
@@ -230,7 +230,7 @@ int bam_hdr_write(BGZF *fp, const bam_hdr_t *h)
     char buf[4];
     int32_t i, name_len, x;
     // write "BAM1"
-    strncpy(buf, "BAM\1", 4);
+    memcpy(buf, "BAM\1", 4);
```
jmarshall: Ugh. If you're going to touch this piece of code anyway, why not actually improve it by nuking `buf` entirely:
`if (bgzf_write(fp, "BAM\1", 4) …`

daviesrob: Good point...

jmarshall:
This seasoned C programmer suspects you should have clarified it to `a + (b? 1 : 0)` rather than the rather different `(a + b)? 1 : 0`.

Hurrah for the nanny state! Compiler warning vindicated!! :rofl:

daviesrob:
Now fixed the right way, thanks.

jkbonfield:
Gah, my mind isn't where it should be I think!  Sorry.  I'll go back to less critical things. :-)

PR #701, opened 2018-05-21
Modifies faidx to work on FASTQ

whitwham:
Adds the ability to read and index FASTQ files in a similar way to FASTA.
FASTQ reading code by @mp15 from #673 used.
Closes #673

Fixes samtools/samtools#804 where a missing gzi file is not built if a fai file is present.

Changes the len error return from -2 to -3 for fai_fetch to differentiate missing entries from out of range numbers.


whitwham:
A few things to sort out on review.

daviesrob:
Merged this part.  `faidx.5` manual page update to follow...

PR #703, opened 2018-05-24
Make htslib a little bit more efficient

daviesrob:
The first two commits here reduce the number of times threads get woken up only to discover there's nothing to do.  Fewer wake-ups means less lock contention.

The third avoids a `memcpy()` and extra call to `fp->backend->write()` when `hwrite()` is called with more data than fits in the buffer and the buffer was empty to begin with.  This seems to happen quite a bit when writing bam files.

The fourth copies the code for handling level 0 (uncompressed blocks) from the libdeflate version of bgzf_compress() into the zlib version as well.  The main benefit is that it avoids some unnecessary memory allocations in `deflateInit2()` and frees in `deflateEnd()` for buffers that are never used.

The fifth is also for uncompressed bgzf, this time when using threads.  It removes the `memcpy()` between the uncompressed and compressed buffers by writing the data into the correct location in the compressed buffer and then calling the thread job to calculate the CRC and surround the data with the normal header and trailer.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -427,27 +427,38 @@ int bgzf_compress(void *_dst, size_t *dlen, const void *src, size_t slen, int le
     z_stream zs;
     uint8_t *dst = (uint8_t*)_dst;
 
-    // compress the body
-    zs.zalloc = NULL; zs.zfree = NULL;
-    zs.msg = NULL;
-    zs.next_in  = (Bytef*)src;
-    zs.avail_in = slen;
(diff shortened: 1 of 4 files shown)


jkbonfield:
It all looks eminently sensible, but try hard as I may I cannot actually find any difference in the speed (converting a cache-fitting u.bam to u.bam).

That said, I'm happy to concede that it may make a difference elsewhere and the thread pool is used by other tools as well.

PR #704, opened 2018-05-24
sam_format1 : replace if/then with switch/case

lindenb:
some minor optimizations : 

in `sam.c / sam_format1`

I replaced the successions of `if/then` with `switch/case`.

replaced the `goto` with a macro.

I also fixed a | FIXME` in the same function.

 >  FIXME: for better performance, put the loop after "if"

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1517,92 +1517,119 @@ int sam_format1(const bam_hdr_t *h, const bam1_t *b, kstring_t *str)
         key[0] = s[0]; key[1] = s[1];
         s += 2; type = *s++;
         kputc('\t', str); kputsn((char*)key, 2, str); kputc(':', str);
-        if (type == 'A') {
-            kputsn("A:", 2, str);
-            kputc(*s, str);
-            ++s;
-        } else if (type == 'C') {
-            kputsn("i:", 2, str);
-            kputw(*s, str);
-            ++s;
-        } else if (type == 'c') {
-            kputsn("i:", 2, str);
-            kputw(*(int8_t*)s, str);
-            ++s;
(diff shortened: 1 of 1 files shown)


daviesrob:
I've had mixed results with this.  For an uncompressed input file with no B-type tags, this PR is about 10% slower on an old Sandy Bridge-era CPU compared to develop.  On a newer Broadwell core, it's about 5% faster.  Looking at the resulting assembly code, the case statement got implemented as an indirect jump.  Presumably the Broadwell core is better at branch predicting this than the Sandy Bridge one.

For the same file with the addition of B tags containing 64 uint16_t values, this pull request is about 3% faster on Sandy Bridge and 6% faster on Broadwell, so it looks like fixing the FIXME did have some effect.

Switch statements can be implemented in a [number of different ways](https://www.codeproject.com/articles/100473/something-you-may-not-know-about-the-switch-statem) including an if-else if-else chain.  It's not always easy to see if it's going to be faster, or if the compiler is making the best choice about which cases to try first.  It might be better to try reordering the existing code so that the most common types seen "in the wild" are tried first, although I expect the gains will be small.

I don't really like replacing the goto with a macro.  It adds more bloat to an already quite bloated function.  Macros also make debugging harder (try debugging problems in khash...)  Using goto is a perfectly good way of dealing with error conditions in C functions.  See http://koblents.com/Ches/Links/Month-Mar-2013/20-Using-Goto-in-Linux-Kernel-Code/ for pro-goto arguments.

Your editor it using tabs.  Please could you replace them with four spaces.  We avoid tabs in htslib as no-one could agree on how long they should be.

PR #706, opened 2018-05-30
Report the reason for failing to open a reference.

jkbonfield:
We previously printed "failed to populate ref" etc, but not why.
This would have been helpful in samtools/samtools#857 to diagnose the
cause of the problem.

### DIFF
--- a/cram/open_trace_file.c
+++ b/cram/open_trace_file.c
@@ -68,6 +68,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <string.h>
 #include <unistd.h>
 #include <limits.h>
+#include <errno.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include "cram/os.h"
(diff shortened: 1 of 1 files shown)


jmarshall:
Coolness. FWIW I rather agree with <https://utcc.utoronto.ca/~cks/space/blog/unix/StandardUnixErrorFormat>'s advocacy of quotes around filenames.

jkbonfield:
Agreed, although what a **can of worms**!  I think I just hunted to find the proper error reporting function and whichever I hit on first didn't have quotes so I copied that style.   Now I look more... what consistency of style? (As ever!)

Htslib appears (via a quick grep so likely erroneous) to have 38 unquoted, 7 with double quotes and 3 with single.  Samtools is 3/42/6 respectively, while bcftools is 212/32/0.  All over the shop, but quite a different "house style" between the library and the command line tools. :-)   I'll let whoever does the merge pick their flavour of the day.

Buck passed! ;-)

daviesrob:
I went with double quotes.

PR #709, opened 2018-06-04
64-bit reference positions - ABI/API placeholder

jkbonfield:
In the future we're going to have to tackle this properly, but this PR basically makes this possible without (I hope!) needing another round of ABI breakage.  To achieve this, it makes pos, mpos and isize all 64-bit in the external/public APIs.

SAM now works natively in 64-bit.  There is place-holder code in CRAM, mainly for testing purposes, taken from io_lib's CRAM4 branch with appropriate ifdefs to block it from normal compilation.  We have no roadmap for CRAM4 yet, but if/when it does appear those ifdefs will become version checks instead.

Caveats

- The SAM CP tag is a non-starter unless we add e.g. `l` BAM encoding for signed 64-bit integer, however I don't think I've ever seen it in the wild.
- We may want to duplicate some functions and have 64-bit named variants, to offer better migration.  I did this with hts_parse_reg as without it so much third-party code dies in horrible ways.  Arguably we could do the same with `hts_readrec_func` prototypes, but these are likely few and far between outside of htslib itself so I deemed it acceptable.
- Structure layout probably isn't efficient now as there may be padding introduced.  We can juggle that or just put in explicit 32-bit reserved ints ready for future expansion.  (We have some ideas on that anyway.)
- Hts_parse_reg64 needs merging with the hts_parse_reg2 from #708.
- Making hts_pair32_t a pair of 64-bit ints is appalling!  However it was vastly less API change this way!  We could bite the bullet and change more API, but note the internal fields are named differently between hts_pair32_t and hts_pair64_t (why?).
- Samtools sort and bedidx code has its own set of issues, as they shift 32 bit positions into a 64 bit int combined with other parameters.

Known bugs (will fix and edit this - just spotted it): multi iterator is still 32-bit.


jmarshall:
`int` is the totally generic integer type, while `int64_t` is rather less so.

Would it be worth introducing a typedef — `hts_pos_t` perhaps — and using it in the API function declarations etc? This would emphasise that these parameters are positions rather than emphasising an implementation detail.

jkbonfield:
I was thinking that while writing this infact, but sadly after I was half way through and lacked will power to go back and changei t all!  I do think it's probably the right thing to do though, with associated PRIhts_pos macro for printing.

daviesrob:
TODO:
- Ensure new header API can handle long references.
- Aux type `l` for int64_t values. (Don't bother with unsigned)

daviesrob:
Rebased with additions to make headers work.

Notes:
- Code using HTSlib will likely need minor changes, even if it only supports 32 bit positions, after this is merged:
  -  Anything that mixes signed and unsigned types like [this naughtiness](https://github.com/samtools/samtools/blob/488752a/bam_mate.c#L317) in Samtools will give the wrong answer due to missing sign extension.
  - printf formats will need to be changed where the item being printed is now 64 bits.
- As upstream code may try to reallocate the `target_len` array (for example [here](https://github.com/samtools/samtools/blob/488752aa75023b12ded1f9e6cab1c9045aa69be5/bam_sort.c#L415) in samtools), trying to change its size is likely to cause much breakage.  A work-around has been put in place where long refs are stored as `UINT_MAX` in `target_len` and the real length is stored elsewhere.  Code that wants to support long references will have to use `sam_hdr_tid2len()` to get the reference length.

We need to test against some htslib-using packages to check that they still work correctly when built against this branch.  I've tried samtools and bcftools, which work bar the items noted above.  I've also tried pysam, which manages fairly well (it builds OK, although a few tests fail due to unrelated issues with illegal headers in a couple of test files).

htslib/hts.h
```diff
@@ -570,10 +570,30 @@ When REST or NONE is used, idx is also ignored and may be NULL.
 #define HTS_FMT_TBI 2
 #define HTS_FMT_CRAI 3
 
+// Almost INT64_MAX, but when cast into a 32-bit int it's
+// also INT_MAX instead of -1.  This avoids bugs with old code
+// using the new hts_pos_t data type.
+#define HTS_POS_MAX ((((int64_t)INT_MAX)<<32)|INT_MAX)
+#define HTS_POS_MIN INT64_MIN
```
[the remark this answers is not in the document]
daviesrob: I think `HTS_POS_MIN` is there for completeness.  It isn't actually returned by any functions in HTSlib, so you would only use it in code upgraded to work with type `hts_pos_t`.  It's also only meaningful for fields like `isize` that could legitimately take a negative value.

header.c
```diff
@@ -831,7 +836,11 @@ int sam_hdr_update_target_arrays(sam_hdr_t *bh, const sam_hrecs_t *hrecs,
             if (!bh->target_name[i])
                 return -1;
         }
-        bh->target_len[i] = hrecs->ref[i].len;
+        if (hrecs->ref[i].len < UINT32_MAX) {
+            bh->target_len[i] = hrecs->ref[i].len;
+        } else {
+            bh->target_len[i] = UINT32_MAX;
```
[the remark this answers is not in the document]
daviesrob: It shouldn't need to because `sam_hdr_tid2len()` uses the `sam_hrecs_t::ref` array if available.  So if `target_len` is being updated from `sam_hrecs_t::ref`, `sdict` isn't going to be used any more and arguably it could actually be freed at this point.

htslib/hts.h
```diff
@@ -570,10 +570,30 @@ When REST or NONE is used, idx is also ignored and may be NULL.
...
+//
+// #define HTS_POS_MAX INT_MAX
+// #define HTS_POS_MIN INT_MIN
+// #define PRIhts_pos PRId32
+// typedef int32_t hts_pos_t;
+
 typedef struct {
-    uint32_t beg, end;
+    //uint32_t beg, end;
+   hts_pos_t beg, end; // sorry for the bad naming: FIXME!
```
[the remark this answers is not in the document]
daviesrob: Possibly yes, although it's exposed in `hts_reglist_t::intervals` and I see that samtools allocates one in [`bed_reglist()`](https://github.com/samtools/samtools/blob/76e7ed6/bedidx.c#L597).

Possibly we should make `hts_pair32_t` a typedef of `hts_pair_pos_t`.

htslib/sam.h
```diff
@@ -1290,6 +1306,8 @@ typedef struct __bam_mplp_t *bam_mplp_t;
     int bam_plp_push(bam_plp_t iter, const bam1_t *b);
     const bam_pileup1_t *bam_plp_next(bam_plp_t iter, int *_tid, int *_pos, int *_n_plp);
     const bam_pileup1_t *bam_plp_auto(bam_plp_t iter, int *_tid, int *_pos, int *_n_plp);
+    const bam_pileup1_t *bam_plp64_next(bam_plp_t iter, int *_tid, hts_pos_t *_pos, int *_n_plp);
```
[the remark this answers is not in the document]
daviesrob: `qpos` is on the read, not the reference.  We're not attempting to support reads longer than 2Gbases at the moment.

daviesrob:
Rebased with more additions:

- hts_reglist_t and associated funtions support 64 bit positions

- 64 bit position support in bgzf_idx_push()

- Read / write bcf1_t::pos as 64 bit for VCF.

- Support VCF headers with contig lengths > 2Gbp.

- Minimal support for 64 bit INFO tags in VCF files.  This is needed for 64 bit END tags in structural variants.  This adds BCF_BT_INT64 so we can store int64_t INFO tags, and BCF_HT_LONG so wa can pass int64_t arrays to / from bcf_get_info_values() and bcf_update_info().

- Various updates to make VCF indexes and tabix work on files with large positions.

- Updates to synced_bcf_reader for 64 bit positions.

- Updates to regidx for 64 bit positions.

- Round-trip and indexing tests for SAM and VCF files with large position values.

Now is the time to seriously kick the tyres on this one, especially the VCF updates.  It's looking ready to be merged, so if anyone has comments it would be best to add them sooner rather than later.

See also samtools/samtools#1107 and samtools/bcftools#1083 which need to be merged **before** this PR so that they continue to work properly.

sam.c
```diff
@@ -3960,7 +4018,7 @@ void bam_plp_destructor(bam_plp_t plp,
  *  Returns BAM_CMATCH, -1 when there is no more cigar to process or the requested position is not covered,
  *  or -2 on error.
  */
-static inline int cigar_iref2iseq_set(uint32_t **cigar, uint32_t *cigar_max, int *icig, int *iseq, int *iref)
+static inline int cigar_iref2iseq_set(uint32_t **cigar, uint32_t *cigar_max, int *icig, int *iseq, hts_pos_t *iref)
 {
     int pos = *iref;
```
whitwham: 64 bit to 32 bit conversion.  Is this right?

htslib/vcf.h
```diff
@@ -680,6 +683,7 @@ set to one of BCF_ERR* codes and must be checked before calling bcf_write().
      *  Returns 0 on success or negative value on error.
      */
     #define bcf_update_info_int32(hdr,line,key,values,n)   bcf_update_info((hdr),(line),(key),(values),(n),BCF_HT_INT)
+    #define bcf_update_info_int64(hdr,line,key,values,n)   bcf_update_info((hdr),(line),(key),(values),(n),BCF_HT_LONG)
```
valeriuo: There should probably be a corresponding **get** macro at line 777:

```c 
#define bcf_get_info_int64(hdr,line,tag,dst,ndst)  bcf_get_info_values(hdr,line,tag,(void**)(dst),ndst,BCF_HT_LONG)

File: sam.c (hunk 5 above)
Comment (daviesrob) [reply sam.c]: I think it gets away with it because `*iref` is set to an offset to the beginning of the read, so should fit within an `int` (barring issues with very big `BAM_CDEL` or `BAM_CREF_SKIP` CIGAR operations).  However it might be better to make `cigar_iref2iseq_set`, `cigar_iref2iseq_next` and `tweak_overlap_quality`, which calls them, use `hts_pos_t` all round as avoiding any possibility of overflow will make the code easier to understand.

File: htslib/vcf.h (hunk 6 above)
Comment (daviesrob) [reply htslib/vcf.h]: Added.  I also decided to make `bcf_update_info_int64()` and `bcf_get_info_int64()` static inline functions instead of macros, so we can ensure that they are used with `int64_t` arrays.

File: htslib/vcf.h [hunk 7]
```diff
@@ -669,26 +669,51 @@ set to one of BCF_ERR* codes and must be checked before calling bcf_write().
...
+    /// Set or update 64-bit integer INFO values
+    /**
+     *  @param hdr:       the BCF header
+     *  @param line:      VCF line to be edited
+     *  @param key:       the INFO tag to be updated
+     *  @param values:    pointer to the array of values. Pass NULL to remove the tag.
+     *  @param n:         number of values in the array. When set to 0, the INFO tag is removed
+     *  @return 0 on success or negative value on error.
+     *
+     *  The @p string in bcf_update_info_flag() is optional,
```
valeriuo: This paragraph can be removed, as this documentation text only refers to `bcf_update_info_int64`.

htslib/vcf.h
```diff
@@ -756,29 +781,56 @@ set to one of BCF_ERR* codes and must be checked before calling bcf_write().
...
-     *  @line:      BCF record
-     *  @tag:       INFO tag to retrieve
-     *  @dst:       *dst is pointer to a memory location, can point to NULL
-     *  @ndst:      pointer to the size of allocated memory
+     *  @param hdr:    BCF header
+     *  @param line:   BCF record
+     *  @param tag:    INFO tag to retrieve
+     *  @param dst:    *dst is pointer to a memory location, can point to NULL
+     *  @param ndst:   pointer to the size of allocated memory
+     *  @return  0 on success
```
valeriuo: There is discrepancy between the return code for success and the paragraph below, which states:
`Returns negative value on error or the number of written values (including missing values) on success.`

htslib/vcf.h
```diff
@@ -756,29 +781,56 @@ set to one of BCF_ERR* codes and must be checked before calling bcf_write().
...
+     *          -1 .. no such INFO tag defined in the header
+     *          -2 .. clash between types defined in the header and encountered in the VCF record
+     *          -3 .. tag is not present in the VCF record
+     *
+     *  Returns negative value on error or the number of written values
```
valeriuo: Same here.

test/sam.c
```diff
@@ -982,11 +984,14 @@ static void samrecord_layout(void)
 
     size_t bam1_t_size, bam1_t_size2;
 
-    bam1_t_size = 36 + sizeof(int) + 4 + sizeof (char *) + sizeof(uint64_t);
+    assert(sizeof(hts_pos_t) == 8 || sizeof(hts_pos_t) == 4);
+    int core_size = sizeof(hts_pos_t) == 8 ? 48 : 36;
+    bam1_t_size = core_size + sizeof (int) + 4 + sizeof (char *) + sizeof(uint64_t);
```
valeriuo: Although a test code, this looks a bit ugly. Maybe it can be changed to something like:
```c
bam1_t_size = core_size + sizeof (int) + sizeof (char *) + sizeof (uint64_t) + (sizeof (uint32_t) << 1);
```
This includes the newly added `mempolicy` member.

daviesrob:
Rebased with documentation fixes, NEWS update and README.large_positions.md file.

PR #710, opened 2018-06-05
Add kbs_resize() to facilitate reusing existing bitsets

jmarshall:
Add n_max field so bitsets can be made smaller (and later enlarged) without involving memory allocation.

Record the last slot's mask of used bits, to fix a bug whereby kbs_next() would return elements beyond ni when filled initialisation or kbs_insert_all() had been used. (`fill` was kind of a mistake :smile: but this would still be necessary to support `kbs_insert_all()` which is a reasonable function to provide…)

This code has been added to my klib bitset pull request attractivechaos/klib#59.

(Travis failure merely reflects the collision with the similarly-named function in _bcf_sr_sort.c_, which is being replaced.)


pd3:
I would like to see this merged, so that this bug fixed more than a year ago can be finally merged too https://github.com/samtools/bcftools/pull/627

daviesrob:
OK, so my plan is to rebase this along with the part of #555 that updates `bcf_sr_sort.c`, and then merge it along with some infrastructure to deal with the ABI break.  The result will look something like [this branch](https://github.com/daviesrob/htslib/tree/abi-break).

The Makefile changes for the ABI break do the following:

- Add a new variable `TWO_TO_THREE_TRANSITION_COUNT` which starts at one and should be incremented each time an ABI breaking change gets merged.
- Set SOVERSION (for the ELF shared library) to `2to3part$(TWO_TO_THREE_TRANSITION_COUNT)`
- Set the dylib compatibility version (for Mac) to `2.$(TWO_TO_THREE_TRANSITION_COUNT)`

Does this look like a good solution?

pd3:
It looks good to me, thank you. I would name the variable more generally, say `ABI_TRANSITION_COUNT`, and name the shared library more concisely, say `2x.$(ABI_TRANSITION_COUNT)`, but none of that matters and I am happy with your version if you like it better.

jmarshall:
Looks good, but I would squash Petr's _bcf_sr_sort.c_ update into my commit (or better to preserve the authorship: amend my commit by applying `s/kbs_resize/my_kbs_resize/` (3 changed lines) to _bcf_sr_sort.c_). At the moment my commit doesn't compile; applying either of these would help bisecting so you wouldn't need to `git bisect skip` if you landed on that commit.

Infrastructure looks good with variables as is (just by eye-balling, I haven't tried compiling on either platform). I assume all this is for temporarily tracking when each new breaker is merged and it'll all evaporate when the next htslib release occurs — and just turn back into one line `LIBHTS_SOVERSION = 3`. So the not very general variable name doesn't matter as it's all temporary.

daviesrob:
OK, I'll try to make a non-breaking version.  Would you mind if I force-pushed the results over this branch?

And yes, the ABI infrastructure will vanish when we've got all the breaking changes in place.

jmarshall:
No problem at all if you do some force-pushing.

jmarshall:
@daviesrob: If you haven't got back to this yet, you might find one or both commits at 
[my **embiggen-kbitset-compilefix** branch](https://github.com/jmarshall/htslib/commits/embiggen-kbitset-compilefix) useful.

daviesrob:
@jmarshall: Thanks, branch was very useful.

PR #711, opened 2018-06-06
Prevent make from working after configure errors

jmarshall:
In samtools/samtools#862 a fairly experienced user managed to not notice the voluminous error message at the end of the voluminous configure output, and continued by running `make` despite configure having ended in error.

With no _config.mk_ etc, `make` carried on and built a default configuration of HTSlib, which did not have desired functionality that would have been --enabled during configure.

This PR makes configuration errors write a _config.mk_ fragment that causes:

```
… … …
checking for curl_easy_pausebork in -lcurl... no
checking for curl_easy_init in -lcurl... yes
configure: error: libcurl library is too old (7.18+ required)

Support for HTTPS and other SSL-based URLs requires routines from the libcurl
library <http://curl.haxx.se/libcurl/>.  Building HTSlib with libcurl enabled
requires libcurl development files to be installed on the build machine; you
may need to ensure a package such as libcurl4-{gnutls,nss,openssl}-dev (on
Debian or Ubuntu Linux) or libcurl-devel (on RPM-based Linux distributions
or Cygwin) is installed.

Either configure with --disable-libcurl or resolve this error to build HTSlib.

$ make
config.mk:1: *** Resolve configure error first.  Stop.
```

You may wish to apply something similar to samtools's _configure.ac_.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -51,7 +51,11 @@ AH_TOP([/* If you use configure, this file provides @%:@defines reflecting your
 dnl Variant of AC_MSG_ERROR that ensures subsequent make(1) invocations fail
 dnl until the configuration error is resolved and configure is run again.
 AC_DEFUN([MSG_ERROR],
-  [echo '$(error Resolve configure error first)' >config.mk
+  [cat > config.mk <<'EOF'
+ifneq ($(MAKECMDGOALS),distclean)
+$(error Resolve configure error first)
+endif
+EOF
    AC_MSG_ERROR([$1], [$2])])
 
 AC_PROG_CC
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 # Configure script for htslib, a C library for high-throughput sequencing data.
 #
-#    Copyright (C) 2015-2017 Genome Research Ltd.
+#    Copyright (C) 2015-2018 Genome Research Ltd.
 #
 #    Author: John Marshall <jm18@sanger.ac.uk>
 #
@@ -32,7 +32,7 @@ AC_CONFIG_HEADERS(config.h)
 m4_include([m4/hts_prog_cc_warnings.m4])
 
 dnl Copyright notice to be copied into the generated configure script
-AC_COPYRIGHT([Portions copyright (C) 2016 Genome Research Ltd.
+AC_COPYRIGHT([Portions copyright (C) 2018 Genome Research Ltd.
 
 This configure script is free software: you are free to change and
 redistribute it.  There is NO WARRANTY, to the extent permitted by law.])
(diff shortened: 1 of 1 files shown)


daviesrob:
A neat solution, although it has the side-effect of breaking `make distclean` so you can't use that as a way of escaping the situation.  It's fairly easy to fix using [MAKECMDGOALS](https://www.gnu.org/software/make/manual/html_node/Goals.html) though, so `config.mk` becomes:

```
ifneq ($(MAKECMDGOALS),distclean)
$(error Resolve configure error first)
endif
```

And testing...

```
$ make
config.mk:2: *** Resolve configure error first.  Stop.
$ make clean
config.mk:2: *** Resolve configure error first.  Stop.
$ make distclean
rm -f test/*.tmp test/*.tmp.* test/tabix/*.tmp.* test/tabix/FAIL*
rm -f *.o *.pico cram/*.o cram/*.pico test/*.o test/*.dSYM version.h
rm -f libhts.so libhts.so.*
rm -f libhts.a bgzip htsfile tabix  test/hts_endian test/fieldarith test/hfile test/sam test/test_bgzf test/test_realn test/test-regidx test/test_view test/test-vcf-api test/test-vcf-sweep test/test-bcf-sr test/test-bcf-translate test/thrash_threads1 test/thrash_threads2 test/thrash_threads3 test/thrash_threads4 test/thrash_threads5 test/thrash_threads6
rm -f config.cache config.h config.log config.mk config.status
rm -f TAGS *.pc.tmp *-uninstalled.pc htslib_static.mk
rm -rf autom4te.cache
```

I can add an extra commit to do this when merging.

jmarshall:
Good point!

I looked at _\<prefix\>/share/autoconf/autoconf/general.m4_'s _AC_INIT_CONFIG_LOG to verify that a simple `echo >config.mk` would be kosher, and the same function suggests a `cat >config.mk <<'EOF'` here document will survive autoconf too.

daviesrob:
Yes, my conclusion was a here-document is the way to go.  I'll push up my change for you to inspect.

jmarshall:
LGTM. It's surprising how few **quoted** here-documents there are in existing autoconf fragments, but autoconf.info _Portable Shell_ doesn't warn against them.

daviesrob:
Quoted here-documents go back to the [original Bourne shell](https://www.in-ulm.de/~mascheck/bourne/v7/) so I think we're on safe ground there.  The `<<-` de-tab feature came a bit later and does get a mention in _Portable Shell_.

Find for the day: [The Traditional Bourne Shell Family](https://www.in-ulm.de/~mascheck/bourne/) listing many variants.

PR #712, opened 2018-06-07
[WIP] Additions to the CRAM header API

valeriuo:
1. Add support for `@HD` **GO** (group order) tag.
2. Preserve order of the lines as they appear in the original file.
3. Add the possibility of removing a key from an existing line.
4. Fix the display of the `@CO` lines in the header dump.
5. Add methods for removing a header line.
6. Add the ability to remove **SQ** and **RG** lines from the dedicated hash tables


jkbonfield:
I haven't gone through all the code yet, but how does the line_order code interact with adding new records?  For example typically we'll have a header file containing say 10 RG lines, 30 SQ lines, etc.  If we add a new RG entry then most likely people would prefer it to be between the RG and SQ lines, along with all the others.

This was one reason why the original code tracked the order of fundamental keys (are RGs before SQs or are SQs before RGs) and the order within each type so we know eg the 5th sequence is chr5, but it couldn't track if we alternated types or had CO comments interspersed at key points.  Fixing that may come at the expense of making other things very hard.  I'm unsure of what's desirable here (other than a basic level of functionality to start with is better than nothing).

Making a proper insertion function is doable, but expensive if it has to go through all records juggling their order.

Maybe the only nod to reordering here should be comments.  Ie each comment is tracked as which record it preceded (or at the end of header), so they don't move, but everything else is mutable and may be collated.  Tricky!

cram/sam_header.c
```diff
@@ -812,36 +1014,30 @@ int sam_hdr_rebuild(SAM_hdr *hdr) {
...
+        type = hdr->line_order[i].type_data;
+        if (EOF == kputc_('@', &ks))
+            return -1;
 
-        t1 = t2 = kh_val(hdr->h, k);
-        do {
-            SAM_hdr_tag *tag;
-            char c[2];
+        strncpy(c, hdr->line_order[i].type_name, 2);
+        if (EOF == kputsn_(c, 2, &ks))
```
jkbonfield: I think this could just be `kputsn_(hdr->line_order[i].type_name, 2, &ks)` instead of copying to `c[2]` first.

PR #713, opened 2018-06-12
Add kstring support to khash.

jkbonfield:
It can be a pain to need to create nul-terminated strings when hashing, if we're just dealing with substrings of known length in a longer string (eg keys in an `@SQ` field).

Instead of passing in a const char * pointer, we instead hash using a kstring_t struct (not pointer, as we wish to store the length in the hash keys array without using external memory).

Technically this wastes a bit of memory as kstring_t has m (malloced size) as well as l (length) and s (string), however kstring is ubiquitous in htslib so it's the logical data type to operate on.

Example usage:

```
#include <stdio.h>
#include <assert.h>
#include <htslib/khash.h>

KHASH_MAP_INIT_KSTR(ks, int)

    khash_t(ks) *h = kh_init(ks);
    int i, idx;

    for (i = idx = 1; i < argc; i++) {
	char *cp = argv[i], *np;
        do {
            np = strchr(cp, ' ');
            kstring_t ks = {np ? np - cp : strlen(cp), 0, cp};
            khiter_t k;
            int ret;

            k = kh_put(ks, h, ks, &ret);
            if (ret == 0) {
                printf("%d\told\t%.*s\t%d\n", ret, (int)ks.l, ks.s, kh_val(h, k));
            } else {
                kh_val(h,k) = idx++;
                printf("%d\t+\t%.*s\t%d\n", ret, (int)ks.l, ks.s, kh_val(h, k));
            }

            while (np && *np == ' ')
                np++;

            cp = np;
        } while (cp);
    }

    kh_destroy(ks, h);

    return 0;
}
```

### DIFF
--- a/htslib/khash.h
+++ b/htslib/khash.h
@@ -128,6 +128,7 @@ int main() {
 #include <stdlib.h>
 #include <string.h>
 #include <limits.h>
+#include <htslib/kstring.h>
 
 /* compiler specific configuration */
 
@@ -400,7 +401,7 @@ static kh_inline khint_t __ac_X31_hash_string(const char *s)
 }
 /*! @function
   @abstract     Another interface to const char* hash function
-  @param  key   Pointer to a null terminated string [const char*]
+  @param  key   Pointer to a nul terminated string [const char*]
   @return       The hash value [khint_t]
  */
 #define kh_str_hash_func(key) __ac_X31_hash_string(key)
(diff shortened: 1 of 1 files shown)


valeriuo:
Merged by https://github.com/samtools/htslib/commit/db40557681fa8bf56ed23a865f0d4ec7bb7edd25

PR #716, opened 2018-06-18
Fix the test run on Windows and AppVeyor.

valeriuo:
Opening a file in binary mode ('b'), prevents the OS from doing line ending translation.


daviesrob:
Looks like fai_build_core() didn't like blank lines with cr-lf endings.  I've put in a fix for that, and included some example data in `test/faidx.fa`.

It might be worth attaching a debugger on Windows to ensure `faidx1()` in `test/sam.c` isn't stripping out the cf-lf endings when it does the file copy.

PR #718, opened 2018-06-22
Auto indexing durng writing of BAM, SAM.gz, CRAM, BCF, VCF.gz

jkbonfield:
I don't have a samtools interface for this yet, but there are new test mechanisms in htslib for validating this.  In samtools I'm unsure how we'd specify it.  Off the top of my head I could think of:

- A new samtools view option to ask for index (plus what type, and size of min_shift).
- output-fmt-option keywords to request indexing, type, options, filename.
- Some automatic defaults? (Probably bad)
- Filename trickery: eg -o foo.bam/csi  (also bad!)

I also added the ability to use BAI and CSI indices for bgzipped SAM while I was at it.  (And indeed the ability to even write it natively.)

Note there are a few questions over the best API for SAM.gz.  Our generic `sam_read1` function takes fp and hdr.  The generic `hts_itr_next` function only passes on fp though (and bgzf fp rather than htsFile fp at that).  the multi-region iterator improves on the type of fp, but it still lacks the header.

We have two choices here.  We could use the data argument to pass over a new struct containing both htsFile and bam_hdr_t, but if we do this in hts layer then we're removing the ability for callers to use this for anything else.  That's perhaps not such a big issue.  Note to do this we'd need to store the header structure somewhere when we initialise the iterator, either in the fp or the itr structure itself both work (I'd favour the latter).

Alternatively we can create a new sam_itr_next2 function which has an additional argument passed in to it (hdr) and we get this to do the struct construction before running hts_itr_next, which doesn't alter the abilities of the underlying function.  However it does make an entirely new API function to call.  This is what I did, but I think perhaps the former may be preferable.

Finally test_view has been upgraded to read/write vcf/bcf.  It fails on vcf.gz reading, but that's due to it only being supported by synced reader.


jkbonfield:
Summary of the changes:

- bgzf_peek to permit SAM header parsing stopping *before* reading the next line.  This is helpful when indexing as otherwise the first sam_idx_init has to have a cached bgzf_tell value copied from the header parsing.  This may mean we can simplify other bits of code, eg the fp->line.l check in sam_read1.  (It's not harmful to keep it though.)

- Addition of idxfp to cram_fd structure, for purposes of writing the index on the fly.  This is because the old cram index functions wrote the index as it read the file, rather than building the index in memory and writing it out at the end.  We could change that, but this was the smaller tweak.

- Various new elements of hts_idx_t.  Specifically tbi_n and last_tbi_tid to track the number of references and the last reference number so we can do the mapping of CSI tids to TBI tids.  (CSI is all, TBI is all covered.)  I don't particularly like this, but couldn't think of anything better.

- New hts_idx_tbi_name function for the purposes of TBI index generation.  Again, we can't do this at the start from the header as we don't know what is covered.  We do it on the fly, storing it in the already existant idx->meta pointer.  We call hts_idx_tbi_name for each and every VCF record (if we're indexing), but it spots the same rname as last time quickly so hopefully this isn't a performance hit.  (I didn't measure it though.)

- Added an hts_idx_amend_last function to tweak the  cached idx->z.last_off value.  Normally we keep calling hts_idx_push repeatedly and it uses the last offset with the offset in this call to get the file extents covered by the record being pushed.  However when we start a new bgzf block, the cached last value is incorrect.  This provides a way to correct this whenever we start a new block.

- hts_idx_fmt function, to return idx->fmt.  This is used in sam_idx_save, but after various tweaks I think it's no longer necessary.  I'll cull it and maybe add a generic format to direct hts_idx_save to use the idx->fmt value.

- htsFile now has an idx pointer (and fnidx).  The fnidx is needed because we only open the file at the time of calling hts_idx_save.  The idx pointer is there so we can accumulate the index while writing each record.  Indeed the presence of it is the determining factor for auto-index or not.  However the fact this now exists may help unify BAM with CRAM and remove the need for the `hts_cram_idx_t *cidx` code.

- sam_itr_queryi2, sam_itr_querys2, sam_itr_next2 functions - see above.  It may be possible to avoid these and hack hts_itr_next instead.

- New sam_idx_init and sam_idx_save functions to go along side the existing bam ones.  These only support CSI and BAI.  Should I add TBI too?  Right now I'm using min_shift as the indicator between CSI/BAI, much as bcftools uses it for CSI/TBI.  If we want all 3 we need a better switch.

- Additional test harness function: test_compare (run command and compare two files).  I can't believe we had no way to compare two files (rather than stdout to file), but I really can't see it.  If this is a duplicate, please say.

- Opening a vcf file now sets fp->format.format, instead of leaving it as some generic text format.  This is needed for detecting the format while creating indices.

- Added some error checking to hts_idx_finish.


Things still to do:
- A few white space / trivial code tweaks.

- Look at why the samtools test harness now fails.

kyleabeauchamp:
FYI there's a related feature request in pysam: https://github.com/pysam-developers/pysam/issues/684

kyleabeauchamp:
Of the 4 options discussed, I too prefer to avoid the last 2.  

"Explicit is better than implicit.": https://www.python.org/dev/peps/pep-0020/

daviesrob:
Rebased with some additions:

- Extra error checking
- Bug-fixes (a couple of which were quite old bugs)
- Improved documentation
- More tests
- Multi-region iterator on SAM

A rather more major change is that the iterator API has reverted to the original one, so no `sam_itr_next2` etc.  This allows existing code to work on SAM with little or no modification, but means a pointer to the header has to be stored in the `htsFile` struct.  This currently only happens for SAM, although CRAM does it implicitly (and always has) by storing header-related data in the `cram_fd`.  BAM escapes for now because a few tools (notably samtools sort) expect to be able to discard the header on some files straight after reading and could use rather more memory if unable to.  The header struct gets a reference count so that it isn't freed prematurely.

NB: This needs a .so number bump on merging.

Issue #587, opened 2017-09-04
Prefix exported symbols


kloetzl:
Hi,

In Debian we recently had the problem, that a tool had a [function called `json` clashing with a value from an enum defined in `hts.h`](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=866646). Already, most functions are prefixed with `hts_`, please consider adding that to other symbols, too. Especially, since `json`, `sam`, etc. are very generic names which may well be used in other contexts. 

Best,
Fabian

daviesrob:
`json` is on the way out, but not immediately as it will change HTSlib's API to do so.  We'll get rid of it next time we have a more major API/ABI breaking release.

kloetzl:
C++ has `enum class` to avoid these kinds of problems. But I am looking forward to the prefixed symbols in the next version.

jmarshall:
HTSlib is a C library so it doesn't really matter what C++ has :smile:

For some reason I thought enumeration constants were in a separate name space (so did not need much prefixing — unlike HTSlib's external functions) and this was not picked up during review when format detection was added in 2014. Bugger. Most of `htsExactFormat` and `htsCompression` are less than ideal…

jmarshall:
I am updating PR #721 at present and am planning to add new entries as `bzip2_compression`, `fastq_format`, etc rather than plain `bzip2` and `fastq`.

Does this seem like a sufficient and good convention?
Will there ever be a good time to change the existing constants?

kloetzl:
Adding more characters definitely helps to prevent a collision.

> Will there ever be a good time to change the existing constants?

I fear the answer to that will always be *no*.

daviesrob:
Would it be better to use an `hts` prefix, resulting in `hts_fmt_sam`, `hts_compress_gzip` etc.?

> Will there ever be a good time to change the existing constants?

While it doesn't solve the immediate problem, we could add new symbols and deprecate the old ones (as was done for `json`).   We could also possibly remove `json` as it's been deprecated for a while now...

jmarshall:
> Would it be better to use an hts prefix, resulting in hts_fmt_sam, hts_compress_gzip etc.?

As usual, it's a tradeoff. It depends how people feel about the hideous ugliness, etc…

jkbonfield:
Release 1.10 is a major step forward in terms of symbol tables. All internal functions are no longer exported from the library.

[...]

Issue #261, opened 2015-08-26
detecting a sam/bam/cram file that is of 0 size.


atks:
It's possible to open a bam/cram file that is of size 0 and then reading it without encountering any errors.

hts_open will still return a file handle
sam_hdr_read will return a struct
the format detected is SAM.

Is there an existing mechanism that I can check to throw an error in such cases?

SamStudio8:
The new quickcheck command may catch this?

atks:
where is the function? I can't find it in the latest pull of htslib.

SamStudio8:
Ah sorry! I was talking about a new samtools subcommand (
https://github.com/samtools/samtools/pull/406) not something in the API...

atks:
thanks for that! it's actually useful, I can go there and check how they do things.  So samtools quickcheck will also be silent about an empty sam/bam/cram file.

jkbonfield:
On Wed, Aug 26, 2015 at 07:58:00AM -0700, Adrian Tan wrote:
> hts_open will still return a file handle
> sam_hdr_read will return a struct
> the format detected is SAM.
> 
> Is there an existing mechanism that I can check to throw an error in such cases?

This is correct though.  A zero length SAM file conforms to the SAM
specification! Pointless though that is...

James

-- 
James Bonfield (jkb@sanger.ac.uk) | Hora aderat briligi. Nunc et Slythia Tova
                                  | Plurima gyrabant gymbolitare vabo;
  A Staden Package developer:     | Et Borogovorum mimzebant undique formae,
https://sf.net/projects/staden/   | Momiferique omnes exgrabure Rathi. 


-- 
 The Wellcome Trust Sanger Institute is operated by Genome Research 
 Limited, a charity registered in England with number 1021457 and a 
 company registered in England with number 2742969, whose registered 
 office is 215 Euston Road, London, NW1 2BE.

SamStudio8:
I believe quickcheck will error on 0-length files as it checks for the
presence of at least one sequence.

jmarshall:
There are improvements (related to #200) to `hts_detect_format()` approaching that could have it return `unknown` or a new `empty` code for 0-length files.

What would you prefer `hts_open()` et al to do for such files?

An empty file is technically a valid SAM file, so the existing behaviour can be construed as correct.

[...]

PR #721, opened 2018-06-26
Text file type detection updates

### LINKED ISSUE #200: htsfile predicts "SAM version 1" for all text files

jmarshall:
Dusting off some old branches, notably to fix #200, making identification of SAM files more accurate. At present, any text file is recognised as SAM, which is less than ideal.

This adds four new entries to `htsExactFormat`, so it may be a good time to review #587 with a view to adding these entries with a suitable prefix from the start.

Unfortunately some code, notably bcftools, has taken advantage of the bug that any text file is recognised as SAM to open arbitrary text files with `hts_open()` in contravention of its documentation:

```
/*! @abstract       Open a SAM/BAM/CRAM/VCF/BCF/etc file
…
*/
htsFile *hts_open(const char *fn, const char *mode);
```

Ideally such code would be rewritten to use `bgzf_open`/`bgzf_getline`/etc instead of misusing `hts_open`/`hts_getline`/etc.

Alternatively the use of `hts_open` to read text files could be blessed. The patch contains workarounds to enable such use (as currently _tabix.c_ is guilty of the same thing), but bcftools may need similar alterations to that in _tabix.c_ to work with this fixed file type detection PR.

### DIFF
--- a/test/sam.c
+++ b/test/sam.c
@@ -1047,6 +1047,7 @@ static void test_empty_sam_file(const char *filename)
 {
     samFile *in = sam_open(filename, "r");
     if (in) {
+        enum htsExactFormat format = hts_get_format(in)->format;
         bam1_t *aln = bam_init1();
 
         sam_hdr_t *header = sam_hdr_read(in);
@@ -1054,6 +1055,9 @@ static void test_empty_sam_file(const char *filename)
             fail("sam_hdr_read() from %s should fail\n", filename);
         if (sam_read1(in, header, aln) >= 0)
             fail("sam_read1() from %s should fail\n", filename);
+
+        if (format != empty_format)
+            fail("detected %s as %d (expected empty_format)", filename, format);
 
         bam_destroy1(aln);
         sam_hdr_destroy(header);
--- a/test/sam.c
+++ b/test/sam.c
@@ -1043,6 +1043,41 @@ static void faidx1(const char *filename)
     fai_destroy(fai);
 }
 
+static void test_empty_sam_file(const char *filename)
+{
+    samFile *in = sam_open(filename, "r");
+    if (in) {
+        bam1_t *aln = bam_init1();
+
+        sam_hdr_t *header = sam_hdr_read(in);
+        if (header)
+            fail("sam_hdr_read() from %s should fail\n", filename);
+        if (sam_read1(in, header, aln) >= 0)
+            fail("sam_read1() from %s should fail\n", filename);
+
+        bam_destroy1(aln);
+        sam_hdr_destroy(header);
+        sam_close(in);
+    }
+    else fail("can't open %s to read as SAM", filename);
+}
+
+static void test_empty_text_file(const char *filename)
+{
+    htsFile *in = hts_open(filename, "r");
+    if (in) {
+        kstring_t str = KS_INITIALIZE;
+        int ret, n = 0;
+        while ((ret = hts_getline(in, '\n', &str)) >= 0) n++;
+        if (ret != -1) fail("hts_getline got an error from %s\n", filename);
+        if (n != 0) fail("hts_getline read %d lines from %s\n", n, filename);
+
+        hts_close(in);
+        free(str.s);
+    }
+    else fail("can't open %s to read as text", filename);
+}
+
 static void check_enum1(void)
 {
     // bgzf_compression() returns int, but enjoys this correspondence
@@ -1078,6 +1113,8 @@ int main(int argc, char **argv)
     use_header_api();
     test_header_pg_lines();
     test_header_updates();
+    test_empty_sam_file("test/emptyfile");
+    test_empty_text_file("test/emptyfile");
     check_enum1();
     check_cigar_tab();
     for (i = 1; i < argc; i++) faidx1(argv[i]);
--- a/test/sam.c
+++ b/test/sam.c
@@ -1062,7 +1062,7 @@ static void test_empty_sam_file(const char *filename)
     else fail("can't open %s to read as SAM", filename);
 }
 
-static void test_empty_text_file(const char *filename)
+static void test_text_file(const char *filename, int nexp)
 {
     htsFile *in = hts_open(filename, "r");
     if (in) {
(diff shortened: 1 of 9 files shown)


daviesrob:
Thanks.  It does cause issues for bcftools, so we'll have to sort those out before merging this.

jmarshall:
The bcftools problem is that it has file arguments that are now recognised as `bed` or `fasta`, that it tries to open using `hts_open()` — which fails because those formats are not in the `case` statements in `hts_open()` and `hts_close()`.

So this is a policy question for the maintainers: should `htsFile` continue to be for SAM/BAM/CRAM and VCF/BCF files exclusively?

If so, the fix for bcftools is to do something different for `bed` files in _synced_bcf_reader.c_'s `bcf_sr_regions_init()` (and maybe a couple of places in bcftools), and to apply a fix like samtools/bcftools#821.

If not, the fix for bcftools is to add `fasta`and `bed` and other text formats to those two `case` statements, and bless the use of `htsFile` for use with `hts_getline` and text input files. (And maybe apply that bcftools PR anyway.)

BTW I have a couple more commits to add to this branch while you folks contemplate that policy question…

jkbonfield:
What's the proposed alternative to `hts_open`?  I guess `bgzf_open` is the closest thing in functionality.  (Indeed I just realised your other PR uses that.)

I'm not sure what the original intention was for hts_open with regards to policy.  It does claim to be "format-neutral I/O", but doesn't really clarify beyond that.

I've always found it a bit confusing that the htsExactFormat enum has both specific file types (bam, sam, bcf, bed, etc) as well as generic ones (binary, text).  The category also has unknown in there, rather than it being an error.  Both of those do rather imply it is intended for generic data, but it's a bit of guesswork.    If it wasn't intended that way, then I'm not convinced the generic ones should exist at all in something claiming to be "exact" format,  although internally they do sometimes get used.  (Eg VCF is sometimes labelled as text and not vcf; I fixed this in the indexing PR.)

So my gut instinct is it should probably work.  Breaking other code though implies this is perhaps best delayed until the ABI breaking stuff lands (even though this is only API breaking, it's easy to lump them together and the timing isn't great for a late change like this).

jmarshall:
The proposed alternative is

> Ideally such code would be rewritten to use `bgzf_open`/`bgzf_getline`/etc instead

…or plain `hopen`/`hgetln`/etc if they don't need to take advantage of gzip/BGZF decompression being done for them.

`text_format`/`binary_format` exist because after `hts_open("wb")` it's impossible to know whether the exact format is going to be `bam` or `bcf` — see 529ca88071e775a488e4108945d5bd863a4190f0. This is called compromising with reality :smile: But in general `hts_detect_format()` is intended to work on any file, which is not necessarily the case for `hts_open()`. The two are separate.

It turns out that that “_hts_open: Open a SAM/BAM/CRAM/VCF/BCF/etc file_” documentation was written by me in 2013 in d038d8ed33a850b1ba2e634d43cd61fb135a1800. The `hts_getline` function was already present at the time (though it was introduced for reading SAM files). So one point of view is that the 2013 documentation over-constrained the function and it would be in keeping with the function as it existed prior to 2013 to relax that to “_Open a sequence data (SAM/BAM/CRAM) or variant data (VCF/BCF) or textual line-orientated file_” (or something shorter!) and thus bless reading line-by-line with `hts_open()`.

If people agree with this solution, I can rework this PR to avoid any breakage, in bcftools or otherwise.

> … until the ABI breaking stuff lands … timing … a late change …

Would you like to give us a hint when ABI breakage is planned to land, and what timing “late” might relate to?

jkbonfield:
I see, I hadn't realised the distinction between what hts_detect_format is meant to work on vs hts_open.  None of this is particularly obvious!

As far as timing, the release candidate is meant to be today, so this is what I mean by late. :-)
The ABI breaking stuff got punted again sadly, but the plan is it'll arrive asap after the release goes out, so develop has the new stuff in for the longest time it can, to give us lots of time to spot potential gotchas.

jmarshall:
I'm planning to rework this as described in https://github.com/samtools/htslib/pull/721#issuecomment-400666150 so as to minimise other changes needed.

jmarshall:
This branch is now finally updated to a point where it is ready for merging. Rebased and conflicts resolved, and two significant changes from previously which mean it is now ready:

* Disallowing empty SAM files (#261) is now implemented: empty files can be opened with `hts_open()` and read with `hts_getline`, but `sam_hdr_read` now fails. Why it was necessary to implement this is now lost in the mists of time, but it has to do with the _xx#blank.sam_ test cases involving empty files — without work one way or the other in this area, the test suite was failing with this PR.

* Reworked to bless using `hts_open` to read arbitrary text files line-by-line, and tests added to ensure that this works. With this done, there are no issues caused for bcftools.

All of the htslib, samtools, and bcftools test suites pass when run against this branch.

jkbonfield:
The xx#blank.sam test was one of mine, lifted over from io_lib.  It was deliberately empty in order to test that case, as it's a valid file, but I'm completely in agreement with taking a stance that valid-or-not, we shouldn't accept it.  The fudge to put a comment in it is good as it still means it tests functionality in e.g. iterators.

Thanks.

test/sam.c
```diff
@@ -1043,6 +1043,41 @@ static void faidx1(const char *filename)
     fai_destroy(fai);
 }
 
+static void test_empty_sam_file(const char *filename)
+{
+    samFile *in = sam_open(filename, "r");
+    if (in) {
+        bam1_t *aln = bam_init1();
```
valeriuo: A check for the _empty_format_ would be a nice addition.
Something like:
```c
if (in->format.format != empty_format)
    fail("file %s is not marked as empty\n", filename);
```

jmarshall: Good idea, thanks.

valeriuo:
Merged into develop

PR #726, opened 2018-06-30
Add new variant type VCF_OVERLAP

pd3:
See also
    https://github.com/samtools/bcftools/issues/736
    https://github.com/samtools/bcftools/issues/817
    https://github.com/samtools/bcftools/commit/db2e2b66f4164e650d627297375c60dfdcca7ad6

### DIFF
--- a/htslib/vcf.h
+++ b/htslib/vcf.h
@@ -132,12 +132,13 @@ extern uint8_t bcf_type_shift[];
 #define BCF_BT_FLOAT    5
 #define BCF_BT_CHAR     7
 
-#define VCF_REF   0
-#define VCF_SNP   1
-#define VCF_MNP   2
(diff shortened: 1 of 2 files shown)


daviesrob:
This needs commit samtools/bcftools@9208dda to be applied to bcftools to make the example in samtools/bcftools#817 work properly (without it you get nothing in the TYPE column for the spanning deletions).  Would you like me to apply a rebased version of it to bcftools at the same time as merging this?

pd3:
That would be great, thank you

daviesrob:
Merged and rebased bcftools update pushed.  You should be able to get rid of the bcftools vcf_update branch now.

pd3:
Thanks!

PR #727, opened 2018-07-04
Update dependencies and htslib*.mk variables

jmarshall:
Update Makefile dependencies to reflect recent #include changes.
Update htslib*.mk to reflect recently-added source files.

Add #include \<config.h\> to three source files where it was missing. (All source files need to include this as a matter of course.)

Since 5c155810344f86948980c2109241410772ceb83e (use hts_log_<level>()), cram/mFILE.c and cram/sam_header.c have included hts_internal.h, but they only need htslib/hts_log.h for this. As this dependency was missed from the Makefile at the time, we might as well correct them to use the more specific header at the same time as fixing the dependencies.


daviesrob:
I thought I'd done this, turns out it was in a branch that hasn't turned into a pull request yet...

I've taken the liberty of adding the copyright update here too.  Thanks for the pointer.

PR #733, opened 2018-07-13
Add aclocal.m4 to .gitignore

ThomasHickman:
This is a generated file from automake, so should be included in the `.gitignore` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /version.h
 
 autom4te.cache
+aclocal.m4
 config.cache
 config.h
 config.h.in


ThomasHickman:
Hmm ok, I was just wondering why I had a `aclocal.m4` in my htslib folder which wasn't ignored by git. I'll close the PR.

jkbonfield:
It's there because `autoreconf` creates it.  This is standard behaviour when checking out from git to get your configure script, so I think the PR is valid.

ThomasHickman:
Just had a look at the README:
> autoconf       # ...and configure script (or use autoreconf to do both)

Surely we should ignore files that are generated by suggested methods?

jkbonfield:
The INSTALL notes actually state to manually do `autoheader` followed by `autoconf`, but `autoreconf` is so ubiquitous as a synonym for running all the autotools programs that I still think it's valid to ignore this file.

Arguably we should be using aclocal (the extra bit that autoconf runs) anyway as this adds PKG_CONFIG path support to the configure script, which we do use in our configure script already. 

I don't understand the insistence on promoting the harder route either.

jmarshall:
“Insistence”? “harder”? :smile:

I imagine _README.md_ was originally written that way so as to act as documentation of which parts of the autotools the project is using. And _aclocal.m4_ would not have been originally listed as ignored because if it did appear (in a non-automake setup like htslib is) it should be consciously either `git add`ed or nuked, as @daviesrob commented.

I'm not really sure why `autoreconf` is insisting on running `aclocal` and creating this.

jkbonfield:
Aclocal does nothing if it finds nothing that needs adding.

Here it is finding our use of PKG_CONFIG and so adding parameters to permit the user to adjust the bin and lib search paths, which isn't a bad thing.  I'm not sure it actually works though as I've given it nothing more than a cursory glance.

Thus the configure script we kept differs if we use the prescribed route vs autoreconf route.

jkbonfield:
Crossed-lines there.   No, we shouldn't add auto-generated files.  We don't add configure, and this falls into the same case.  We do state it is something to ignore though.

jmarshall:
We're agreeing that there is a bug here; but _aclocal.m4_ is in a different category from _configure_.

_configure_ is generated from other files within the HTSlib distribution. Indeed it should not be checked in (in 2018; in the old days there was an argument for checking it in to control for different versions of autoconf).

_aclocal.m4_ OTOH is a cache of m4 functions from outwith HTSlib, which at least in theory people might not have or might have borken versions of. (Compare samtools/samtools@30bc5e2dbcbc13ca772a607f5a2fe1cfe12be709's commit message.) So there are reasons to consider checking it in.

sb10:
Having aclocal.m4 committed means that you end up with dirty builds. I sort of followed the above discussion, but didn't understand the conclusion. What benefit does it give anyone to provide a file that they will generate for themselves?

PR #741, opened 2018-07-23
Repeated calling of bcf_empty() must not segfault

pd3:
Set `free()`-ed pointers to NULL, calling `bcf_empty()` repeatedly should not result in segfault.


pd3:
Forgot to mention that this bcftools commit https://github.com/samtools/bcftools/commit/8c1f50fa330915fd4d59ac1a61f52b75ae4febfe depends on this pull request.

PR #744, opened 2018-07-24
Revise bgzf library to support multi-member gzip files

### LINKED ISSUE #742: bgzf gzip-read feature does not support multi-member (concatenated) gzip files
The bgzf library that is part of htslib is supposed to support reading blocked-gzipped, gzipped, and uncompressed data, with the format of the input data being detected at runtime.

This mostly works. However, the decompression of gzip-format files fails when those files have multiple "members", each with its own zlib header and constituting its own zlib stream.

The [gzip RFC](https://tools.ietf.org/html/rfc1952) specifies that gzip files may have multiple members:

```
2.2. File format

      A gzip file consists of a series of "members" (compressed data
      sets).  The format of each member is specified in the following
      section.  The members simply appear one after another in the file,
      with no additional information before, between, or after them.
```

A side effect of this is that concatenating two gzip files results in a valid gzip file (and is probably the easiest way to get a multi-member file).

When the bgzf gzip decompressor reaches the end of a gzip member, it uses this code:
https://github.com/samtools/htslib/blob/a4910bf6e704fc64ad8787499f72ac69c9699470/bgzf.c#L614-L619

The `inflate()` call is going to return `Z_STREAM_END` (which is >0, and which is not checked for), and will leave the start of the next stream "available" in the input buffer, if present, while not necessarily filling the output buffer. The current code takes the ending data, but ignores the return code.

The next time it tries to read, it discards the unused data in the input buffer, throws a new input buffer at `inflate()`, notices that this time it got no more output data, and signals an EOF. So the next gzip member isn't decompressed, and in fact some of its data is discarded.

It looks like, borrowing from https://stackoverflow.com/a/17822217, the right thing to do here is to call `inflateReset()` whenever a zlib stream ends, to clear the `Z_STREAM_END` condition, so we can decompress the remainder of the input buffer as a new stream if possible.

We might also want to change the input code, which assumes that if there is any room left in the output buffer, it means the input buffer was totally consumed and needs to be refilled. That's not true when `Z_STREAM_END` happens. There's also another positive error code, `Z_NEED_DICT`, that zlib can emit and that isn't checked for.

I'm going to see if I can put together a PR to resolve this.

adamnovak:
This fixes #742 so that the gzip reader in the bgzf library supports files with multiple members, as allowed by the gzip spec.

I eliminated the manual gzip header parsing code in favor of letting zlib parse the header itself. I also eliminated the `cached` flag to `inflate_gzip_block`, and changed the semantics of the function so that it will always use up any existing input data before going to fetch more.

We can also now detect and warn about truncated gzip files, if the truncation happens in the middle of a member.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -201,6 +201,8 @@ static const char *bgzf_zerr(int errnum, z_stream *zs)
         return "progress temporarily not possible, or in() / out() returned an error";
     case Z_VERSION_ERROR:
         return "zlib version mismatch";
+    case Z_NEED_DICT:
+        return "data was compressed using a dictionary";
     case Z_OK: // 0: maybe gzgets error Z_NULL
     default:
         snprintf(buffer, sizeof(buffer), "[%d] unknown", errnum);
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
Will these changes work with libdeflate, which is one of the support ZLIB implementations?

adamnovak:
You mean the libdeflate at https://github.com/ebiggers/libdeflate?

That project doesn't implement the `inflate()`/`deflate()` API that's used throughout the BGZF code. It also looks like the htslib makefile is requiring actual `zlib`, or at least something that answers to its API:

https://github.com/samtools/htslib/blob/a285d851a47d6d6add989d07aa5b344481c47294/configure.ac#L203-L216

I haven't rewritten the BGZF code on top of the libdeflate API, or put in a compatibility layer to make that easier. But since libdeflate advertises support for GZIP, if and when the BGZF code is moved over, it should work fine.

kyleabeauchamp:
Thanks, this clarifies my understanding: HTSLib requires zlib and may ALSO link to libdeflate---but still requires zlib for BGZF and other portions.

adamnovak:
More specifically, it looks like the block-based compression and decompression of actual blocked-gzipped data can use libdeflate, but the non-blocked gzip IO (which I am touching) still uses zlib exclusively.

libdeflate is apparently designed for block operations rather than streaming operations, so it may not be desirable to convert the streaming gzip IO over to it.

kyleabeauchamp:
Yep, that is my understanding as well---I just didn't realize HTSLib was doing both types of operations.  Regardless, I'll update the conda recipe for HTSLib to make this clear.

jkbonfield:
Correct - libdeflate is non-streaming, so we only use it in portions.  Fortunately these are the most widely used portions and the speed benefit is simply huge.

bgzf.c
```diff
@@ -593,36 +595,67 @@ static int inflate_block(BGZF* fp, int block_length)
...
+            fp->errcode |= BGZF_ERR_ZLIB;
+            return -1;
+        } else if ( ret == Z_STREAM_END ) {
+            // we finished a GZIP member
+            if ( input_eof ) {
+                // we are done with the last stream in the file
+                break;
+            } else if (fp->gz_stream->avail_in > 0) {
+                // try and read another GZIP member in the remaining data
+                int reset_ret = inflateReset(fp->gz_stream);
```
jkbonfield: I'm a little concerned about the case where the boundary of two gzip streams occurs precisely at the point where avail_in becomes zero.

However after some binary searching I have a test file that has this case and it works, despite not calling inflateReset.  The reset is necessary in other cases though, so this has my scratching my head.

Did you think about this case?

jkbonfield: Specfically `if (hpeek(fp->fp, &c, 1) == 1) {` instead of `avail_in > 0` will cause the reset to apply in this corner case too.

However I still don't quite understand why it's not needed.

adamnovak: I did consider that case; the reason it works as is is that once `inflate` starts returning `Z_STREAM_END`, it will continue to return `Z_STREAM_END` when called again, until you `inflateReset` it.

So when the gzip stream ends, we call `inflateReset` if we have more data available right now, and skip it if we don't. Then the next time we go through, if there's more data, we'll put it in the input buffer, and when `inflate` returns `Z_STREAM_END` we'll see that we have more data available and reset it.

I don't think `hpeek(fp->fp, &c, 1) == 1` will work as a condition, because that doesn't catch the case when there's data in the input buffer but nothing more in the file. But `hpeek(fp->fp, &c, 1) == 1 || avail_in > 0` would probably work, and save a repeated `inflate` call.

adamnovak:
I've modified the logic controlling the decompressor reset to incorporate @jkbonfield's suggestion.

bgzf.c
jkbonfield: Aye realised I meant "> 0 || peek" shortly after posting, but then was in a meeting.   Thanks for the explanation as to how come it still worked.

jkbonfield:
Many thanks @adamnovak for the bug report and fix.

luyang93:
I try to reproduce the problem.
I gzip two fastq into two seperate gz file, the cat them to one gz file.
Then, I use pysam to read the merged file and count read_num.the number is less than that I use to genearate gz file.
but bgzip is equal

So,  I want to ask, the fix hasn't been merged to master branch?

jmarshall:
The HTSlib fix (this PR, 744) was merged to HTSlib **master** and is present in releases starting from 1.10.

Pysam releases and builds are still using htslib 1.9 so do not contain this fix. pysam-developers/pysam#905 was recently merged, so [pysam](https://github.com/pysam-developers/pysam)'s **master** branch now uses htslib 1.10.2 and a new pysam release will be made fairly soon. In the meantime, you can build pysam **master** yourself if you wish.

PR #745, opened 2018-07-27
BGZF CRC checking on decode now also applies when threaded.

jkbonfield:
I accidentally added the CRC32 check to the wrong function (inflate_block rather than bgzf_uncompress) back in 44c6a36f2894b45ad6cb491833d27c6a2106410f, meaning it only applied on non-threaded decodes.  This has now been moved. CRC checks are done inside the child threads for extra speed.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -516,7 +516,9 @@ static int deflate_block(BGZF *fp, int block_length)
 
 #ifdef HAVE_LIBDEFLATE
 
-static int bgzf_uncompress(uint8_t *dst, size_t *dlen, const uint8_t *src, size_t slen) {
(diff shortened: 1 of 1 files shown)


bgzf.c
```diff
@@ -531,12 +531,18 @@ static int bgzf_uncompress(uint8_t *dst, size_t *dlen, const uint8_t *src, size_
         return -1;
     }
 
+    uint32_t c1 = libdeflate_crc32(0L, (unsigned char *)dst, *dlen);
```
valeriuo: A minor note: `libdeflate_crc32 ` first argument is a 32 bit `int`, so a simple `0` would do.

PR #746, opened 2018-07-27
Fix possible thread pool lockup, and tidy up mutex locking

daviesrob:
Fix thread pool lock-up that caused 1 in 20 `bambi i2b` processes to get stuck.

Also tidy up some mutex locking to remove double locks and simplify `tpool_worker()`.


jkbonfield:
Fixed spacing issues and merged as 33a275b

PR #748, opened 2018-07-30
do not override build variables

wookietreiber:
When you set CC, CPPFLAGS and LDFLAGS with `=` the users choices will be overridden. The user can then only override this with `make -e` or hack the Makefile. CPPFLAGS and LDFLAGS are empty, so there is no need at all to set these. If CC is set by the user, e.g. to another compiler, `?=` should be used instead of `=` to provide the default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,20 +22,18 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
-CC     = gcc
+CC     ?= gcc
 AR     = ar
 RANLIB = ranlib
 
 # Default libraries to link if configure is not used
 htslib_default_libs = -lz -lm -lbz2 -llzma
 
-CPPFLAGS =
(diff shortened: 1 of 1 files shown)


daviesrob:
If you want to set `CC`, `CFLAGS`, `CPPFLAGS` or `LDFLAGS` you should do so when running `configure`.  The settings get propagated through to the Makefile.

wookietreiber:
@daviesrob This is exactly **what I do**. There are, however, a lot of tools that include htslib and simply use the Makefile. I know that these tools are not directly the concern of htslib, but the changes I proposed do not hurt either.

jmarshall:
If the changes proposed did not hurt, they would probably have been accepted on any of the last dozen occasions you or your associates proposed them.

Have you considered instead evangelising how to use `make` properly to the maintainers of these other tools?

jkbonfield:
Closing as this is an unsupported, and unexpected, way of controlling make variables.  Eg try automake/autoconf standard tools and you'll find "make CC=foo" changes CC but "CC=foo make" does not.  We're following best practices here.

PR #754, opened 2018-08-07
Make bgzip --help and tabix --help write to standard output

jmarshall:
Minor change to make these utilities' `--help` options follow the GNU conventions. And so be conveniently greppable.


daviesrob:
Thanks for the patch, now merged.

PR #757, opened 2018-08-10
Escape reserved and unsafe characters in bucket.

whitwham:
Percent encode s3 bucket paths (if not already percent encoded).

Fixes samtools/samtools#775.


whitwham:
PR #839 superseded this pull request.

PR #758, opened 2018-08-10
tabix coordinates fix

### LINKED ISSUE #753: Inconsistent tabix results between index-jumping and streaming
Querying a tabix-indexed file yields different results, depending on whether the `-R` or the `-T` options is used to supply the query. Judging from the documentation, I would expect to get the same results in both cases. It seems that streaming (`-T`) through the file incorrectly outputs adjacent regions as well.

A small example:

`test.txt.gz`:
```
chr1	1	Site1
chr1	3	Site2
chr1	4	Site3
chr1	6	Site4
chr1	7	Site5
chr1	7	Site6
chr1	12	Site7
chr1	18	Site8
```

`query.txt`:
```
chr1	7	7
```

Query with index-jumping:

```sh
tabix -R query.txt test.txt.gz
```

```
chr1	7	Site5
chr1	7	Site6
```

Query with streaming:
`chr1:6` gets returned as well, this looks wrong.

```sh
tabix -T query.txt test.txt.gz
```

```sh
chr1	6	Site4
chr1	7	Site5
chr1	7	Site6
```

Generation of the tabix file used above:
```sh
bgzip -c test.txt > test.txt.gz
tabix -b 2 test.txt.gz
```

Tested with: htslib v1.8 and v1.9

valeriuo:
Use 0-based exclusive representation of coordinates from regions and targets files.
Fixes #753

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,12 @@
 Noteworthy changes in release a.b
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* tabix uses an internal 0-based exclusive representation of the regions (-R)
+  and targets (-T) files and consequently enforces stricter checks at both ends
+  when detecting overlaps. Thus, some of the resulted lines obtained by using
+  a targets file may be removed now, especially the ones overlapping at the
+  boundary of a region.
+
 Noteworthy changes in release 1.9 (18th July 2018)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/regidx.c
+++ b/regidx.c
@@ -266,7 +266,7 @@ int regidx_overlap(regidx_t *idx, const char *chr, uint32_t from, uint32_t to, r
     for (i=ireg; i<list->nregs; i++)
     {
         if ( list->regs[i].start > to ) return 0;   // no match
-        if ( list->regs[i].end >= from && list->regs[i].start <= to ) break; // found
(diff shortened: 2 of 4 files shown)


pd3:
I haven't used the old htslib/regidx for a long while now, but this does not seem correct. As far as I remember, and also as the description in regidx.h suggests, it internally uses 0-based inclusive coordinates. Do you really need to modify regidx, cannot this be fixed in tabix??

If changes to regidx should be made, maybe it is time to finally replace htslib/regidx with the much improved bcftools/regidx.

valeriuo:
We can fix it just from tabix, by changing the coordinate comparison logic, but we wanted to work with only two formats (1-based inclusive for CLI regions and TAB region files and 0-based exclusive for genomic positions and BED region files). Having another intermediate representation (0-based inclusive) seemed unnecessary. Unless there is valid reason for keeping it. 

I agree that we should unify the regidx code and have a single homogeneous interface, but it should probably be part of a separate effort.

pd3:
I never liked the half-exclusive BED format. Although I understand why it's liked by some, I find it much less intuitive and therefore more error prone. I prefer 1-based inclusive coordinates for users and 0-based inclusive coordinates internally.

daviesrob:
Unfortunately it looks like the zero-based inclusive range is part of out API, as it's in `htslib/regidx.h`, so I guess we're stuck with it.  I guess we'll just have to fix the test instead and document that the inputs to it are expected to be in slightly inconsistent formats.

PR #759, opened 2018-08-13
Fix memory leak in knetfile.c

dalexander:
This fixes a memory leak found by fuzzing tools at Google.  Here's the comment from the team that found this:

fp->size_cmd is set to allocated memory in kftp_parse_url() but this is not freed by knet_close(). The fix should be safe in other cases as these allocate the memory of fp via calloc leaving fp->size_cmd equal to NULL by default, and free(NULL) is a nop (https://linux.die.net/man/3/free).

### DIFF
--- a/knetfile.c
+++ b/knetfile.c
@@ -597,6 +597,7 @@ int knet_close(knetFile *fp)
 	free(fp->host); free(fp->port);
 	free(fp->response); free(fp->retr); // FTP specific
 	free(fp->path); free(fp->http_host); // HTTP specific
+	free(fp->size_cmd);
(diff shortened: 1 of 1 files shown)


daviesrob:
Thanks.  It might be a good idea to make a similar pull request on [klib](https://github.com/attractivechaos/klib), which is the upstream source of `knetfile.c`.

I should also mention that `knetfile` will at some point be removed from HTSlib as it can't handle `https://` URLs.  After that, libcurl (via `hfile_libcurl.c`) will be the default way of accessing remote data.

PR #761, opened 2018-08-14
Update regidx - bug fixes and new features. Breaks API

pd3:
This regidx code lived for years in bcftools alongside the deprecated htslib/regidx. It would be nice to make the new features available in htslib.


daviesrob:
It looks like these have diverged over time, so each version is better in different areas.  Some merging will be needed to get an optimal solution here.

daviesrob:
Various small things like using `hts_parse_decimal()` instead of `strtod()`, `isspace_c()` instead of `isspace()`, `hts_log_error()` instead of `frpintf(stderr, ...)`, `HTSLIB_REGIDX_H` include guard instead of `__REGIDX_H__` (which is illegal according to the C99 standard).  There are also changes to error checking which would need to be looked into.

pd3:
I changed these. Which error checks do you mean specifically?

jkbonfield:
It looks like the main API breakage here is regidx_parse_f typedef change.  While ABI wise this isn't any issue - a pointer to one function is, realistically at least, going to be passed in much the same as a pointer to a different one, it does break the API if old code expects the function prototype to be different.

Given the change is from reg_t to explicit beg and end, is there a reason to actually change this prototype?

I did a github code search for regidx_init and found a myriad copies of htslib itself (so not affected) and some bcftools ones.  Most of those, if not all, also have copies of htslib embedded - I guess if you're going down that style of programming it's all or nothing for dependencies.  However I did find one usage:

https://github.com/freeseek/mocha/blob/e2d21d189c575916d8e23a062fc399d808bca7e8/vcfmocha.c#L3198

So if there's one, there may be more out there.  I think we should just keep the API as it is and change the rest of it.  If you really need a new function that has explicit beg and end arguments then it can be done as a new regidx_init2 (etc) function.  The code changes required in bcftools would then be minimal.

pd3:
The `reg_t` type disappeared from public view, `regitr_t` changed too. All but the necessary structures were made opaque in order to avoid the complications which prevent us from making regular updates.

Given that you found a single use in a single project whereas there are 23 uses of the new code in bcftools, 1 in annot-regs and more in non-released projects, I'd argue that a fair way forward is to switch to the new API. It was deliberately designed so that future updates cannot stall like this.

jkbonfield:
I'm not arguing against that and a new better designed API is clearly a big win.  I'm simply arguing against removing the old API at the same time given apparently it has user(s).

We've regularly added new functions (see `sam_index_build`*) with improved APIs.

pd3:
I understand. My point is that some users will have to change the code either way if this pull request is incorporated in any form. Either it will be the 23+1 documented cases in bcftools and annot-regs or some other hypothetical users.

Funnily enough, even the one you found in fact uses the **new** API, taken directly from bcftools code base, not the one living in htslib!
:-))

regidx.c
```diff
@@ -244,48 +329,136 @@ void regidx_destroy(regidx_t *idx)
...
+            memset(list->idx+old_midx, 0, sizeof(uint32_t)*(midx-old_midx));
+        }
+        if ( ibeg==iend ) {
+            if ( !list->idx[ibeg] ) list->idx[ibeg] = j + 1;
+        } else {
+            for (k=ibeg; k<=iend; k++)
+                if ( !list->idx[k] ) list->idx[k] = j + 1;
+        }
+        if ( list->nidx < iend+1 ) list->nidx = iend+1;
+    }
```
pd3: This does not seem safe for 64-bit positions.

jkbonfield: Agreed. reg_t is defined as:

```
#define iBIN(x) ((x)>>13)

typedef struct
{
    hts_pos_t beg, end;
}
reg_t;
```

Hence ibeg and iend should be declared hts_pos_t above too.  Good spot.

regidx.c
```diff
@@ -244,48 +329,136 @@ void regidx_destroy(regidx_t *idx)
...
+        uint32_t i = list->idx[ibeg];
+        if ( !i )
+        {
+            int iend = iBIN(end);
+            if ( iend > list->nidx ) iend = list->nidx;
+            for (i=ibeg; i<=iend; i++)
+                if ( list->idx[i] ) break;
+            if ( i>iend ) return 0;
+            i = list->idx[i];
+        }
```
jkbonfield: Similar to @pd3's comment above,  `ibeg` and `iend` come from `iBIN` so should be `hts_pos_t` too, as should that `uint32_t i` too.

PS.  For what it's worth, 31-bits + 13 bits shift equates to 16 GB before overflow.  That's conceivable for chromosome length on some organisms if I recall.

daviesrob: I think you may be out by a factor of 1024 or so.  `2^(31+13) = 17592186044416`, so it should be good for up to 17 Tbases.  This limit is [here in the header file](https://github.com/samtools/htslib/blob/7ab4cac/htslib/regidx.h#L82) and is enforced in `regidx_push` so it shouldn't be possible to exceed this.

I think 17Tbases is likely to be enough for now.  We will be forced to amend it should anything be discovered that needs a bigger number.

jkbonfield: Oops yes you're right.  I miscounted my units.

In that case agreed.  It's not a practical concern, and even if it ever became one it's internal to the code and not an API/ABI problem.

pd3: The maximum coordinate 1<<44 should be probably lowered. As it is now, keeping an index like this could require more than 8*((1<<44)>>13)/1e9 = 17GB. One of the largest known genomes (Picea abies) has 20e9 bp. Even if this was in a single chromosome, 1<<35 would suffice while keeping the memory in the order of tens of MB.

daviesrob: The actual amount of memory used depends on the highest position in the region list, so normally it shouldn't be too big.  However there is a potential for overflow on 32-bit platforms when calculating the size of the memory buffer, so we probably do need to cut it down a bit more.  1 Tbp should be plenty big enough, even if the estimated genome size of [_Polychaos dubium_](https://www.ncbi.nlm.nih.gov/pubmed/16701223) turns out to be real.  That would give a maximum index size of 128Mb.

daviesrob:
Limit changed.  I was slightly out with the maximum index size - it's 512Mbytes (I forgot to multiply by `sizeof(uint32_t)`).  It still won't overflow, and normally the index should be much smaller.

In the future we may want to check how well this index works on some realistic examples.  It might be possible to tweak it to be more efficient on cases like very large references.

pd3:
Because of the possible overflow, `idx` and related variables must become 64 bit integers as well. Therefore in your calculation multiply by `sizeof(uint64_t)`.

daviesrob:
@pd3 No need.  `idx` is an index into the list of regions, so if you're still happy with a maximum of 2G regions for each chromosome it doesn't need to get any bigger.

pd3:
@daviesrob Uh, forgot about the `>>13` bit, sorry.

regidx.c
```diff
@@ -294,23 +467,30 @@ int regidx_parse_bed(const char *line, char **chr_beg, char **chr_end, reg_t *re
...
-    if ( !*se ) { hts_log_error("Could not parse bed line: %s", line); return -2; }
 
     *chr_beg = ss;
     *chr_end = se-1;
 
+    if ( !*se )
+    {
+        // just the chromosome name
+        *beg = 0;
+        *end = MAX_COOR_0;
```
jkbonfield: This line can cause high memory usage, but only in an very esoteric scenario.

If we have a region file which is a mix of whole chromosomes and specific regions in those same chromosomes, then `reglist_build_index_` is called and it allocates around a Gb per chromosome.  I ran it on some mouse genome data with all the chromosomes listed twice and it started swapping the 48Gb memory on my system.

However this is VERY specific.  I expect either we'll be specifying regions which are whole chromosomes and not parts of them - in which case we have just 1 entry in the list and the index isn't built - or we'll be specifying lots of regions in each chromosome in which case the end coordinate is known and it works fine (as before).

It's only the mix of whole chromosomes plus at least one sub-region in that same chromosome which triggers the indexing code.

I'm inclined to just reduce the value of MAX_COOR_0 a bit (say 64Gb instead of 1Tb) and just leave it at that.  If it ever becomes an issue we can deal with it differently later, or if we feel it is a concern, merge anyway and raise an issue so we can track it.

This is not a regression as the current code cannot handle this mix of regions anyway.

daviesrob: Size has been reduced, and I've also changed the way the memory gets allocated so it may be a bit more efficient.

daviesrob:
Rebased so AppVeyor build works.

PR #772, opened 2018-09-21
Improve c++ compatibility of htslib header files

### LINKED ISSUE #771: kstring.h error: initializer-string for array of chars is too long
When trying to install another program (pindel), there is an error related to kstring.h

```
make -C src pindel
make[1]: Entering directory '/mnt/projects/cwrussell/MAE/svre/pindel/Pindel/src'
g++  -I/mnt/projects/cwrussell/MAE/svre/pindel/htslib -Wall -g -c -O3 -fopenmp pindel.cpp -o pindel.o
In file included from /mnt/projects/cwrussell/MAE/svre/pindel/htslib/htslib/khash.h:131:0,
    from pindel.h:34,
    from pindel.cpp:35:
/mnt/projects/cwrussell/MAE/svre/pindel/htslib/htslib/kstring.h: In function ‘int kputuw(unsigned int, kstring_t*)’:
/mnt/projects/cwrussell/MAE/svre/pindel/htslib/htslib/kstring.h:221:9: error: initializer-string for array of chars is too long [-fpermissive]
         "90919293949596979899";
         ^
```

Changing line 211 in kstring.h from `static const char kputuw_dig2r[200] =` to `static const char kputuw_dig2r[201] =` allows the pindel installation to occur without error, although I'm unsure of any other consequences related to this change.

daviesrob:
Allow room in kputuw_dig2r for the nul byte on the end of the string used to initialize it.  gcc doesn't mind it being dropped but g++ complains.

Fix missing stdint.h limit macros when including htslib/kstring.h and htslib/vcf.h in a c++ compilation.  The C99 standard says c++ implementations should only define them if __STDC_LIMIT_MACROS
is also defined.  C++11 later put them back, but older versions of libc may not provide them without this.

As we can't be sure we're the first to include stdint.h, defining it in our header files may be too late. Instead suitable definitions are added for the macros that are used by each header.

Fixes #771 (kstring.h error: initializer-string for array of chars is too long)

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -1731,7 +1731,7 @@ int vcf_hdr_write(htsFile *fp, const bcf_hdr_t *h)
 
 void bcf_enc_vint(kstring_t *s, int n, int32_t *a, int wsize)
 {
-    int32_t max = INT32_MIN + 1, min = INT32_MAX;
+    int32_t max = INT32_MIN, min = INT32_MAX;
     int i;
     if (n <= 0) bcf_enc_size(s, 0, BCF_BT_NULL);
     else if (n == 1) bcf_enc_int1(s, a[0]);
@@ -1742,13 +1742,13 @@ void bcf_enc_vint(kstring_t *s, int n, int32_t *a, int wsize)
             if (max < a[i]) max = a[i];
             if (min > a[i]) min = a[i];
         }
-        if (max <= INT8_MAX && min >= INT8_MIN + 8) {
+        if (max <= BCF_MAX_BT_INT8 && min >= BCF_MIN_BT_INT8) {
             bcf_enc_size(s, wsize, BCF_BT_INT8);
             for (i = 0; i < n; ++i)
                 if ( a[i]==bcf_int32_vector_end ) kputc(bcf_int8_vector_end, s);
                 else if ( a[i]==bcf_int32_missing ) kputc(bcf_int8_missing, s);
                 else kputc(a[i], s);
-        } else if (max <= INT16_MAX && min >= INT16_MIN + 8) {
+        } else if (max <= BCF_MAX_BT_INT16 && min >= BCF_MIN_BT_INT16) {
             uint8_t *p;
             bcf_enc_size(s, wsize, BCF_BT_INT16);
             ks_resize(s, s->l + n * sizeof(int16_t));
--- a/htslib/kstring.h
+++ b/htslib/kstring.h
@@ -151,7 +151,8 @@ static inline char *ks_release(kstring_t *s)
 
 static inline int kputsn(const char *p, size_t l, kstring_t *s)
 {
-    if (l > SIZE_MAX - 2 - s->l || ks_resize(s, s->l + l + 2) < 0)
+	size_t new_sz = s->l + l + 2;
+	if (new_sz < s->l || ks_resize(s, new_sz) < 0)
 		return EOF;
 	memcpy(s->s + s->l, p, l);
 	s->l += l;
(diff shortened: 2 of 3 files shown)


jmarshall:
The C99 standard was [full of crap when it tried to tell the C++ committee what to do](https://stackoverflow.com/a/8132440), so the comment doesn't need to be so detailed — at this point, disabling these macros in the absence of `__STDC_LIMIT_MACROS` is just a bug in old compilers/C libraries.

What about just avoiding using `INTx_MAX` etc in HTSlib's public header files? Certainly for the _vcf.h_ ones, just writing out the numerical values in `#define bcf_int8_vector_end` etc wouldn't be much of a loss.

Defining `SIZE_MAX` itself will lead to ill-formed programs if something includes `<stdint.h>` **after** `"htslib/kstring.h"`. It's a somewhat theoretical problem (you currently get warnings only if you're using `-Wsystem-headers`, which most people won't do), but could be avoided by instead defining a local name within the header:

```c
#ifdef SIZE_MAX
#define SIZE_MAX_KS SIZE_MAX
#else
#define SIZE_MAX_KS ((size_t) -1)
#endif
…
// Use SIZE_MAX_KS within htslib/kstring.h
…
#undef SIZE_MAX_KS  // Undef it at the end of the header file
```

(Undeffing it afterwards so that other code (especially third-party code) just uses the standard nomenclature.)

daviesrob:
Yes, the `vcf.h` bit is quite ugly, so it might be better if it defined its own MAX and MIN macros.

I don't see why including `stdint.h` again should be a problem.  `htslib/kstring.h` will already have included it, and we only create our own `SIZE_MAX` if `stdint.h` failed to make one.  As we've already included `stdint.h`, it shouldn't do anything if included again later.

I'd rather not have to jump through hoops in the rest of the file, just because of an ancient c/c++ incompatibility.

jmarshall:
> I don't see why including stdint.h again should be a problem. htslib/kstring.h will already have included it

Oops — you are quite right that _<stdint.h>_ was never likely to operate like _<assert.h>_/`NDEBUG` despite the C people asserting that it had a somewhat similar control macro.

It would make me nervous in general to define names that are reserved to the compiler, especially in a public header, but YMMV.

daviesrob:
Mk.2 version (just pushed) avoids the problem by not using `SIZE_MAX` at all.  According to [Godbolt](https://godbolt.org/), the new version of the overflow checks in `kstring.h` may actually be a bit more efficient.

vcf.c
```diff
@@ -1731,7 +1731,7 @@ int vcf_hdr_write(htsFile *fp, const bcf_hdr_t *h)
 
 void bcf_enc_vint(kstring_t *s, int n, int32_t *a, int wsize)
 {
-    int32_t max = INT32_MIN + 1, min = INT32_MAX;
+    int32_t max = BCF_MIN_BT_INT32 + 1, min = BCF_MAX_BT_INT32;
```
jmarshall: These are boundaries on the **unencoded** `int32_t` values, so should just be

    int32_t max = INT32_MIN, min = INT32_MAX;

and on the lines below you need to lose the `+ 8`.

valeriuo: Thanks for spotting that. I was under the impression they were match for match.

jmarshall: Thanks — looks good to me now.

valeriuo:
Closed by 57fe419344cb03e2ea46315443abd242489c32f2 and e85b599fb0af89570e1d546852a7f201530d7ab4.

PR #773, opened 2018-09-26
[travis] Add Travis CI badge

Fohlen:
Hello there! Thanks for creating htslib, it is super useful.
When looking at the GitHub repo I noticed that you have Travis CI enabled, however it is not indicated in the README. Usually people tend to judge the status of a project easily by something like CI, which is why I thought I'd add the missing Travis badge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![Build Status](https://travis-ci.org/samtools/htslib.svg?branch=master)](https://travis-ci.org/samtools/htslib)
+
 HTSlib is an implementation of a unified C library for accessing common file
 formats, such as [SAM, CRAM and VCF][1], used for high-throughput sequencing
 data, and is the core library used by [samtools][2] and [bcftools][3].


Fohlen:
Additionally this could also use the AppVeyor badge, maybe

valeriuo:
Hi! Yes, the AppVeyor badge should be there too. However, they should be linked to the build status on develop, as master is the release branch and it is guaranteed to have a working build. Thus, the badges should work as an indication to users who want the latest code, without waiting for the official release.

Also, could you add a download button?

[Github All Releases](https://github.com/samtools/htslib/releases/latest)
`[Github All Releases](https://github.com/samtools/htslib/releases/latest)`

Fohlen:
Hey there @valeriuo! These are good suggestions. I added them to README and believe now it's good to be merged. Cheers.

valeriuo:
@Fohlen it appears the AppVeyor public style badges are a bit unreliable. The one you've added showed the build as failing, which was not correct. I've added the private tokenized version and this one is ok.

Fohlen:
@valeriuo I noticed that as well but with the private token it works. What needs to be done in order to bring this forward?

PR #777, opened 2018-10-04
Bug fix to cram index query for unmapped reads.

jkbonfield:
Fixes https://github.com/samtools/samtools/issues/958

### DIFF
--- a/cram/cram_index.c
+++ b/cram/cram_index.c
@@ -229,7 +229,7 @@ int cram_index_load(cram_fd *fd, const char *fn, const char *fn_idx) {
             goto fail;
 
         e.end += e.start-1;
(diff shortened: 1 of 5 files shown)


jkbonfield:
I found another bug sadly.  Using a second region in conjunction with "*", samtools works without the multi-region iterator but not with it on CRAM.  It works both with and without on BAM.

    $ samtools view -c -M ~jkb/scratch/data/9827_2#49.cram "GL000217.1" "*"

    [E::cram_read_container] Container header CRC32 failure

Previously this didn't error, but it did give incorrect results due to the bug we are fixing here.

jkbonfield:
Rebased and merged as adab0cbbd9535fea7d7d3c53491593eedd0f12dd

PR #781, opened 2018-10-11
On travis, add gcc-8 tests, remove MacOS gcc and try to tidy matrix

daviesrob:
Add travis tests that use gcc-8 on Linux.  These take about 50 seconds longer than the installed gcc due to the extra apt install needed, so currently there are only two of them.

Remove the gcc tests on MacOS.  It wasn't really running gcc, it was clang with a gcc front-end.

Put all the tests in the 'matrix:' section instead of having them added implicitly.  This is so we get the right settings for each case and can have more control over exactly which tests get run.

The tests get queued in the same order that they appear in the matrix section.  As Mac jobs tend to be a bit slower to start they've been moved up a bit in the hope that they will tile better with the Linux ones and give a faster run time overall.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,52 @@
 # Control file for continuous integration testing at http://travis-ci.org/
 
 language: c
-compiler:
-  - clang
-  - gcc
-
-os:
-  - linux
-  - osx
-
-env:
-  - USE_CONFIG=no
-  - USE_CONFIG=yes
 
 matrix:
   include:
-  - compiler: gcc
+  - compiler: gcc-8
     os: linux
-    env: USE_CONFIG=yes USE_LIBDEFLATE=yes
+    env: USE_CONFIG=yes CC=gcc-8 AR=gcc-ar-8
+    addons:
+      apt:
+        sources:
+          - ubuntu-toolchain-r-test
+        packages:
+          - gcc-8
+
+  - compiler: gcc-8
+    os: linux
+    env: USE_CONFIG=yes USE_LIBDEFLATE=yes CC=gcc-8 AR=gcc-ar-8
+    addons:
+      apt:
+        sources:
+          - ubuntu-toolchain-r-test
+        packages:
+          - gcc-8
+
+  - compiler: clang
+    os: osx
+    env: USE_CONFIG=no
+
+  - compiler: clang
+    os: osx
+    env: USE_CONFIG=yes
+
   - compiler: clang
     os: osx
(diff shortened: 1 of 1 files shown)


junaruga:
> These take about 50 seconds longer than the installed gcc due to the extra apt install needed

There is a cache feature on Travis for installed application, though I do not know the detail.
https://docs.travis-ci.com/user/caching/

daviesrob:
Travis recommend not caching [Debian packages](https://docs.travis-ci.com/user/caching/#things-not-to-cache) because it takes as long to get them from the cache as it does to download them from the apt repository.  It's not a major problem though as after a bit of pruning this version runs in about the same time as the old one.

junaruga:
What is the reason of that this PR is not merged yet?

I just found an idea to improve the running time of Travis regular jobs.

## Run heavy job for only Travis `cron` mode

If we care the total running time, we can adopt below way.
Maybe we can run heavy and secondary important jobs for only Travis `cron` mode like this with stage [1] and conditional build [2][3].

Set like this.

```
matrix:
  include:
    - some task 1 without stage
      ...
    - same task 2 without stage
       ...
    - stage: secondary
      compiler: gcc-8
      ...

stages:
  - name: secondary
    if: type = cron
```

We might need to replace `matrix - include` -> `jobs - include`.

## Using YAML anchor and reference feature

It's just for refactoring. We can use anchor and references= of YAML feature like this. See the document [4] and the good example [5].
Travis is using YAML 1.1.

```
    - compiler: gcc-8
      env: USE_CONFIG=yes CC=gcc-8 AR=gcc-ar-8
      addons: &addons_gcc8
        apt:
          sources:
            - ubuntu-toolchain-r-test
          packages:
            - gcc-8

    - compiler: gcc-8
      os: linux
      env: USE_CONFIG=yes USE_LIBDEFLATE=yes CC=gcc-8 AR=gcc-ar-8
      addons: *addons_gcc8
```

[1] https://docs.travis-ci.com/user/build-stages
[2] https://docs.travis-ci.com/user/conditional-builds-stages-jobs/
[3] https://docs.travis-ci.com/user/conditions-v1
[4] https://en.wikipedia.org/wiki/YAML#Advanced_components
[5] https://github.com/Microsoft/GSL/blob/master/.travis.yml

valeriuo:
Merged as b034a41be952dff6a1afd267f2dea5a6e29ab8e9

junaruga:
Thank you for merging!

PR #783, opened 2018-10-17
Add documentation to sam_read1 and sam_write1 functions

### LINKED ISSUE #782: sam_write1 return
Hello,

I'm having an issue properly catching a write error in bgzf  ([E::bgzf_flush] File write failed (wrong size)). Is the documentation on line 391 from sam_read1 (394) or sam_write1? 

The documents lead me to believe that a return of < -1 is an error code, but I think it might be == -1, after reading through the function.

Can you clarify this for me?

Documentation:
https://github.com/samtools/htslib/blob/c93b6735a8ebddc47a31a325f8c15313d82ae6d5/htslib/sam.h#L391 

function call:
https://github.com/samtools/htslib/blob/c93b6735a8ebddc47a31a325f8c15313d82ae6d5/htslib/sam.h#L394

valeriuo:
Fix #782

### DIFF
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -387,10 +387,19 @@ int sam_index_build3(const char *fn, const char *fnidx, int min_shift, int nthre
     int sam_parse1(kstring_t *s, bam_hdr_t *h, bam1_t *b) HTS_RESULT_USED;
     int sam_format1(const bam_hdr_t *h, const bam1_t *b, kstring_t *str) HTS_RESULT_USED;
 
-    /*!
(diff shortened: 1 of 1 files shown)


jmarshall:
A blank line between `int sam_read1(…` and the following function's documentation would be good.

Similarly to the already documented functions, there's no need to put the function name on the `///` line — if documentation is one day ground from these header annotations, doxygen will pick out the function name itself.

What does “fully or partially” mean? What does “placeholder” imply?

Issue #784, opened 2018-10-25
Samtools not seeing cram index file when generating pre-signed AWS url


mebbert:
Hello,

A large sequencing consortium recently released data for ~5000 new whole genomes that are hosted on `AWS`. They require the downloader to specify `--request-payer requester` for billing. Per the very helpful tip from @delagoya at #346, I was able to generate a pre-signed URL using `boto3`.

I only want to download specific regions from all samples, so I'm using `samtools view`. Unfortunately, `samtools` is complaining that `[main_samview] random alignment retrieval only works for indexed BAM or CRAM files`, eventhough I downloaded the `.crai` file for the associated `.cram` file *before* attempting to access the region of interest. `samtools` specifications state that it will look for a local `.crai` file. NOTE: I do receive the header for the `.cram` file.

Can anyone tell me why `samtools` doesn't see the local index file?

Here is my request and directory contents:

### Command
    samtools view -h 'https://[bucket].s3.amazonaws.com/path/to/cram/[sample].cram?AWSAccessKeyId=XXXXXXXXXXX&x-amz-request-payer=requester&Expires=XXXXXXXXXXX&Signature=XXXXXXXXXXXXX' chrX:XXXXXX-XXXXXX > test.sam

### Directory contents (after running command)

> [sample].cram.crai
> test.sam

I even tried mimicking the directory structure in the bucket (`path/to/cram/`) within the working directory, in case `samtools` was looking there. No dice.

I would appreciate any help.

jkbonfield:
I'm not sure samtools supports this yet.  #346 is a proposal, rather than something that was implemented, but that aside there is also a bug still.

I experimented with some 1000 genomes public S3 data and there's clearly something broken here.

```
samtools view 'https://1000genomes.s3.amazonaws.com/data/NA12878/alignment/NA12878.alt_bwamem_GRCh38DH.20150718.CEU.low_coverage.cram' chr21:10013337-10013337
# 26 records returned
```

Now put any old ?a=b http parameter on and the entire behaviour changes.  It no longer does a range query, but downloads the complete file to <blah>.cram?a=b.crai.    This is clearly a bug somewhere in htslib.  (I can verify this with htsfile -c too).

jkbonfield:
ARGH I see it now.

Normally doing a range query on http://<server>/<path>.cram will download http://<server>/<path>.cram.crai first.  It does this by simply adding .crai onto the filename.

[...]

PR #788, opened 2018-10-30
New filename extension modification API function in htslib/hfile.h 

jmarshall:
As noted in #784, HTSlib would benefit from an URL-aware function to add/replace filename extensions.

`hisremote()` is precedent for _hfile.h_ functions not directly associated with I/O, and similarly `haddextension()` can reuse the same URL-detection infrastructure — so _hfile.c_ is a good place for it. So here is a suggested API.

I previously tried to keep `kstring_t` out of _hfile.h_ :smile:, but with the first commit enabling better forward `struct kstring_t` declarations it's less invasive.

This will misrecognise filenames like `foo:bar?baz` as URLs, but there's not much that can be done about it (the library doesn't know which if any file should exist, so `access(2)` doesn't help), such filenames deserve what they get, and `./foo:bar?baz` is a workaround to ensure it's recognised as a file path.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # Makefile for htslib, a C library for high-throughput sequencing data formats.
 #
-#    Copyright (C) 2013-2017 Genome Research Ltd.
+#    Copyright (C) 2013-2018 Genome Research Ltd.
 #
 #    Author: John Marshall <jm18@sanger.ac.uk>
 #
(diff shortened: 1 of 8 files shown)


jkbonfield:
Thanks.  Rebased and merged as c852a3fea3ab143d61d4e5b818c50ea6cc189edc.

I'll look at fixing #784 now.

PR #790, opened 2018-11-01
Use new haddexension API when fetching remote indices.

### LINKED ISSUE #784: Samtools not seeing cram index file when generating pre-signed AWS url

jkbonfield:
Specifically range querying e.g. https://server/a/b/c.bam?param=val
will now retrieve https://server/a/b/c.bam.bai?param=val and locally
write this to c.bam.bai.

Fixes #784

I confess however it is not without problem, but whether it is a problem that needs fixing in this PR or is a separate issue I do not know.  Specifically file:// URIs are questionable in their use of percents.  https://tools.ietf.org/html/rfc8089 states that UTF8 files SHOULD be percent encoded, but then goes on to discuss heuristics and explicit knowledge for decisions when not to percent encode.  I assume percent itself however must always be encoded...

The very first file I chose to test file: on happened to be `file:///nfs/sam_scratch/jkb/data/9827_2#49.bam`.  It worked before, and doesn't now due to the `#`.  Furthermore, we cannot escape that with %23 instead.

I don't know what the best solution is here (other than to not use file: URIs).  Possible choices are to not do the ?# trimming on file URIs (technically we should), or to do so local hackery trying with/without and see what works (messy).

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -2775,33 +2775,46 @@ int hts_itr_multi_next(htsFile *fd, hts_itr_multi_t *iter, void *r)
 /**********************
  *** Retrieve index ***
  **********************/
+// Local_fn and local_len will return a sub-region of 'fn'.
+// Eg http://elsewhere/dir/foo.bam.bai?a=b may return
+// foo.bam.bai via local_fn and local_len.
+//
 // Returns -1 if index couldn't be opened.
 //         -2 on other errors
-static int test_and_fetch(const char *fn, const char **local_fn)
+static int test_and_fetch(const char *fn, const char **local_fn, int *local_len)
 {
     hFILE *remote_hfp;
     FILE *local_fp = NULL;
     uint8_t *buf = NULL;
     int save_errno;
+    kstring_t s = {0};
 
     if (hisremote(fn)) {
         const int buf_size = 1 * 1024 * 1024;
         int l;
-        const char *p;
-        for (p = fn + strlen(fn) - 1; p >= fn; --p)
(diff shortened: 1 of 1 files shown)


valeriuo:
Merged into #870

PR #791, opened 2018-11-06
Allow splitting of MD5 string in cram REF_PATH URLs

daviesrob:
When expanding file names in REF_PATH, it's possible to substitute %Ns (where N is a number) to include the next N characters of the reference hex MD5 checksum. Using %s with no digit adds all of the remaining characters.

For some reason, http:// etc. URLs in REF_PATH only allowed the %s substitution.  This pull request allows the %Ns version as well, with the minor difference that N can only be a single digit (to avoid
URL percent-encodings like %25 that are followed by an 's' character from being misinterpreted).

This is implemented by making find_file_url() use the same expand_path() function as find_file_dir().  expand_path() gets an extra parameter so the URL version can limit the number
of digits between the '%' and 's'.  find_file_url() gets some extra error handling code to free up the memory returned by expand_path().

### DIFF
--- a/cram/open_trace_file.c
+++ b/cram/open_trace_file.c
@@ -168,45 +168,48 @@ char *tokenise_search_path(char *searchpath) {
     return newsearch;
 }
 
+static char *expand_path(char *file, char *dirname, int max_s_digits);
+
 mFILE *find_file_url(char *file, char *url) {
-    char buf[8192], *cp;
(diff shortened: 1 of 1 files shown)


jkbonfield:
Rebased and merged as 226b4a8f9d08383955b3311d73001cddde6e3eec

PR #792, opened 2018-11-07
CRAM encoding now checks MD/NM for validity.

jkbonfield:
If the MD / NM values are incorrect, then we now detect that
auto-generation would yield something different and store them
verbatim so the round trip matches.

The missing component though is still a means to detect whether a
record had NM/MD at all.  Sadly this is not possible within CRAM
currently unless we invent a new tag type.


cram/cram_encode.c
```diff
@@ -2530,6 +2543,8 @@ static int process_one_read(cram_fd *fd, cram_container *c,
     ref = c->ref;
     cr->flags       = bam_flag(b);
     cr->len         = bam_seq_len(b);
+    if (!bam_aux_get(b, "MD"))
+        MD = NULL;
```
daviesrob: I think you need
```
else
    MD->l = 0;
```
here otherwise MD from all previous reads get carried over into the next one.

cram/cram_encode.c
```diff
@@ -2808,6 +2847,9 @@ static int process_one_read(cram_fd *fd, cram_container *c,
         fake_qual = spos;
         cr->aend = fd->no_ref ? apos : MIN(apos, c->ref_end);
         cram_stats_add(c->stats[DS_FN], cr->nfeature);
+
+        if (MD && ref)
+            kputuw(apos - MD_last - MD_last, MD);
```
daviesrob: Should this just be `apos - MD_last`?

test/test.pl
```diff
@@ -350,6 +352,41 @@ sub test_view
...
+        $cram = "$base.tmp.cram";
+
+	# Forcibly store MD and NM and don't auto-generate.
+	# ALL NM/MD should match and be present only when originally present
+        testv $opts, "./test_view -o store_nm=1,store_md=1 -t $ref -C $sam > $cram";
+        testv $opts, "./test_view -i decode_md=0 -D $cram > $cram.sam_";
+        testv $opts, "./compare_sam.pl  $sam $cram.sam_";
+
+        # Skip auto-MD generation; check MD iff in output file
+        testv $opts, "./test_view -t $ref -C $sam > $cram";
```
daviesrob: The bugs in `process_one_read()` are causing all MD:Z to be stored verbatim, whether needed or not.  This test needs to be beefed up to detect that.

Presumably when this PR is working, the `xx#MD.sam` round-tripped with `decode_md=0` should contain no `MD:Z` tags, and `xx#MD2.sam` should have one for each read?

cram/cram_encode.c
jkbonfield: Yes, this got moved and lost in the translation from dstring to kstring sadly.

cram/cram_encode.c
```diff
@@ -2097,23 +2105,28 @@ static char *cram_encode_aux(cram_fd *fd, bam_seq_t *b, cram_container *c,
         // MD:Z
         if (aux[0] == 'M' && aux[1] == 'D' && aux[2] == 'Z') {
             if (cr->len && !fd->no_ref && !(cr->flags & BAM_FUNMAP) && !verbatim_MD) {
-                while (*aux++);
-                continue;
+                if (MD && strncasecmp(MD->s, aux+3, orig + aux_size - (aux+3)) == 0) {
```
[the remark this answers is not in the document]
jkbonfield: It always finishes MD with number of bases left, and even the zero length string will get MD:Z:0.  However it doesn't harm to add `&& MD->s` in there too.

PR #796, opened 2018-11-19
Add fuzz target that opens sequences files.

markus-kusano:
hts_open_fuzzer.c is a fuzz target (see https://llvm.org/docs/LibFuzzer.html).
This is an entry point passing an arbitrary data buffer into htslib. The buffer
is saved within a temporary file and opened as a sequence file.

The htslib Makefile was updated to build hts_open_fuzzer.c as a .o file.
In order to build as a standalone binary we need to link in libFuzzer,
which (I think) would require clang. Building simply as a .o file
atleast provides a sanity check that the fuzzer builds.


daviesrob:
Thanks for this.  It looks like I'll need to get myself an up to date clang to see if I can make myself a working fuzzer.

Meanwhile, it currently breaks the test harness because it ignores the return value from the various `*_write()` functions, which generates an error via function annotations and building with `-Wall -Werror`.  Normally programs should bail out of the read-write loop when these functions error.  Did you ignore this deliberately so that it continues through the input, or can we add the tests to see if they worked?

Does the file need to be written out to disk?  If not, it's possible to use an in-memory hfile to access the data. It's not too elegant and you need to take a copy as `hclose()` will try to free it, but it does appear to work.  `LLVMFuzzerTestOneInput` starts like this:

```
int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
    htsFile *ht_file;
    hFILE *memfile;
    uint8_t *copy;
    copy = malloc(size);
    if (copy == NULL)
        return 0;
    memcpy(copy, data, size);

    memfile = hopen("mem:", "rb:", copy, size);
    if (memfile == NULL) {
        //fuzzer_release_tmpfile(temporary_file);
        free(copy);
        return 0;
    }

    ht_file = hts_hopen(memfile, "data", "rb");
    if (ht_file == NULL) {
        hclose(memfile);
        return 0;
    }

```

and the rest is the same.  It gets rid of all the temporary file stuff, and presumably allows faster fuzzing.

On a more pedantic note, could you reindent the file to have four space indents to match the rest of HTSlib?  There should also be no tab characters, but I think you're OK there at the moment.

Would it be a good idea to put this in a `test/fuzz` directory, on the assumption that other files (e.g. test cases) may join it at some point?

markus-kusano:
Thanks for the comments and please take a look at the new changes :)

> It looks like I'll need to get myself an up to date clang to see if I can make myself a working fuzzer.

Let me know if you need help. I haven't actually tested getting this thing working fully with clang within htslib's build system so apologies if it doesn't work.

> Meanwhile, it currently breaks the test harness because it ignores the return value from the various *_write() functions, which generates an error via function annotations and building with -Wall -Werror. Normally programs should bail out of the read-write loop when these functions error. Did you ignore this deliberately so that it continues through the input, or can we add the tests to see if they worked?

Ah, this was sloppy on my part :) I was ignoring them out of laziness. I've added some asserts() and explicit errors checks. I didn't use `goto cleanup` in `view_vcf()` and `view_sam()` which may have made things a bit messier. I'm happy to look into changing this if you'd like. It seems that the tests are passing now.

> Does the file need to be written out to disk? If not, it's possible to use an in-memory hfile to access the data. It's not too elegant and you need to take a copy as hclose() will try to free it, but it does appear to work. LLVMFuzzerTestOneInput starts like this:

Thanks for pointing this out and providing the example code! I had no idea there was memfile support already. I've switched over to using this.

> On a more pedantic note, could you reindent the file to have four space indents to match the rest of HTSlib? There should also be no tab characters, but I think you're OK there at the moment.

Not pedantic at all :) thanks for bringing that up. I've switched to four spaces.

> Would it be a good idea to put this in a test/fuzz directory, on the assumption that other files (e.g. test cases) may join it at some point?

Good idea, I've moved the file into test/fuzz. You're right, we should create an initial set of test inputs for the fuzzer (these can just be any sequence files).

markus-kusano:
Just bumping this to make sure it didn't get lost somewhere :)

daviesrob:
Don't worry it's not forgotten.  I've managed to make a working fuzzer now (after a bit of fun building `clang`).  It almost instantly discovered a bcf file bug, plus a couple of unrelated problems related to memory allocations in the vcf/bcf reader when I tried to minimize the file that caused the crash.  I've now got a fix for these problems, and have been trying to tame vcf.c's habit of calling `exit()` or `abort()` on bits of file that it doesn't like (somewhat extensive and long overdue changes required).  Hopefully with these fixes libfuzzer will run for a bit longer before stopping.

I should be able to squash + merge this soon, hopefully later this week.

daviesrob:
Squashed and merged as 4eebd00.

I'll see if I can tweak the Makefile (and possibly configure) a bit to get the fuzzer to build more easily.  It looks like the best solution is to have `CFLAGS` include `-fsanitize=address,fuzzer-no-link`, `LDFLAGS` include `-fsanitize=address` and then update the `hts_open_fuzzer` rule to add `-fsanitize=fuzzer` itself.

jmarshall:
(BTW the Makefile _test/fuzz/hts_open_fuzzer.o_ rule is missing `$(htslib_hfile_h) $(htslib_hts_h) $(htslib_sam_h) $(htslib_vcf_h)` and _hts_open_fuzzer.c_ is missing copyright boilerplate and `#include <config.h>`.)

daviesrob:
Good point.

@markus-kusano Are the Google legal people happy with having this under the same MIT/Expat license as the same as HTSlib, and who should I put on the Copyright line?

markus-kusano:
Google is fine with this being under MIT/Expat. I've updated test/fuzz/hts_open_fuzzer.c to include the Copyright boiler plate. I've assigned the file's Copyright to "Google LLC" and put myself (kusano@google.com) as the author. Let me know if this isn't OK and we can change it.

Regarding the Makefile, that sounds like it'll work. One thing to consider is that OSS-Fuzz has its own build-script system (this is used to allow for multiple fuzzing engines to be used within the same scripts). You can see the description here: https://github.com/google/oss-fuzz/blob/master/docs/new_project_guide.md#buildsh.

Let me know if you need help with getting the Makefile setup with fuzzing, or with getting things setup with OSS-Fuzz: I can get someone from the OSS-Fuzz team to help out.

I'm glad to hear that the fuzzer was useful. When I was developing the fuzzer on my side I also ran into some of these issues (the aborts, some bugs in file-format parsers.). It sounds like they are the same bugs but if you want I can send over the test inputs causing them. Also, once this is into OSS-Fuzz we can just directly upload the test cases, and it'll send out an email with the bugs.

daviesrob:
Thanks, that looks better.  It's now merged as f0b83374.

I'll see how it gets on with my latest attempts at bug-fixing its previous discoveries.  I'll happily accept any other test cases you have as i guess it'll save some time trying to rediscover them.

PR #804, opened 2018-12-11
Add checks for missing copyright boilerplate, tabs and trailing spaces

daviesrob:
Adds scripts `check_copyright.pl` and `check_spaces.pl` to look for missing copyright boilerplate, tabs and trailing spaces.  They can be run using `make maintainer-check` in the source tree.  One of the travis runs is updated to do this so we can spot any problems in pull requests (the rest are left unchanged so we also get to see that the PR otherwise works).  The copyright check is very basic - it just looks for the copyright line and the first few words of the license text - but it should be good enough.

The scripts are currently not part of `make test` so that the normal test harness doesn't break due to random bits of script lying around in source trees that will never by pushed upstream.

Also includes fixes for problems in the existing source found while developing the scripts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ language: c
 
 matrix:
   include:
+  - compiler: gcc
+    os: linux
+    env: DO_MAINTAINER_CHECKS=yes USE_CONFIG=no
+
   - compiler: gcc-8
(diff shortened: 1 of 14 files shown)


valeriuo:
Merged into develop.

PR #808, opened 2018-12-19
Fix seq_nt16_table[(unsigned char) foo] cast

jmarshall:
`sam_cap_mapq()` might be called with arbitrary data in _ref_, including characters '\x80'..'\xFF' that are negative on many platforms.

All uses of `seq_nt16_table[]` in htslib are now cast appropriately.

OTOH there are lots of `seq_nt16_table[(int) ref[x]]` in the SAMtools codebase still. For most of these, `ref` has come from `fai_fetch()` et al, which returns only `isgraph()` characters, so for the typical POSIX/C locale problematic characters won't occur. (Not that this makes such code right…)

All this somewhat suggests that HTSlib could usefully have a wrapper function around `seq_nt16_table` that does this casting for you…


daviesrob:
Thanks for the patch.  Yes, a wrapper around these (and other look-up tables) would be a good idea.

PR #811, opened 2018-12-30
Check if fp is NULL in tabix file_type()

### LINKED ISSUE #810: Invalid Address Deference in function file_type in tabix.c
Hi,

We found an Invalid Address Deference problem in function elf_end in libelf of the latest htslib code base. I have confirmed them with Address Sanitizer, too.

Here are the POC files. Please use " ./tabix $POC " to reproduce this bug. 
[POC.zip](https://github.com/samtools/htslib/files/2717183/POC.zip)

The ASAN dumps the stack trace as follows:

```
ASAN:DEADLYSIGNAL
==14393==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000044 (pc 0x0000004ecd16 bp 0x7fff191771f0 sp 0x7fff191771a0 T0)
    #0 0x4ecd15 in file_type htslib_new/tabix.c:82:45
    #1 0x4f0be3 in main htslib_new/tabix.c:473:17
    #2 0x7fbb4dbf582f in __libc_start_main /build/glibc-Cl5G7W/glibc-2.23/csu/../csu/libc-start.c:291
    #3 0x41a928 in _start (htslib_new/build/bin/tabix+0x41a928)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV htslib_new/tabix.c:82:45 in file_type
==14393==ABORTING
```

daviesrob:
hts_open() returns NULL if it does not recognise the file type of the input.  Add a check for this in file_type() to fix a segfault that would otherwise occur when it tries to dereference the
result from hts_open().

Fixes #810 

This just bails out if hts_open() fails, which is the minimal fix.  It will only fail on files that would previously have made it crash.

Ideally this function should be rewritten to only open the input file once, but that's a rather bigger job.

### DIFF
--- a/tabix.c
+++ b/tabix.c
@@ -79,6 +79,15 @@ int file_type(const char *fname)
     else if (l>=4 && strcasecmp(fname+l-5, ".cram") == 0) return IS_CRAM;
 
     htsFile *fp = hts_open(fname,"r");
+    if (!fp) {
+        if (errno == ENOEXEC) {
+            // hts_open() uses this to report that it didn't understand the
+            // file format.
+            error("Couldn't understand format of \"%s\"\n", fname);
+        } else {
+            error("Couldn't open \"%s\" : %s\n", fname, strerror(errno));
+        }
+    }
     enum htsExactFormat format = fp->format.format;
     hts_close(fp);
     if ( format == bcf ) return IS_BCF;


valeriuo:
Merged as 7ff15099cbccf9d624da13d0a9cc4f974045c414

PR #812, opened 2019-01-03
Header API for SAM/BAM/CRAM

valeriuo:
Management methods:
- `bam_hdr_init` - create a new header
- `bam_hdr_destroy` - free the resources used by the header
- `bam_hdr_read` - read the header from a BGZF file
- `bam_hdr_write` - write the header to a BGZF file
- `bam_hdr_dup` - duplicate the header structure
- `bam_hdr_length` - return the length of the header text
- `bam_hdr_str` - return the header text as a string
- `bam_hdr_nref` - return the number of references (`@SQ` lines)
- `bam_hdr_incr_ref` - increment the reference counter on the header structure

Line level methods:
- `bam_hdr_add_lines` - add a group of lines from a char buffer
- `bam_hdr_add_line` - add a single line with individual key:value pairs
- `bam_hdr_find_line_id` - find a line specified by a key:value identifier
- `bam_hdr_remove_line_id` - remove a line specified by a key:value identifier
- `bam_hdr_remove_line_pos` - remove a line specified by position
- `bam_hdr_update_line` - update a line (works for tags)
- `bam_hdr_keep_line` - remove all lines of the same type except the current one
- `bam_hdr_remove_lines` - remove all lines, except the one specified by the hash set arg
- `bam_hdr_count_lines` - count the header lines
- `bam_hdr_add_pg` - add a new `@PG` line
- `bam_hdr_find_hd` - return the `@HD` line
- `bam_hdr_update_hd` - `@HD` line update specific macro 

Tag level methods:
- `bam_hdr_find_tag_id` - find a tag in a line specified by a key:value identifier
- `bam_hdr_find_tag_hd` - same as above for `@HD` line
- `bam_hdr_find_tag_pos` - find a tag in a line specified by position
- `bam_hdr_remove_tag_id` - remove a tag from a line specified by a key:value identifier
- `bam_hdr_remove_tag_hd` - same as above for `@HD` line


header.c
```diff
@@ -323,6 +323,45 @@ static int bam_hrecs_remove_hash_entry(bam_hrecs_t *hrecs, int type, bam_hrec_ty
...
+    int update_first_line = 0;
+
+    // First line seen
+    if (!hrecs->first_line) {
+        hrecs->first_line = h_type->global_next = h_type->global_prev = h_type;
+        return;
+    }
+
+    // @HD goes at the top (unless there's one already)
+    if (h_type->type == hd_type && hrecs->first_line->type != hd_type) {
```
valeriuo: We should probably prevent the addition of a second `@HD` line.

htslib/sam.h
```diff
@@ -272,28 +289,336 @@ typedef struct {
...
+ */
+int bam_hdr_name2ref(bam_hdr_t *bh, const char *ref);
 
-    bam_hdr_t *bam_hdr_init(void);
-    bam_hdr_t *bam_hdr_read(BGZF *fp);
-    int bam_hdr_write(BGZF *fp, const bam_hdr_t *h) HTS_RESULT_USED;
-    void bam_hdr_destroy(bam_hdr_t *h);
-    int bam_name2id(bam_hdr_t *h, const char *ref);
-    bam_hdr_t* bam_hdr_dup(const bam_hdr_t *h0);
+/// Alias of bam_hdr_name2ref(), for backwards compatibility.
```
valeriuo: We should probably ditch (deprecate) the use of `bam_hdr_t.sdict` altogether and make `bam_name2id` an actual alias of `bam_hdr_name2ref`. Currently, they have different implementations.

header.c
```diff
@@ -0,0 +1,2084 @@
...
+            return -1;
+        }
+    }
+
+    bh->hrecs = hrecs;
+
+    if (hrecs->refs_changed >= 0 && rebuild_target_arrays(bh) != 0)
+        return -1;
+
+    bam_hdr_link_pg(bh);
```
valeriuo: Calling `bam_hdr_link_pg` would result in resetting `bh->text`, which is not intuitive when calling `bam_hdr_parse`, or some of its callers that do not modify the header, such as `bam_hdr_name2ref`.
For example, the following code would result in a null text field, which is probably not the expected behaviour.
```
        samFile *in = sam_open("test_header.sam", "r");
        bam_hdr_t *h = sam_hdr_read(in);	
        int id = bam_hdr_name2ref(h, "ref2");
        printf("id=%d, header=%p\n", id, h->text);

Perhaps, a better place for the `bam_hdr_link_pg` call would be inside `bam_hdr_rebuild`.

**Comment by daviesrob** [comment]
Added `bam_hdr_find_line_pos()`, which is useful for `samtools merge`.

All these `_pos()` functions take 1-based indexes.  What is the reason for that?  I'm tempted to change them to zero-based as it's annoying to have to put `+1` when they are used in a normal for-loop, or when accessing an `@SQ`line with a known `tid`.

**Comment by valeriuo** [comment]
It is just a matter of convention. I was thinking that people see the lines more as 1<sup>st</sup> `@SQ`, 2<sup>nd</sup> `@SQ`, ... and not like array indices. But if you think it makes more sense to change to a 0-based numeration, I have nothing against.

File: header.c [hunk 4]
```diff
@@ -564,6 +564,22 @@ static int bam_hrecs_remove_line(bam_hrecs_t *hrecs, const char *type_name, bam_
     return 0;
 }
 
+// Paste together a line from the parsed data structures
+static int build_header_line(const bam_hrec_type_t *ty, kstring_t *ks) {
+    bam_hrec_tag_t *tag;
+    int r = 0;
+    char c[2]= { ty->type >> 8, ty->type & 0xff };;
```
valeriuo: Extra semicolon.

header.c
```diff
@@ -1032,18 +1036,32 @@ int bam_hdr_find_line_id(bam_hdr_t *bh, const char *type,
...
+    if (build_header_line(ty, ks) < 0) {
+        return -2;
+    }
+
+    return 0;
+}
+
+int bam_hdr_find_line_pos(bam_hdr_t *bh, const char *type,
+                          int pos, kstring_t *ks) {
+        bam_hrecs_t *hrecs;
```
valeriuo: Different indentation.

header.c
```diff
@@ -1032,18 +1036,32 @@ int bam_hdr_find_line_id(bam_hdr_t *bh, const char *type,
...
+    if (!bh || !type)
```
valeriuo: We should check for null `type` in `bam_hdr_find_line_id:1026`.

jkbonfield:
While we're here, we should make bam_hdr_len and related functions return size_t (or ssize_t if they need negatives for errors).  See https://github.com/samtools/samtools/issues/1066 for reasons why.

This also means changing the l_text field, or at least having a separate copy that the new API is using so the bam_hdr_len function can return large headers.

sam.c
```diff
@@ -1097,13 +1102,13 @@ static bam_hdr_t *sam_hdr_sanitise(bam_hdr_t *h) {
     if (last != '\n') {
         hts_log_warning("Missing trailing newline on SAM header. Possibly truncated");
 
-        if (h->l_text == UINT32_MAX) {
-            hts_log_error("No room for extra newline");
-            bam_hdr_destroy(h);
-            return NULL;
-        }
+        if (h->l_text < 2 || i >= h->l_text - 2) {
+            if (h->l_text >= SIZE_MAX - 2) {
```
jkbonfield: SIZE_MAX again...

jkbonfield: Given this is a continual problem, maybe we should just fix the C++ brokenness.  Eg:

```
#ifndef SIZE_MAX
#  define SIZE_MAX ((size_t)-1)
#endif
```

Although signed overflow is undefined, promotion rules are different (I think!).  This is making assumption on 2s complement arithmetic, but that's an assumption we're safe with.

jmarshall:
It is great that HTSlib will finally have a header API :tada:

It is a little bit of a shame that the new functions are of the form `bam_hdr_xyz()` when HTSlib's convention for generic SAM/BAM/CRAM functions is a `sam_` prefix. (cf e.g. `bam_hdr_read()` which works for BAM files only vs. `sam_hdr_read()` which is for any of SAM/BAM/CRAM.)

Adding a new swathe of header API seems like an opportunity to make the canonical header struct be `sam_hdr_t` (with a compatibility `bam_hdr_t` typedef) and have all the new functions be called `sam_hdr_xyz()` and perhaps provide `sam_hdr_abc()`/`bam_hdr_abc()` names for existing `bam_hdr_` functions.

Was this `sam_hdr_` naming approach considered during development of this header API?

If they might not have time/interest/priorities to do it themselves, would the maintainers be amenable to pull requests making the new header API functions be named `sam_hdr_` etc?

daviesrob:
@jmarshall Yes it was, it changed in 7f876cf.  I don't particularly want to change it back at this point.

I think the `sam_hdr_` prefix has a few clashes with pre-existing bits of API, which made it not work quite so well.

jmarshall:
I can see why having successfully merged this after so many years of the saga, you would be less than keen to reopen it again. :smile:

But if the maintainers can bear to look at it for the sake of HTSlib's `sam_` convention, #887 is a proposal for doing this.

There were no troublesome clashes with pre-existing API bits and pieces; I suspect that problem may have been limited to this PR when it was in its intermediate state with various parts still up in the air.

PR #814, opened 2019-01-16
Define PATH_MAX as it is done in cram/open_trace_file.c

mr-c:
Define PATH_MAX as it is done in cram/open_trace_file.c if
not existent.  The definition should be removed from this C file since the
header file affects all three C files where this definition is used.

Co-authored-by: Svante Signell <svante.signell@gmail.com>

Originally from https://sources.debian.org/patches/htslib/1.9-9/define_PATH_MAX.patch/

### DIFF
--- a/cram/open_trace_file.c
+++ b/cram/open_trace_file.c
@@ -72,9 +72,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <sys/types.h>
 #include <sys/stat.h>
 #include "cram/os.h"
-#ifndef PATH_MAX
-#  define PATH_MAX 1024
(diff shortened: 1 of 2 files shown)


daviesrob:
It looks like this patch hasn't been needed since commit 5586168b.  While that commit uses `FILENAME_MAX`, which [glibc claims is a bad idea](https://www.gnu.org/software/libc/manual/html_node/Limits-for-Files.html) because it might be very large, it turns out that Debian GNU/hurd actually sets it to 1024.  I certainly managed to get the develop branch to work without needed any modifications.

This is possibly a bit neater, but it would need to revert the change in 5586168b as well.  And as mentioned in the message for that commit, the real solution would be to use dynamic allocations so we don't need PATH_MAX any more.

PR #817, opened 2019-01-22
Fix bgzip index creation when using multiple threads

### LINKED ISSUE #793: bgzip empty index with pipes
Hi, when using the following command with the htslib 1.9 suite, the resulting compressed FASTA gets created as well as the index file, but the latter is empty (actually 8 bytes).

    cat input.fna | bgzip -@ 20 -l 9 -i -I input.fna.gzi > input.fna.gz

I have to run a second pass, it works.

    bgzip -r input.fna.gz

I believe this is a bug and should be fixed. Piping is important when streaming large FASTA from the network, for instance.

Best,
Johannes

valeriuo:
Fixes #793

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -1119,10 +1119,29 @@ static void *bgzf_mt_writer(void *vp) {
     mtaux_t *mt = fp->mt;
     hts_tpool_result *r;
 
+    if (fp->idx_build_otf) {
+        fp->idx->moffs = fp->idx->noffs = 1;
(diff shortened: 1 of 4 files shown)


daviesrob:
Rebased with minor fixes to avoid data races.  I've also added a fix to `bgzip -b` which wasn't honouring the `-I` option and would segfault if reading from `stdin` because it didn't have a filename to add a suffix to; and some tests for basic `bgzip` functionality.

PR #819, opened 2019-01-25
Fix data race accessing fd->unsorted

daviesrob:
fd->unsorted needs to be protected by fd->ref_lock.

Not sure if this is the most optimal solution, but it does work.


jkbonfield:
Rebased and merged as 0a2cf81edde3b9c35cc82c29350af4c071eb2c39

PR #820, opened 2019-02-01
Prevent vcf_parse_format int overflow on huge input

mlin:
Upgrade variables in an internal data structure from int to size_t. The subsequent code multiplies/adds these with other values in order to calculate buffer sizes and positions. When we parse extremely large VCF sites (e.g. *N*>100k samples, *k*>100 alternate alleles, perhaps a O(*Nkk*) PL field for good measure) these operations were liable to overflow 32-bit int (>2GiB buffers) causing segfaults.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -2052,7 +2052,8 @@ uint8_t *bcf_fmt_sized_array(kstring_t *s, uint8_t *ptr)
  ********************/
 
 typedef struct {
-    int key, max_m, size, offset;
+    int key, max_m;
+    size_t size, offset;
     uint64_t is_gt:1, max_g:31, max_l:32;
     uint32_t y;
     uint8_t *buf;


daviesrob:
I don't think this is a complete solution:

1. It will still overflow on 32-bit platforms.
2. `max_m` is still an int, so it's possible that [`max_m << 2`](https://github.com/samtools/htslib/blob/4a519668/vcf.c#L2200) could sign-extend resulting in a big number that then overflows when multiplied.
3. `size` gets passed to `bcf_enc_size()` at some points (which sadly takes an `int`), so can't be bigger than `INT_MAX`.

I think a full fix for `vcf_parse_format()` would include upgrading counting variables `m`, `l` and `g` (possibly more) to `size_t` so they can't overflow; checking that the resulting `max_m` etc. are small enough that `size` can't get bigger than `INT_MAX`; checking that the product `v->n_sample * f->size` can't overflow and can be added to `mem->l`; and checking that [ks_resize](https://github.com/samtools/htslib/blob/4a519668/vcf.c#L2200) actually worked.  Other functions will likely need similar improvements as well.

valeriuo:
Fixed by https://github.com/samtools/htslib/pull/1104

PR #823, opened 2019-02-08
Make hts_hfile method visible outside htslib.

valeriuo:
Needed by https://github.com/samtools/samtools/pull/1007

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -978,7 +978,7 @@ const char *hts_format_file_extension(const htsFormat *format) {
     }
 }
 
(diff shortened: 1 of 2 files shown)


jmarshall:
Making this public is a very bad idea, as it will facilitate screwing up any buffering that the `htsFile` is doing. See https://github.com/samtools/htslib/pull/338#issuecomment-307844332 and the following comments.

As was the case back then, what you really want is a proper `hts_seek()` function.

PR #826, opened 2019-02-14
Fix leak of hts_itr_t in hts_itr_query

xunjieli:
When hts_itr_query() returns NULL, it should also free the hts_itr_t object if that object is previously allocated.


daviesrob:
Yes, there was some obvious leakiness there.
Thanks for the patch.

Issue #827, opened 2019-02-20
libhts with libdeflate: tabix compressed index not byte identical


mr-c:
This broke the pysam tests: https://github.com/pysam-developers/pysam/issues/761

mr-c:
I get the same result when I build htslib with libdeflate and GCS+S3 turned off. Next to see what happens without libdeflate (did some other dependency induce the change?)

mr-c:
Removing libdeflate restores the previous format, so it is indeed the culprit

mr-c:
Here's the diff from running `tabix` on https://github.com/pysam-developers/pysam/blob/master/tests/tabix_data/example.gtf.gz

```
6,10c6,10
< 00000050: 6042 1803 17bf c580 0ab8 90d8 1fa5 2074  `B............ t
< 00000060: cc3c 08fd 7219 84d6 dc06 a1fb b7a3 d27f  .<..r...........
< 00000070: 7640 6874 3731 23f9 0366 6fb4 1684 0e81  v@ht71#..fo.....
< 00000080: ba17 c667 6058 c6c7 80e4 5e98 7a98 b830  ...g`X....^.z..0
< 00000090: 032a e041 3613 4a3b 4950 8746 b809 0200  .*.A6.J;IP.F....
---
> 00000050: 6042 3207 267e 8b01 1570 21b1 3f4a 41e8  `B2.&~...p!.?JA.
> 00000060: 9879 10fa e532 08ad b90d 42f7 6f47 a5ff  .y...2....B.oG..
> 00000070: ec40 7523 cc0d cc48 fe80 d91b ad05 a143  .@u#...H.......C
> 00000080: a0ee 85f1 1918 96f1 21bb 17a6 1e26 2e8c  ........!....&..
> 00000090: e65e 1e64 33a1 b493 0475 6884 9b20 0000  .^.d3....uh.. ..
```

(top is from a `tabix` compiled with libdeflate, bottom is without libdeflate)

mr-c:
However, comparing the `gunzip`'d output confirms that the indexes themselves are the same. Maybe a note should be added to the docs about this?

mp15:
Using libdeflate changes the compressed file itself yes? Still same decompression algorithm but diff compression algorithm (similar to changing the level parameter). Ergo the sequences pointed to in the tabix index will have moved and thus doing a straight binary comparison of the two tabix files doesn’t tell us much. If you want to diff them you need something that understands the internals of tabix.

[...]

PR #828, opened 2019-02-20
note the effects of different compression libs

mr-c:


jmarshall:
This is not limited to libdeflate — different versions of system zlib (or different distro-specific patches to it) or running it on different platforms might well cause different compressed bytestreams too. How deterministic is zlib? Who knows. The conservative assumption is to assume it's not.

For most HTSlib users all this is irrelevant anyway. It's only people trying to compare compressed files e.g. in test suites who might need a small note about this.

mr-c:
@jmarshall Feel free to add to the branch https://github.com/mr-c/htslib/invitations

jkbonfield:
Even with the same deflate it's possible to alter the compression level programatically or eg via `samtools view -O bam,level=9`.  We've never stated the binary content matters and must be kept identical.

This is a storm in a teacup over a poorly conceived test.

jmarshall:
Good point about level settings, and agreed about teacups. There would be some scope for adding something like the following to the README:

```
Many of these file formats and associated index formats are compressed, 
which in this implementation involves the use of system or external
compression libraries such as zlib.  Hence the exact bytes in compressed
files output by HTSlib may vary across different compression libraries or 
with different compression level settings, so tasks requiring byte-for-byte 
equivalence across repeated runs should compare decompressed file contents.
```

However I am in two minds about whether this would be useful information for the reproducible science crowd or would end up pointlessly scaring the horses…

mr-c:
> teacups

FYI, not trying to make a storm, just trying to be helpful

> However I am in two minds about whether this would be useful information for the reproducible science crowd or would end up pointlessly scaring the horses…

+1 for a note from a reproducible science booster

@jmarshall I adopted your suggested wording and moved it to the end of the document

jmarshall:
> I adopted your suggested wording and moved it to the end of the document

If the benefits of a note like this outweigh the downsides, the place for it would be README rather than INSTALL — as the latter is less likely to be read by end users and might not even be installed by binary packages. For example, some outfit named Debian don't install INSTALL, though actually they don't install the README either ([#922860](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=922860))…

mr-c:
>  the place for it would be README rather than INSTALL

Done :-)

jkbonfield:
We felt that this is likely to not help many people (to be frank I expect most people wouldn't read it anyway) and may cause more confusion given there are several formats supported by htslib which have differing ways to uncompress and perform comparisons.  Plus there are potential differences even with filenames encoded in some situations, or `@PG` headers.

Rather than muddy the waters further we decided to just omit this.  Sorry for that, but thanks for your time.

mr-c:
> We've never stated the binary content matters and must be kept identical.

Can this be made explicit?

jkbonfield:
What problem are you trying to address here exactly?

If we look at, e.g. the gzip man page it doesn't state that the output will be different if you use -1 to -9 because it's obvious.  Similarly you're not going to find warnings in e.g. evince, about the impact of linking against a newer zlib (eg CloudFlare's or Intel's) on the bit-exactness of any PDFs saved.

Even zlib itself has on occasion tweaked its algorithm to improve compression ratios (eg compare output from 1.1.0 with 1.2.0), which means compressed files differ - obviously.

jmarshall:
> What problem are you trying to address here exactly?

Look back at the original commit:

> Note: turning on --with-libdeflate will cause any compressed output, such as Tabix indices, to have different bytes, though the underlying uncompressed data streams will be unchanged.

He's trying to prevent others from getting the same surprise that Debian got, when turning on libdeflate led to test failures in an unfortunately-written test suite. A laudable goal in itself, mired by: it's not just libdeflate / maybe it shouldn't be a surprise (or at best a “oh yeah; d'oh; move on”) / those in line to get this surprise are not the bulk of HTSlib users / any brief wording has the potential to be confusing.

jkbonfield:
I remembered it being something to do with a poorly written test, but I'm sure this was actually mentioned at some point here.  It seems to have vanished since then and that original commit doesn't mention testing either.  I assume there was a link to an external issue at one point which has since been edited away.

Regardless, adding to the README file isn't going to fix those tests.  Nor is it likely to stop people making the same mistake again either.  More documentation isn't always good, especially in a README file.  People stop reading once it gets too long, so excess test reduces the impact of the rest of the file.

Anyway, decision has been made.  No change.

jmarshall:
Nothing but agreement from me :smile:. It was #827 — contemporaneous, so the unstated link was obvious at the time.

PR #833, opened 2019-03-04
Fixed deadlock in thread pool / bgzf error handling.

### LINKED ISSUE #831: Deadlock in multithreaded bam write
We observe deadlocks in bam_write1() function when htsFile is in MT mode. The problem is observed consistently with htslib 1.5 and htslib 1.9 on ubuntu 18.04, compiled with gcc7, but is believed to apply to all pthread implementations. The deadlock is triggered by running out of disk space and presumably by a premature destroy of the conditional variable after a failed call to either hwrite() or hflush() in bgzf_mt_writer().

jkbonfield:
In the thread pool, we now set "no_more_input" flag in
hts_tpool_process_destroy as it was possible for deadlocks to occur
with another process hammering the input queue while attempting to
reset and flush prior to destroying the queue.  The shutdown flag
isn't quite sufficient here as that also changes the flush behavior.
We wish to flush correctly (on valid termination), but do not wish to
add more items to the input queue.

In the bgzf side, I also improved the error handling so that
fp->errcode isn't set within bgzf_mt_writer (no intervening lock so
that is a data race).  It now handles this properly with the join
retval.

Fixes #831

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -152,7 +152,7 @@ struct __bgzidx_t
 
 void bgzf_index_destroy(BGZF *fp);
 int bgzf_index_add_block(BGZF *fp);
-static void mt_destroy(mtaux_t *mt);
+static int mt_destroy(mtaux_t *mt);
 
 static inline void packInt16(uint8_t *buffer, uint16_t value)
 {
@@ -805,7 +805,8 @@ int bgzf_read_block(BGZF *fp)
                 if (fp->uncompressed_block == NULL) return -1;
                 fp->compressed_block = (char *)fp->uncompressed_block + BGZF_MAX_BLOCK_SIZE;
             } // else it's already allocated with malloc, maybe even in-use.
-            mt_destroy(fp->mt);
+            if (mt_destroy(fp->mt) < 0)
+                fp->errcode = BGZF_ERR_IO;
             fp->mt = NULL;
             hts_tpool_delete_result(r, 0);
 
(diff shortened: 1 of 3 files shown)


bgzf.c
```diff
@@ -1723,17 +1729,29 @@ ssize_t bgzf_raw_write(BGZF *fp, const void *data, size_t length)
     return ret;
 }
 
+// Helper function for tidying up fp->mt and setting errcode
+int bgzf_close_mt(BGZF *fp) {
```
jmarshall: Can `bgzf_close_mt` be static?

jkbonfield: Yes. Will do.

bgzf.c
```diff
@@ -805,7 +805,8 @@ int bgzf_read_block(BGZF *fp)
                 if (fp->uncompressed_block == NULL) return -1;
                 fp->compressed_block = (char *)fp->uncompressed_block + BGZF_MAX_BLOCK_SIZE;
             } // else it's already allocated with malloc, maybe even in-use.
-            mt_destroy(fp->mt);
+            if (mt_destroy(fp->mt) < 0)
+                fp->errcode = BGZF_ERR_IO;
```
jkbonfield: Hmm I'm wondering if this has the same missing mutex as the bgzf_mt_writer code's previous attempt at setting errcode.  I didn't validate this using libdeflate.  However it does seem to not produce whinges from -fsanitize=thread (unlike the original bgzf_mt_writer func).  So I'm assuming it's not in a situation that can trigger data races.

Issue #764, opened 2018-08-28
S3 buckets with dot in name


andrew-slater:
I'm able to run htsfile successfully on an s3 bucket without a dot in the name:
```
$ ./htsfile -h s3://omicsofttest1/10samp.10kvar.vcf.gz
##fileformat=VCFv4.2
##FILTER=<ID=PASS,Description="All filters passed">
```
But when the bucket has a dot in the name, it does not work:
```
$ ./htsfile -h s3://east.test.omicsoft/Andrew/10000samp.1mvar.vcf.gz
htsfile: can't open "s3://east.test.omicsoft/Andrew/10000samp.1mvar.vcf.gz": Input/output error
```
With more verbose, seems the problem is with the bucket name in the domain:
```
$ ./htsfile -h -v -v -v -v -v -v s3://east.test.omicsoft/Andrew/10000samp.1mvar.vcf.gz
[D::init_add_plugin] Loaded "knetfile"
[D::init_add_plugin] Loaded "mem"
[D::init_add_plugin] Loaded "/home/andrews/Temp/hgslib-1.9/install_dir/libexec/htslib/hfile_libcurl.so"
[M::hfile_gcs.init] version 1.9
[D::init_add_plugin] Loaded "/home/andrews/Temp/hgslib-1.9/install_dir/libexec/htslib/hfile_gcs.so"
[M::hfile_s3.init] version 1.9
[D::init_add_plugin] Loaded "/home/andrews/Temp/hgslib-1.9/install_dir/libexec/htslib/hfile_s3.so"
* About to connect() to east.test.omicsoft.s3.amazonaws.com port 443 (#0)
*   Trying 54.231.49.218...
* Connected to east.test.omicsoft.s3.amazonaws.com (54.231.49.218) port 443 (#0)
* Initializing NSS with certpath: sql:/etc/pki/nssdb
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* Server certificate:
*       subject: CN=*.s3.amazonaws.com,O=Amazon.com Inc.,L=Seattle,ST=Washington,C=US
*       start date: Sep 22 00:00:00 2017 GMT
*       expire date: Jan 03 12:00:00 2019 GMT
*       common name: *.s3.amazonaws.com
*       issuer: CN=DigiCert Baltimore CA-2 G2,OU=www.digicert.com,O=DigiCert Inc,C=US
* NSS error -12276 (SSL_ERROR_BAD_CERT_DOMAIN)
[...]

Issue #492, opened 2017-03-17
S3 access not working (as expected) when paths contain "#"


dkj:
Whilst I might expect to have to escape a `#` in an `http://` or `https://` URI, I don't for an `s3://` URI. The s3cmd tool does not expect such escaping.
```
$ samtools/1.4/bin/samtools view -H s3://npg_cloud_realign_wip/SC_WES_INT5823952/17626_5#1/17626_5#1.cram
[E::hts_open_format] fail to open file 's3://npg_cloud_realign_wip/SC_WES_INT5823952/17626_5#1/17626_5#1.cram'
samtools view: failed to open "s3://npg_cloud_realign_wip/SC_WES_INT5823952/17626_5#1/17626_5#1.cram" for reading: Permission denied
$ samtools/1.4/bin/samtools view -H s3://npg_cloud_realign_wip/SC_WES_INT5823952/17626_5%231/17626_5%231.cram | wc -l
3406

$ s3cmd info s3://npg_cloud_realign_wip/SC_WES_INT5823952/17626_5#1/17626_5#1.cram | grep size:
   File size: 3418351545
$ s3cmd info s3://npg_cloud_realign_wip/SC_WES_INT5823952/17626_5%231/17626_5%231.cram | grep size:
ERROR: S3 error: 404 (Not Found)
```

daviesrob:
I suspect s3cmd may contain code to do the escaping for you.  We will take a look and see if something similar can be done in htslib.

daviesrob:
Fixed by #839

Issue #477, opened 2017-02-27
writing to s3: urls not working


mcshane:
Based on comments from @daviesrob in samtools meeting.

daviesrob:
Fixed by #839

Issue #454, opened 2017-01-11
Support AWS Signature Version 4 authentication


jenniferliddle:
Amazon S3 uses two types of signature, called type 2 and type 4.
htslib currently only supports type 2. Later regions in S3 (for example, eu-west-2 (ie London)) only support type 4.
HTSlib needs to support type 4 authentication in order to be used with more recent regions on S3.

jmarshall:
Amazon says all regions (even the early ones) do support type 4, but I suppose there might be local installations that only support type 2.

I think it might be appropriate to use type 4 if you have a `region` (or `bucket_location` for _.s3cfg_) specified in your config file, or type 2 otherwise.

daviesrob:
Fixed by #839

PR #839, opened 2019-03-20
Update S3 code to use version 4 signatures.

### LINKED ISSUE #776: write to s3 buckets
I've got samtools/bcftools compiled to use bam/vcf files directly from s3 buckets. Is it possible to write back to S3 buckets as well or only to local disk? use case: read VCF from a public S3 bucket and deliver the filtered VCF to another S3 bucket (private). (the script itself does not run on a AWS instance, but on our  local university server)

whitwham:
This change will default to using version 4 signatures over the current version 2 ones.  It will also allow the writing of files to S3 repositories. Fixes #776, #764, #492, #477 and #454.

Documentation of new and old environmental variables and of general s3 htslib usage will be done in a separate pull request.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # Makefile for htslib, a C library for high-throughput sequencing data formats.
 #
-#    Copyright (C) 2013-2018 Genome Research Ltd.
+#    Copyright (C) 2013-2019 Genome Research Ltd.
 #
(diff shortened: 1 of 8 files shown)


daviesrob:
Squashed and rebased, with more use of virtual-hosted buckets given that Amazon want to deprecate the path-style ones.

Has been tested against Amazon and CEPH S3 gateway, both with and without a dot in the bucket name.

PR #843, opened 2019-03-24
Fixes #842, Support vcf hrec deletion and re-add

### LINKED ISSUE #842: bcf_hdr_remove leaves hash behind -- cannot re-add same ID
Preface: This came up when merging vcfs -- I need to remove exiting `##contig` lines and replace them with new `##contig` lines (potentially with new lengths, etc.) 

After removing a key, it can never be re-added.

Brief example:
```
#include <stdio.h>

#include <htslib/vcf.h>

int main(void)
{
    vcfFile *infile = bcf_open("/tmp/input.vcf", "r");
    vcfFile *outfile = bcf_open("/tmp/out.vcf", "w");

    bcf_hdr_t* infile_hdr = bcf_hdr_read(infile);
    bcf_hdr_t* outfile_hdr = bcf_hdr_dup(infile_hdr);

    bcf_hdr_remove(outfile_hdr, BCF_HL_CTG, "chr3");
    bcf_hdr_append(outfile_hdr, "##contig=<ID=chr333,length=333>");
    bcf_hdr_append(outfile_hdr, "##contig=<ID=chr3,length=123");

    bcf_hdr_write(outfile, outfile_hdr);

    bcf_close(infile);
    bcf_close(outfile);

    return 0;
}
```

In this example, chr3 must exist in input VCF; *it will never appear in output VCF* (while chr333 will).

I believe what is happening is that `bcf_hdr_remove` does not remove the key from the dictionary, (`kh_del`?) although it does destroy the hrec. 

https://github.com/samtools/htslib/blob/2cd99cf333938a9eb70393bc59379e1d6ea66b37/vcf.c#L876-L916

Thereafter, `bcf_hdr_register_hrec` sees the key as already existing, so returns 0 to caller `bcf_hdr_add_hrec`, which returns 0 to caller `bcf_hdr_append`.

https://github.com/samtools/htslib/blob/2cd99cf333938a9eb70393bc59379e1d6ea66b37/vcf.c#L497-L520

jblachly:
Simple fix -- delete the key representing the deleted hrec from the dictionary.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -887,6 +887,7 @@ void bcf_hdr_remove(bcf_hdr_t *hdr, int type, const char *key)
             vdict_t *d = type==BCF_HL_CTG ? (vdict_t*)hdr->dict[BCF_DT_CTG] : (vdict_t*)hdr->dict[BCF_DT_ID];
             khint_t k = kh_get(vdict, d, key);
             kh_val(d, k).hrec[type==BCF_HL_CTG?0:type] = NULL;
+            kh_del(vdict, d, k);
         }
         else
         {
--- a/vcf.c
+++ b/vcf.c
@@ -888,6 +888,7 @@ void bcf_hdr_remove(bcf_hdr_t *hdr, int type, const char *key)
             khint_t k = kh_get(vdict, d, key);
             kh_val(d, k).hrec[type==BCF_HL_CTG?0:type] = NULL;
             kh_del(vdict, d, k);
+            free((char*) kh_key(d,k));
         }
         else
         {


daviesrob:
This isn't looking quite so easy.  The patch causes memory leaks when running `test/test-vcf-api` (and also quite a few of the bcftools tests):
```
==6090==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 10 byte(s) in 3 object(s) allocated from:
    #0 0x43df20 in __interceptor_strdup llvm-7.0.0.src
/projects/compiler-rt/lib/asan/asan_interceptors.cc:447
    #1 0x5b3917 in bcf_hdr_register_hrec htslib/vcf.c:6
24
    #2 0x5b0cda in bcf_hdr_add_hrec htslib/vcf.c:672
    #3 0x5b5a21 in bcf_hdr_parse htslib/vcf.c:794
    #4 0x5e384d in bcf_hdr_dup htslib/vcf.c:3348
    #5 0x52f3a2 in bcf_to_vcf htslib/test/test-vcf-api.c:191
    #6 0x5317fe in main htslib/test/test-vcf-api.c:425
    #7 0x2b7f1bf467ec in __libc_start_main /build/eglibc-FTTGU2/eglibc-2.15/csu/libc-start.c:226

Direct leak of 7 byte(s) in 1 object(s) allocated from:
    #0 0x43df20 in __interceptor_strdup llvm-7.0.0.src/projects/compiler-rt/lib/asan/asan_interceptors.cc:447
    #1 0x5b1a3e in bcf_hdr_register_hrec htslib/vcf.c:514
    #2 0x5b0cda in bcf_hdr_add_hrec htslib/vcf.c:672
    #3 0x5b5a21 in bcf_hdr_parse htslib/vcf.c:794
    #4 0x5e384d in bcf_hdr_dup htslib/vcf.c:3348
    #5 0x52f3a2 in bcf_to_vcf htslib/test/test-vcf-api.c:191
    #6 0x5317fe in main htslib/test/test-vcf-api.c:425
    #7 0x2b7f1bf467ec in __libc_start_main /build/eglibc-FTTGU2/eglibc-2.15/csu/libc-start.c:226
```

I've tried freeing the keys in the obvious place, but that leads to use-after-free bugs when running the bcftools tests.  It looks like more work is needed to track down exactly what's going on here and how to do this safely.

I detected the problems by enabling address sanitizer when configuring both htslib and bcftools:

```
./configure CFLAGS='-g -fsanitize=address' LDFLAGS='-fsanitize=address'
```

pd3:
Can you try calling `bcf_hdr_sync()` before adding anything new?

jblachly:
I confirm that a call to `bcf_hdr_sync` after `bcf_hdr_remove` does not fix the problem.

jblachly:
Unfortunately ASAN apparently not supported on my Mac (CLANG docs say up to OS X 10.11, which is a few versions old now). I'll try to look at this on linux machine at work.
```
jsbmbp13:~/Documents/Development/github.com/samtools/htslib-jblachly/test$ export ASAN_OPTIONS=detect_leaks=1
jsbmbp13:~/Documents/Development/github.com/samtools/htslib-jblachly/test$ ./test-vcf-api
==36601==AddressSanitizer: detect_leaks is not supported on this platform.
Abort trap: 6
jsbmbp13:~/Documents/Development/github.com/samtools/htslib-jblachly/test$ uname -a
Darwin jsbmbp13.local 17.5.0 Darwin Kernel Version 17.5.0: Mon Mar  5 22:24:32 PST 2018; root:xnu-4570.51.1~1/RELEASE_X86_64 x86_64
```

As for the error itself, I am fairly confident that `bcf_hdr_remove` is mistaken in free'ing the hrec without also removing the key from vdict, although I haven't read the codebase well enough yet to understand the implications throughout of fixing this.  Thanks for your help

jblachly:
Still a simple fix :-)
 
This will do it:

```
diff --git a/vcf.c b/vcf.c
index 9ed2814..03f4d93 100644
--- a/vcf.c
+++ b/vcf.c
@@ -887,6 +887,8 @@ void bcf_hdr_remove(bcf_hdr_t *hdr, int type, const char *key)
             vdict_t *d = type==BCF_HL_CTG ? (vdict_t*)hdr->dict[BCF_DT_CTG] : (vdict_t*)hdr->dict[BCF_DT_ID];
             khint_t k = kh_get(vdict, d, key);
             kh_val(d, k).hrec[type==BCF_HL_CTG?0:type] = NULL;
+           kh_del(vdict, d, k);
+           free((char*) kh_key(d,k));
         }
         else
         {
```

You could also:
```
+            const char *str = kh_key(d, k);
+           kh_del(vdict, d, k);
+           free((char*)str);
```
if for some reason freeing key after bucket deletion is not safe, I haven't studied khash enough yet to know. 

If someone will sign off on this I'll update the pull request.

Best wishes

jblachly:
@pd3 @daviesrob To be clear, the above patch fixes the memory leaks detected by ASAN.
Thanks to @daviesrob for showing me how to configure and run ASAN-enabled htslib

whitwham:
Hello @jblachly.  I have been testing your commit and it breaks tests in bcftools. The problem is that the vdict entry is used for more than just header info.  In the bcftools tests the QS header is removed but the QS values are needed for later calculations (in mcall).  The QS entry in vdict is used to find those values, if it is removed then the lookup, and the program, fails.

@daviesrob and I have come up with an alternative solution that I will check in as soon as I have finished testing.

jblachly:
@whitwham I am  trying to understand these related projects so please do not take offense: it seems like an error to remove a tag from header then to rely on its value later?

Why is the QS tag removed in  `bcftools/vcfcall.c init_data:654`  ?

It seems like the most parsimonious fix for the htslib bug is to free the key and value rather than add additional layers of complexity. Thanks for helping me understand. Best wishes

pd3:
The bcftools code should create a copy of the header for writing and remove the QS tag from the copy. Currently it seems to incorrectly rely on the existence of the key in the header dictionary.

whitwham:
@jblachly No offence taken, bcftools is not an area I have worked on before.  As for the QS tag removal, I don't know the reasoning behind it or why it is done the way it is.  It seems to be using the QS values to decide what entries to output but does not put any of the QS values in that output.

pd3:
QS are temporary values used for (and only for) calling. They are generated by `bcftools mpileup` and are removed by `bcftools call`.

daviesrob:
`bcftools` is taking advantage of an implementation quirk which allows you to remove a header record but then continue to read the rest of the data as if the header record still exists.  Arguably this is not a good idea and `bcftools` should be fixed, but we also want older versions of `bcftools` to still work with the latest `htslib` so we can't lightly remove the quirky behaviour.

#847 fixes the problem you described in issue #842 by allowing `bcf_hdr_register_hrec()` to replace headers that have been removed.  It works when tested against the program you included in #842.  Could you give it a try and let us know if it works for you as well, please?

jblachly:
@daviesrob @whitwham Using patch in #847 and the small test vcf now demonstrates correct behavior (can re-add contig entries [did not test with other INFO types etc]).

In the end, I still vote for the two line fix (and attendant fix of bcftools as suggested by @pd3) on the basis of simplicity. As htslib is in the middle of a 2 to 3 transition where ABI breakages are expected this may be the opportune time to make the easy fix.  But, it is of course your call and I am grateful for all your work, and help, in all cases.

daviesrob:
Changing the ABI doesn't imply breaking backwards compatibility.  Unfortunately this solution causes two problems in `bcftools` - the disappearing `QS` tag in `bcftools call` and a use-after-free bug in `bcftools annotate`.  The second one happens because the pointer to the hash table key is also copied into the `bcf_hdr_t.id` field [here](https://github.com/samtools/htslib/blob/9af328a/vcf.c#L491).

#847 doesn't have these problems and solves your original problem, so I'll go for that one.  It may be possible to revisit this later if we manage to come up with a better fix.

daviesrob:
Closing, as #847 has now been merged.

PR #844, opened 2019-03-26
Set different integer types when decoding NM tag from CRAM

jfarek:
When the NM tag is decoded in `cram_decode_seq()`, the tag integer type is set to `I`, regardless of NM value. This commit allows setting the decoded NM tag integer type as `C`, `S` or `I`, depending on NM value, which mirrors how `sam_parse1()` encodes auxiliary tag integer values.


daviesrob:
Thanks.  It's only a small memory / file size saving, but I guess every little helps...

PR #846, opened 2019-03-27
Silence -Wstrict-prototypes warnings

jmarshall:
Not sure if these days you prefer trivia like this as a PR or committed straight to **develop**…


daviesrob:
I tend to prefer PRs these days - they're a bit more visible than a push to develop.

PR #850, opened 2019-04-05
Display an error message about the compression mismatch

### LINKED ISSUE #841: Uninformative error message from tabix
Tabix fails for gzip-compressed files with an uninformative error message
````
tabix -s1 -b2 -e2 test.gz 
tbx_index_build failed: test.gz
````
This could be addressed in `tbx_index_build3()` by returning a different error code or at least printing to stderr.

valeriuo:
Fixes #841


jmarshall:
This function is not just used by the **tabix** utility.

It would be better to return a distinct error code (either via return value or errno) and print a message after this is called in _tabix.c_. See also pysam-developers/pysam#760.

Issue #693, opened 2018-04-25
Duplicate symbols with libbwa


rob-p:
Hi,

  First, thank you all for htslib.  This is an amazing resource and incredibly useful library.  I have a project that includes uses both libbwa and libhts.  However both of these (linked statically) seem to include the same symbol `_ksprintf`.  The specific error is:

```
duplicate symbol _ksprintf in:
    ../../external/install/lib/libhts.a(kstring.o)
    ../../external/install/lib/libbwa.a(kstring.o)
ld: 1 duplicate symbol for architecture x86_64
```
any suggestions about how to avoid this?

Thanks!
Rob

daviesrob:
I'm having trouble getting this to happen, at least with my small test program.  How did you build htslib and bwa, and what are you trying to build that joins them together?

valeriuo:
Also, are you building from Xcode and do you have [-ObjC](https://developer.apple.com/library/content/qa/qa1490/_index.html) set in your linker flags?

rob-p:
Hi @daviesrob,

  Thanks for the quick reply.  I build htslib with a regular `./configure --prefix=<INSTALL_DIR> && make install`.  I am building these together for our project [Salmon](https://github.com/COMBINE-lab/salmon).  We use some internal functions of bwa, so we depend on `libbwa.a`.  Currently, we use Staden's `io_lib` for SAM and BAM parsing, but I would like to move to htslib, as it is a more canonical SAM/BAM parser, with more features, that is updated more frequently. The build process first builds `libbwa.a` then builds `libhts.a`, and then they are both linked together when building the Salmon binary.  This problem popped up only when attempting to switch to `htslib`, as `io_lib` does not include the `_ksprintf` symbols.

@valeriuo : I am building from the command line using CMake and Make (in this case with the default OSX compiler --- i.e. Apple Clang).  I don't have -ObjC set.   All of the code I'm building is either C or C++ (specifically C++11).

Thanks,
Rob

rob-p:
Hi all,

  I realized the awesome [SeqLib](https://github.com/walaj/SeqLib#integrating-into-build-system) library links both libbwa and libhts.  So, I dug into things a little bit, and realized they actually had to hack the kstring.c/kstring.h in bwa [as so](https://github.com/walaj/bwa/commit/c02766e3c34ac3f4af9842e20a54b7f9f4b36d0b).  The fact that you have to actually modify the source of one or the other library is not the most satisfying solution, but it does, of course, work in my case.

[...]

PR #851, opened 2019-04-09
Allow making KSORT functions static

jmarshall:
Consider the following program:

```c
#include "htslib/ksort.h"
#include "htslib/hts.h"

/* Compile and link with
        gcc -o foo foo.c libhts.a -lbz2 -llzma -lz
*/

typedef struct {
    int x;
} foo;

#define bar(a,b) ((a).x < (b).x)

KSORT_INIT_GENERIC(uint16_t)
KSORT_INIT(_off, foo, bar)
KSORT_INIT(_off_max, foo, bar)

int main() {
    hts_version();  // pull in hts.o
    errmod_cal(NULL, 0, 0, NULL, NULL);  // pull in errmod.o
}
```

This fails to link due to 24 duplicate `ks_*` functions, as the ksort usage in _errmod.c_ and _hts.c_ uses non-`hts_`-prefixed public namespace functions. It would be possible to make those HTSlib source files use ksort functions with names like `…_htslib_internal_…` which would not collide with sensible third-party code, but it's about as easy to add `static` support to _htslib/ksort.h_.

See also PacificBiosciences/pbmm2#11 (via [this tweet](https://twitter.com/jomarnz/status/1095272592582328321)); if static ksort had been available, this would have been a good way to fix this rather than renaming away from the collision.

### DIFF
--- a/errmod.c
+++ b/errmod.c
@@ -30,7 +30,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include "htslib/ksort.h"
 #include "htslib/hts_os.h" // for drand48
 
-KSORT_INIT_GENERIC(uint16_t)
+KSORT_INIT_STATIC_GENERIC(uint16_t)
 
 struct errmod_t {
     double depcorr;
--- a/hts.c
+++ b/hts.c
@@ -1282,8 +1282,8 @@ int hts_check_EOF(htsFile *fp)
 
 #define pair64_lt(a,b) ((a).u < (b).u)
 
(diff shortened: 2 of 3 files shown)


daviesrob:
Thanks.  All good for minimising our symbol leakage.  Which reminds me that we should also put in a fix for #693 as well (that one is about `kstring`).

jmarshall:
Cheers. Re that `kstring` one: I reckon HTSlib is the 800-pound gorilla here, and that one should be worked around inside libbwa.

PR #855, opened 2019-04-15
Add extra target for shared library

valeriuo:
Add extra target `hts.dll.a` for building htslib as a shared library. Needed for compiling bcftools plug-ins on Windows.

### DIFF
--- a/htslib.mk
+++ b/htslib.mk
@@ -152,7 +152,7 @@ $(HTSDIR)/config.h:
 $(HTSDIR)/libhts.a: $(HTSLIB_ALL)
 	+cd $(HTSDIR) && $(MAKE) lib-static
 
-$(HTSDIR)/libhts.so $(HTSDIR)/libhts.dylib: $(HTSLIB_ALL)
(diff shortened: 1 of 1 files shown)


daviesrob:
Cygwin builds make `libhts.dll.a`.  Should we include that to the list as well?

valeriuo:
I think Cygwin uses **libdl** (`dlopen`) when building its shared libraries, so there is no need for `libhts.dll.a`, but we could add it for symmetry.

jmarshall:
When I added the Cygwin build (2cddfb3c75887ca02d58708e5519c6d62992634c), making an import library appeared to be the standard practice on Cygwin, so HTSlib's makefile creates _libhts.dll.a_.

And one presumes that on these two platforms, third-party code links against these import libs rather than the _.dll_ itself — so these are the right targets to be adding to this line?

So yes it should be on the list; but please put these minor platforms at the end of the list rather than the start, i.e., so/dylib/cyg/windows, the same ordering as for the _libhts.xx_ rules in the main makefile.

PR #858, opened 2019-04-17
Fix builds on gcc older than 4.6

### LINKED ISSUE #857: htslib/sam.h update breaks compatibility with gcc 4.4.7
With one of the latest updates I cannot compile HTSlib with gcc 4.4.7 anymore. The error is the following:
```
$ gcc -Wall -g -O2 -I.  -c -o hts.o hts.c
In file included from ./cram/cram_samtools.h:61,
                 from cram/cram.h:45,
                 from hts.c:41:
./htslib/sam.h:63: error: redefinition of typedef ‘bam_hdr_t’
htslib/hts.h:209: note: previous declaration of ‘bam_hdr_t’ was here
```
The problem seems to be that in `htslib/hts.h` you have one definition of `bam_hdr_t`:
```
typedef struct bam_hdr_t bam_hdr_t;
```
While in `htslib/sam.h` you have a different definition of `bam_hdr_t`:
```
typedef struct bam_hdr_t {
    int32_t n_targets, ignore_sam_err;
    uint32_t l_text;
    uint32_t *target_len;
    int8_t *cigar_tab;
    char **target_name;
    char *text;
    void *sdict;
    uint32_t ref_count;
} bam_hdr_t;
```
I don't really understand the reason why older version of gcc don't see this as an issue. Is it possible to restructure the code to restore compatibility with gcc 4.4.7?

daviesrob:
4c2c536 added a typedef of bam_hdr_t in htslib/hts.h which made old versions of gcc complain about "redefinition of typedef 'bam_hdr_t'". [Versions of gcc after 4.6 don't mind.](https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=ce3765b)

Fix by referring to the struct instead of the typedef.

Fixes #857 (htslib/sam.h update breaks compatibility with gcc 4.4.7)

### DIFF
--- a/htslib/hts.h
+++ b/htslib/hts.h
@@ -42,6 +42,7 @@ extern "C" {
 typedef struct BGZF BGZF;
 #define HTS_BGZF_TYPEDEF
 #endif
+struct bam_hdr_t;
 struct cram_fd;
 struct hFILE;
 struct hts_tpool;
@@ -206,8 +207,6 @@ typedef struct htsFormat {
 struct __hts_idx_t;
 typedef struct __hts_idx_t hts_idx_t;
 
-typedef struct bam_hdr_t bam_hdr_t;
-
 // Maintainers note htsFile cannot be an opaque structure because some of its
 // fields are part of libhts.so's ABI (hence these fields must not be moved):
 //  - fp is used in the public sam_itr_next()/etc macros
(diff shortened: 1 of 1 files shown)


htslib/hts.h
```diff
@@ -206,7 +206,7 @@ typedef struct htsFormat {
 struct __hts_idx_t;
 typedef struct __hts_idx_t hts_idx_t;
 
-typedef struct bam_hdr_t bam_hdr_t;
+struct bam_hdr_t;
```
jmarshall: There's an (alphabetical) list of forward struct declarations further up (line 45ff) that this could be grouped with.

daviesrob:
I have grouped it...

PR #859, opened 2019-04-18
Make kvsprintf format checking platform configurable

valeriuo:
Helps to clear build warnings on MinGW.


jmarshall:
The klib _k*.h_ headers and source files are self-contained and mostly reflect their separate upstream editions at <https://github.com/attractivechaos/klib>. So rather than adding a `#include` of an HTSlib header, it would be better to add `#ifdef __MINGW_PRINTF_FORMAT`/etc to the code defining `KS_ATTR_PRINTF` here.

valeriuo:
Thank you! Amended.

jmarshall:
LGTM. Thanks for indulging the pedant :smile:

PR #860, opened 2019-04-18
 Add a strict C99 build to Travis

jmarshall:
Assuming that the intention is for HTSlib to stay C99-compatible, rather than quietly start using C11 functionality without necessarily noticing…

Add a `gcc -std=c99 -pedantic` Travis build, but first fix various C99 problems that have accumulated.

This would have caught #857.

The first two commits are mostly no-brainers. The third one (alter `cram_codec` to have a `union {…} u` member rather than an unnamed member) is annoyingly invasive, but hopefully @jkbonfield doesn't mind adding `u.` too much; sadly this commit is necessary to make `-std=c99 -pedantic -Werror` useable.

You can see the warnings (and an error) from a strict C99 Travis run on **develop** prior to #858 (and prior to the fixes in this PR) at [Travis build 36](https://travis-ci.org/jmarshall/htslib/builds/521685049), jobs 9 and 10.


daviesrob:
Thanks.  Merged with a minor update to remove some Sun Studio compiler pragmas that are no longer needed now that the union has a name.

PR #861, opened 2019-04-18
Support for non-standard index names

pd3:
This modifies the hts_open() and hts_idx_load() functions to allow
non-standard index file names.

The advantage of this approach compared to the samtools -X solution
    https://github.com/samtools/samtools/issues/485
    https://github.com/samtools/samtools/commit/651bf1469d1abd86a78b5e57c29b24a93598b0c3
is that it works transparently for all file types and all programs
without the need to add new options.

The disadvantage are possible conflicts, but the risk can be minimized by using an
unlikely delimiter. In this commmit the string ##idx## is used (see HTS_IDX_DELIM in
htslib/hts.h).

Example of usage:
````
    tabix file.vcf.gz##idx##crazy-index-file-name 1:10000060-10000060
````

After merging, this should address https://github.com/samtools/bcftools/issues/984

### CODE REVIEW

**Comment by daviesrob** [comment]
[Appveyor](https://ci.appveyor.com/project/samtools/htslib/builds/23949382?fullLog=true) wasn't happy with this:

```
test_index:
	C:/projects/htslib/tabix C:/msys64/tmp/DGi6DKZ10W/index.vcf.gz##idx##C:/msys64/tmp/DGi6DKZ10W/index.vcf.gz.tbi 1:10000060-10000060
bash: C:/projects/htslib/tabix: Bad address
.. failed ...
```

I'm guessing that MinGW's command-line handling didn't like the odd-looking filename.  It's possible quoting may fix it, but if not you might want to temporarily make it print out what it thinks the two filenames are to see what's going on.

File: hts.c [hunk 1]
```diff
@@ -421,6 +421,9 @@ htsFile *hts_open_format(const char *fn, const char *mode, const htsFormat *fmt)
     if (fmt && fmt->format != unknown_format)
         *mode_c = "\0g\0\0b\0c\0\0b\0g\0\0"[fmt->format];
 
+    char *fnidx = strstr(fn, HTS_IDX_DELIM);
+    if ( fnidx ) *fnidx = 0;
```
jmarshall: This is an invalid modification of a const string.

valeriuo:
A [potential fix](https://github.com/valeriuo/htslib/commit/b2aa72b245f0c3147085106293de943b32c0d52c) for the failing test.

daviesrob:
Squashed and merged as 7e583cce, removing an accidental change to `htslib.mk` and with a fix to stop `rmme` from possibly being used for error reporting when NULL.

I've also added 5c7ff9c6 which makes the error message from `hts_open()` a bit more useful, and passes the correct file name to `hts_idx_load2()` so the warning about stale indexes works properly.

PR #864, opened 2019-04-23
Fix configure --disable-werror (trivial m4 fix)

jmarshall:
AC_ARG_ENABLE's third argument, ACTION-IF-GIVEN, includes when the argument is given as --enable-werror=no / --disable-werror. The fourth argument continues the existing default of enable=no.


daviesrob:
Thanks for the fix.

PR #867, opened 2019-05-01
Fix undefined behavior

SoapGentoo:
## Fix undefined behavior in left-shift

UBSAN complains in

    bcf_sr_sort.c:188:14: runtime error: left shift of 1 by 31
    places cannot be represented in type 'int'

which involves signed overflow, and is hence undefined behavior.
The function `pairing_score` is only ever used in an unsigned
context, hence perform calculations using unsigned types, which
will not overflow/wrap.

## Fix unaligned access

UBSAN complains about unaligned access in

    cram/cram_encode.c:2638:20: runtime error: store to misaligned address
    0x619000000f9d for type 'uint16_t', which requires 2 byte alignment

which is valid, as the current code contains undefined behavior.
Using `memcpy` avoids this undefined behavior and every modern
C compiler nowadays turns `memcpy` calls into the most optimal
fetch instruction for the target platform.

See also
* https://blog.regehr.org/archives/1520
* https://pzemtsov.github.io/2016/11/06/bug-story-alignment-on-x86.html
* https://blog.quarkslab.com/unaligned-accesses-in-cc-what-why-and-solutions-to-do-it-properly.html

### DIFF
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -2632,10 +2632,14 @@ static int process_one_read(cram_fd *fd, cram_container *c,
 
         int l2 = cr->len / 2;
         unsigned char *from = (unsigned char *)bam_seq(b);
-        uint16_t *cpi = (uint16_t *)cp;
         cp[0] = 0;
         for (i = 0; i < l2; i++)
-            cpi[i] = le_int2(code2base[from[i]]);
+        {
+            // do not replace this memcpy by creating a misaligned
+            // pointer to cp using a uint16_t*, that is UB
+            const uint16_t cpi = le_int2(code2base[from[i]]);
+            memcpy(cp + i*sizeof(uint16_t), &cpi, sizeof(uint16_t));
+        }
         if ((i *= 2) < cr->len)
             cp[i] = seq_nt16_str[bam_seqi(bam_seq(b), i)];
     }
--- a/bcf_sr_sort.c
+++ b/bcf_sr_sort.c
@@ -147,7 +147,7 @@ static int multi_is_subset(var_t *avar, var_t *bvar)
     }
     return 0;
 }
-int32_t pairing_score(sr_sort_t *srt, int ivset, int jvset)
+uint32_t pairing_score(sr_sort_t *srt, int ivset, int jvset)
 {
     varset_t *iv = &srt->vset[ivset];
     varset_t *jv = &srt->vset[jvset];
@@ -185,7 +185,7 @@ int32_t pairing_score(sr_sort_t *srt, int ivset, int jvset)
     for (i=0; i<iv->nvar; i++) cnt += srt->var[iv->var[i]].nvcf;
     for (j=0; j<jv->nvar; j++) cnt += srt->var[jv->var[j]].nvcf;
 
-    return (1<<(28+min)) + cnt;
+    return (1u<<(28+min)) + cnt;
 }
 void remove_vset(sr_sort_t *srt, int jvset)
 {


cram/cram_encode.c
```diff
@@ -2632,10 +2632,14 @@ static int process_one_read(cram_fd *fd, cram_container *c,
...
-        uint16_t *cpi = (uint16_t *)cp;
         cp[0] = 0;
         for (i = 0; i < l2; i++)
-            cpi[i] = le_int2(code2base[from[i]]);
+        {
+            // do not replace this memcpy by creating a misaligned
+            // pointer to cp using a uint16_t*, that is UB
+            const uint16_t cpi = le_int2(code2base[from[i]]);
+            memcpy(cp + i*sizeof(uint16_t), &cpi, sizeof(uint16_t));
+        }
```
jkbonfield: This entire block of code is inside "HTS_ALLOW_UNALIGNED".

The whole purpose of the ifdef is to permit use of undefined behaviour as if it has been manually shown to behave in the way we wish on a specific platform (so far - it can always change with newer compilers!).

I'd need to check that it does actually work sensibly and isn't slowing things down on older compiler versions.

jkbonfield: I just did a check and the original version of particular loop only is around 10% faster on old gcc (4.6.3) and 23% faster on Intel cc (15.0.0).  Clang (3.7.0 to 7.0.0) showed no difference in performance, as did a more recent gcc (7.4.0).  Disabling HTS_ALLOW_UNALIGNED is around 3x slower, so clearly the idea is useful as a concept and the difference between memcpy vs unaligned copy is sometimes real, but not significant.

Benchmarking an entire uncompressed BAM to uncompressed CRAM shows the entire translation slows up by around 10% without the unaligned copies, but even with icc the difference between memcpy vs int16 copying is only around 1% so it looks to be a change we can live with.

(And obviously no effect whatsoever on modern gcc/clang.)

SoapGentoo: @jkbonfield I still think being 100% correct with C is important, as you can never forsee when future GCC or Clang releases might do some funky optimizations based on the type punning here.

jkbonfield: Agreed.  It's possible the memcpy could speed up other systems too.  The question really is therefore whether it is better to completely remove the HTS_ALLOW_UNALIGNED ifdefs and always use that route.

daviesrob: Actually, I think a shift-and-mask solution would be better here.  Nowadays it will usually optimise to the obvious load instruction, but if it doesn't the resulting code should be quicker than a call to `memcpy()`.  It's also endian-agnostic so the `le_int2()` part could be removed too.  Or it could be recoded to use the `le_to_u16()` function from `htslib/hts_endian.h` which will do the naughty cast if it can get away with it (i.e. on x86 or x86-64) or use shifting otherwise.

If playing with UBSAN, you should compile with '-DHTS_ALLOW_UNALIGNED=0' or you're likely to get false positives all over the place.

jkbonfield: Workable snippet of code that looks portable and is fast:

```
    // Convert seq 2 bases at a time for speed.
    static const char code2base[512] =
	"===A=C=M=G=R=S=V=T=W=Y=H=K=D=B=N"
	"A=AAACAMAGARASAVATAWAYAHAKADABAN"
	"C=CACCCMCGCRCSCVCTCWCYCHCKCDCBCN"
	"M=MAMCMMMGMRMSMVMTMWMYMHMKMDMBMN"
	"G=GAGCGMGGGRGSGVGTGWGYGHGKGDGBGN"
	"R=RARCRMRGRRRSRVRTRWRYRHRKRDRBRN"
	"S=SASCSMSGSRSSSVSTSWSYSHSKSDSBSN"
	"V=VAVCVMVGVRVSVVVTVWVYVHVKVDVBVN"
	"T=TATCTMTGTRTSTVTTTWTYTHTKTDTBTN"
	"W=WAWCWMWGWRWSWVWTWWWYWHWKWDWBWN"
	"Y=YAYCYMYGYRYSYVYTYWYYYHYKYDYBYN"
	"H=HAHCHMHGHRHSHVHTHWHYHHHKHDHBHN"
	"K=KAKCKMKGKRKSKVKTKWKYKHKKKDKBKN"
	"D=DADCDMDGDRDSDVDTDWDYDHDKDDDBDN"
	"B=BABCBMBGBRBSBVBTBWBYBHBKBDBBBN"
	"N=NANCNMNGNRNSNVNTNWNYNHNKNDNBNN";

    uint16_t *c2bi = (uint16_t *)code2base;
    for (i = 0; i < l2; i++)
	memcpy(&cp[i*2], &c2bi[from[i]], 2);
```

It needs the sneaky cast to 16bit in order to generate decent code with gcc, but with it it's basically the same as the original on modern gcc and clang, which is pretty much sufficient for our needs.  I don't know why the cast helps.  An explicit *2 in the indexing isn't as quick, but that's all the cast is doing.  It's not failing ubsan with those casts either.

I've no idea how this would work on things like ARM.  Hopefully it'd generate something sensible.  Anyway, I'll merge this in tomorrow with some tidying up.

SoapGentoo: The cast to `uint16_t*` is still kinda iffy, C99:

    6.3.2.3/7 A pointer to an object or incomplete type may be converted to a pointer
    to a different object or incomplete type. **If the resulting pointer is not correctly
    aligned for the pointed-to type, the behavior is undefined**. Otherwise, when
    converted back again, the result shall compare equal to the original pointer.
    When a pointer to an object is converted to a pointer to a character type, the
    result points to the lowest addressed byte of the object.

Why not just treat the `code2base` array as an array `uint16[256]`, which is guaranteed to be correctly aligned. The problem then only becomes the `cpi` pointer and its (likely) alignment violation in a `uint16_t`.

jkbonfield: It started off as a uint16_t, but I had complaints it wasn't clear. :/

daviesrob: [Godbolt](https://godbolt.org/) says the cast is unnecessary if you use:

```
memcpy(cp + i * 2, &code2base2[(size_t) from[i] * 2], 2);
```

This only matters with gcc.  clang knows that from[i] can be treated as unsigned, as it'll never be big enough to become negative.  It also unrolls the loop at higher optimisation.

jkbonfield: Older clangs and iccs (maybe modern, I only tried 15) don't optimise well with the size_t, but it's good enough.  This is over-optimisation really as all of it is massively faster than the naive nibble by nibble approach.  I'll just go with what's standard and ditch the ifdef.

jkbonfield:
Rebased and merged with some tidyup as 0241d07

PR #869, opened 2019-05-08
Reduce namespace pollution by making some private functions static

jmarshall:
These functions are used only within their respective source files and have no declarations in any header files, public or otherwise. Mark them as `static`, which they always should have been.

In particular, this makes static a number of private functions that have no `hts_internal_` or similar prefix and so pollute the caller's global namespace.

A second commit removes _cram/files.c_, moving the in-use subset of its functions to be `static` in the single file that uses them. If in future other source files want to use them, they should be moved to some toplevel utility file or the other source should write them out in place (as `cram_populate_ref()` already does).

Fortunately the soversion is being bumped anyway, so you won't have the Debian folks complaining that these vital not-declared-anywhere functions are having their symbols removed from _libhts.so_. :smile:


daviesrob:
Thanks.  I'm all for being tidy.

I do have a branch somewhere that adds `-fvisibility=hidden` when building the shared library and adds infrastructure to say exactly what should be exported.  I should really have a go at reviving so we can definitively say what is and is not in the external API.

jmarshall:
Thanks.

I forgot to mention (for the record) that nuking _cram/files.c_ removed fallback definitions for `S_ISDIR`/`S_ISREG` for ancient Concentrix, but if this was going to cause a problem then it would already have been a problem as `S_ISREG` is already used in _cram/cram_io.c_.

jkbonfield:
I'm pretty sure no one is going to be trying to build on that ancient system!  This is just a hang-over from the LMB days.  The Alliant was our main server pre-Alpha days, but had a rather broken OS in many ways.

I certainly had no objections to tidying up this legacy cruft, although there's a part of me that takes pride in how long ancient code can cling on to existence. :-)

PR #870, opened 2019-05-10
Improve index file search and load

valeriuo:
1. First search for a local index file with any of the possible extensions (`*.bam.csi`, `*.csi`, `*.bam.bai`, `*bai`).
2. If a file is found, load it or report an error.
3. If no index file is found and the alignment file is remote, try downloading it from the same server.
4. Before beginning the full download, check if the format is correct and only proceed if it is a known index format.
5. Load the downloaded file.

Fixes https://github.com/samtools/samtools/issues/1045

### DIFF
--- a/hfile.c
+++ b/hfile.c
@@ -689,7 +689,7 @@ static hFILE *hopen_fd_fileuri(const char *url, const char *mode)
 
 #if defined(_WIN32) || defined(__MSYS__)
     // For cases like C:/foo
-    if (url[0] == '/' && url[2] == ':' && url[3] == '/') url++;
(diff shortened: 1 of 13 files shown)


valeriuo:
Add method `hts_idx_stream` in order to create an index without downloading the remote index file.
Add methods `sam_index_load3` and `tbx_index_load3` to allow the user no to download the index file when performing an indexed query.

daviesrob:
I've added a commit to stop `vcf_hdr_read()` from complaining about missing indexes whenever it gets called.  Also some other minor fixes - file: URL handing; a memory leak that appeared in `idx_test_and_fetch()` that appeared when the `htsaddextension()` commit was merged.

PR #871, opened 2019-05-13
S3 multi region support

dbjohnson:
@daviesrob, this PR addresses your last comments from https://github.com/samtools/htslib/pull/680.  Direct S3 access is such an awesome feature, it's a shame not to be able to support multiple regions.

> As we're using s3cmd's configuration file, we should try to behave as much as possible like it does. In this case, it seems to use bucket_location so set the region that it tries to download from / upload to rather than region (which I don't think it understands).
> 
> I think this would be OK with a rebase to current develop and that one change.

Note that  [s3cmd usage](https://s3tools.org/usage) indicates either `bucket_location` or `region` are acceptable to specify the bucket region.  The only change I've made to @mwchristiansen's original PR is  to use the `bucket_location` key.  It's also rebased to match the current `develop` branch. 

Please let me know if there's anything else I can do to support getting this merged.  

Thanks!

### DIFF
--- a/hfile_s3.c
+++ b/hfile_s3.c
@@ -378,7 +378,7 @@ static hFILE * s3_rewrite(const char *s3url, const char *mode, va_list *argsp)
     if (ad->id.l == 0)
         parse_ini("~/.s3cfg", profile.s, "access_key", &ad->id,
                   "secret_key", &ad->secret, "access_token", &ad->token,
-                  "region", &ad->bucket_location || &ad->region,
+                  "bucket_location", &ad->region,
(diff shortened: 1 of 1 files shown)


daviesrob:
#839 has added this functionality, along with the ability to follow both 301 redirections (for path-style buckets) and extract the correct region out for 400 responses (for version 4 signatures with virtual-hosted buckets).  Hopefully that should have solved the problem for you.

dbjohnson:
awesome - thanks @daviesrob and @whitwham!

PR #876, opened 2019-05-31
bgzip '-F, --on-the-fly': writes an incremental index file as input is read.

circulosmeos:
Intended for possible long stdin inputs (e.g. logs), this allows
for quick random access reads of the still-growing compressed file.

I'm planning to use bgzip for compressed logging, so this feature is a must to exploit the advantages of gzipped blocks from bgzip.

I think I haven't break anything, as multithreading (*-@*) is excluded, and modifier (*-F, --on-the-fly*) must be explicitly set, but :
* **hseek()** is used (to overwrite the number of index entries so far, and later positioning at the end again)
* **bgzf_index_dump_hfile()** and **bgzf_index_dump()** are modified with an extra parameter (this could be omitted, but this way the index is only appended, not totally overwritten on each update).

Maybe an easier way is possible.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -1988,7 +1988,7 @@ static char * get_name_suffix(const char *bname, const char *suffix)
     return buff;
 }
 
-int bgzf_index_dump_hfile(BGZF *fp, struct hFILE *idx, const char *name)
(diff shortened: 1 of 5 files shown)


circulosmeos:
Ups, incorrect version, sorry: I'll make another pull request to be clean

PR #877, opened 2019-05-31
bgzip '-F, --on-the-fly': writes an incremental index file as input is read.

circulosmeos:
Intended for possible long stdin inputs (e.g. logs), this allows
for quick random access reads of the still-growing bgzip-compressed file.

I'm planning to use bgzip for compressed logging, so this feature is a must to exploit the advantages of gzipped blocks from bgzip.

I think I haven't broken anything, as multithreading (*-@*) is excluded, and modifier (*-F, --on-the-fly*) must be explicitly set, but :
* **hseek()** is used (to overwrite the number of index entries so far, and later positioning at the end again)
* **bgzf_index_dump_hfile()** and **bgzf_index_dump()** are modified with an extra parameter (this could be omitted, but this way the index is only appended, not totally overwritten on each update).

Maybe an easier way is possible.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -1981,14 +1981,15 @@ static inline int hwrite_uint64(uint64_t x, hFILE *f)
 
 static char * get_name_suffix(const char *bname, const char *suffix)
 {
-    size_t len = strlen(bname) + strlen(suffix) + 1;
+    size_t len = strlen(bname) + (suffix?strlen(suffix):0) + 1;
     char *buff = malloc(len);
     if (!buff) return NULL;
-    snprintf(buff, len, "%s%s", bname, suffix);
+    snprintf(buff, len, "%s%s", bname, (suffix?suffix:""));
     return buff;
 }
 
-int bgzf_index_dump_hfile(BGZF *fp, struct hFILE *idx, const char *name, int last_index_pointer)
+// ctx is not NULL only with `--on-the-fly` index generation
+int bgzf_index_dump_hfile(BGZF *fp, struct hFILE *idx, const char *name, BGZF_IDX_CONTEXT *ctx)
 {
     // Note that the index contains one extra record when indexing files opened
     // for reading. The terminating record is not present when opened for writing.
@@ -2013,7 +2014,7 @@ int bgzf_index_dump_hfile(BGZF *fp, struct hFILE *idx, const char *name, int las
     if (fp->mt && fp->idx)
         fp->idx->noffs--;
 
-    if ( fp->mt ) {
+    if ( fp->mt || ctx == NULL ) {
         if (hwrite_uint64(fp->idx->noffs - 1, idx) < 0) goto fail;
         for (i=1; i<fp->idx->noffs; i++)
         {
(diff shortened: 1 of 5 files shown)


jkbonfield:
I haven't investigated the source code yet, but if it periodically seeks back to overwrite the existing index then are there race conditions that mean a user of the index may get a partially written file?  I don't see any locking, but that's also only really useful if the application using the index also bothers to check the lock.

circulosmeos:
> I haven't investigated the source code yet, but if it periodically seeks back to overwrite the existing index then are there race conditions that mean a user of the index may get a partially written file? I don't see any locking, but that's also only really useful if the application using the index also bothers to check the lock.

As it is, the index is written incrementally, that means that previous *{ caddr, uaddr }* pairs are left as they are, and only new ones are appended to the file. Also the counter *fp->idx->noffs* of pairs is written at the beginning of the file. This all needs *hseek()* calls.

I understand your reasoning, but my intended use would be the casual use by a human operator who would need to access contents of the still growing bgzip-compressed log file. In this case any (im?)possible error would be mitigated by the fact that she can retry the operation in case of error: sysadmins are conscious of these difficulties with still opened files :-o
Also please note that the incremental index will be generated **only** if requested with *-F*: the usual behaviour remains untouched (if I'm correct).

I'm not sure, but introducing locking code seems a little complex for all this... Maybe a note in the man would suffice (btw I've just noticed that I haven't modified the man page). 
But I'm open to your suggestions.

PS: I'm planning also a bgztail Perl script who would be like "tail -f" for text files - this is not possible with gzip as it is a continuous stream. I suppose that wouldn't fit here, as it is a very specific use. And Perl, not C.

jkbonfield:
Agreed in principle, but I'm afraid it's likely to happen next release as we need to get the current one sorted asap.

circulosmeos:
thanks for your feedback!
> Agreed in principle, but I'm afraid it's likely to happen next release as we need to get the current one sorted asap.

no problem! this is provided just in case it is useful - as it is for me :-)

circulosmeos:
I've modified the pull request with better code: now the new parameter passed to `bgzf_index_dump_hfile()` and `bgzf_index_dump()` is a pointer to a context (struct) that is NULL for the case in which `--on-the-fly` is not used.

daviesrob:
I think a better way to do this might be to add the fields in `BGZF_IDX_CONTEXT` to the existing `struct __bgzidx_t` and provide a new function `bgzf_index_build_init2()` that allows you to set the index filename and enable the incremental write feature.  Then you could put the index update into `bgzf_index_add_block()` which would remove the need to call `bgzf_index_dump()`.  As a side benefit, it should work in multi-threaded mode too.  `bgzf_index_dump()` could also detect that an incremental write was going on and perform the finalisation step (and its API would not need to change).

circulosmeos:
Hi @daviesrob , your comments are wise: I don't have enough knowledge of this project to do these changes now... (specially the bgzf threaded part is a total unknown for me) but I'll try to think about all this. At first I tried to make minimum changes not to mess the whole project up :-)

circulosmeos:
As a related work, I've developed [`gztool`, a command line tool that creates little file indexes for gzip files](https://github.com/circulosmeos/gztool/), so they can be accessed at any random position without any time penalty (after the index is created).

I've tried to make the gzip index format "compatible" with bgzip indexes in that a gzip index file starts with a 0x0 uint64_t that in bgzip index format means that no bgzip-index-pairs follow, so (as they have the same extension `.gzi`) both formats could *live* together.

Anyway, actually `gztool` does not understand correctly bgzipped files. May be I'll try to add this feature in the future - though any pull request would be great! :-)

daviesrob:
Glad to see you have a solution.  I'm closing this now as presumably it's no longer needed - and we would most likely do this in a slightly different way, as described [in the comment above](https://github.com/samtools/htslib/pull/877#issuecomment-507233204).

circulosmeos:
Hope you will consider the on-the-fly functionality in the future: it reduces to zero the penalty for index creation among other benefits.
Thanks for your time!

PR #879, opened 2019-06-07
Add extra kstring functions

daviesrob:
KS_EMPTY     -  Initializer for structure assignment
  ks_init()    -  Initializer for pointed-to kstrings
  ks_expand()  -  Increase kstring capacity by a given amount
  ks_reset()   -  Reset kstring length to zero
  ks_free()    -  Free the underlying buffer
  ks_destroy() -  Free both the buffer and the pointed-to kstring

Brings together some kstring-related macros and functions that have been added in various other parts of the code-base.

### DIFF
--- a/htslib/kstring.h
+++ b/htslib/kstring.h
@@ -68,11 +68,11 @@
 /* kstring_t is a simple non-opaque type whose fields are likely to be
  * used directly by user code (but see also ks_str() and ks_len() below).
  * A kstring_t object is initialised by either of
- *       kstring_t str = { 0, 0, NULL };
- *       kstring_t str; ...; str.l = str.m = 0; str.s = NULL;
+ *       kstring_t str = KS_INITIALIZE;
+ *       kstring_t str; ...; ks_initialize(&str);
  * and either ownership of the underlying buffer should be given away before
  * the object disappears (see ks_release() below) or the kstring_t should be
- * destroyed with  free(str.s);  */
+ * destroyed with  ks_free(&str) or free(str.s) */
 #ifndef KSTRING_T
 #define KSTRING_T kstring_t
 typedef struct kstring_t {
@@ -118,6 +118,22 @@ extern "C" {
 }
 #endif
 
+/// kstring initializer for structure assignment
+#define KS_INITIALIZE { 0, 0, NULL }
+
+/// kstring initializer for pointers
+/**
+   @note Not to be used if the buffer has been allocated.  Use ks_release()
+   or ks_clear() instead.
+*/
+
+static inline void ks_initialize(kstring_t *s)
+{
+    s->l = s->m = 0;
+    s->s = NULL;
+}
+
+/// Resize a kstring to a given capacity
 static inline int ks_resize(kstring_t *s, size_t size)
 {
 	if (s->m < size) {
@@ -132,14 +148,48 @@ static inline int ks_resize(kstring_t *s, size_t size)
 	return 0;
 }
 
+/// Increase kstring capacity by a given number of bytes
+static inline int ks_expand(kstring_t *s, size_t expansion)
+{
+    size_t new_size = s->l + expansion;
+
+    if (new_size < s->l) // Overflow check
+        return -1;
+    return ks_resize(s, new_size);
+}
+
+/// Returns the kstring buffer
 static inline char *ks_str(kstring_t *s)
 {
 	return s->s;
 }
 
+/// Returns the kstring buffer, or an empty string if l == 0
+/**
+ * Unlike ks_str(), this function will never return NULL.  If the kstring is
+ * empty it will return a read-only empty string.  As the returned value
+ * may be read-only, the caller should not attempt to modify it.
+ */
+static inline const char *ks_c_str(kstring_t *s)
+{
+    return s->l && s->s ? s->s : "";
+}
+
 static inline size_t ks_len(kstring_t *s)
 {
 	return s->l;
+}
+
+/// Reset kstring length to zero
+/**
+   @return The kstring itself
+
+   Example use: kputsn(string, len, ks_clear(s))
+*/
+static inline kstring_t *ks_clear(kstring_t *s)
+{
+    s->l = 0;
+    return s;
 }
 
 // Give ownership of the underlying buffer away to something else (making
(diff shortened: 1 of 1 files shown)


jmarshall:
Are you planning to submit these upstream to [klib](https://github.com/attractivechaos/klib)? (Here is what happened last time `KS_EMPTY`/`ks_init()`-style facilities were suggested: attractivechaos/klib#21.)

* It would be good to update the comment above `#ifndef KSTRING_T` to show the new facilities.

* The name `ks_clear()` would be more familiar to people used to C++ strings than `ks_reset()`.

    Personally I'm not sure hiding `ks_reset` in the arguments of other kstring functions is a good idea; usage like the example would be clearer as
```c
  ks_clear(s);
  kputsn(string, len, s);
```

* Is `ks_free()` for use when the caller's kstring is going out of scope, or also for use when you want to restart using the kstring afresh? If the former, then `if (s) free(s->s);` suffices; if the latter, then `if (s) { free(s->s); ks_init(s); }` would be clearer (or maybe the caller having to write `ks_free(&mykstr); ks_init(&mykstr);` would be clearer still).

* The proposed `ks_init()` and `ks_destroy()` are not inverses of each other, which I think differs from the conventions of other klib/htslib `foo_init`/`foo_destroy` functions. (Yes, it's kind of a shame we've used `foo_init` everywhere rather than `foo_alloc`/`foo_new` or similar.)

* I have realised (since the `ks_resize` PR mentioned above) that the “give me the string value even if the kstring might be empty and never been malloced” function I really want is

```c
  static inline const char *ks_cstr(const kstring_t *s)
  {
        return (s->l > 0)? s->s : "";
  }
```

jkbonfield:
It's a good idea to tidy these up and at least offer better interfaces.

I think my personal preference would be to have KS_INIT (or KS_INITIALIZER) and ks_init for the initialisers.  Eg see the `pthread_mutex_lock` man page for examples of similar things elsewhere.  Using both empty and init words implies to me they're doing something different, which they're not.

Personally I don't see the point of `ks_destroy`.  We don't have a `ks_create` or `ks_alloc` that allocates the original structure, so it is assumed this is under the control of the calling code.  For symmetry, the destruction should be too.

@jmarshall I disagree regarding ks_free wanting a separate ks_init if you want to reuse it.  It's safe that ks_free as written here cannot be called twice and generate a double-free.  Clearing the values is a tiny amount of extra work in the situation where you didn't care.

daviesrob:
I don't really plan to push these upstream, but maybe given the kstream incompatibility mentioned in attractivechaos/klib#21 we should make them something like `KS_EMPTY_STRING` / `ks_empty_string()` or `KS_INITIALIZER` / `ks_initializer()` and remove the definition in `cram/sam_header.h`.

`ks_clear()` is probably better than `ks_reset()`.  I rather like the idea of putting it in the arguments of other functions, though, as otherwise you may as well  just write:
```c
s.l = 0;
kputsn(str, len, &s);
```
as is frequently done now.  There's nothing to stop it from being used on its own, anyway.

All the setting stuff to zero in `ks_free()` should be optimized out if not needed.  Given this it seems reasonable to provide the additional protection against accidental double free or use-after-free bugs.

I'll get rid of `ks_destroy()` and look into adding a `ks_c_str()`.

jmarshall:
> @jmarshall I disagree regarding ks_free wanting a separate ks_init if you want to reuse it. It's safe that ks_free as written here cannot be called twice and generate a double-free. Clearing the values is a tiny amount of extra work in the situation where you didn't care.

Okay, so you disagree with the “clearer still” bit and have an answer to the question (being the latter); so you might well still agree with the suggested clearer code for the latter (which is likely more optimisable).

> Edit: I see why you added ks_destroy now. kseq.h uses it, but it also defines ks_init to do the memory allocation too which you are not. The two need to be symmetric with each other.

As I had forgotten and @daviesrob has realised, _kseq.h_ means functions with those two particular names are a non-starter anyway.

Yes, fine to have the option to compose `ks_clear()` into the arguments of other functions or not. Thanks for looking at the `ks_c_str()` thing.

jkbonfield:
I was assuming we'd ditch the cram sam_header.c definitions in a separate commit once this lands anyway (or do it as a second commit in this PR itself).  Similarly for ksinit and ksfree in hfile_s3.c.  So we don't need to worry about our other internal-only functions clashing on names.

jkbonfield:
>     `if (s) { free(s->s); ks_init(s); }` would be clearer (or maybe the caller having to write `ks_free(&mykstr); ks_init(&mykstr);` would be clearer still).

I didn't twig you were referencing to replacement code for the ks_free function and took it as a question over why the caller couldn't just do those two inline calls because, well, that's the way icky way it's traditionally been done!

So for clarity, yes your suggestion for the implementation of ks_free is clearer than calling `ks_release`.

daviesrob:
Updated.  On discussion we decided to go for `KS_INITIALIZE` / `ks_initialize()`.

htslib/kstring.h
```diff
@@ -132,16 +148,50 @@ static inline int ks_resize(kstring_t *s, size_t size)
...
 static inline size_t ks_len(kstring_t *s)
 {
 	return s->l;
 }
 
+/// Reset kstring length to zero
+/**
+   @return The kstring itself
+
+   Example use: kputsn(string, len, ks_reset(s))
```
jmarshall: s/reset/clear/ here too (and there is an excess space in `kstring_t * ks_clear` below).

htslib/kstring.h
```diff
@@ -68,11 +68,11 @@
...
  * used directly by user code (but see also ks_str() and ks_len() below).
  * A kstring_t object is initialised by either of
- *       kstring_t str = { 0, 0, NULL };
- *       kstring_t str; ...; str.l = str.m = 0; str.s = NULL;
+ *       kstring_t str = KS_INITIALIZE;
+ *       kstring_t str; ...; ks_initialize(&str);
  * and either ownership of the underlying buffer should be given away before
  * the object disappears (see ks_release() below) or the kstring_t should be
- * destroyed with  free(str.s);  */
+ * destroyed with  ks_free(&str);  */
```
jmarshall: Might be fair to bow to legacy reality with

    destroyed with  ks_free(&str)  or  free(str.s)

htslib/kstring.h
```diff
@@ -154,6 +204,13 @@ static inline char *ks_release(kstring_t *s)
 	return ss;
 }
 
+/// Safely free the underlying buffer in a kstring.
+static inline void ks_free(kstring_t *s)
+{
+    if (s) free(s->s);
```
jmarshall: This function is now in two minds whether `s!=NULL` is a precondition. Either nuke `if (s)` or put the `ks_initialize` inside the if too.

daviesrob: Should be fixed now.

jmarshall: Thanks :+1:

daviesrob:
Squashed for easier merge.

PR #880, opened 2019-06-07
Auto index

jkbonfield:
3 things, 2 small 1 not.

1. Added address sanitizer to travis config.  It appears to be an oversight that we only did this for samtools and not here.

2. Made hts_open_format behave as expected; where the `fmt` argument takes precedence over mode string.

3. Fixed BGZF auto-indexing so it works when running with multiple encoder threads.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -105,6 +105,18 @@ enum mtaux_cmd {
     CLOSE,
 };
 
+// When multi-threaded bgzf_tell won't work, so we delay the hts_idx_push
+// until we've written the last block.
+typedef struct {
+    int tid, beg, end, is_mapped;  // args for hts_idx_push
+    uint64_t offset, block_number;
+} hts_idx_cache_entry;
+
+typedef struct {
+    int nentries, mentries; // used and allocated
+    hts_idx_cache_entry *e; // hts_idx elements
+} hts_idx_cache_t;
+
 typedef struct bgzf_mtaux_t {
     // Memory pool for bgzf_job structs, to avoid many malloc/free
     pool_alloc_t *job_pool;
@@ -133,6 +145,12 @@ typedef struct bgzf_mtaux_t {
     pthread_mutex_t command_m; // Set whenever fp is being updated
     pthread_cond_t command_c;
     enum mtaux_cmd command;
+
+    // For multi-threaded on-the-fly indexing. See bgzf_idx_push below.
+    pthread_mutex_t idx_m;
+    hts_idx_t *hts_idx;
+    uint64_t block_number, block_written;
+    hts_idx_cache_t idx_cache;
 } mtaux_t;
 #endif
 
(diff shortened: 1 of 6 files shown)


bgzf.c
```diff
@@ -1172,9 +1268,18 @@ static void *bgzf_mt_writer(void *vp) {
             fp->idx->offs[ fp->idx->noffs-1 ].caddr = fp->idx->offs[ fp->idx->noffs-2 ].caddr + j->comp_len;
         }
 
+        // Flush any cached hts_idx_push calls
+        bgzf_idx_flush(fp);
```
daviesrob: Lack of error handling here...

bgzf.c
```diff
@@ -150,6 +168,84 @@ struct __bgzidx_t
...
+        mt->block_written++;
+        return 0;
+    }
+
+    pthread_mutex_lock(&mt->idx_m);
+
+    hts_idx_cache_entry *e = mt->idx_cache.e;
+    int i;
+
+    for (i = 0; i < mt->idx_cache.nentries && e[i].block_number == mt->block_written; i++) {
```
daviesrob: Is it ever possible that this could get out of synch?  I.e. `e[i].block_number < mt->block_written` which would result in the data staying queued forever...

I'm not sure if it could, but it might be an idea to put in an assert so we get to find out about it if it ever happens.

hts.c
```diff
@@ -424,6 +424,18 @@ htsFile *hts_open_format(const char *fn, const char *mode, const htsFormat *fmt)
         *mode_c = format_to_mode[fmt->format];
     }
 
+    // If we really asked for a compressed text format but didn't
+    // specify 'z' format,  add it anyway.
+    if (strchr(mode, 'w') && fmt && fmt->compression == bgzf) {
+        if (fmt->format == sam || fmt->format == vcf || fmt->format == text_format) {
+            char *cp;
+            if ((cp = strchr(smode, 'b')))
```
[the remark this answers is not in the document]
jkbonfield: Hmm, staring at the contorted logic a bit longer I think you're right.  It's not obvious at first glance though. :-)

Thanks.

bgzf.c
jkbonfield: In theory it can't happen as both producer and consumer are incrementing their own variables.  An assert isn't a bad idea though just in case something whacky happens that I didn't think about.

hts.c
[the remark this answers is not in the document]
jkbonfield: Unsure whether to also add in tests above:

```
	else if (*cp == 'z')
            fmt_code = 'z';
```

That would mean hts_open_format with mode "wz" an fmt as SAM with no_compression set wouldn't write compressed data.  However I'm unsure if we care.  It follows the logic of the format taking precedence I guess so it's done, but I can see maybe people wanting an explicit compression in the mode string to be honoured for ease of coding (why add it otherwise)?

jkbonfield:
Updated and fixed a long standing bug with ultra-long format strings.  Not that we'd ever use them, but the code claims to handle 100 chars but didn't quite.

htslib/bgzf.h
```diff
@@ -416,6 +417,16 @@ typedef struct BGZF BGZF;
...
 
+
+    /*
+     * Buffers up arguments to hts_idx_push for later use, once we've written all bar
+     * this block.  This is necessary when multiple blocks are in flight (threading).
+     *
+     * Returns 0 on success,
+     *        -1 on failure
+     */
+    int bgzf_idx_push(BGZF *fp, hts_idx_t *hidx, int tid, int beg, int end, uint64_t offset, int is_mapped);
```
jmarshall: Is this function for public use? If not, it could go in _hts_internal.h_ or so.

jkbonfield: Good point.  Thanks.

hts.c
```diff
@@ -438,6 +444,13 @@ htsFile *hts_open_format(const char *fn, const char *mode, const htsFormat *fmt)
     fp = hts_hopen(hfile, fn, smode);
     if (fp == NULL) goto error;
 
+    // htsExactFormat goes from a nice exact input format (bam, sam)
+    // to a woefully inexact one (binary, text).  Honour the original one.
+    if (fp->is_write && fmt &&
+        (fmt->format == bam || fmt->format == sam ||
+         fmt->format == vcf || fmt->format == bcf))
+        fp->format.format = fmt->format;
+
```
jmarshall: >  // htsExactFormat goes from a nice exact input format (bam, sam)
>  // to a woefully inexact one (binary, text).  Honour the original one.

It'd be great to do this without the editorialising. The reason for `binary_format`/`text_format` has been explained before, e.g., in https://github.com/samtools/htslib/pull/721#issuecomment-400666150.

What's really going on here is the exactness is lost because `smode` (given to `hts_hopen()`) is an approximation of the format. Ideally this would be solved with some refactoring so no approximating was necessary; but more simple would be to track the real exactFormat on line 424 above (alongside `*mode_c = format_to_mode[fmt->format]`) and set `fp->format.format` accordingly here `if (fp->is_write && that_code_path_was_taken)`.

That would avoid this test for `bam`/`sam`/`vcf`/`bcf` and so make this work for e.g. `bed` as well.

jkbonfield: I'll rewrite the comment and add bed, although I don't think it's editorial - it's 100% factual.  We can't really get rid of the rather fragile set of formats until we fix other things, so this is a pragmatic workaround for now.  I deliberately avoided copying over format for everything beyond `unknown_format` because I confess I didn't really understand the implications of types on things like json or the index formats.  It's rather twisty logic in places, so played it safe with the bits I can actually test don't cause breakage.

If you think it's safe we could get rid of the whole format checking bit and just use `&& fmt->format > unknown_format`, but I don't think our test harness would cover a lot of it as it doesn't have web stuff in there.

Regarding PR #721, the last we heard was you were going to rework it.  Did that get shelved?  It'd definitely be great to have this tidied up.

I also see we stated "the release candidate is meant to be today" in June 2018 and the new stuff would arrive asap afterwards. Um... no comment!

jmarshall: “nice” and “woefully” are hardly 100% factual.

Yes, PR #721 is back on my to-do list to bring up to date during the current apparent HTSlib frenzy :smile:.

IIRC the web stuff only occurs during reading, so fixups that occur for is_write are fine for web stuff. Point is: for writing, `fp->format.format` should end up being either exactly the same as or compatible with the `fmt->format` argument, otherwise there is a bug in the `smode` shoehorning or in `hts_hopen()`. So if line 424 `*mode_c = …` was triggered and we're writing, then it is safe to set `fp->format.format` to the same thing that `*mode_c` was based on, i.e., the input `fmt->format` argument.

I think a big part of the problem here is that it's unclear how `hts_open_format`'s `mode` and `fmt` arguments are supposed to overrule each other. (The documentation just says TODO…)

Most sensible might be to have `hts_open_format(fn, mode, NULL)` call `hts_open(fn, mode)` effectively; otherwise when `fmt` is non-NULL `mode` is just examined for `r` or `w` and otherwise ignored. But I haven't worked that through…

htslib/bgzf.h
```diff
@@ -36,6 +36,7 @@
 #include <sys/types.h>
 
 #include "hts_defs.h"
+#include "hts.h"
```
jmarshall: Also remove the new `#include` in _bgzf.h_.

jkbonfield:
Removed the extra include from bgzf.h, squashed and rebased off develop.

jkbonfield:
Argh, will fix.  One of those days...

daviesrob:
Some fixes to commit 1 ended up in commit 2 , to the point where none of commit 1 was left.  I've pushed them back to where they make more sense.

PR #881, opened 2019-06-14
Put hts_ prefix in front of various klib functions

### LINKED ISSUE #693: Duplicate symbols with libbwa

daviesrob:
Put hts_ prefix in front of various klib functions and add wrappers to preserve the existing API.  This is so we don't get clashes with the Attarctive Chaos versions of these functions, in case anyone links against both.

Fixes #693 (Duplicate symbols with libbwa)

N.B. Breaks ABI, so `TWO_TO_THREE_TRANSITION_COUNT` needs to be incremented on merge (which should not be done as a fast-forward).

### DIFF
--- a/htslib/kfunc.h
+++ b/htslib/kfunc.h
@@ -35,13 +35,19 @@ extern "C" {
  * \log{\Gamma(z)}
  * AS245, 2nd algorithm, http://lib.stat.cmu.edu/apstat/245
  */
-double kf_lgamma(double z);
+static inline double kf_lgamma(double z) {
+    double hts_kf_lgamma(double z);
+    return hts_kf_lgamma(z);
+}
 
 /* complementary error function
  * \frac{2}{\sqrt{\pi}} \int_x^{\infty} e^{-t^2} dt
  * AS66, 2nd algorithm, http://lib.stat.cmu.edu/apstat/66
  */
-double kf_erfc(double x);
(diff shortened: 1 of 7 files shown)


jmarshall:
IMHO this complexity is a mistake. HTSlib is the 800-pound gorilla here.

#693 could be very easily worked around in bwa, as suggested in https://github.com/samtools/htslib/issues/693#issuecomment-482032058.

People using the Attractive Chaos versions of gamma functions etc who add HTSlib usage to their project can be expected to use the HTSlib-supplied versions of such functions and remove their own copied-in klib headers.

And rather than contemplate such changes  in _knetfile.[ch]_, perhaps it is time to remove them from HTSlib (as services are said to be moving to HTTPS, which knetfile does not support, and knetfile has long been superseded by hfile_libcurl in HTSlib).

PR #884, opened 2019-06-18
Make sam_open_mode() work with ##idx## index delimiter

daviesrob:
Also sam_open_mode_opts().

Makes explicitly naming an index compatible with choosing the output format by extension.


valeriuo:
Merged as f4cabfa4ca192b0cea8d8842ae26cddd5d1fa4e2

PR #887, opened 2019-06-26
Proposal: Use sam_hdr_ prefix for new header API functions

jmarshall:
With the exception of some early functions, HTSlib's convention for generic SAM/BAM/CRAM functions is a `sam_` prefix. (cf e.g. `bam_hdr_read()` which works for BAM files only vs. `sam_hdr_read()` which is for any of SAM/BAM/CRAM.)

For your consideration, this PR renames the newly added public header API functions to use a `sam_hdr_` prefix rather than the `bam_hdr_` prefix of PR #812. If the maintainers' consensus is that they would like to go down this route, I would be happy to add commits to rename the new internal functions similarly and/or to add renaming shims for e.g. `sam_hdr_init()` and `sam_hdr_t`, i.e., things that have long had a somewhat unfortunate `bam_hdr_` prefix.

I have also tried out the renaming on samtools/samtools#998; that compiles cleanly, and both htslib and samtools test suites pass.

(In https://github.com/samtools/htslib/pull/812#issuecomment-504064969 Rob alluded to clashes with pre-existing bits of API. I think that problem must have been mostly limited to an intermediate version of the API. Once cram.h's `sam_hdr_length`/`_str` declarations were tidied up, this PR had no such problems.)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -48,8 +48,8 @@ Noteworthy changes in release a.b
     to size_t, to allow for very large headers in SAM files.  The text
     and l_text fields have been left for backwards compatibility, but
     should not be accessed directly in code that uses the new header API.
-    To access the header text, the new functions bam_hdr_length() and
-    bam_hdr_str() should be used instead.
+    To access the header text, the new functions sam_hdr_length() and
+    sam_hdr_str() should be used instead.
 
   - bcf_index_load() no longer tries the '.tbi' suffix when looking for
     BCF index files (.tbi indexes are for text files, not binary BCF).
--- a/NEWS
+++ b/NEWS
@@ -54,9 +54,14 @@ Noteworthy changes in release a.b
   - bcf_index_load() no longer tries the '.tbi' suffix when looking for
     BCF index files (.tbi indexes are for text files, not binary BCF).
 
-* A new header API has been added to HTSlib, allowing header data to
-  be updated without having to parse or rewrite large parts of the header
-  text.  See htslib/sam.h for function definitions and documentation.
+* A new SAM/BAM/CRAM header API has been added to HTSlib, allowing header
+  data to be updated without having to parse or rewrite large parts of the
+  header text.  See htslib/sam.h for function definitions and documentation.
+
+  The header typedef and several pre-existing functions have been renamed
+  to have a sam_hdr_ prefix: sam_hdr_t, sam_hdr_init(), sam_hdr_destroy(),
+  and sam_hdr_dup().  (The existing bam_hdr_-prefixed names are still
+  provided for compatibility with existing code.)
 
 * SAM format reading and writing is now much faster.  Compared to revision 1.9,
   reading uncompressed SAM may be up to 100% faster; writing up to 80%.  The
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -1221,7 +1221,7 @@ static int cram_decode_seq(cram_fd *fd, cram_container *c, cram_slice *s,
     int decode_md = s->decode_md && s->ref && !has_MD && cr->ref_id >= 0;
     int decode_nm = s->decode_md && s->ref && !has_NM && cr->ref_id >= 0;
     uint32_t ds = s->data_series;
-    bam_hrecs_t *bfd = sh->hrecs;
+    sam_hrecs_t *bfd = sh->hrecs;
 
     if ((ds & CRAM_QS) && !(cf & CRAM_FLAG_PRESERVE_QUAL_SCORES)) {
         memset(qual, 255, cr->len);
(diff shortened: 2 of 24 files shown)


jkbonfield:
Anything that avoids the nonsense of schizophrenic name changes like this would be appreciated:

```
bam_hdr_t *h;
...
h = sam_hdr_read(fp);
...
bam_hdr_destroy(h);
```

For frell's sake - allocate with sam and destroy with bam.  Ugh.

I think the problem is the structure is bam_hdr_t which isn't really correct.  Our functions should reflect the structure, and the structure is a SAM header (the textual format which is shared between SAM, BAM and CRAM) and not labelling it as a BAM specific structure name.

Unfortunately we also have bam1_t for the bam structure, but that - more or less - actually is BAM format in memory.

The original API (from io_lib and later htslib/cram) had `sam` prefixes and it's still my preferred naming.  Our commit didn't mention what clashes we were attempting to resolve, but it's possible it was linking old samtools vs modern htslib.  (We know the new one will link.)  Did you test that combination?  Maybe Rob can remember the reasoning.  I alas cannot.

jmarshall:
> it's possible it was linking old samtools vs modern htslib. (We know the new one will link.) Did you test that combination?

Yep, that compiled without warnings and passed its tests. (I half-expected samtools to define its own `sam_hdr_foo` function that would conflict, but there was no such problem. Also no bcftools problems.)

daviesrob:
I think I would like to see the structure renamed `sam_hdr_t` (with typedef for backwards compatibility) and the remaining `bam_hdr_` functions get `sam_hdr_` equivalents.  If we want to go down this route we should try to end up with something fully consistent.

I think the name clashing may well have been resolved when the parsed headers struct got renamed to `bam_hrecs_t` and lots of internal functions got renamed to start `bam_hrecs_` instead of `sam_hdr_`.  A good spot on the `struct sam_hdr`, though, which is a bit I obviously missed when doing this and definitely wrong.

jmarshall:
Thanks Rob. That sounds like a likely theory about the intermediate name clashing.

Agreed re fully consistent. I'm happy to add to this PR to rename the struct to `sam_hdr_t` and similarly the remaining `bam_hdr_` functions, with compatibility typedefs and static inline wrappers. I'd be able to get that done over the weekend.

daviesrob:
Yes, that would be fine, thanks.

daviesrob:
Any progress on this yet?

jmarshall:
Yep, sorry about the delay: day job got busy, and there's a couple of internal functions to disentangle. Should be ready tomorrow.

In the meantime, for mulling on, here are a couple of minor new-API problems I've noticed so far:

* `bam_hdr_length()` returns `size_t` but has a -1 on error return code — should this be `ssize_t`?

* `bam_hdr_remove_lines()` has a `void *` parameter, but doesn't describe what sort of a khash you're supposed to give it. If anybody's going to be able to use this function, it needs a better parameter type (probably not possible) or some example code in the function's docs.

daviesrob:
On `bam_hdr_length()` - yes it should be `ssize_t` although that does introduce some trickyness the the header `l_text` being `size_t`.  It's therefore possible to overflow it (at least on 32-bit platforms) although I suspect if the header gets that big other problems are likely to show up pretty quickly.

On `bam_hdr_remove_lines()` - yes, I was thinking of making a typedef, but then forgot about it.  Possibly it might be easier to make it take a list of strings and have it build the hash table itself.

jmarshall:
This should be ready to go now.

You might or might not want to take the final commit (452353f9d4774a2d559a222ef24d29c95dc44a7f) which renames `bam_hrecs_*` to `sam_hrecs_*` and similarly for `bam_hrec_*`. These identifiers are all internal, so it doesn't matter too much for the API; probably mostly it just comes down to how used you folks are to the status quo for these identifiers.

I left the struct tag as `struct bam_hdr_t` for compatibility with wrappers such as pysam and Bio::DB::HTS. Actually by inspecting the code (untested): it doesn't matter for Bio::DB::HTS, as that just uses the `bam_hdr_t` typedef; for pysam, their .pxd file says `ctypedef struct bam_hdr_t` but then they mostly bundle particular HTSlib versions and tweak their cython code accordingly anyway. Hmmm… not sure whether it matters for these or other wrappers…

jmarshall:
Deserves a _NEWS_ entry, I guess…

jmarshall:
_NEWS_ entry added and struct tag renamed.

jmarshall:
Sigh… that _NEWS_ update caused a merge conflict so CI couldn't run. Added a commit merging from **develop** so we can observe the green tick :smile:

That `[OMIT ME]` commit should be omitted when you merge this. Or let me know if you're unconcerned about seeing CI on the final state of this PR and I'll remove the `[OMIT ME]` commit from the PR.

daviesrob:
Looks good.  Would it be possible to rebase the whole thing and force push the result onto your existing branch?

jmarshall:
No problem, will do. So that's a yes to `bam_hrecs_`→`sam_hrecs_` as well?

daviesrob:
Yes, consistency for the win.  Plus it's internal so shouldn't break anything.

jmarshall:
Rebased on top of current **develop**.

daviesrob:
Now merged, thanks very much.

PR #890, opened 2019-07-08
Make BAM files with no SQ lines in the text header work again

daviesrob:
It's possible to make a BAM file with entries in the list of references that don't have corresponding `@SQ` lines in the text header.  This makes them work with the new header API.

While BAM files like this are hopefully rare, some frameworks (e.g. pysam) currently rely on being able to do a similar thing by putting entries in the `target_name` and `target_len` header arrays without adding text `@SQ` lines.  This allows them to still work with the new header code.

Also includes a commit to make some header string pointers const.  They're pointers into the string pool, and shouldn't be modified after creation.


header.c
```diff
@@ -801,12 +825,108 @@ static int rebuild_target_arrays(sam_hdr_t *bh) {
...
+            if (!hrecs->ref[i].name) continue;
+            k = kh_get(m_s2i, hrecs->ref_hash, hrecs->ref[tid].name);
+            if (k < kh_end(hrecs->ref_hash)) kh_del(m_s2i, hrecs->ref_hash, k);
+        }
+        hrecs->nref = 0;
+        return -1;
+    }
+}
+
+static int add_stub_ref_sq_lines(sam_hrecs_t *hrecs) {
```
valeriuo: A comment that explains why this method is needed would be welcome.

valeriuo:
`sam_hrecs_update_hashes` should also be called from `sam_hdr_update_line`. Otherwise, an update of the reference sequence length, for example, following padding/de-padding, could leave discrepancies in the header.

daviesrob:
Updated following comments.

It turns out that `sam_hdr_update_line()` could make some fairly complicated changes, notably by renaming things.  This should not work for `@SQ` and `@RG` records, more-or-less.  It will refuse to rename `@PG` records because it's a bit too complicated.

header.c
```diff
@@ -127,8 +168,30 @@ static int sam_hrecs_update_hashes(sam_hrecs_t *hrecs,
...
-        sam_hrec_tag_t *tag;
+        sam_hrec_tag_t *tag = sam_hrecs_find_key(h_type, "ID", NULL);
         int nrg = hrecs->nrg;
+        khint_t k;
+
+        if (!tag) {
+            hts_log_error("Header includes @RG line with no ID: tag");
+            return -1;  // ID should be present, according to spec.
+        }
+        assert(tag->str && tag->len >= 3);
```
valeriuo: Since the tag is parsed and checked here, there is no need to do it again at line 211. So the code from 205-228 could be replaced with 
```
        hrecs->rg[nrg].name = tag->str + 3;
        hrecs->rg[nrg].name_len = tag->len - 3;
        hrecs->rg[nrg].ty   = h_type;
        hrecs->rg[nrg].id   = nrg;

        int r;
        k = kh_put(m_s2i, hrecs->rg_hash, hrecs->rg[nrg].name, &r);
        if (-1 == r) return -1;
        kh_val(hrecs->rg_hash, k) = nrg;

daviesrob: Yes, that does make it a bit simpler.

daviesrob:
Rebased with suggested simplification.

PR #891, opened 2019-07-08
Fix sam_parse1() / cigar_tab[] race condition

jmarshall:
[Something that I noticed while going over the header code…]

Since 3bb8fd1b1a940ca6c30f744451e58bafdf177461 all SAM file readers share a static `cigar_tab` that is reinitialised each time a `sam_hdr_t` is given to `sam_parse1()` for the first time. If multiple SAM files are being read by a multithreaded program, in theory this could cause spurious parse failures if `cigar_tab` is reinitialised while another thread is parsing a CIGAR string.

This could be improved by only actually writing to `cigar_tab` the first time `cigar_tab_init()` is called, but to be really threading-correct I think this would need a mutex. Easier just to do what should have been done years ago and write out the contents of the table explicitly and make it `const`! :smile:

Also a small commit to fix a dependency that caused spurious test failures while testing this as one of the test programs wasn't getting rebuilt.

This PR removes `sam_hdr_t`'s `cigar_tab` field on the basis that surely no-one is using it (as it's not set up until after you've already read the first read). If you prefer, I can redo this to just change it to `const int8_t *cigar_tab` (preferably) and initialise it to the prebuilt table early on.

This PR makes the table public, mostly so that _test/sam.c_ can check it's correct. I guess that's not really necessary as other test cases will fail if their CIGARs can't be parsed so it could get away with just making sure there are test SAM files that exercise all the operators. But it fills a gap anyway…

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ lib*.so.*
 /test/test_bgzf
 /test/test_index
 /test/test_kstring
+/test/test-parse-reg
 /test/test_realn
 /test/test-regidx
 /test/test-vcf-api
--- a/Makefile
+++ b/Makefile
@@ -445,7 +445,7 @@ test/sam.o: test/sam.c config.h $(htslib_hts_defs_h) $(htslib_sam_h) $(htslib_fa
 test/test_bgzf.o: test/test_bgzf.c config.h $(htslib_bgzf_h) $(htslib_hfile_h) $(hfile_internal_h)
 test/test_kstring.o: test/test_kstring.c config.h $(htslib_kstring_h)
 test/test-parse-reg.o: test/test-parse-reg.c $(htslib_hts_h) $(htslib_sam_h)
-test/test-realn.o: test/test_realn.c config.h $(htslib_hts_h) $(htslib_sam_h) $(htslib_faidx_h)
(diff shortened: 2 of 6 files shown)


daviesrob:
I'd be inclined to leave it as a `const int8_t *` and mark it as deprecated, just in case anyone is using it directly (admittedly unlikely).

jmarshall:
While leaving it as always-`NULL` (as per `sdict`), or filling it in?

daviesrob:
Filling it in would be a bit more friendly.  `sdict` being `void *` was difficult to use outside of HTSlib.  `cigar_tab` on the other hand does have usable contents.

jmarshall:
Now marked as deprecated in the header; bit of a pain as we need to suppress `HTS_DEPRECATED` in _sam.c_ to initialise it.

This facilitates existing code that was using `cigar_tab` in their own functions.

BTW both versions of the PR do not support lunatic code that was previously updating `hdr->cigar_tab` so that `sam_parse1()`/`sam_read1()` would read records using their own bespoke CIGAR operators, as I'd rather have `sam_parse1()` use the fixed table directly rather than via a `h->cigar_tab` indirection. I don't think HTSlib needs to support that level of craziness…

daviesrob:
Thanks, now merged.

PR #894, opened 2019-07-17
[header API] Change return value of sam_hdr_length

valeriuo:
Make `sam_hdr_length` return **SIZE_MAX** in case of error, thus matching its return type range.

The underlying error would be logged by the called method that produced it.


daviesrob:
Rebased.  It looks like a sam_hdr_remove_lines documentation fix got mixed in with this, but shouldn't be a problem as it looks OK.  I've added `@code ... @endcode` around the example.

PR #896, opened 2019-07-18
Make CRAM multi-ref mode more transitive.

jkbonfield:
There was already code to automatically enable it, but now it has code to automatically disable again when appropriate.  This can have a considerable size reduction for some data sets (due to re-enabling AP_delta mainly).

I've checked it using `-fsanitize=thread` too to validate the new lock.

### DIFF
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -3135,7 +3135,7 @@ int cram_put_bam_seq(cram_fd *fd, bam_seq_t *b) {
             multi_seq = 1;
         } else if (fd->multi_seq == 1) {
             pthread_mutex_lock(&fd->metrics_lock);
-            if (fd->last_RI_count <= c->max_slice) {
+            if (fd->last_RI_count <= c->max_slice && fd->multi_seq_user != 1) {
                 multi_seq = 0;
                 hts_log_info("Multi-ref disabled for next container");
             }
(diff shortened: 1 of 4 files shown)


cram/cram_encode.c
```diff
@@ -3150,7 +3164,9 @@ int cram_put_bam_seq(cram_fd *fd, bam_seq_t *b) {
          * multiple sequences per container we emit the small partial
          * container as-is and then start a fresh one in a different mode.
          */
-        if (multi_seq) {
+        if (multi_seq == 0 && fd->multi_seq == 1 && fd->multi_seq_user == -1) {
+            fd->multi_seq = -1;
+        } else if (multi_seq || fd->multi_seq == 1) {
```
valeriuo: These two conditions can both be true at the same time. Is this the intended behaviour?

cram/cram_encode.c
```diff
@@ -3165,8 +3165,11 @@ int cram_put_bam_seq(cram_fd *fd, bam_seq_t *b) {
          * container as-is and then start a fresh one in a different mode.
          */
         if (multi_seq == 0 && fd->multi_seq == 1 && fd->multi_seq_user == -1) {
+            // User selected auto-mode, we're currently using multi-seq, but
+            // have detected we don't need to.  Switch back to auto.
             fd->multi_seq = -1;
         } else if (multi_seq || fd->multi_seq == 1) {
+            // We either detected multi-seq needed or user explicitly asked for it.
```
valeriuo: Should it be then: `else if (multi_seq || fd->multi_seq`**_user**` == 1)` ?

jkbonfield: A very good point.  I clearly confused myself with that comment.  I'll try and test this and figure out what I intended when I wrote it!

jkbonfield: Ok it's necessary, but it could be either `fd->multi_seq` or `fd->multi_seq_user` as they're the same thing in this condition.  It's basically there because `fd->multi_seq` could (perhaps should) be set to zero but we're soldiering on using multi-ref mode because that's what we've been told to do.

However I reworked it so that the `multi_seq` variable assignment itself checks `fd->multi_seq_user` which avoids the need for this extra OR clause at all.

valeriuo:
Merged as 45f657ccb063f7ce3db1a925b470359943d80f9c and 53c045975124ebebf09b2a261967585cb93c1c22

PR #900, opened 2019-07-24
Restore support for QNAMEs longer than 251 characters

### LINKED ISSUE #520: l_qname overflows for QNAMEs of 252 or more characters

jmarshall:
Another ABI breaker that was in danger of being forgotten about while the ABI change window is open (leastwise I for one had forgotten about it!): change `l_qname` to `uint16_t` so QNAME length can be returned to the spec's 254 from 079c7b81d3eb740f575d6ee07f8d0df36932559e's 251 — see #520.

As the soversion is being bumped, we have the luxury of fixing this by simply enlarging the `l_qname` field. It needs to represent the values 0…256 so most of the new bits are wasted, but ¯\\\_(ツ)\_/¯ and maybe one day we may have a format wanting longer QNAMEs…

Adds test cases with long read names. I wish I had though of such test cases when I first implemented the align-the-CIGAR-data thing in 5d114ebd8e9b80622769b8e575c5a9359cd51273!

Fixes #520. Supersedes and closes #523.

### DIFF
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -188,10 +188,9 @@ typedef struct {
     int32_t pos;
     uint16_t bin;
     uint8_t qual;
-    uint8_t l_qname;
-    uint16_t flag;
-    uint8_t unused1;
     uint8_t l_extranul;
+    uint16_t flag;
+    uint16_t l_qname;
     uint32_t n_cigar;
     int32_t l_qseq;
     int32_t mtid;
--- a/NEWS
+++ b/NEWS
@@ -52,6 +52,12 @@ Noteworthy changes in release a.b
     sam_hdr_str() should be used instead.  The old cigar_tab field is
     now marked as deprecated; use the new bam_cigar_table[] instead.
 
+  - The bam1_core_t structure's l_qname and l_extranul fields have been
+    rearranged and enlarged; l_qname still includes the extra NULs.
+    (Almost all code should use bam_get_qname(), bam_get_cigar(), etc,
+    and has no need to use these fields directly.)  HTSlib now supports
+    the SAM specification's full 254 QNAME length again.  (#520)
+
   - bcf_index_load() no longer tries the '.tbi' suffix when looking for
     BCF index files (.tbi indexes are for text files, not binary BCF).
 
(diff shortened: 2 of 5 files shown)


htslib/sam.h
```diff
@@ -188,10 +188,9 @@ typedef struct {
     int32_t pos;
     uint16_t bin;
     uint8_t qual;
-    uint8_t l_qname;
-    uint16_t flag;
-    uint8_t unused1;
     uint8_t l_extranul;
+    uint16_t flag;
+    uint16_t l_qname;
     uint32_t n_cigar;
```
jmarshall: This diff is pretty unhelpfully displayed. Equivalently and more clearly:
```diff
     int32_t pos;
     uint16_t bin;
     uint8_t qual;
-    uint8_t l_qname;
+    uint8_t l_extranul;
     uint16_t flag;
-    uint8_t unused1;
-    uint8_t l_extranul;
+    uint16_t l_qname;
     uint32_t n_cigar;
     int32_t l_qseq;
     int32_t mtid;
```

jmarshall:
I suppose, to be precisely accurate, it needs to represent 4…256 inclusive.

So we could put `+ 4` in the various `bam_get_cigar` macros:

```c
#define bam_get_cigar(b) ((uint32_t*)((b)->data + (b)->core.l_qname + 4)
// etc
```

and `l_qname` would store 0…252 and thus still fit in `uint8_t`! However this would be at the cost of breaking bad third-party code that does this computation itself rather than using `bam_get_cigar()` etc, which was part of the reason for counting the extra NULs in `l_qname` in the first place…

daviesrob:
I think `uint16_t` is the easiest solution.  #709 really messes with the contents of `bam1_core_t` so more rearrangement is going to happen anyway if we want to keep the structure hole free.

Possibly we need some checks in `bam_write1()` as it's possible to programmatically make read names that don't fit in a BAM file.  In theory we could also relax the limit in SAM (and CRAM), although I suspect it would be a bad idea.

daviesrob:
Thanks, long QNAMEs are now restored.

PR #901, opened 2019-07-24
Support seeking in uncompressed files with bgzf_seek and bgzf_tell

### LINKED ISSUE #899: bgzf_seek on uncompressed data always seeks to the beginning of the block
The `bgzf_*` functions handle uncompressed files by presenting them as a series of maximum-sized blocks. In this way, virtual offsets as used by the BGZF layer correspond to actual offsets in the file. As far as I can make out, `bgzf_tell` and `bgzf_seek` should work correctly on such files.

However, this is not the case. `bgzf_tell` works fine, and `bgzf_seek` appears to work (i.e. a subsequent `bgzf_tell` gives you back the virtual offset you seeked to), but the next read will be fulfilled from the beginning of the block that contains the position you tried to go to, instead of from the actual position.

Here's where we set up the seek (setting `block_address` and `block_offset` and noting that we need to load the block):

https://github.com/samtools/htslib/blob/49058f4de0f91fb6200f722c79cbb4336587ac9c/bgzf.c#L1971-L1977

But here's where we load the block. If any data is actually read (**EDIT: This is wrong; we're really checking if any data was read since the most recent seek. I was bumping myself back to the beginning of the block in my own code.**), `block_offset` is cleared out, dumping us back at the beginning of the block instead of where we were supposed to be going.

https://github.com/samtools/htslib/blob/49058f4de0f91fb6200f722c79cbb4336587ac9c/bgzf.c#L976-L993

It looks like this may have been this way [since 2013](https://github.com/samtools/htslib/commit/9ec7860705a613c659992c437379a2f1374c71ac). This may be a contributor to the problems encountered in #388.

I want seeking in uncompressed files to work for vg, which uses HTSlib's BGZF layer in its own IO, so I am going to see if I can put together a patch.

adamnovak:
This modifies bgzf_tell and bgzf_seek to pass through the real offset as the
virtual offset in uncompressed files. It also makes sure to consistently use
only block_address and not block_offset when passing around the seek location
in uncompressed files.

This fixes #899.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -972,6 +972,9 @@ int bgzf_read_block(BGZF *fp)
  single_threaded:
     size = 0;
 
+    int64_t block_address;
+    block_address = bgzf_htell(fp);
+
     // Reading an uncompressed file
     if ( !fp->is_compressed )
     {
@@ -986,15 +989,13 @@ int bgzf_read_block(BGZF *fp)
             fp->block_length = 0;
             return 0;
         }
-        if (fp->block_length != 0) fp->block_offset = 0;
-        fp->block_address += count;
+        if (fp->block_length != 0) fp->block_offset = 0; // Seeks in uncompressed files do not use block_offset.
         fp->block_length = count;
+        fp->block_address = block_address;
         return 0;
     }
 
-    // Reading compressed file
-    int64_t block_address;
-    block_address = bgzf_htell(fp);
+    // Reading non-block compressed file
     if ( fp->is_gzip && fp->gz_stream ) // is this is an initialized gzip stream?
     {
         count = inflate_gzip_block(fp);
@@ -1986,7 +1987,13 @@ int64_t bgzf_seek(BGZF* fp, int64_t pos, int where)
         fp->errcode |= BGZF_ERR_MISUSE;
         return -1;
     }
-    return bgzf_seek_common(fp, pos >> 16, pos & 0xFFFF);
+    if (fp->is_compressed) {
+        // Virtual offsets are packed block start and offset
+        return bgzf_seek_common(fp, pos >> 16, pos & 0xFFFF);
+    } else {
+        // Virtual offsets are real
+        return bgzf_seek_common(fp, pos, 0);
+    }
 }
 
 int bgzf_is_bgzf(const char *fn)
--- a/htslib/bgzf.h
+++ b/htslib/bgzf.h
@@ -201,7 +201,7 @@ typedef struct BGZF BGZF;
      * call to bgzf_seek can be used to position the file at the same point.
      * Return value is non-negative on success.
      */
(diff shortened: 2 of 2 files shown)


daviesrob:
This takes the decision that indexes on uncompressed files without BGZF wrappers should store the real file offset instead of `real_offset << 16`.

The [SAM specification](http://samtools.github.io/hts-specs/SAMv1.pdf) is a bit ambiguous on this.  Section 4.1.1 defines the virtual file offset in terms of BGZF blocks as:

> Each  virtual  file  offset  is  an  unsigned  64-bit  integer,  defined  as: `coffset<<16|uoffset`, where `coffset` is an unsigned byte offset into the BGZF file to the beginning of a BGZF block, and `uoffset` is an unsigned byte offset into the uncompressed data stream represented by that BGZF block.

Section 5.2 refers to file positions using:

> (Virtual) file offset of the [... location ...]

Should we take the brackets around "virtual" as meaning we can use the real offset on non-BGZF files, or should we shift it by 16 bits and always set `uoffset` to zero?  An opinion from @jkbonfield, @jmarshall  and @yfarjoun (with SAM/BAM format maintainer hats on) would be welcome.

jkbonfield:
I wouldn't say that's ambiguous.  If `coffset` matches the uncompressed offset then the maths is still valid.  Although there is flexibility in how we subdivide offset into `coffset` and `uoffset` when creating the virtual offset, the other way around is well defined.

Eg pos 100000 could be coffset 100000, uoffset 0 or coffset 65536 uoffset 34464, but practically speaking it makes no difference as whichever virtual offset we get, we can read the uncompressed file (Seek to `virt_offset>>16`, read a block, and then skip `virt_offset&0xffff `always works).

adamnovak:
@jkbonfield So you would want BGZF to give virtual offsets instead of real offsets for uncompressed data, to match what it does for compressed data? I can make this PR do that instead.

jkbonfield:
Well that's my opinion, but we should wait and see what others think first, in particular Petr who wrote this code in the first place.

jmarshall:
With my maintainer hat on, my opinion is at https://github.com/samtools/htslib/pull/904#issuecomment-543166029.

My initial preference is for using the real offset for uncompressed files (I don't think the spec is ambiguous; it doesn't profess to say what “virtual file offsets” are for anything other than BGZF files).

I would also accept defining “virtual file offset” for uncompressed files as being `(rawoffset & ~64K, rawoffset % 64K)`, i.e., consider the file to consist of “blocks” of exactly 64K, as this is:
not completely pointless; consistent with virtual file offsets for BGZF files; uniquely defined (i.e. there's a 1-to-1 mapping between virtual and real offsets).

I'm less pleased with “shift it by 16 bits and always set uoffset to zero” as it fails the pointlessness test. :smile:

daviesrob:
Closing as replaced by #904, which is now merged.

PR #902, opened 2019-07-26
sam_parse1(): return to allowing FLAGs in hex/octal; avoid overflow

jmarshall:
[More things noticed while going over the header code…]

Avoid overflow in `STRTOL64()` and `STRTOUL64()` (and similar functions in _cram/cram_index.c_) by ensuring that the subtraction occurs first. Compiling with `-fsanitize=undefined` (for the unsigned ones, use `-fsanitize=unsigned-integer-overflow` as well) demonstrates these overflows.

At the moment this doesn't matter too much for the _sam.c_ functions as they are used only up to 32 bits, but this prepares them for actual 64-bit use. Also changed the return type of `STRTOUL64()` to be unsigned — @jkbonfield was it intentional to use `int64_t` for `STRTOUL64` (performance reasons?) or is changing this to `uint64_t` okay?

Prior to the recent 9f28a4f1bee95eb0842e92a8a994c34bdfb657ae, `sam_parse1()` used `strtol(p, &p, 0)` to parse SAM's FLAG field, which thus accepted decimal/hex/octal. HTSJDK has reimplemented the historical samtools decimal/hex/symbolic-letters FLAG field (cf old `samtools view -x -X`; see samtools/htsjdk#232, which implements decimal/hex/octal/symbolic), and it is planned to propose hex/symbolic-letters for inclusion in the SAM spec sometime (probably at a point when HTSlib at least has the infrastructure to implement it!). So it would be a shame to remove HTSlib's existing hex/octal FLAG parsing ability at this point.

This PR adds a `parse_sam_flag()` function to parse decimal/hex/octal while preserving James's fast implementation for the common decimal case. The function also provides a slot for the eventual implementation of parsing symbolic-letters FLAGs.

(Prior to that change, B-arrays were also parsed with `strtol(…, 0)` so accepted decimal/hex/ octal, but HTSJDK doesn't accept that, there's no precedent or desire to extend the spec to allow hex/etc there, and non-array `i` tag values were always decimal-only so this appeared to be unintentional; hence this PR leaves the new decimal-only behaviour for B-arrays as is.)

### DIFF
--- a/cram/cram_index.c
+++ b/cram/cram_index.c
@@ -98,8 +98,10 @@ static int kget_int32(kstring_t *k, size_t *pos, int32_t *val_p) {
     if (p >= k->l || !(k->s[p] >= '0' && k->s[p] <= '9'))
         return -1;
 
-    while (p < k->l && k->s[p] >= '0' && k->s[p] <= '9')
-        val = val*10 + k->s[p++]-'0';
+    while (p < k->l && k->s[p] >= '0' && k->s[p] <= '9') {
+        int digit = k->s[p++]-'0';
+        val = val*10 + digit;
+    }
 
     *pos = p;
     *val_p = sign*val;
@@ -121,8 +123,10 @@ static int kget_int64(kstring_t *k, size_t *pos, int64_t *val_p) {
     if (p >= k->l || !(k->s[p] >= '0' && k->s[p] <= '9'))
         return -1;
 
-    while (p < k->l && k->s[p] >= '0' && k->s[p] <= '9')
-        val = val*10 + k->s[p++]-'0';
+    while (p < k->l && k->s[p] >= '0' && k->s[p] <= '9') {
+        int digit = k->s[p++]-'0';
+        val = val*10 + digit;
+    }
 
     *pos = p;
     *val_p = sign*val;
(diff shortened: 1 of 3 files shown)


jkbonfield:
I can't recall why I used signed instead of unsigned, but it's probably just a cut and paste bug.  Regardless I think we don't plan to really care about 63 vs 64 bit quantities and are likely to limit ourselves to 63 arbitrarily (it's needed in TLEN for example).  However overflow bugs do indeed need fixing.  Thanks.

I wasn't even aware we supported hex flags and failed to spot the strtol 0 vs 10 case.  I also wasn't aware of plans to add it to the spec (indeed I assumed it was removed from samtools output precisely because we had no plans to change the spec).  I've no objection to adding them back though if that's our long term goal.

jmarshall:
Agreed re 2<sup>63</sup> should be enough for anyone :smile:

I think the old `samtools view -x -X` were removed from samtools simply because the HTSlib implementation of SAM I/O didn't support them (samtools/samtools@296d0294ca1c7de6dc0ef94b6db98bb94c9a4f9b). @nh13 and I have been fairly quiet about our cunning plan to re-add this to htslib+samtools (I've been a bit slow on that one…) and then as it would be in both major implementations seems fairly logical to formalise it in the spec; but it's all there in that very long HTSJDK thread…

daviesrob:
Thanks. I'm certainly in favour of allowing octal and hex flags.

PR #904, opened 2019-07-31
Allow seeking to virtual offsets in uncompressed streams

### LINKED ISSUE #899: bgzf_seek on uncompressed data always seeks to the beginning of the block

adamnovak:
Patch suggested by @jkbonfield

This is an alternative to #901 that will also fix #899. But in this version, virtual offsets in uncompressed files are virtual: they consist of a block start address and an offset in the block.

Many virtual offsets can point to the same byte in the file, because a block could start at any byte in an uncompressed file. However, the library will read the file in max size blocks starting from 0, so unless the user is synthesizing virtual offsets there will still be a canonical set of block start points.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -1818,8 +1818,12 @@ int bgzf_flush_try(BGZF *fp, ssize_t size)
 
 ssize_t bgzf_write(BGZF *fp, const void *data, size_t length)
 {
-    if ( !fp->is_compressed )
+    if ( !fp->is_compressed ) {
+        size_t push = length + (size_t) fp->block_offset;
+        fp->block_offset = push % BGZF_MAX_BLOCK_SIZE;
+        fp->block_address += (push - fp->block_offset);
         return hwrite(fp->fp, data, length);
+    }
 
     const uint8_t *input = (const uint8_t*)data;
     ssize_t remaining = length;
@@ -1841,8 +1845,12 @@ ssize_t bgzf_write(BGZF *fp, const void *data, size_t length)
 
 ssize_t bgzf_block_write(BGZF *fp, const void *data, size_t length)
 {
-    if ( !fp->is_compressed )
+    if ( !fp->is_compressed ) {
+        size_t push = length + (size_t) fp->block_offset;
+        fp->block_offset = push % BGZF_MAX_BLOCK_SIZE;
+        fp->block_address += (push - fp->block_offset);
         return hwrite(fp->fp, data, length);
+    }
 
     const uint8_t *input = (const uint8_t*)data;
     ssize_t remaining = length;
--- a/bgzf.c
+++ b/bgzf.c
@@ -974,6 +974,9 @@ int bgzf_read_block(BGZF *fp)
  single_threaded:
     size = 0;
 
+    int64_t block_address;
+    block_address = bgzf_htell(fp);
+
     // Reading an uncompressed file
     if ( !fp->is_compressed )
     {
@@ -989,14 +992,12 @@ int bgzf_read_block(BGZF *fp)
             return 0;
         }
         if (fp->block_length != 0) fp->block_offset = 0;
-        fp->block_address += count;
+        fp->block_address = block_address;
         fp->block_length = count;
         return 0;
     }
 
     // Reading compressed file
-    int64_t block_address;
-    block_address = bgzf_htell(fp);
     if ( fp->is_gzip && fp->gz_stream ) // is this is an initialized gzip stream?
     {
         count = inflate_gzip_block(fp);
--- a/test/test_bgzf.c
+++ b/test/test_bgzf.c
@@ -31,6 +31,7 @@ DEALINGS IN THE SOFTWARE.
 #include <errno.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <inttypes.h>
 #include <fcntl.h>
 #include "htslib/bgzf.h"
 #include "htslib/hfile.h"
@@ -242,6 +243,39 @@ static int try_bgzf_index_dump(BGZF *fp, const char *bname, const char *suffix,
     return 0;
 }
 
+static int64_t try_bgzf_tell(BGZF *fp, const char *name, const char *func) {
+    int64_t told = bgzf_tell(fp);
+    if (told < 0) {
+        fprintf(stderr, "%s : %s %s : %s\n",
+                func, "Error telling in",
+                name, strerror(errno));
+        return -1;
+    }
+
+    return told;
+}
+
+static int64_t try_bgzf_tell_expect(BGZF *fp, int64_t expected, const char *name, const char *func) {
+    int64_t told = try_bgzf_tell(fp, name, func);
+    if (told != expected) {
+        fprintf(stderr, "%s : Unexpected value (%" PRId64 ") from bgzf_tell on %s; "
+                "expected %" PRId64 "\n",
+                func, told, name, expected);
+        return -1;
+    }
+    return told;
+}
+
+static int try_bgzf_seek(BGZF *fp, int64_t pos, int whence,
+                          const char *name, const char *func) {
+    if (bgzf_seek(fp, pos, whence) < 0) {
+        fprintf(stderr, "%s : Error from bgzf_seek(%s, %" PRId64 ", %d) : %s\n",
+                func, name, pos, whence, strerror(errno));
+        return -1;
+    }
+    return 0;
+}
+
 static int try_bgzf_useek(BGZF *fp, long uoffset, int where,
                           const char *name, const char *func) {
     if (bgzf_useek(fp, uoffset, where) < 0) {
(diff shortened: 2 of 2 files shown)


mr-c:
Hello, I am packaging vg for Debian. If this PR is accepted and part of a future htslib release, that would be a lot better for us. Thanks!

valeriuo:
The fix is good. I just added another [test](https://github.com/valeriuo/htslib/commit/7f6fe16f4895068a62b9b2c2a64c51cb42f67314) to capture the change in `bgzf_read_block`, but I could not push it to your branch, for lack of permission.

adamnovak:
OK @valeriuo, I've force-pushed your improved and rebased version, with the additional tests.

jkbonfield:
I'm happy with the PR, but given this work is so we can construct an index on uncompressed data it has an implication for on-disk file formats.  We have a GA4GH file formats meeting today so I'll just double check it there and then hopefully we're good to go.

jkbonfield:
> We have a GA4GH file formats meeting today so I'll just double check it there and then hopefully we're good to go.

There wasn't complete agreement on which method to use here sadly.  A reminder to @jmarshall  and @yfarjoun to add your thoughts here please.

To get you both up to speed, the issue is a file formats one because any indexing done with this code will become a file on disk and hence stray into GA4GH land.  The possible alternatives for handling indexing of totally uncompressed files.

1.  The  addresses in the CSI index aren't virtual addresses but absolute addresses.  This breaks CSI (I think) as it claims all addresses are virtual.  It's simple to understand though and every location has precisely one address (itself).  The implementation of this is #901, which works by having a switch in both `bgzf_seek` and `bgzf_tell` to check whether or not the file is uncompressed.

2. The addresses are virtual offsets, stored in imaginary 64k blocks.  This is a degenerate form of existing compression with a 1:1 compression ratio (ie none at all) and I think it's compatible with the CSI spec.   Uncompressed offset p yields address `((p & ~(64k-1))<<16) + (p%64k)`.   The custom logic on compressed vs uncompressed in code moves from needing modified seek/tell functions and into read/write functions instead.  See #904 (this PR).

3. As per 2, but with just have virtual addresses always being `uncompressed_address<<16` with the offset component always being zero.  Maybe simpler; I haven't really thought through the implications. 

All 3 of these are exposed in the data written to CSI.  John I know had some concerns over there being multiple encodings for the same virtual address with uncompressed data, and hence favoured option 1, but Yossi then pointed out the same is already true for bgzf virtual addresses anyway.  I cannot see which case that is though.  Is it where the address is either the first byte of the next block of 1 byte beyond the previous block?

Either way we want to make a release and we'd like this or the other PR in it so need a decision soon.  We'll make it without feedback if necessary. :-)

adamnovak:
I also don't see how you accomplish aliasing with BGZF virtual
offsets. You aren't supposed to be allowed to address actual bytes in
the next block, are you?

jmarshall:
Surely merging this is premature — it is listed for discussion on this afternoon's formats meeting agenda.

jkbonfield:
Sorry too late.

It was scheduled for discussed in the last meeting, which we did, and it was stated at the time we'd merge it in the absence of any more feedback.  I explicitly didn't want to raise it again today if there were no more comments because that would invite a delay of another month before merging.

jmarshall:
Point of order. It's listed on the agenda, and people who sent their apologies for the last meeting have been researching their opinion on that basis.

jkbonfield:
There are comments in previous meeting minutes from both you and Yossi on this topic.  We can't delay merging features in htslib indefinitely.

If you wish to raise it again at the upcoming meeting, please add it to the agenda (it's in previous actions only at present, listed as bring brought to the attention of maintainers but not with any specific action for a review).

Note it's in develop now, not master.  If you have a desire to change it to e.g. always have 0 for the uncompressed offset component then you're free to make a PR for us to consider, but do it promptly!  We're not going to delay the release for this.

jmarshall:
Question for people wanting this functionality of seeking in uncompressed streams: if HTSlib finally got the `hts_seek(htsFile *fp, off_t pos)` function that it's been missing for years, would that solve your use case? Or do you specifically want to do this in the context of a `BGZF*`? If the latter, is this because you're doing so in the context of iterators and BAI/CSI indexes?

In terms of https://github.com/samtools/htslib/pull/904#issuecomment-534020186, there is no breakage to BAI/CSI in any of the cases. BAI and CSI are formats that describe a mapping from bins to “virtual file offsets”. In a separate section, the specification defines what a “virtual file offset” is for a BGZF-compressed file. It is undefined in the specifications what a virtual file offset is for an uncompressed file, so we would be free to define it as we chose, either the natural thing (for uncompressed, the “virtual file offset” is just the real file offset) or by extending the definition for BGZF to be the same thing without blocks (allowing cases (2) and (3) and everything in between, in principle — it appears that this PR would have bgzf_tell return (2) and bgzf_seek accept (2) or (3) or anything in between, happily enough).

IMHO the natural thing to use for uncompressed files would be the real file offset, as there is no uncertainty over the values. I don't see what advantage reusing BGZF's virtual layout has other than saving two `if` statements in HTSlib. If HTSlib wants to stay with this implementation, IMHO the right thing from a specifications perspective might be for the spec to remain agnostic on what virtual file offsets for non-BGZF files are.

Alternatively: thank you for clarifying that the proposal is (2) rather than (3). This can be construed as effectively dividing the uncompressed file into 64K blocks and using the address/offset notation to encourage keeping the underlying raw `read(2)` calls aligned with disk blocks. So that would be (rather old-school) justifiable reasoning for the spec and perhaps I have talked myself into accepting (2)… :smile:

(I too am pretty sure that the aliasing Yossi was referring to was the block boundary one, i.e., that `(block_address,0)` can be occasionally referred to as `(prev_block_address,prev_block_length)`. I've always had the vague feeling that this is a buglet in the tools that generate such degenerate virtual file offsets.)

jkbonfield:
Firstly I'm sure no one would object to a flexible `hts_seek` function, so if you get time to dust off that code then it'd be great, thanks.

Regarding this particular issue though, @daviesrob reminded me that htsjdk's CRAM uses .bai indices too.  In that situation the top 48 bits are the absolute file offset of the container and the bottom 16 bits are the "landmark" index.  CRAM's containers are larger than 64k so it's essentially indirecting via a lookup table to go to the Nth slice number (which practically speaking is normally zero).  That's documented in the CRAM spec too.

So basically we have
- BGZF: top 48-bits = absolute file offset, bottom 16 bits = offset into uncompressed
- CRAM: top 48-bits = absolute file offset, bottom 16 bits = offset into landmarks array
- RAW: top 48-bits = absolute file offset, bottom 16 bits = offset added to file offset.

There's some symmetry there and it may make a more general BAI/CSI documentation cleaner this way.

Also regarding option 2/3 above, there's one useful thing I learnt from ISO world which is our specifications describe how to decode the files, not how to encode them.  It's irrelevant whether an encoder chose to use method 2 or method 3, the decode process is exactly the same.  For this reason I think we should be agnostic.

adamnovak:
@jmarshall 
> Question for people wanting this functionality of seeking in uncompressed streams: if HTSlib finally got the hts_seek(htsFile *fp, off_t pos) function that it's been missing for years, would that solve your use case? Or do you specifically want to do this in the context of a BGZF*? If the latter, is this because you're doing so in the context of iterators and BAI/CSI indexes?

My use case for this is using the BGZF machinery to implement my own optionally-block-compressed, seekable format, not to read/write any of the formats that HTSlib already supports. Basically I'm doing BAM and BAM indexing for graph-structured references, in the vg project. Looking at [the descriptions of the `hts_*` functions](https://github.com/samtools/htslib/blob/7fc833d07462ff59dfbad30e8f517f4301eed433/htslib/hts.h), it looks like there isn't a generic `hts_read` for reading bytes, for example, and I see code reaching down into `htsFile*` [to access the internal BGZF when writing data](https://github.com/samtools/htslib/blob/d59c3645f798169f7fde53dbf5c6d3e095465e8c/sam.c#L688).

PR #905, opened 2019-08-01
[header API] Add methods `sam_hdr_line_index` and `sam_hdr_line_name`.

valeriuo:
`@SQ`, `@RG` and `@PG` lines have dedicated look-up tables and arrays, that should be exposed via the header API for faster retrieval of line position and identifier string.

### DIFF
--- a/header.c
+++ b/header.c
@@ -1572,12 +1572,15 @@ int sam_hdr_count_lines(sam_hdr_t *bh, const char *type) {
     case 'S':
         if (type[1] == 'Q')
             return bh->hrecs->nref;
+        break;
     case 'R':
         if (type[1] == 'G')
             return bh->hrecs->nrg;
+        break;
     case 'P':
         if (type[1] == 'G')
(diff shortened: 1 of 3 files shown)


header.c
```diff
@@ -1595,6 +1595,80 @@ int sam_hdr_count_lines(sam_hdr_t *bh, const char *type) {
...
+    case 'R':
+        if (type[1] == 'G' && pos < hrecs->nrg)
+            return hrecs->rg[pos].name;
+    case 'P':
+        if (type[1] == 'G' && pos < hrecs->npg)
+            return hrecs->pg[pos].name;
+    default:
+        break;
+    }
+
```
daviesrob: There should be a fall-back here for types not covered in the switch statement, along similar lines to [the one in `sam_hdr_count_lines`](https://github.com/samtools/htslib/blob/53be0f9/header.c#L1585).  (Also needed for `sam_hdr_line_index`).

PR #906, opened 2019-08-05
Improve BGZF multi-thread command state machine

daviesrob:
It was possible for bgzf_mt_reader() to catch its own signal on mtaux_t::command_c, resulting in a deadlock as the signal was not delivered to the thread that it was intended for.  This recently manifested as a problem with the Appveyor tests, which were frequently failing to complete due to a random lock-up when reading a BAM file in multi-threaded mode.

Fix by adding extra values for mtaux_t::command, which can be used to detect when  mtaux_t::command_c needs to be re-signalled.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -101,7 +101,9 @@ typedef struct bgzf_job {
 enum mtaux_cmd {
     NONE = 0,
     SEEK,
+    SEEK_DONE,
     HAS_EOF,
+    HAS_EOF_DONE,
     CLOSE,
 };
 
@@ -1408,10 +1410,10 @@ static int bgzf_check_EOF_common(BGZF *fp)
 static void bgzf_mt_eof(BGZF *fp) {
     mtaux_t *mt = fp->mt;
 
-    mt->command = NONE;
     pthread_mutex_lock(&mt->job_pool_m);
     mt->eof = bgzf_check_EOF_common(fp);
     pthread_mutex_unlock(&mt->job_pool_m);
+    mt->command = HAS_EOF_DONE;
     pthread_cond_signal(&mt->command_c);
 }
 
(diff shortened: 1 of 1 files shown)


bgzf.c
```diff
@@ -1922,10 +1935,24 @@ int bgzf_check_EOF(BGZF *fp) {
...
         fp->mt->command = HAS_EOF;
         pthread_cond_signal(&fp->mt->command_c);
         hts_tpool_wake_dispatch(fp->mt->out_queue);
-        pthread_cond_wait(&fp->mt->command_c, &fp->mt->command_m);
+        do {
+            pthread_cond_wait(&fp->mt->command_c, &fp->mt->command_m);
+            switch (fp->mt->command) {
+            case HAS_EOF_DONE: break;
+            case HAS_EOF:
+                pthread_cond_signal(&fp->mt->command_c);
```
jkbonfield: Maybe add some comments, such as "we caught the signal intended for bgzf_mt_reader, so send it again".

Also I wonder if the logic of *received in error, resend signal* should be in mt_reader instead.  Wouldn't this avoid the need for additional states in the state machine?  If mt_reader gets a NONE command then it knows it was its response and it can be ignored (resent).  That would keep the code receiving replies (main thread) simple.

daviesrob: The same logic is [here](https://github.com/samtools/htslib/blob/5ebfddd/bgzf.c#L1567) in bgzf_mt_reader.  It can't signal on NONE because if it did it would get into an endless loop of signalling itself.  Signalling on `SEEK_DONE / HAS_EOF_DONE` it should get out of the loop as the other thread will at some point be scheduled and reset the command to `NONE`.

The logic here should also deal with the possibility of spurious wake-up, if an old signal was still pending for some reason.  Without this it's possible that the value of `fp->mt->eof` could be read before `bgzf_mt_reader()` has had a chance to change it.

daviesrob: Comments also added.

jkbonfield: Not an endless loop - it's only endless if no one else is also trying to listen, which is an error in itself.  It's a loop of trying to not win the race, albeit perhaps not very efficient. :)  That said, if you win it once maybe there is some condition that makes you repeatedly win so you'd need some mitigation for that which becomes more complicated and possibly fragile.  Ie point taken.

Two communication channels (sender and receiver) would be cleanest, but it's also a more invasive change and more unknowns to worry about so this is sufficient.

daviesrob: When `bgzf_mt_reader()` reaches EOF on a long file, it's likely that `mt->command` will be `NONE` and there will be no other listeners.  If it got past `pthread_cond_wait()` due to a pending signal and it signalled on NONE itself then it would start looping until the reader thread got around to signalling `CLOSE`.

PR #910, opened 2019-08-12
Fixes to issues found by fuzzing

daviesrob:
- Stops `hts_open_fuzzer` from spamming log files with useless output.

- Updates `hts_open_fuzzer` so that SAM/BAM/CRAM headers get parsed, which should exercise more code paths.

- Fixes numerous bugs (see individual commits).

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -372,8 +372,8 @@ int ltf8_decode(cram_fd *fd, int64_t *val_p) {
 
 int ltf8_decode_crc(cram_fd *fd, int64_t *val_p, uint32_t *crc) {
     unsigned char c[9];
-    int64_t val = (unsigned char)hgetc(fd->fp);
-    if (val == -1)
+    int64_t val = hgetc(fd->fp);
+    if (val < 0)
         return -1;
 
     c[0] = val;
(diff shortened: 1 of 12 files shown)


cram/cram_io.c
```diff
@@ -384,79 +384,98 @@ int ltf8_decode_crc(cram_fd *fd, int64_t *val_p, uint32_t *crc) {
...
+        uval = (uval<<8) | c[2];
+        uval = (uval<<8) | c[3];
+        uval = (uval<<8) | c[4];
+        uval = (uval<<8) | c[5];
+        uval = (uval<<8) | c[6];
+        uval = (uval<<8) | c[7];
+        uval = (uval<<8) | c[8];
         *crc = crc32(*crc, c, 9);
-        *val_p = val;
+        *val_p = c[1] < 0x80 ? uval : -((int64_t) (0xffffffffffffffffULL - uval)) - 1;
```
jkbonfield: This line specifically is pointless.  It's not undefined behaviour (although it is probably implementation defined, but we can live with that).  This is simply an unsigned to signed integer conversion, covered by point 3.

Let's not needlessly obfuscate our code, without comment, for no reason.



>     6.3.1.3 Signed and unsigned integers
> 
>     1 When a value with integer type is converted to another integer type other than _Bool, if the value can be represented by the new type, it is unchanged.
> 
>     2 Otherwise, if the new type is unsigned, the value is converted by repeatedly adding or subtracting one more than the maximum value that can be represented in the new type until the value is in the range of the new type.
> 
>     3 Otherwise, the new type is signed and the value cannot be represented in it; either the result is implementation-defined or an implementation-defined signal is raised.

cram/cram_io.c
```diff
@@ -531,22 +531,24 @@ int int32_get_blk(cram_block *b, int32_t *val) {
...
-         b->data[b->byte  ]        |
-        (b->data[b->byte+1] <<  8) |
-        (b->data[b->byte+2] << 16) |
-        (b->data[b->byte+3] << 24);
+    uint32_t v =
+         ((uint32_t) b->data[b->byte  ])        |
+        (((uint32_t) b->data[b->byte+1]) <<  8) |
+        (((uint32_t) b->data[b->byte+2]) << 16) |
+        (((uint32_t) b->data[b->byte+3]) << 24);
+    *val = v < 0x80000000U ? v : -((int32_t) (0xffffffffU - v)) - 1;
```
jkbonfield: Again, `*val = v` is sufficient here.

valeriuo:
Merged into develop

PR #912, opened 2019-08-16
When not using configure, don't define HAVE_LZMA_H on macOS

jmarshall:
Extend #628 so the stub header is used on macOS also when configure is not used. (Thus fixing non-configure builds on macOS without Brew/etc-supplied _lzma.h_.)

This is PR arq5x/bedtools2#709 ported to upstream HTSlib.

Vanilla macOS includes /usr/lib/liblzma.dylib but no LZMA header files. Running configure detects this situation and uses os/lzma_stub.h instead of externally-supplied LZMA headers when they are missing. This patch makes non-configure builds also use os/lzma_stub.h on macOS. (Unlike
via configure, this will **always** use the stub on macOS even if you have lzma.h e.g. via Brew, but the stub header remains compatible.)

Also remove `HAVE_FSEEKO` and the corresponding configure.ac test: `fseeko()` no longer appears in HTSlib since da5c0c7b9aa1710b70ceb2a1097101d1df1c115a.


daviesrob:
Thanks, have merged.

PR #916, opened 2019-08-20
Multi-threaded SAM reading and writing.

jkbonfield:
An update of the old SAM multi-threading branch, with support for the new features that have been merged during the last year:  auto-indexing and header API (plus the SAM parser speed up itself which was previously in the earlier PR but is now rebased away).

Some benchmarks with samtools view -@8 9827_2#49.sam > /dev/null

```
Samtools 1.9:    2m18s
Samtools-dev:    1m5s
This PR:         0m14s  
```

Also tried samtools view -u -@8 9827_2#49.u.bam > /dev/null (uncompressed BAM):

```
This PR:         0m24s 
```

It is assumed the slower time is due to CRC checking, although this build didn't
include libdeflate.

Fixes #786 (replaces it).


jkbonfield:
Comedy benchmarks, just because it surprised me. :-)  This build has libdeflate support.  It's a test using a pointless view and pipe, to see how efficient SAM is as a pipe interchange format compared to uncompressed BAM.

```
$ time samtools view -u -@8 ~/scratch/data/9827_2#49.bam | samtools flagstat -@8 - 
56463236 + 0 in total (QC-passed reads + QC-failed reads)
0 + 0 secondary
0 + 0 supplementary
269248 + 0 duplicates
55357963 + 0 mapped (98.04% : N/A)
56463236 + 0 paired in sequencing
28231618 + 0 read1
28231618 + 0 read2
54363468 + 0 properly paired (96.28% : N/A)
55062652 + 0 with itself and mate mapped
295311 + 0 singletons (0.52% : N/A)
360264 + 0 with mate mapped to a different chr
300699 + 0 with mate mapped to a different chr (mapQ>=5)

real	0m21.505s
user	2m15.200s
sys	0m18.401s

$ time samtools view -h -@8 ~/scratch/data/9827_2#49.bam | samtools flagstat -@8 - 
56463236 + 0 in total (QC-passed reads + QC-failed reads)
0 + 0 secondary
0 + 0 supplementary
269248 + 0 duplicates
55357963 + 0 mapped (98.04% : N/A)
56463236 + 0 paired in sequencing
28231618 + 0 read1
28231618 + 0 read2
54363468 + 0 properly paired (96.28% : N/A)
55062652 + 0 with itself and mate mapped
295311 + 0 singletons (0.52% : N/A)
360264 + 0 with mate mapped to a different chr
300699 + 0 with mate mapped to a different chr (mapQ>=5)

real	0m18.026s
user	3m3.699s
sys	0m19.549s
```

SAM is more CPU time (unexpected) but oddly lower wall-clock time.  I'm not entirely sure why, but I'm guessing the bam data shuffling (disk layout vs memory layout) is in the worker threads for SAM and the main thread for BAM?  Either way it's not a sensible suggestion, but nice to know we don't have a hit any more provided we have sufficient CPUs to parallelise with.

Final piece of comedy...

```
-rw-r--r-- 1 jkb team117 6548191924 Jan 13  2014 /nfs/users/nfs_j/jkb/scratch/data/9827_2#49.bam
-rw-r--r-- 1 jkb team117 6188816754 Aug 19 14:33 /nfs/users/nfs_j/jkb/scratch/data/9827_2#49.sam.gz
```

Die BAM DIE! ;-)

[ Yes, this is a joke really - the single threaded CPU cost still makes BAM the better format overall. ]

hts.c
```diff
@@ -428,6 +428,9 @@ htsFile *hts_open_format(const char *fn, const char *mode, const htsFormat *fmt)
     if (strchr(mode, 'w') && fmt && fmt->compression == bgzf) {
         if (fmt->format == sam || fmt->format == vcf || fmt->format == text_format)
             *mode_c = 'z';
+    } else if (!*mode_c && !strchr(smode, 'z') &&
+               strlen(fn) > 3 && strcmp(fn+strlen(fn)-3, ".gz") == 0) {
+        *mode_c = 'z';
```
jmarshall: This should be in `sam_open_mode()` instead. Samtools uses that function so adding "z" based on the filename in `sam_open_mode()` should also have the effect of permitting samtools to output compressed SAM.

jkbonfield: Thanks.  I'll take a look.

I confess I get a bit confused at times as to which function is the official blessed way to open files.  The htslib test_view program for example uses hts_open to read and write files, and this has been working with "z" for ages.

I'd be surprised if z wasn't permitted with sam_open already.  This change wasn't to add that, but to *automatically* add it if the filename ended in .gz.

jmarshall: Perhaps it's slightly unfortunately named: `sam_open_mode()` doesn't open files!

Instead it constructs an “open mode” string based on the filename given to it, to be later given as the mode parameter to `hts_open`/`sam_open` of whatever flavour.

TODO it's got no doco blurb in _htslib/sam.h_…

jkbonfield: Updated `sam_open_mode`, `sam_open_mode_format` and "in for a penny, in for a pound" `hts_parse_format` too.

This means we can now do e.g. "samtools view -O sam.gz,level=1 foo.bam | whatever".  I'm not sure that's actually helpful, but it's consistent.

sam.c
```diff
@@ -1948,33 +2569,98 @@ int sam_read1(htsFile *fp, sam_hdr_t *h, bam1_t *b)
...
+                    fp->fp.bgzf->seeked = 0;
+                }
+            }
+
+            if (!fd->h) {
+                fd->h = h;
+                fd->h->ref_count++;
+                // Ensure hrecs is initialised now as we don't want multiple
+                // threads trying to do this simultaneously.
+                sam_hdr_fill_hrecs(fd->h);
```
daviesrob: Error check needed here.

sam.c
```diff
@@ -1948,33 +2569,98 @@ int sam_read1(htsFile *fp, sam_hdr_t *h, bam1_t *b)
...
+            if (!gb) {
+                hts_tpool_result *r = hts_tpool_next_result_wait(fd->q);
+                fd->curr_bam = gb = (sp_bams *)hts_tpool_result_data(r);
+                hts_tpool_delete_result(r, 0);
+            }
+            if (!gb)
+                return fd->errcode ? -2 : -1;
+            bam1_t *b_array = (bam1_t *)gb->bams;
+            if (fd->curr_idx < gb->nbams)
+                bam_copy1(b, &b_array[fd->curr_idx++]);
```
daviesrob: Error check needed here.

sam.c
```diff
@@ -1948,33 +2569,98 @@ int sam_read1(htsFile *fp, sam_hdr_t *h, bam1_t *b)
...
-int sam_format1(const sam_hdr_t *h, const bam1_t *b, kstring_t *str)
+static int sam_format1_append(const bam_hdr_t *h, const bam1_t *b, kstring_t *str)
 {
     int i;
     uint8_t *s, *end;
     const bam1_core_t *c = &b->core;
 
-    str->l = 0;
+    if (c->l_qname == 0)
+        abort();
```
daviesrob: Better to return an error here.

sam.c
```diff
@@ -2176,27 +2870,95 @@ int sam_write1(htsFile *fp, const sam_hdr_t *h, const bam1_t *b)
...
+                    if (!(gb->bams = calloc(NB, sizeof(*gb->bams)))) return -1;
+                    gb->nbams = 0;
+                    gb->abams = NB;
+                    gb->fd = fd;
+                    fd->curr_idx = 0;
+                    fd->curr_bam = gb;
+                }
+            }
+
+            bam_copy1(&gb->bams[gb->nbams++], b);
```
daviesrob: Error check needed here.

test/test.pl
```diff
@@ -864,7 +864,10 @@ sub test_logging
   print "$test:\n";
   print "\t$cmd\n";
   my ($ret,$out) = _cmd($cmd);
-  if ( $ret ) { failed($opts,$test); }
+  if ( $ret ) {
+      print $out;
```
daviesrob: Useful :smile:

sam.c
```diff
@@ -1924,6 +1937,614 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
...
+    if (!fd)
+        return NULL;
+
+    fp->state = fd;
+    fd->fp = fp;
+
+    return fd;
+}
+
+// Destroys the state produce by sam_state_create.
```
daviesrob: This comment might be better two lines down, after the function prototypes.

sam.c
```diff
@@ -1924,6 +1937,614 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
...
+                    i++;
+                if (i < gl->data_size)
+                    i++;
+
+                if (fp->format.compression == bgzf) {
+                    if (bgzf_write(fp->fp.bgzf, &gl->data[j], i-j) != i-j) {
+                        pthread_mutex_lock(&fd->command_m);
+                        fd->errcode |= BGZF_ERR_IO;
+                        pthread_mutex_unlock(&fd->command_m);
+                        fd->errcode |= BGZF_ERR_IO;
```
daviesrob: Odd duplicated line here.  Presumably the one that updates `fd->errcode` with the mutex locked is the right one.  There are other example of this elsewhere.  Also, `fd->errcode` is not consistently locked everywhere.  Maybe a static inline to access it safely would be a good idea?

Finally, it sometimes gets `errno` values (e.g. `EIO`), and others (like here) `BGZF_ERR_IO`.  Which should it be?

sam.c
```diff
@@ -1924,6 +1937,614 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
...
+    // Use a block of SAM strings we had earlier if available.
+    pthread_mutex_lock(&fd->lines_m);
+    if (fd->lines) {
+        gl = fd->lines;
+        fd->lines = gl->next;
+    } else {
+        gl = calloc(1, sizeof(*gl));
+        if (!gl) {
+            fd->errcode = ENOMEM;
+            return NULL;
```
daviesrob: Mutex needs to be unlocked here.

sam.c
```diff
@@ -1924,6 +1937,614 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
...
+        }
+
+        l->serial = fd->serial++;
+        //fprintf(stderr, "Dispatching %p, %d bytes, serial %d\n", l, l->data_size, l->serial);
+        hts_tpool_dispatch(fd->p, fd->q, sam_parse_worker, l);
+    }
+
+    hts_tpool_dispatch(fd->p, fd->q, sam_parse_eof, NULL);
+
+    // At EOF, but there is potential for the caller to then do another seek.
```
daviesrob: Seek isn't supported.

sam.c
```diff
@@ -1924,6 +1937,614 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
...
+    if (l) {
+        l->next = fd->lines;
+        fd->lines = l;
+    }
+    free(line.s);
+
+    return NULL;
+}
+
+// Runs in the main thread.
```
daviesrob: Runs in its own thread...

sam.c
```diff
@@ -1924,6 +1937,614 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
...
+            pthread_cond_signal(&fd->command_c);
+            ret = -fd->errcode;
+            if (!ret) hts_tpool_wake_dispatch(fd->q); // unstick the reader
+            pthread_mutex_unlock(&fd->command_m);
+
+            if (fp->is_write) {
+                // Dispatch the last partial block.
+                sp_bams *gb = fd->curr_bam;
+                if (!ret && gb && gb->nbams > 0)
+                    hts_tpool_dispatch(fd->p, fd->q, sam_format_worker, gb);
```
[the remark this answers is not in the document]
jkbonfield: Only on malloc failure.  I'll add a check

sam.c
jkbonfield: I'll use EIO, ENOMEM, etc and avoid the BGZF_ERR codes as they're a bit unhelpful.

However arguably it's redundant as there's no way to return errors other than pass and fail from the high level functions anyway.  Still, it's there now so I may as well keep it and we can figure out how best to deal with global errors later.  (Maybe just set the global errno at the point of top-level failing would be best.)

I did try to figure out the return codes for some of our functions, but they're mostly undocumented with <0 being error, despite actually having different values for different things.

sam.c
```diff
@@ -1924,6 +1937,614 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
...
+        memcpy(l->data, line.s, line_frag);
+
+        l->data_size = line_frag;
+        size_t nbytes;
+    longer_line:
+        if (fp->is_bgzf)
+            nbytes = bgzf_read(fp->fp.bgzf, l->data + line_frag, l->alloc - line_frag);
+        else
+            nbytes = hread(fp->fp.hfile, l->data + line_frag, l->alloc - line_frag);
+        l->data_size += nbytes;
```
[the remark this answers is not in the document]
jkbonfield: Moved, but it's not a bug - we break immediately and it's unused after that anyway.

jkbonfield:
Thanks for the feedback.  The error cases should now be more robust.  I found and fixed an issue in the thread pool too.

There are still some problems with bgzf/BAM here which can be fixed in a separate PR.  My method of testing was:

```
(for m in `seq 500000 -1000 1`;do echo -n "$m      ";ulimit -v $m;./test/test_view -@4 ~/scratch/data/9827_2#49.1m.sam 2>/dev/null > /dev/null;echo $?;done)
```

On BAM files this crashes at some points due to poor error recovery.

sam.c
```diff
@@ -1924,13 +1938,655 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
...
+        //
+        // However this is an API change so for now we copy.
+
+        char *nl = strchr(cp, '\n');
+        nl = nl ? nl : cp_end;
+        if (*nl) *nl++ = '\0';
+        kstring_t ks = {nl-cp, gl->alloc, cp};
+        if (sam_parse1(&ks, fd->h, &b[i]) < 0) {
+            sam_state_err(fd, EIO);
+            return NULL;
```
daviesrob: This could leak a lot of memory on error:

- `sp_lines *gl` passed in as input
- `gl->data`
- `sp_bams *gb`
- `gb->bams`
- `gb->bams[i]->data`

All these `return NULL`s should go to some clean-up code instead, which should either free the data, or possibly put `gl` and `gb` on to the appropriate free lists in the `SAM_state` so they can be reused or (more likely) cleaned up later.

sam.c
```diff
@@ -1924,13 +1938,655 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
...
+    }
+    gl->serial = gb->serial;
+    gl->next = NULL;
+
+    kstring_t ks = {0, gl->alloc, gl->data};
+
+    for (i = 0; i < gb->nbams; i++) {
+        if (sam_format1_append(fd->h, &gb->bams[i], &ks) < 0) {
+            sam_state_err(fd, EIO);
+            return NULL;
```
daviesrob: Similar problem here - something needs to take care of `gl` and `gb` to prevent them from leaking.

sam.c
```diff
@@ -1924,13 +1938,655 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
...
+static void *sam_dispatcher_write(void *vp) {
+    htsFile *fp = vp;
+    SAM_state *fd = fp->state;
+    hts_tpool_result *r;
+
+    // Iterates until result queue is shutdown, where it returns NULL.
+    while ((r = hts_tpool_next_result_wait(fd->q))) {
+        sp_lines *gl = (sp_lines *)hts_tpool_result_data(r);
+        if (!gl)
+            continue;
```
daviesrob: Shouldn't it give up here, as if `gl == NULL`, `sam_format_worker()` has likely failed?

(Aside: It might make this easier to follow if the `hts_tpool_next_result_wait()` calls had comments to say which function(s) the result may have come from.)

jkbonfield: I honestly cannot remember why I made it continue on empty results now, as it's lost on the mists of time.  That dates back to the original PR from last year.

I'll try changing it and see if anything breaks!  It's not obvious why it would be beneficial to continue.

htslib/kstring.h
```diff
@@ -140,8 +140,8 @@ static inline int ks_resize(kstring_t *s, size_t size)
 		char *tmp;
 		kroundup_size_t(size);
 		tmp = (char*)realloc(s->s, size);
-		if (!tmp)
-			return -1;
+		if (!tmp && size)
```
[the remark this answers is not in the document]
daviesrob: Actually, this can't happen if input `size` is 0, because `s->m` and `size` are both unsigned so `s->m < size` will always be false.  However it can if `size` is greater than `SIZE_MAX / 2` as `kroundup_size_t(size)` then overflows and sets `size` to zero.  So this should actually be an out-of-memory error, given that you can never allocate a buffer that big.

This kroundup overflow is likely to be a problem elsewhere too.

sam_internal.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include "htslib/sam.h"
+
+// Used internally in the SAM format multi-threading.
+int sam_state_destroy(samFile *fp);
+int sam_set_thread_pool(htsFile *fp, htsThreadPool *p);
+int sam_set_threads(htsFile *fp, int nthreads);
+
+#ifdef __cplusplus
+}
```
jmarshall: This is missing its `extern "C" {` companion :smile:

daviesrob: Thanks.  Strictly I don't think it's needed given that this is an internal header and trying to compile the files that use it with a c++ is likely to be a heroic exercise, but I've added it for completeness.

jmarshall: That's true too. I checked the other *_internal.h headers before deciding between suggesting adding `extern "C" {` or removing this one…

PR #917, opened 2019-08-22
Validate record range when constructing BAI/CSI indexes

jmarshall:
Inserting malformed ranges into the index leads to a broken index for which `hts_itr_query()` produces incorrect results for nearby query regions. This pretty much can't happen for SAM/BAM/CRAM (the CIGAR calculation produces a length >= 0), but does happen for VCF/BCF when the input record has an invalid INFO/END field that's < POS.

In particular, this will make `bcftools index` refuse to index files where INFO/END is being used for a different purpose — see samtools/hts-specs#425. Previously it would happily produce an index that didn't work properly.

Instead of this PR, it would be possible to validate that `rlen >= 0` in the _vcf.c_ code that computes `rlen` from an input INFO/END field value. However that would mean that HTSlib would refuse to read in these delly/gnomad/etc records at all, which would not be reasonable. Better just to refuse to build an index, as indexing is what gets broken by these misused INFO/END fields.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -1583,6 +1583,11 @@ int hts_idx_push(hts_idx_t *idx, int tid, int beg, int end, uint64_t offset, int
         hts_log_error("Unsorted positions on sequence #%d: %d followed by %d", tid+1, idx->z.last_coor+1, beg+1);
         return -1;
     }
+    else if (end < beg) {
+        // Malformed ranges are errors. (Empty ranges (beg==end) are unusual but acceptable.)
+        hts_log_error("Invalid record on sequence #%d: end %d < begin %d", tid+1, end, beg+1);
+        return -1;
+    }
     if ( tid>=0 )
     {
         if (idx->bidx[tid] == 0) idx->bidx[tid] = kh_init(bin);


daviesrob:
Thanks.  I think refusing to read the data would be going too far, so this is a reasonable solution.  We'll have to make sure we deflect any complaints about refusing to index to the projects that produce the out-of-spec data.

jmarshall:
Good point about deflecting complaints. Maybe a NEWS entry? Workshopping the error message?

    [E::hts_idx_push] Invalid record on sequence #1: end 1386324 < begin 1386372
    Note: This can be caused by INFO/END fields that are not compliant with the VCF specification
    index: failed to create index for "gnomad_v2_sv.sites.vcf.gz"

daviesrob:
Maybe we should add something to `vcf.c` where it calls `hts_idx_push()` (three places) with a vcf-specific version of the error message?  That way you wouldn't be confused by a "VCF" error if you attempted to add a negative-length region for some other file type (the actual route how this may happen is left to the imagination).

jmarshall:
Those three places implement online indexing and `bcftools index foo.bcf`.
For `bcftools index bar.vcf.gz` the pushing actually occurs in `tbx_index()`.

Perhaps a better route would be to record in the `hts_idx_t` what kind of file is being indexed (when that information is available). e.g. `htsFormatCategory` and/or `htsExactFormat` could be recorded by shoehorning them (appropriately `<<` shifted) into `hts_idx_init()`'s `fmt` argument or `hts_idx_init2()` could be invented with additional parameters.

PR #918, opened 2019-08-27
Various minor tidy ups

jmarshall:
Various minor tidy ups, as described in the commit messages.

Avoid some reserved `_FOO` macro names, rename recently-added test data files so they are git-ignored and `make testclean`ed, reduce _htslib/faidx.h_'s includes from `<inttypes.h>` to `<stdint.h>` (htslib/samtools/bcftools test suites all still pass, and this \#include was a recent addition on **develop**; cf https://github.com/samtools/samtools/issues/1079#issuecomment-514565628).


daviesrob:
Looks good, thanks.

PR #919, opened 2019-08-28
Added some missing bam editing functions.

jkbonfield:
I use these in the samtools sort minimiser and also some in Crumble.  Samtools fastq also has its own code for similar things.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -661,6 +661,102 @@ static int bam_write_idx1(htsFile *fp, const bam1_t *b) {
     return ret;
 }
 
+/*!
+ * @abstract
+ * Extracts the sequence (in current alignment orientation) from
+ * a bam record and places it in buf, which is nul terminated.
+ *
+ * @param b     The bam structure
+ * @param buf   A buffer at least b->core.l_qseq+1 bytes long
+ */
+void bam_to_seq(bam1_t *b, char *buf) {
+    int i;
+    uint8_t *seq = bam_get_seq(b);
+    for (i = 0; i < b->core.l_qseq; i++)
+        buf[i] = seq_nt16_str[bam_seqi(seq, i)];
+    buf[i] = 0;
+}
+
+/*!
+ * @abstract
+ * Writes a new sequence, of length b->core.l_qseq, to a BAM record.
+ *
+ * If a sequence of a new length is required the caller must first make
+ * room for it by updating the bam1_t struct.
+ *
+ * @param b     The bam structure
+ * @param buf   A buffer at least b->core.l_qseq bytes long
+ */
+void seq_to_bam(bam1_t *b, char *buf) {
+    int i;
+    uint8_t *seq = bam_get_seq(b);
+    for (i = 0; i < b->core.l_qseq; i++)
+        bam_set_seqi(seq, i, seq_nt16_table[(unsigned char)buf[i]]);
+}
+
+/*!
+ * @abstract Reverse complements a BAM record.
+ *
+ * It's possible to do this inline, but complex due to the 4-bit sequence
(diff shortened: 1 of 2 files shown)


sam.c
```diff
@@ -661,6 +661,102 @@ static int bam_write_idx1(htsFile *fp, const bam1_t *b) {
...
+    }
+    if (i ==j)
+        seq[i] = comp[(unsigned char)seq[i]];
+
+    seq_to_bam(b, seq);
+
+    if (seq != seq_)
+        free(seq);
+
+    b->core.flag ^= 0x10;
```
jmarshall:
```suggestion
    b->core.flag ^= BAM_FREVERSE;
```

sam.c
```diff
@@ -661,6 +661,102 @@ static int bam_write_idx1(htsFile *fp, const bam1_t *b) {
...
+ * @abstract
+ * Writes a new sequence, of length b->core.l_qseq, to a BAM record.
+ *
+ * If a sequence of a new length is required the caller must first make
+ * room for it by updating the bam1_t struct.
+ *
+ * @param b     The bam structure
+ * @param buf   A buffer at least b->core.l_qseq bytes long
+ */
+void seq_to_bam(bam1_t *b, char *buf) {
```
jmarshall:
```suggestion
void seq_to_bam(bam1_t *b, const char *buf) {
```

sam.c
```diff
@@ -661,6 +661,102 @@ static int bam_write_idx1(htsFile *fp, const bam1_t *b) {
...
+ *
+ * @param b     The bam structure
+ * @param buf   A buffer at least b->core.l_qseq+1 bytes long
+ */
+void bam_to_seq(bam1_t *b, char *buf) {
+    int i;
+    uint8_t *seq = bam_get_seq(b);
+    for (i = 0; i < b->core.l_qseq; i++)
+        buf[i] = seq_nt16_str[bam_seqi(seq, i)];
+    buf[i] = 0;
```
jmarshall:
```suggestion
    buf[i] = '\0';
```

jmarshall:
One of my long-stalled PRs has something somewhat similar, suggesting a table for helping with complementing bases:

```c
/*! @abstract Table for complementing a 4-bit encoded nucleotide.
*/
extern const int seq_nt16_complement[];
```

It wasn't very clear reading this which way around `bam_to_seq` and `seq_to_bam` are (and the latter doesn't exactly fit the prefix naming conventions) — maybe we can come up with better function names, e.g.,

```
void bam_decode_seq(bam1_t *b, char *seq);
void bam_update_seq(bam1_t *b, const char *newseq);
int bam_revcomp(bam1_t *b);
```

YMMV as to whether it's important to say _reverse_ in `bam_complement()` :smile:

It is also important to document exactly what gets (reverse) complemented by the latter function — compare samtools/htsjdk#1375.

jkbonfield:
It actually started off as `bam_reverse_complement` and I change it because I expected people to whinge about the long function names. hah.  (You can see the original code in the minimiser PR still.)

Also, isn't bam_to_seq and seq_to_bam kind of obvious due to the use of "_to_".  It's left to right.
If we keep with bam prefix then it becomes a bit unclear IMO as we have to somehow imply "bam->seq" and "bam<-seq" in English.

jmarshall:
Despite the `_to_` but aided by the parameter being identically `char *buf` in both cases, what I said is that I didn't find it very obvious. Hence the suggestion of `bam_decode_seq` and `bam_update_seq` as being both even more obvious as to what they do and in line with the naming of other similar existing `bam1_t` manipulation functions.

jkbonfield:
Closing as vetoed.

PR #923, opened 2019-09-04
More documentation + bam_copy1() and bam_dup1() memory checks

daviesrob:
- Remove #ifdef BAM_NO_ID, which doesn't seem to be used and would probably break things for anyone who did define it.

 - Add doxygen documentation for some bam1_t related functions.

 - Add allocation failure checks to bam_copy1() and bam_dup1(). bam_copy1() gets annotated HTS_RESULT_USED to encourage checking of the return value.


valeriuo:
Merged into develop

Issue #129, opened 2014-09-09
Using bgz extension instead of gz for bgziped files


hguturu:
Would make it very clear what type of file is being dealth with. Additionally, can avoid issues with things like vim being able to read  bgziped files saved with .gz extensions and resaving them as regular gzip.

I believe minimal changes to bgzip.c to autoappend .bgz instead of .gz and tabix.c to auto detect files with .bgz extensions.

I can submit a pull request to these if needed.

wkretzsch:
bgzipped files are gzip compliant.  Why not tell vim to gzip everything with bgzip instead?

hguturu:
vim was just an example, not the motivation. The more general motivation is
bgzip is slightly different than gzip and there doesn't appear to a reason
to conflate the two. Especially since sometimes its preferred to just have
pure gzip (no index will be built/the file has nothing to do with genomics).

wkretzsch:
Yep, you're right. Even different implementations of the same standard (for example .lz and .lzma [here](https://en.wikipedia.org/wiki/List_of_archive_formats#Compression_only)) appear to get different file endings.

jrandall:
For `bgzip`, we could make the ".bgz" extension an optional (non-default) filename. 

For the rest of `htslib`/`samtools`/`bcftools`, we should make sure that ".bgz" files are recognized as bgzip for input and output filenames.

davidpratella:
.bgz are still not recognized when using bgzip. 
I needed to modify all my file name from .bgz to gz, which is not an intuitive thing since i haven't found any information about this little detail...
I guess i'm not the only one to download file with ".bgz" extension from database (e.g. gnomAD vcf).
And i guess it's an easy modification to do 

Thank you

pd3:
Why do you need to rename your files? The suffix name does matter for the operation of the program.

davidpratella:
i have to rename them from ".bgz" to ".gz" so bgzip can work, otherwise i get this error "unknown suffix -- ignored "
but since the tool is called bgzip (and not gzip), it would make sense (for me) that the extension could be ".bgz"

jkbonfield:
That's likely a bug in bgzip as it ought to be using the magic number instead of filename.

[...]

PR #927, opened 2019-09-17
Make bgzip -d work with non-.gz extensions (and minor nearby tidy ups)

jmarshall:
Without commenting on whether `.bgz` is an extension that should be encouraged :smile:, this fixes the usability problem in #129 by making the following work:

```
bgzip -d gnomad.exomes.r2.1.1.sites.vcf.bgz
```

As noted on that issue, `tabix gnomad.exomes.r2.1.1.sites.vcf.bgz 1:12000-13000`  already works fine.

Also various minor tidy ups nearby that were noticed in the course of doing this.


bgzip.c
```diff
@@ -312,8 +296,17 @@ int main(int argc, char **argv)
             }
             else {
                 const int wrflags = O_WRONLY | O_CREAT | O_TRUNC;
+                char *name = argv[optind];
```
daviesrob: Is there a danger here that if you tried to `bgzip -d -f something.bam.csi` then you might overwrite `something.bam` unintentionally?  Or given that it's necessary to use the `-f` option for this to happen, do we not worry as it was explicitly requested?

jmarshall: Good point, I guess that is a new danger. I guess the possible answers are

1. Whitelist known compression extensions, namely `.gz` and `.bgz` (and …?)

2. Blacklist known index extensions, e.g., `.bai`, `.csi`, etc

3. Accept any extensions like this code does, and depend on the `-f` failsafe; if they say yes to that, they really do mean it.

4. Status quo, which is basically "only `.gz` is whitelisted"

Hmmm… option 3 and change to one of the others if problems ensue?
I guess (4) suggests that (1) is the minimal change…

daviesrob: Sticking to `.gz` and `.bgz` would be safest.  Or we could make it politely refuse if the output file exists even if you've used `-f` _unless_ one of the known-safe extensions is used (and allow something like `-f -f` for "I really really want to do this").

jmarshall: PR updated to do (1) essentially — asks you both “are you sure about this funny extension” and “are you sure about overwriting” as appropriate, and both questions can be suppressed with enough `-f -f`.

jmarshall:
Thanks for the review — updated accordingly.

Also added one more nearby bug fix. It's a doozy!

The code currently works due to a superposition of two pieces of luck and the pooled allocator not memsetting the newly `pool_free()`ed block to garbage. The bug fix is needed for an upcoming PR (to implement `@SQ-AN`; um, hope nobody else has been working on that :thinking:).

daviesrob:
As far as I know, no-one is looking at `@SQ-AN`, so feel free to carry on.

daviesrob:
Fixes now merged, thanks for all the hard work :smiley:

Issue #490, opened 2017-03-14
lzma, bz2 dependency


willmclaren:
Does support for lzma and bz2 have to be on by default?

Adding these dependencies has broken a *lot* of Ensembl's installation code and Travis and Docker builds.

It appears the fix is not a simple one, as we are now required to run `autoheader` and `autoconf`, which are not default packages in many setups, before running `./configure` and then `make`. 

Previously you could `git clone`, `make` and be off and running.

htslib has gained so much traction partly because of its simplicity, ease of installation and lack of external dependencies.

lairdm:
I've just checked and this issue would be solved by the maintainers running autoreconf and including aclocal.m4, config.h.in, and configure with the distro. It's been far too many years since I've played with the auto* tool chain, but these scripts and templates are usually included in a package and would allow users to simply ./configure the package to build it without the auto* steps first. I already have a fork of the repo which I'm not ready to merge back so I can't quickly make this change and do a PR, but if anyone else wanted to quickly do this change it'd solve/simplify part of the build issue quite quickly.

mp15:
I have tested on Ubuntu Precise (coming up to 5 years old) and you can still git clone and make and be off and running. Could you be a bit more specific about your configuration so we know what the problem is? You still don't have to run configure unless you want to disable our default dependancies (zlib, lzma, bz2 and curses) or to enable libcurl (recommended) or s3 support. I know Ensembl has a lot of older boxes, but none of them are due to live past April AFAIK?

As for LZMA and bz2 they are required for CRAM decoding, if you don't have them you simply cannot decode a lot of the CRAMs out there. We do not want to put out a library which cannot do the task it is designed for by default. Given that they're installed by default on most distributions which are still within the security support window I would be inclined to say is is fair to assume that -llzma and -lbz2 will work in 95% of cases (sadly not xz on OS X but meh).

willmclaren:
@mp15 the issue is not so much users' systems as Travis and the like. See this failed build as an example:

https://travis-ci.org/Ensembl/ensembl-variation/jobs/210885028

It's based on Precise too. This will also happen to anything else using a bare-bones virtualised environment, such as Docker.

We've reverted to using 1.3.2 for now (though there will probably be some places we miss this change and still get bitten).

I guess we shouldn't have been depending on master/develop retaining the same set of dependencies in the first place.

jkbonfield:
[...]

PR #929, opened 2019-09-18
Adds CRAM 3.1 support (experimental)

jkbonfield:
This is here as demonstration of how to add this into Htslib (in addition to the existing support in Scramble).

It uses htscodecs as a submodule, but compiles and links the code directly into htslib so there isn't a second library hanging around.  This now means scramble and htslib share the same codec implementation and we don't accidentally only fix bugs in one place.

Unlike Scramble there is nothing to actively warn that this is a technology test of CRAM 3.1 and not an official GA4GH standard.  I could add this, but I wasn't envisioning merging it until it's an adopted  standard anyway.  (It's sufficient to be able to state there is support for this format in Samtools and hence maybe get more people looking into assessing the new spec.)


jkbonfield:
Bug-fixed windows / macos (limited stack sizes) and rebased off develop.

Some examples:

Develop vs this PR using an existing local v3.0 CRAM.  Tests read this file (a 3.0 file of 10 million NovaSeq records) and convert it to a new BAM/CRAM file and then read it back again to get encode and decode times.

| Branch      | CRAM vers | Size            | Enc(s) | Dec(s)
| ------------- | --------------- | -------------- | ------ |-----
| develop     | (as BAM)    | 576789115 | 53.2 | 4.3
| develop     | 3.0              | 208635462 | 40.0 | 12.9
| htscodecs | 3.0              | 208392952 | 38.5 | 12.0
| htscodecs | 3.1 normal  | 175696160 | 36.4 | 10.8
| htscodecs | 3.1 small     | 166172362 | 88.8 | 39.0
| htscodecs | 3.1 archive  | 162697852 | 123.2 | 38.5

jkbonfield:
**Happy Birthday** :-)

Given 1.11 is imminent and this PR has missed two (lengthy!) releases already, I am bumping this and request please that it is reviewed *early* in the 1.12 release cycle.   If not I'll continue with my previous assertion that all unreviewed PRs of mine after 1 year get merged anyway under the implicit assumption that noone objects. ;-)

daviesrob:
The main issue with this pull request is how we manage the new forced dependency on the `htscodecs` library.  It at least needs additional configure options to allow linking against an external htscodecs, as [distributors like Debian will want to include it that way](https://packages.debian.org/sid/libhtscodecs-dev).  It also currently forces a dependency on autotools which may be unpopular with projects that embed htslib and [expect to be able to just use `make`](https://github.com/samtools/htslib/issues/490); and it uses recursive make, which is known to cause problems.

This will also need to be checked carefully to ensure it doesn't introduce any regressions, especially on memory safety and handling invalid inputs.  Given the size of this change, that might take a while.

jkbonfield:
The main purpose for htscodecs as far as I am concerned is to avoid duplication between io_lib and htslib.  I need both because io_lib is much faster moving and has more experimental features (eg it's had CRAM 3.1 and 4 for several years).

It doesn't necessarily have to be explicitly done as a separate library, or submodule.  Eg the files could be copied in (klib style), provided we're committed to make changes in the upstream project instead of just forking and diverging.

Htscodecs has been through rounds of sanitizers and fuzzing, but yes it may not be as robust as the OSSFuzz hammering.  I know it'll take time, but that's why I'm asking for it to be consider at the *start* of the next release cycle.  It's always been "this is too complex, so we'll punt it to the next one" which basically means never unless we act early on.  Hence this reminder.

jkbonfield:
Also, to be frank do we want to encourage continued "cp -r htslib ." cargo-cult style of development?  It's a total nightmare for getting bug fixes out there.  As long as we make it easy for people to do things the Right Way then it's not so bad.  Autotools also isn't that bad from a user perspective - you don't have to install anything (unlike cmake or the myriad of other flavour of the month make replacements).  If it's built using --disable-shared then most of the libtool monstrosities go away. :)

jkbonfield:
Rebased and added some updates to bring this in line with io_lib's CRAM 4 implementation.  I've verified that scramble and test_view can now read and write each others files with compatibility between them.  CRAM 4 is even more experimental than 3.1, but we may as well at least be consistent.   This now also brings this PR in line with the draft cram4.tex specification, which had a few tweaks and adjustments since Sep 2019 (when this PR was made).

The old branch, before rebasing, can be found here: https://github.com/jkbonfield/htslib/tree/htscodecs.old
(That may be useful if you wish to replay the latest commits on this against your own rebase of the original PR.)

jkbonfield:
(Compare with 2nd to top post above, but note these are different files than before.)

Current htscodecs timings - ie this PR as it now stands, with updated htscodecs submodule too.

10 million NovaSeq records:

| Branch        | CRAM vers       | Size           | Enc(s) | Dec(s)            
| ------------- | --------------- | -------------- | ------ |-----              
| develop       | BAM/libdeflate  | 539562772      | 44.9   | 5.3               
| develop       | 3.0             | 223704658      | 32.9   | 13.3              
| htscodecs     | 3.0             | 223876657      | 31.7   | 10.6              
| htscodecs     | 3.1             | 186554165      | 34.4   | 9.4               
| htscodecs     | 4.0             | 182186308      | 32.2   | 10.7              
| htscodecs     | 3.1 small       | 175986355      | 78.4   | 37.8              
| htscodecs     | 4.0 small       | 170854620      | 75.0   | 36.1              
| scramble      | 3.0             | 223761677      | 33.4   | 10.3              
| scramble      | 3.1             | 186429613      | 33.8   | 8.3               
| scramble      | 4.0             | 182080941      | 32.3   | 8.2               

It's unclear why scramble's 3.1 is comparable encode speed to 3.0, but not htslib's.  It's of note though that scramble's system time is vastly smaller, which is mainly due to it using mallopt to workaround inefficiencies in glib's malloc (which htslib does not do).

On 10 million recs from the HiSeq 2k data (9827_2#49.bam):

| Branch        | CRAM vers       | Size           | Enc(s) | Dec(s)            
| ------------- | --------------- | -------------- | ------ |-----              
| develop       | BAM/libdeflate  | 1163050644     | 50.6   | 7.9               
| develop       | 3.0             | 573365519      | 36.5   | 14.4              
| htscodecs     | 3.0             | 573317290      | 34.4   | 12.7              
| htscodecs     | 3.1             | 551563284      | 39.1   | 13.3              
| htscodecs     | 4.0             | 549018451      | 38.3   | 11.9              
| htscodecs     | 3.1 small       | 388439096      | 111.1  | 63.3              
| htscodecs     | 4.0 small       | 385983116      | 108.2  | 63.5              
| scramble      | 3.0             | 573349740      | 35.3   | 11.0              
| scramble      | 3.1             | 551636226      | 39.5   | 11.1              
| scramble      | 4.0             | 549116518      | 38.3   | 11.5

jkbonfield:
Some more benchmarks, this time 10 million NovaSeq records:

```
CRAM 3.0        Read(s)   Read/Write(s)                                         
---------------------------------------                                         
Develop         13.43     40.11                                                 
Htscodecs       10.60     38.25                                                 
```

I post these as data to consider for how to progress.  @daviesrob  I know your preference is to keep cram/rANS_static.c and have htscodecs as an optional build.  IMO this is a big mistake.  We don't need more variants and ways to build sub-optimal libraries.  Either we support CRAM, or we don't.  I'd rather us have a `--disable-cram` option that to have two different choices of codecs.  It's more work for us to maintain and I see zero benefit.  (We've been down this road before with knet + curl as choices, and it's caused many more problems than it solved.)  It'll also lead to poor comparisons in benchmarks run by other people.

The other thing is there is a catch-22 issue here.  CRAM 3.1 spec has gained minimal support elsewhere, partly because it's an abstract notion and isn't supported by the main packages.  However this PR has basically been ignored for well over a year because CRAM 3.1 is experimental and hasn't been adopted by GA4GH yet.  The only way to break this cycle is to *lead* the community, not sit back and hope it'll all just go away.  That way just gives power to the likes of MPEG fostering commercial alternatives that may be able to legitimately compare against CRAM 3.0 and conveniently pretend 3.1 doesn't exist.

As you may gather, I'm getting somewhat annoyed by the endless delays.  *PLEASE* consider merging this.  We're more or less half way through our mythical 3 month release cycle.  I don't wish there to be yet another excuse of "we can't merge it this late" as happened for the last 2 releases.  The sooner it is merged, the sooner it gets into the standard fuzzing and testing regimes and the more confident we'll be that it is solid.

cram/cram_encode.c
```diff
@@ -822,89 +782,147 @@ static int cram_encode_slice_read(cram_fd *fd,
...
+        }
+        if (fd->level > 6) {
+            qmethod  |= (1<<FQZ_c) | (1<<FQZ_d);
+            qmethodF |= (1<<FQZ_c) | (1<<FQZ_d);
+        }
+    }
 
+    pthread_mutex_lock(&fd->metrics_lock);
+    for (i = 0; i < DS_END; i++)
+        fd->m[i]->stats = c->stats[i];
```
daviesrob, line 858: This is leading to threading problems.  `fd->m[i]` is shared between all threads, so they fight over which `c->stats[i]` should be in there.  When the [pointer is used in `cram_compress_block2()`](https://github.com/samtools/htslib/blob/18f16253/cram/cram_io.c#L1958) it may get stats for the wrong container, or in the worst case the memory may have been released already resulting in a use-after-free bug.

Either `cram_compress_block2()` needs an extra parameter for `c->stats[type]` (or `c->stats[type]->nvals`); or the adjustments to `method` need to be done in this function so there is no need to pass the stats to `cram_compress_block2()`.  After that I think it should be possible to remove `cram_metrics::stats` as I don't think it's used anywhere else.

jkbonfield: Wrong container doesn't matter much as the metrics are on-going trends that aggregate data across containers.  However you're correct that having a pointer to it is bad as they may go out of scope.

I think the correct answer is to set a flag (or boost a counter) in metrics so it doesn't need the pointer, but metrics analysis can still use it for biasing which codecs to try.

jkbonfield: This definitely looks like a real problem, although I've tried very hard to recreate it and with 100s of tests still cannot.

Anyway, I've now replaced it with a boolean flag which is all we really need.

jmarshall:
> It uses htscodecs as a submodule, but compiles and links the code directly into htslib so there isn't a second library hanging around. This now means scramble and htslib share the same codec implementation and we don't accidentally only fix bugs in one place.

The question of how to incorporate building htscodecs into HTSlib's build infrastructure comes down to this question about whether there should be an htscodecs entity that is versioned independently from htslib, whether to facilitate distribution packagers making such an independent library, and how to make the codec source code available — both in a Git checkout and in an HTSlib source tarball.

FWIW I agree that avoiding having an independently versioned second library for this is the desirable and simple approach. The purpose of the separate htscodecs repository is simply to share the code between HTSlib and e.g. io_lib, not to have them both depend on another separate moving target.

I'm not usually a fan of Git submodules: they make the subproject operate in absolute lockstep with the superproject, so are not a good fit for e.g. the samtools/bcftools relationship with htslib (in which many adjacent versions of htslib will do for a particular samtools/bcftools commit). However for this case, in which htslib wants to share a particular htscodecs commit with scramble, a Git submodule is **exactly** what is desired.

Now, **how to build the codecs** and link their object code directly into _htslib.a/.so_? HTSlib has its own very capable build system already, so the simplest thing (from HTSlib's point of view) would be to have the submodule's _.c_ files built using rules in HTSlib's _Makefile_. As the htslib git checkout would reference a particular commit of the codecs submodule, there is no difficulty in having the list of codec source files located in an HTSlib file that would need updating whenever the list of source files in the submodule changes (which would be infrequent anyway). In fact, probably the best way to do this would be to have a new HTSlib source file, e.g., _codecs.c_, that includes all the desired htscodecs source files:

```c
// codecs.c -- HTSlib source file that pulls in code from the htscodecs submodule

#include <config.h>  // i.e. *HTSlib's* config.h

// Any other needed setting up

#include "htscodecs/arith_dynamic.c"
#include "htscodecs/fqzcomp_qual.c"
#include "htscodecs/pack.c"
#include "htscodecs/rANS_static.c"
#include "htscodecs/rANS_static4x16pr.c"
#include "htscodecs/rle.c"
#include "htscodecs/tokenise_name3.c"
```

Then HTSlib's _Makefile_ would just list _codecs.o_ in `LIBHTS_OBJS` just like any other HTSlib source file — and would not itself need any lists of codecs source files etc. Or, indeed, any further complexity at all.

And that's it. No need for HTSlib's _INSTALL_ or _configure.ac_ to learn about automake or libtool or do anything at all. Not using htscodecs's makefile or build system, as we are literally using only the codec source code from the subproject. Not running or using the htscodecs test suite either, but the expectation would be that people doing development on the codecs themselves would be doing it from a separate checkout of htscodecs itself (rather than as a submodule) so would be building and testing using htscodecs's own build system and test suite — quite separately from HTSlib; and then when complete the new commits could be git fetched/pulled into HTSlib's referenced htscodecs commit.

Finally, **what about HTSlib release source tarballs**? There is no git checkout to use to supply files from a submodule, so _c-maint/mktarball_ would need to be extended to copy the htscodecs files (or just the _.c_/_.h_ files actually needed and the basic README, LICENSE etc) from the referenced repository and commit into a newly mkdired subdirectory within the tarball it's creating. Then the existing _codecs.c_ and its build rules would work exactly as they do in a git checkout.

There are a couple of minor organisational changes that would facilitate all this:

* I guess before this is merged to samtools/htslib, it would be desirable to move the codecs repository to the **samtools** organisation. IMHO `samtools/hts-codecs` would be a slight improvement on `htscodecs`, with the hyphen paralleling the one in `samtools/hts-specs`.

* It would be preferable for _mktarball_ to copy the needed source files into a top-level directory, say _htslib-1.x/codecs/*.[ch]_, so that they're not nested inside a tree of otherwise empty directories. Therefore ideally the codec source code would be at the top-level directory of the submodule, so that `#include "codecs/foo.c"` (for a submodule path of `codecs`) would be the right path for both a git checkout and a release tarball.

jmarshall:
The downside of this is that the codecs would be built in HTSlib by a different system from the one they are built by in the native htscodecs project; in particular, with different compile flags than as determined by the _htscodecs_ build system. OTOH they would be built with the same compile flags as the rest of HTSlib — so there are pros and cons here; it's a tradeoff. My suspicion is that the differences would be minimal anyway.

The big advantage is that htscodecs is free to use automake and libtool (boo, hiss :smile:) if it wants to, and this will have no effect on HTSlib. However perhaps it would be worth revisiting the htscodecs build system to focus on just test harnesses for the various codecs.

[Rob]
>  The main issue with this pull request is how we manage the new forced dependency on the htscodecs library. It at least needs additional configure options to allow linking against an external htscodecs, as distributors like Debian will want to include it that way.

[James]
> It also doesn't necessarily have to use the build system or be a library. It could be done within the htslib Makefile […] That won't be flavour of the month with Debian though.

Even though Debian have already jumped the gun and drafted a shared library package from James's draft repository, HTSlib and htscodecs have a choice about whether to facilitate that. It's really a question of managing expectations.

Debian perfectly reasonably likes to debundle separate libraries so that the subproject can be independently updated and the superproject gain from that (getting bug fixes etc).

However one point of view here is that htscodecs really isn't a separate library — it's just a collection of source files. There's no intrinsic reason why HTSlib or htscodecs need to facilitate using or building htscodecs as a library, when they claim that htscodecs is just a git-native way of sharing some code between HTSlib and io_lib — a more reliable form of klib's copying files around between projects.

The claim is that by only having HTSlib (and io_lib) with released versions, and each HTSlib release determining the htscodecs submodule commit that it's used with, maintenance and support are simplified. (E.g. the need to maintain yet another stable API is avoided entirely!) Behaviour from bug reports is easier to understand because there isn't a random combination of htslib version and codec version.

This point of view will be more palatable if htscodecs doesn't present itself as a library. I haven't really looked at how io_lib uses it, but if the htscodecs build system was changed to just build a bunch of _*.o_ files that can be linked with their test harnesses, then it would be clear that this code was not intended to be separated as a library. There may be temporary squealing :smile: but I suspect packagers would come around to just using HTSlib's / io_lib's native inclusion of these source files. Particularly as HTSlib/io_lib releases should be frequent enough to keep up with the (in the fullness of time) likely small infrequent updates to the codecs.

jkbonfield:
> Even though Debian have already jumped the gun and drafted a shared library package from James's draft repository, HTSlib and htscodecs have a choice about whether to facilitate that. It's really a question of managing expectations.

It's used by a released version of io_lib, so they haven't jumped the gun.

This problem has mainly come about due to our tardiness.  I wasn't willing to tie my own releases of io_lib (libstaden_read1 on Debian) to wait *years* for htslib PRs to get reviewed and with a lack of any feedback here I just went with something that worked for me.  I'd rather not rename it now either, but it does need moving to the samtools organisation (that's not something I have permission to do).  It was named after htslib, which doesn't have a hyphen.
 
> However one point of view here is that htscodecs really isn't a separate library — it's just a collection of source files. There's no intrinsic reason why HTSlib or htscodecs need to facilitate using or building htscodecs as a library, when they claim that htscodecs is just a git-native way of sharing some code between HTSlib and io_lib — a more reliable form of klib's copying files around between projects.

It can be thought of either way.  I'm open to all ways for htslib to use it.  My aim is simply to avoid duplication.  I've had enough of having to support two variants of the same code, and occasionally finding subtle bugs that we'd already fixed in the other copy.  One source location is absolutely the right solution.

As for the tarball, I think it should just be bundled in with it.  Definitely for samtools/bcftools, but maybe also for htslib.

daviesrob:
I'm just getting around to working out how to integrate this into HTSlib without the need for a recursive make (which I'm not in favour of due to the pain it has caused in SAMtools).  My current thinking is to have configure options that would allow it to be built using either bundled htscodecs code, or linked to an external library as Debian would like.  Some other projects work this way, for example mplayer can build its own copy of various dependencies or link to them externally.  It would be useful as long as it doesn't add too much complication if it means distributors don't start adding big patches so they can package htscodecs externally.

James should be able to transfer the htscodecs repository to the samtools organisation now.

jkbonfield:
My former copy of this is now https://github.com/jkbonfield/htslib/tree/htscodecs.jkb.  The new htscodecs here is @daviesrob's copy with a revised build system.

Comparing my copy of this PR and Rob's rebased/rejigged copy I have spotted some discrepancies.

1. Why does 9dfbf8d7 bring in a second copy of the `cram_content_type` enum?

2. Why has cram/rANS_byte.h reappeared in htslib.mk?  It wasn't there in my branch, but is there in this one, yet it's an old line according to git blame.  Is this a case of going back and rebasing everything again but failing to remove something during a conflict?  It shouldn't be there as the file is now htscodecs/htscodecs/rANS_byte.h.

3. What happened to https://github.com/jkbonfield/htslib/commit/0dfea109?  It worked around stack size issues on threaded Mac implementations, upping the storage for threads to match the primary thread stack size.  It was certainly necessary when I did it, but does it no longer need to be done (either due to MacOS changes or due to the code no longer needing it due to use of thread-local-storage instead?

daviesrob:
To answer the questions:

1. I don't think there should be a second copy of `cram_content_type`.  The one that was always there (and commented out) in `cram/cram_structs.h` isn't touched by the new set of commits.
2. `cram/rANS_byte.h` probably failed to be removed during a rebase.
3. It was replaced by 9ffa293 which should work on both Macs and non-Macs with undersized thread stack allocations.

jkbonfield:
Regarding 1 - I'm only reporting the differences between my branch and your branch.  If it's not in your branch it's because you either added it there, or I deleted it and you decided to not include that commit.  Hence my question.

Anyway, who is tidying up these rebase problems?  Me, or you as part of the ongoing code review?

jkbonfield:
Actually maybe I'm wrong about 1.  Git diff was rather hiding (by dint of no colouration and being harder to read) that the change is the difference between removing the extra copy of the struct and simply commenting out the struct.  I don't know how we managed to get this difference during rebasing, but it's an irrelevance.

So that really just leaves a git rm left.

daviesrob:
You removed it, but then it was already commented out.  I think the only reason it was left there in residual form was so anyone looking would know where it had gone.

If you remove 1 and 2 with individual commits, they can easily be squashed into place if we decide to later.

jkbonfield:
Ignoring 1 (it doesn't matter much either way; I simply initially failed to spot the "/*" in the github diff output).

2 has been removed once more.

3 as you correctly point ouy was subsequently avoided by a commit added elsewhere so became redundant.

I've also fixed the issue with discussed in the meeting, splitting `cram_block_method` to internal and external.

@jmarshall may wish to review the addition of HTS_COMPAT there and suggest a better name if you disagree with it.  The intention is clear - to head off future pain at the pass, before it arrives as it inevitably will!  I'm hoping we've headed it off early enough. :-)

daviesrob:
I think the symbols used internally in the `cram` directory should be renamed, rather than the ones that were exposed externally.  That way we would avoid any need for `HTS_COMPAT` - although we might want to deprecate the whole of `cram_block_method` as it's not entirely clear how it could be used outside of HTSlib itself.

jkbonfield:
I'm not sure I follow.  What renaming?  More than has been done in the last commit, or is this just cross-wires?

At present, I don't believe cram_block_method is used externally.  I can't see the type in any public prototypes or structures, making it a rather pointless orphan.  So I'd be in favour of completely removing it if you're willing to accept that.

In future though it will indeed be useful once we expose the CRAM API more widely.  (Long term project, to facilitate container and slice level manipulation / data gathering; eg io_lib's cram_size, cram_filter, etc tools.)

daviesrob:
`cram_block_method` appears to be useless, so the best solution might be to hide it in `htslib/cram.h`.  As discussed earlier, we would need to invoke a cake penalty if it turns out that someone was using it and raises an issue about it going away.

If going the `HTS_COMPAT` route, it should probably follow the format of [`HTS_VERSION` in `htslib/hts.h`](https://github.com/samtools/htslib/blob/7ca2b49f/htslib/hts.h#L435).

Also, you accidentally reverted the submodule by about four commits when you pushed 435caac.  To reduce the chance of this sort of accident, we might want to set `fetchRecurseSubmodules` in `.gitmodules` (see `git help gitmodules`).

jmarshall:
Nice, I'm quite liking this build approach. Inclined to agree with Rob on the `HTS_COMPAT` stuff but I haven't ruminated on it much as yet. Some initial unrelated comments:

* The `htscodecs_…_h` variables and `htscodecs/…/*.o …/*.pico` rules currently added to _Makefile_ would perhaps be better in _htscodecs_bundled.mk_ (their empty compatriots are already in _htscodecs_external.mk_).

* Could _htscodecs_bundled.mk_ be located in the htscodecs repository and included from there, rather than duplicating this information into HTSlib?

jkbonfield:
Fix the submodule commit hash.  Also revised the htslib/cram.h.

It no longer has an #else clause, as we can add that if and when we ever actually need it.  I'm now confident this enum has no *legitimate* use as it never occurs as a type anywhere else in the public API.  It's basically just an orphan, presumably left in by accident.

However for belt and braces it's ifdefed out with the option of reenabling it again, should someone be automatically parsing all structs/enums and doing "weird stuff" with them (eg pysam?  Maybe @jmarshall knows more what impact there would be to removing an enum on pysam), so we have a trivial way to ensure compatibility.

I updated the HTS_COMPAT to use the numbering system used in HTS_VERSION.  Thanks for that.  I'd completely forgotten about the existance of it.

jkbonfield:
Oops... of course I broke it.  Not having a good day ;).  Fix inc

jkbonfield:
There is one weakness with this new build system though - it doesn't run any of the htscodecs tests, unlike my original PR.

Given how quick they are to run, I think we should be building and running those tests.  You could argue maybe that they've already been tested in htscodecs, but I'd feel happier for it to test here too given you routinely do all sorts of whacky system tests on htslib (eg SunStudio compiler, mips, sparc, 32-bit builds, etc).  We don't have any of these things available in the CI system.

daviesrob:
I don't think it should be too hard to get the new build system to run the htscodecs tests.

I put the variables and rules in the Makefile mainly because that's where everything else is, so the thought was that they would be easier to find.  I think they could be moved into `htscodecs_bundled.mk`, and the tests would probably need to be there.  `htscodecs_bundled.mk` seemed to be a bit HTSlib-specific which is why I put it there, but there's no reason why it (or parts of it) couldn't be put into htscodecs itself it we decide they're more generally useful.

jmarshall:
For **pysam** there would be no impact from removing this enum.

Pysam contains cythonic declarations of HTSlib facilities in _.pxd_ files. These are hand-written — there's no magical _.h_ parsing or anything going on — and are somewhere between exhaustive equivalents and just declaring the things that pysam actually uses. I believe presence of a declaration in a _.pxd_ file just means it's usable from the pysam-internal _.pyx_ source files, it doesn't mean it's exposed to pysam users.

For `cram_block_method` in particular: _pysam/libchtslib.pxd_ declares `cram_block_method` and its enumerators (without the `s/ERROR/BM_ERROR/` fix) but they have never been used in pysam. If they were used in some _.pyx_ source file, then removing the enum from HTSlib would lead to ordinary C compilation errors when the C code generated from that _.pyx_ was compiled. However as they are unused pysam would just continue to compile just fine if the enum were removed from HTSlib. The unused _pysam/libchtslib.pxd_ declaration is decoupled from the _htslib/cram.h_ one and could be deleted from pysam either before or after the equivalent was deleted from HTSlib — it doesn't matter either way.

I might just go ahead and proactively delete it from pysam's _.pxd_ file anyway :smile:

jmarshall:
Having reread the latest edition of the branch, FWIW :+1: to just nuking `cram_block_method` from _htslib/cram.h_ entirely and not needing to invent `HTS_COMPAT` just yet. There is precedent for this, e.g., 2fb3b7e178304eff76381c630e54e6408a94f834. (I would chip in to the cake fund in the unlikely event that proved necessary :smile:)

jmarshall:
`$(HTSDIR)/cram/rANS_byte.h` has reappeared in _htslib.mk_'s `$(HTSLIB_ALL)` again…

daviesrob:
Yes, also the htscodecs submodule got reverted again.  But don't worry, I have a few updates to fix that which I'll add soon.

jkbonfield:
I don't understand where that's coming from.  I did a git pull in the submodule to ensure it was up to date so I've no idea how it can magicall revert when making an unrelated change.

daviesrob:
The submodule should now be back in the correct position, and I've put in a `.gitmodules` option that should keep it up to date through pulls.

@jkbonfield I think your problem might be that your local checkout has your own fork of htscodecs in it, and not the samtools/htscodecs one.  I had to edit a file to get my copy to use the correct upstream.

I've also got it to run the htscodecs tests.  Some shenanigans was necessary to get them all to run (it turns out `getopt()` needs `__USE_POSIX2` in strict-c99 mode).  It might be possible to simplify the integration with a few small htscodecs changes.

jkbonfield:
> @jkbonfield I think your problem might be that your local checkout has your own fork of htscodecs in it, and not the samtools/htscodecs one. I had to edit a file to get my copy to use the correct upstream.

Ah yes.  It was surprisingly hard to change that!  I assumed git pull would also be pulling over the submodule meta-data, but apparently not.  In the end I gave up and blew away the entire .git/modules directory and started again. :-)   Moving module location is probably something that's pretty abnormal to be fair.

> I've also got it to run the htscodecs tests. Some shenanigans was necessary to get them all to run (it turns out `getopt()` needs `__USE_POSIX2` in strict-c99 mode). It might be possible to simplify the integration with a few small htscodecs changes.

That's fine.  I expected we'd uncover various corner cases as it got further integrated.

jmarshall:
> (it turns out getopt() needs __USE_POSIX2 in strict-c99 mode).

The error message is quite amusing (`implicit declaration of function ‘getopt’; did you mean ‘getgid’?` — what? I included the POSIX header `<unistd.h>` so why are you giving me one POSIX function but not another?!), but glibc's _include/features.h_ has the answers: the difference is between POSIX 1003.1 and 1003.2, and it needs `_POSIX_C_SOURCE >= 2` and/or `_XOPEN_SOURCE` defined.

And in my configured builds, I mostly get `_XOPEN_SOURCE=600` from _config.h_. This suggests two things:

1. HTSlib is using a bunch of POSIX functions, so it probably does need to define `_XOPEN_SOURCE` in general.

   _configure.ac_ does so as a side-effect of something about `PTHREAD_MUTEX_RECURSIVE` on “some broken glibcs” but “`PTHREAD_MUTEX_RECURSIVE`” and “some broken” are probably massive oversummaries.

2. The source files in _htscodecs/tests/*.c_ need to have `#include <config.h>` added.

daviesrob:
@jmarshall Correct - and I had to set `_POSIX_C_SOURCE` because the Appveyor build adds `-D_XOPEN_SOURCE=600` to the compile command line, so trying to set the latter resulted in a redefinition error.

While adding the tests, I also got reminded of another reason why the htscodecs rules need to be in the Makefile.  `htscodecs_bundled.mk` gets injected into the samtools and bcftools makefiles so that they can properly track htscodecs-inside-htslib dependencies.  As the build rules themselves aren't appropriate for samtools or bcftools, they have to live elsewhere.

daviesrob:
I've added a couple of updates so running configure or make will ask you to update the submodule if it's required but not present.  It's possible that the Makefile could update the submodule itself if it detects you have a git checkout, but I decided not to attempt that in case it ran into problems with concurrent builds.

daviesrob:
I've squashed in a couple of the bugfix commits, and rebased so this will merge cleanly again.

daviesrob:
The samtools and bcftools tests should now work with the submodule, so I'll merge this tomorrow morning unless there are any last-minute hitches.

PR #931, opened 2019-09-20
Add support for SQ-AN alternative sequence names

jmarshall:
Support alternative sequence names as per samtools/hts-specs#100.

This is a work-in-progress for testing purposes, as `sam_hdr_update_line()` does not yet handle AN altnames. Everything else should be ready to go though.

(View the `sam_hrecs_remove_hash_entry()` diff with -w as the bulk of the code is unchanged other than its indentation level.)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -93,6 +93,10 @@ Noteworthy changes in release a.b
   operations.  Trying to do this (for example with an iterator range request)
   will result in the SAM readers dropping back to single-threaded mode.
 
+* HTSlib now supports the @SQ-AN header field, which lists alternative names
+  for reference sequences.  This means given "@SQ SN:1 AN:chr1", tools like
+  samtools can accept requests for "1" or "chr1" equivalently.  (PR #931)
+
 * Saving a file with a '.sam.gz' suffix will write BGZF-compressed SAM format.
 
 * hts_detect_format() no longer identifies arbitrary text files as SAM.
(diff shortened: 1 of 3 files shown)


daviesrob:
I've added a commit to adjust how duplicates between `SN:` and `AN:` tags are handled, in case someone makes a header like this:
```
@SQ	SN:CHROMOSOME_I	LN:1009800	AN:1,foo,CHROMOSOME_II
@SQ	SN:CHROMOSOME_II	LN:5000
@SQ	SN:CHROMOSOME_III	LN:5000	AN:CHROMOSOME_I
@SQ	SN:CHROMOSOME_IV	LN:5000	AN:CHROMOSOME_IV
@SQ	SN:CHROMOSOME_V	LN:5000	AN:CHROMOSOME_V
```

This didn't cause a problem for SAM or BAM, but with CRAM the CHROMOSOME_II altname on the first entry masked the primary name on the second, effectively making it disappear.

It also fixes a few other edge cases, like what happens when you delete an `AN:` tag with the same name as `SN:` in it.

jmarshall:
Thanks for those updates, Rob. Spec says “no duplicates across all SN and AN entries” which is a good formal rule but agree samtools style is to be a bit more lenient.

Though it does open it up to problems like the one described in your last paragraph but glad you’ve caught that one!

daviesrob:
Yes, while the specification disallows it, I think it's good to have a way of handling the problem when someone inevitably ignores the rule.

PR #934, opened 2019-09-24
Add test_view -v option to increase verbosity

daviesrob:


daviesrob:
Rebased with added help message.

PR #939, opened 2019-09-27
Change upload part size.

### LINKED ISSUE #938: large BAM to CRAM conversion hits S3 chunk limit
(How) can I increase chunk size when converting a 200Gb BAM to CRAM on S3 so the chunk number stays under 10K?

`samtools view -C s3://bucket/bam/x.bam -o s3://bucket/x.cram -T x.fa`

\<Error>\<Code>InvalidArgument\</Code>\<Message>Part number must be an integer between 1 and 10000, inclusive\</Message>\<ArgumentName>partNumber\</ArgumentName>\<ArgumentValue>10001\</ArgumentValue>\<RequestId>

whitwham:
Added a way of setting the upload part size.

Use the HTS_S3_PART_SIZE environment variable to set the upload size.  Minimum part size is still 5Mb.

Fixes #938.

### DIFF
--- a/hfile_s3_write.c
+++ b/hfile_s3_write.c
@@ -88,6 +88,10 @@ Andrew Whitwham, January 2019
 #define S3_MOVED_PERMANENTLY 301
 #define S3_BAD_REQUEST 400
 
+// Lets the part memory size grow to about 1Gb giving a 2.5Tb max file size.
+// Max. parts allowed by AWS is 10000, so use ceil(10000.0/9.0)
+#define EXPAND_ON 1112
+
 static struct {
     kstring_t useragent;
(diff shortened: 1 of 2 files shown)


whitwham:
Part sizes will now grow over time. Every 1200 parts the part size will double.  This allows files of up to 2Tb to be uploaded at a maximum cost of 1Gb internal memory.  The setting of HTS_S3_PART_SIZE will disable this ability.

daviesrob:
Rebased and squashed into two commits.  The growth rate has been changed slightly to double every 1112  (10000/9 rounded up) parts which allows it to store files of up to 2.5 Tbytes for the same maximum part size.

PR #942, opened 2019-10-02
Added bam_set_qname() function

jenniferliddle:
I needed this function for bambi, so it seemed sensible to add it to htslib

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1,6 +1,6 @@
 /*  sam.c -- SAM and BAM file I/O and manipulation.
 
-    Copyright (C) 2008-2010, 2012-2018 Genome Research Ltd.
+    Copyright (C) 2008-2010, 2012-2019 Genome Research Ltd.
     Copyright (C) 2010, 2012, 2013 Broad Institute.
 
     Author: Heng Li <lh3@sanger.ac.uk>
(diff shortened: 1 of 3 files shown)


sam.c
```diff
@@ -685,6 +685,38 @@ static int bam_write_idx1(htsFile *fp, const bam1_t *b) {
...
+
+    uint8_t *new_data = calloc(1, rec->m_data);
+    if (!new_data) return -1;
+    memcpy(new_data, qname, new_len);
+    uint8_t *p = new_data + new_len;
+    for (int n=0; n<extranul; n++) *p++=0;
+    memcpy(p, rec->data + rec->core.l_qname, rec->l_data - rec->core.l_qname);
+
+    free(rec->data); rec->data=new_data;
+    rec->l_data = new_data_len;
```
daviesrob: You shouldn't be doing this `calloc()` as you've already resized the bam data.  Instead, just shift everything after the name with `memmove()`:
```c
// Make room
if (new_len + extranul != rec->core.l_qname)
    memmove(rec->data + new_len + extranul, rec->data + rec->core.l_qname, rec->l_data - rec->core.l_qname);
// Copy in new name and pad if needed
memcpy(rec->data, qname, new_len);
for (int n = 0; n < extranul; n++) rec->data[new_len + n] = '\0';
```

Note that this does introduce an extra bit of memory shuffling in the case where `realloc_bam_data()` had to copy everything to a new location.  That could be avoided, but it gets a bit more complicated.  You'd have to:

- Check if the data fits in the existing area.
   - If yes, do the `memmove()` above.
   - If no, you have to `malloc()` a new buffer, copy in the new name, do any NUL padding and copy the rest of the data.

If you allocate a new data buffer, you should only free the old one if `(bam_get_mempolicy(b) & BAM_USER_OWNS_DATA) == 0`.  You also need to change the ownership of the data using:
```c
bam_set_mempolicy(b, bam_get_mempolicy(b) & (~BAM_USER_OWNS_DATA));
```
(See [`sam_realloc_bam_data()`](https://github.com/samtools/htslib/blob/f8512fb/sam.c#L375).

Issue #695, opened 2018-04-27
version 1.8 drops symbol cram_nop_decode_reset without bumping soversion


tillea:
Hi,
I'm about to build Debian packages of htslib 1.8.  I realised that in file cram/cram_codecs.c function

    void cram_nop_decode_reset(cram_codec *c) {}

was droped.  Well, droping an empty function looks harmless but you never know what programmers out there did with this function.  So its de facto an ABI change which should be reflected in bumping soversion.
Kind regards, Andreas.

jkbonfield:
While technically available in the symbol table, this is an internal function with no public prototype.  Yes you could go grubbing around the library to find symbols you can utilise, but it's not something supported and you'd be on your own if you did so.

Hence we don't regard this as an ABI change as it isn't part of our public ABI.

mapreri:
Yes, GCC and clang follow the same specs (after all, clang aims at being pretty mucha a drop in replacement in most cases...), ISTR msvc has a slightly different syntax, but supporting both is not particularly difficult.

Could you please consider reopening and retitling this bug report?

Issue #311, opened 2015-12-24
Missing symbols in htslib 1.3


satta:
Hi,

I'm trying to update Debian's htslib package to the recently released version 1.3. This worked nicely with the exception of a few exported symbols having disappeared from the shared library, most of which were either made static, replaced by a macro or renamed. This would cause the ABI version to be bumped in Debian, which is probably not intended for a minor version update like the new 1.3. I was able to add some  wrappers for most of them (see attached [fixup_symbols.patch.txt](https://github.com/samtools/htslib/files/71600/fixup_symbols.patch.txt)) to ensure they are still in the library for backwards compatibility.

Three more puzzle me a bit:
```
libhts.so.1 libhts1 #MINVER#
- MD5_Final@Base 1.0
- MD5_Init@Base 1.0
- MD5_Update@Base 1.0
+#MISSING: 1.3-1# MD5_Final@Base 1.0
+#MISSING: 1.3-1# MD5_Init@Base 1.0
+#MISSING: 1.3-1# MD5_Update@Base 1.0
```

It appears like at some earlier point in time libhts exported these OpenSSL functions in its own symbol table, and now that the implementation changed in 1.3 they are gone. I have simply wrapped these as well (though I doubt anyone linking to the shared htslib library uses them) -- is the new internal implementation really binary compatible to the OpenSSL one? Can I do this?

I would be glad to get some advice whether my patches make sense... thanks in advance!

Happy holidays,
Sascha

jmarshall:
`bcf_empty1()` was previously not declared or even mentioned in a header file, so was not useable in previous htslib releases.  Effectively `bcf_empty()` and `bcf_empty1()` have been **freshly added** to the public htslib API in 1.3.  See d4482eb578aada0080eb30239ff57e9f6a6de93b for details.

`int32_get()` and `int32_put()` were internal functions private to the library.  They were not declared in any header file and were not part of the API.  Similarly `hts_idx_load_local()` is not declared in any header and is not part of the API, so making it `static` was intended and does not affect either public API or ABI.

`MD5_Update()`/`MD5_Final()`/`MD5_Init()` were badly-named internal functions private to the library.  Defining these functions with their OpenSSL names (and via implementations unlikely to be using the context data compatibly) was a bug, and removing them fixed the bug.  See 3accd1934472e2bdd501404f4c1a2e570f868cd1 and PR #180 for details.

[...]

PR #946, opened 2019-10-07
Ensure only symbols we want are exported by the shared library

### LINKED ISSUE #616: increment SOVERSION please
Hello,

I was working on fixing build issues with htslib 1.5 and python-pysam 0.12 on Debian and discovered that symbols were removed between htslib 1.4.1 and 1.5. Which should've triggered a SOVERSION increase.

Looks like there's other symbols that were removed in 1.6. I found debug_vbuf, hopen_json_redirect which were public and implemented in 1.5.

Would it be reasonable to update the SOVERSION number to 4 for 1.6? and I can set it to 3 for Debian's 1.5?

daviesrob:
Restricts the symbols exported by `libhts.so` to just the ones in the ABI.  There are a few reasons for doing this:

1. It defines exactly what is in the ABI.  Packagers inspecting the library for changes will no longer be mislead by changes to non-ABI functions (See #311, #616 and #695) .  Library users will be less tempted to use non-ABI functions.
2. It reduces the risk of non-ABI functions being interposed by other libraries.
3. It makes the dynamic symbol table smaller, so it takes up less memory and links faster.
4. It eliminates an indirect jump when using functions that are not exported.

(See [Ulrich Drepper's "How to Write Shared Libraries"](https://software.intel.com/sites/default/files/m/a/1/e/dsohowto.pdf) for more information).

Symbols are changed to be hidden by default (using `-fvisibility=hidden`) and attributes are used to mark the ones that should be exported.  The attributes are mostly in the `htslib/*.h` header files.  A few exceptions are in `.c` files - these are  functions that are used externally (mostly by plug-ins) but not currently in the public headers; and data exports declared `extern` in the headers.

This has been tested against shared library builds of `samtools` and `bcftools` to ensure all the symbols they need are available, along with the ones required for the plug-in shared libraries.

**After merging this, maintainers would need to check that the attributes are added for any functions added to the ABI.  Symbols in the static library do not get hidden, so there is a risk that static builds work while dynamic ones don't.**  It would be nice to have an automated test for this, but I haven't (yet) thought of an easy way to do it.

Currently builds without `configure` will still export everything.  This could be changed by adding `-fvisibility=hidden` to the defaults in the `Makefile`.  It would break `gcc` before version 4.0.0, but I'm not sure if anything that old can build HTSlib now anyway.

Fixes #616


hfile.c
```diff
@@ -212,6 +214,7 @@ int hfile_set_blksize(hFILE *fp, size_t bufsiz) {
 }
 
 /* Called only from hgetc(), when our buffer is empty.  */
+HTSLIB_EXPORT
```
valeriuo: Why does this method need to be exported, since it is only used internally by `hread`, inside which it is already declared `extern`?
Same for `hread2`, `hputc2`, `hwrite2`, `hputs2`.

daviesrob: It's declared in this way because it needs to be called from external code but only via `hread` and not directly.  Declaring it `extern` isn't enough to have it exported - it needs the annotation, which has to be in the C file as the only declaration in the header file is hidden in the `hread` function definition.

jkbonfield: What's missing from the above explanation is that it's called from a static inline, and thus external code is essentially calling these internal functions (hence the HTSLIB_EXPORT).

However that means hread2 etc are also part of the external API and external ABI and should be documented as such.  The comment "Called only from hgetc()" is misleading - it's only called by our code there, but people are free to call it anywhere as it is a public function.

jmarshall: `hgetc2`/`hread2`/etc are part of the **ABI** (as you say, as they are invoked by an **inline** API function so calls to these functions are directly burnt into client executables), but they are not part of the **API** (as there is no file-scope declaration of them in a public header). The comment is accurate as the only declaration of `hgetc2()` has a scope limited to the body of the `hgetc()` function — so it truly can only be called from that function.

People could call it from anywhere by writing their own declaration for it, but that would be cheating, could be done for any (visible) function, and is not an activity that HTSlib (or any sensible library) supports.

jkbonfield: Fair enough.   Makes sense now I look again.

Thanks.

jkbonfield:
As expected it does what we intended, tested with

```
#include <stdio.h>

extern char *cram_block_method2str(int _);

int main(void) {
  printf("%s\n", cram_block_method2str(1));
  return 0;
}
```

However with straight "make" this program builds and runs OK.  With "./configure; make" it doesn't.  We shouldn't be exasperating the dual build system even more given the number of https protocol errors we get demonstrating some people use the former rather than the latter.

Personally I'd just ditch the non-configure route entirely, but if we must keep it please add `-fvisibility=hidden` to the default CFLAGS in Makefile so we have the same behaviour both ways.

htslib/vcf.h
```diff
@@ -427,18 +449,31 @@ set to one of BCF_ERR* codes and must be checked before calling bcf_write().
...
+    /** @param line   Line to write
+        @param fp     File to write it to
+        @return 0 on success; -1 on failure
 
+        @note No checks are done on the line being added, apart from
+              ensuring that it ends with a newline.  This function
+              should therefore be used with care.
+    */
+    HTSLIB_EXPORT
+    int vcf_write_line(htsFile *fp, kstring_t *line);
```
jkbonfield: This function didn't previously have a public API.

Hence this isn't just tying down the API but also expanded it.  I don't mind, but please mention this in the commit message along with a reason (eg who uses it).

It's not used by Samtools or Bcftools.

m4/hts_hide_dynamic_syms.m4
```diff
@@ -0,0 +1,65 @@
+dnl @synopsis HTS_HIDE_DYNAMIC_SYMBOLS
+dnl
+dnl Turn on compiler options that prevent unwanted symbols from being exported
+dnl by shared libraries.
+dnl
+dnl @author Rob Davies <rmd@sanger.ac.uk>
+dnl @license MIT/Expat
+dnl
+dnl Copyright (C) 2018 Genome Research Ltd.
```
jkbonfield: Wrong year. :-)

(Or have we really had a branch for a year awaiting a PR to be made from it?)

jkbonfield:
Barring trivial documentation things and a default in Makefile, it seems to work and I like the idea. So +1

htslib/vcf.h
jmarshall: I see _bcftools/vcfconcat.c_ has declared it itself since 2014. Aaargh!

jmarshall:
Re https://github.com/samtools/htslib/pull/946#issuecomment-541789723

> However with straight "make" this program builds and runs OK. With "./configure; make" it doesn't. We shouldn't be exasperating the dual build system even more given the number of https protocol errors we get demonstrating some people use the former rather than the latter.

Enabling HTTPS even in plain-vanilla non-configure builds was a good idea because it enables significant user-visible behaviour — i.e. HTTPS works.

This doesn't really produce a difference visible to most end-users, and some platforms won't understand `-fvisibility=hidden` so adding it to the default CFLAGS could cause new bug reports. It's really only people distributing built libhts.so's to others who care about this — i.e. distribution packagers — and they can be expected to run configure.

(Personally, I would have expected them to implement this themselves and carry it in their local patches for a while to road test it, but Rob is more generous than me :smile:)

jkbonfield:
The way I see it there are two schools of build configuration.  There's ./configure with a bunch of options for things that are non-standard or optional, and Makefile with manual editing for things that are non-standard / optional.  Basically the same thing via two different routes. 

All I'm proposing is that the default action on both matches up.  We saw how many problems we generated by having Makefile not use curl and the configure route to use it.  It took *years* of badgering to get that changed, and not just as we're about to harmonise configure and non-configure routes we're proposing once again making them differ (albeit less severely).

Gcc, clang and icc all understand `-fvisibility=hidden` so the platforms out there that don't are going to be esoteric.   If people want a build that copes with their unusual setups, they can use the configure script  (it's what it's for!), edit the Makefile, or just override CFLAGS on the make command line.

htslib/hts_os.h
```diff
@@ -26,14 +26,23 @@ DEALINGS IN THE SOFTWARE.  */
...
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-extern void hts_srand48(long seed);
-extern double hts_erand48(unsigned short xseed[3]);
-extern double hts_drand48(void);
-extern long hts_lrand48(void);
+HTSLIB_EXPORT
```
valeriuo: Do these methods have to be exported from both the `*.h` and the `*.c` files (`os/rand.c` and `hts_os.c`)?

daviesrob: I think it's because `hts_os.c` doesn't include `htslib/hts_os.h` (and possibly shouldn't due to `#define`s in the latter file.

jmarshall:
I think it would be good to think further about how to test this. At present, bgzip etc, samtools, and bcftools are all built statically so adding a new public function but forgetting to annotate it for export would not be detected until the distro packagers made their releases, long after an HTSlib release.

Samtools and bcftools travis could do a build in which they build specifically against shared libhts.so with hiddenness, so breakage would be detected for functions that {sam,bcf}tools use.

We could add a _htslib/test/test_api.c_ that was specifically linked against _libhts.so_ only, that called every function in the API. But that would basically be a list of API functions, and if HTSlib wanted to contain a list of API functions then that would be a way to do `HTS_EXPORT` that was less invasive to the public headers…

jkbonfield:
That's a good point, but by luck we already do this in samtools (although not bcftools).  I added an explicit test to ensure that samtools can build against an installed htslib to act as a test bed for the htslib install procedure and to spot forgetting to include header files.  (It did find a bug too.)

Samtools in that scenario will link against the dynamic library I think.  We should do the same for bcftools.

However it's best to be able to do this test internal to htslib instead of relying on another package.  The problem is how to derive such a test that isn't a tautology.  Eg grepping all the HTS_EXPORT and calling them wouldn't test anything at all, whereas "do samtools and bcftools build" is a real world test.

For what it's worth I tried pysam and STAR using htslib develop and they work, although it's unclear in pysam's case how much of that is due to the bundled htslib subdir (which cannot be trivially removed).

daviesrob:
I think maintainer vigilance may have to be the first line of defence, but it might be possible to write something to scan the header files for function prototypes and `extern` variables and then check that they are in the shared library.  It might not catch everything, but it should make a big dent in the problem.

For example, `ctags` can extract a list of function prototypes:
```
$ ctags -f - -n --language-force=c --c-kinds=p htslib/*.h
bam_aux2A	htslib/sam.h	1285;"	p
bam_aux2Z	htslib/sam.h	1292;"	p
bam_aux2f	htslib/sam.h	1278;"	p
        ... 410 lines skipped ...
tbx_seqnames	htslib/tbx.h	111;"	p
vcf_format	htslib/vcf.h	341;"	p
vcf_parse	htslib/vcf.h	338;"	p
```

This could be compared with the contents of the shared library, e.g.:

```
$ comm -23 <(ctags -f - -n -I HTS_RESULT_USED -I HTS_DEPRECATED+ -I HTS_FORMAT+ --language-force=c --c-kinds=p htslib/*.h | awk '{ print $1 }' | sort -u) <(nm -D -g libhts.so | awk '$2 == "T" { print $3 }' | sort -u)
knet_win32_destroy
knet_win32_init
```

Unfortunately `ctags` isn't quite as good for `extern` variables as for some reason it reports `struct` forward declarations in the same category.  It's possible other C file scanners may do better.

kevinjacobs-progenity:
> For what it's worth I tried pysam and STAR using htslib develop and they work, although it's unclear in pysam's case how much of that is due to the bundled htslib subdir (which cannot be trivially removed).

I've been waiting to adapt pysam to use htslib develop.  Are the API changes mostly done?

jkbonfield:
Yes.  There's some region parsing code (regidx https://github.com/samtools/htslib/pull/761) that needs to go in, and we may have another minor 64-bit thing to consider, but basically it's pretty much there already so you can definitely get a head-start by testing it out now.

daviesrob:
Rebased with some Windows fixes and a Makefile target that checks (partially) that the shared library is exporting the symbols that we want it to.  The target has been added to all the Travis and Appveyor tests.

I switched to a perl script to scan the public headers for exported functions, as it wasn't possible to get a suitable version of `ctags` on all the platforms we want to check.  It finds everything `ctags` did at the moment.  We can replace it with something better in the future if necessary.

Makefile
```diff
@@ -273,6 +276,13 @@ print-config:
 	@echo LIBS = $(LIBS)
 	@echo PLATFORM = $(PLATFORM)
 
+# These definitions are needed to get the right __declspec() annotation on
+# Windows.  See htslib/hts_defs.h
+$(LIBHTS_OBJS) : BUILDING_LIBRARY = -DHTS_BUILDING_LIBRARY
+$(LIBHTS_OBJS:.o=.pico) : BUILDING_LIBRARY = -DHTS_BUILDING_LIBRARY
+$(PLUGIN_OBJS) : BUILDING_LIBRARY = -DHTS_BUILDING_LIBRARY
+$(PLUGIN_OBJS:.o=.pico) : BUILDING_LIBRARY = -DHTS_BUILDING_LIBRARY
```
jmarshall: iMHO the low-tech approach of putting `#define HTS_BUILDING_LIBRARY` at the top of all the relevant _*.c_ files would be better than cluttering the Makefile and build logs by having this explicit on the command line.

daviesrob:
Rebased for latest HTSlib API additions, and made to use the low-tech `HTS_BUILDING_LIBRARY` approach.

While testing on Windows, I discovered issue #964 where registering an external plug-in on a statically linked program caused a segmentation fault.  This seems to be a problem with how dll linkage works on Windows, though, and not caused by any changes in this pull request.

PR #948, opened 2019-10-14
Ensure qname is properly NUL terminated when reading BAM

daviesrob:
Prevents possible invalid read when printing names.  This fixes up the bad records, which makes the patch a bit bigger than one that just rejected them would be.  If extranuls are available it just takes one, otherwise it has to expand the record a bit to add a NUL on the end of the name (and extranul becomes 3 as this will only happen if l_qname was originally a multiple of 4).

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -537,6 +537,24 @@ static void swap_data(const bam1_core_t *c, int l_data, uint8_t *data, int is_ho
     for (i = 0; i < c->n_cigar; ++i) ed_swap_4p(&cigar[i]);
 }
 
+// Fix bad records where qname is not terminated correctly.
+static int fixup_missing_qname_nul(bam1_t *b) {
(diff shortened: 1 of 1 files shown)


jkbonfield:
We need to find out what the recommended way is of referring to OSS fuzzing bugs.  I've been quoting the oss-fuz bug ID, but I don't know if there is meant to be some formal language that it can parse.  Any guidance on the blurb they gave you?

daviesrob:
OSS fuzz [say](https://google.github.io/oss-fuzz/architecture/) to put "Credit to OSS-Fuzz" in the commit message.  Adding the id for our information is probably useful.

valeriuo:
Merged into develop as d59c3645f798169f7fde53dbf5c6d3e095465e8c

PR #949, opened 2019-10-14
2 vcf fixes (security issue + an abort)

jkbonfield:
As found by OSS fuzzer.  See commits for details.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -2285,7 +2285,7 @@ static int vcf_parse_format(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v, char *p
         } else
         {
             hts_log_error("The format type %d is currently not supported", f->y>>4&0xf);
-            abort(); // I do not know how to do with Flag in the genotype fields
+            return -1;
         }
         align_mem(mem);
         f->offset = mem->l;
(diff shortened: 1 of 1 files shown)


vcf.c
```diff
@@ -2668,18 +2668,22 @@ int vcf_parse(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v)
...
     ret = 0;
 
+ end_with_ret:
  err:
     free(flt_a);
     free(val_a);
+
+    // A catch-all for any broken data we failed to notice in parsing.
+    if (ret == 0)
+        ret = bcf_record_check(h, v, 1);
```
daviesrob: I don't think we should be calling this:

1. As you've noticed, it's not a quick function to run
2. Records made by `vcf_parse` should always pass this function.  If not, `vcf_parse` has a bug which should be fixed, leadint to...
3. It's hiding the real problem in `vcf_parse`

In the case detected, the real problem is too make ALT entries, leading to a wrap-around of `v->n_allele` (which is only 16 bits).  I have a branch which catches this case (and also wrap-around of `v->n_info` which could happen for a similar reason).

jkbonfield: I confess it's partly because I got lost in a maze of code that didn't appear to ever set key anywhere (it clearly must do, but it eludes me), so the catch all was the quick fix to get something done before the release.

If you have a branch already, please turn it into a PR.

daviesrob:
Merged (with modifications) as 92ea05a, 12eb1e8 and 3d59a6a

Issue #794, opened 2018-11-08
HTS_VERSION macro


jkbonfield:
Htslib exports an `hts_version` function, those code is:

```
const char *hts_version()
{
    return HTS_VERSION;
}
```

where HTS_VERSION is in version.h, eg currently mine is `#define HTS_VERSION "1.9-63-g5d8bf3f"`.

We need to ability to be able to detect this at configure time, preferably without needing to compile and execute a test program so we permit cross-compiling.  As this isn't even a static variable, it's hard to glean this from symbol table grepping.

Cleaner would simply be moving HTS_VERSION to a public header file instead.  Perhaps with HTS_VERSION_MAJOR and HTS_VERSION_MINOR numerical equivalents too.  See ZLIB_VER macros in zlib.h for an example.

Even the string form alone is parseable from bourne shell.  Eg for a.b.c style versions

```
      v1=`expr "$v" : '\([[0-9]]*\)'`
      v2=`expr "$v" : '[[0-9]]*\.\([[0-9]]*\)'`
      v3=`expr "$v" : '[[0-9]]*\.[[0-9]]*\.\([[0-9]]*\)'`
```

jmarshall:
The common need is for HTSlib-using programs to be able to report the version of HTSlib that they are _running against_, as it's this that determines the behaviour observed and in particular what library bugs may have been fixed. Hence an `hts_version()` function whose result is determined by code in the shared library being _run against_ — which might be a later version than the one the program was originally _compiled against_. See c2e93911859de493a055ce21d37be8239b971d34, which added the function and removed an embryonic `#define`.

The slightly-less common need is for programs to be able to enquire about HTSlib's capabilities at compile time, and emit an understandable `#error` or (if they are overachievers) adjust their code accordingly. Projects that use configure can write a configure test to do a trial compilation of the capability they're interested in (even when cross-compiling), but not everybody is using configure.

When I added `hts_version()` back in 2013, I anticipated that interesting new HTSlib capabilities that people would want to enquire about would often be accompanied by `#defines` that they could coincidentally check for (e.g., fancy array resizing is advertised by `#ifdef HTS_RESIZE_CLEAR`), but this has mostly not been the case — most interesting stuff is in new API functions and invisible to the preprocessor.

[...]

Issue #559, opened 2017-07-04
Is hts_os.h part of the public API?


jmarshall:
Is it really intended that the recently added _htslib/hts_os.h_ is part of HTSlib's public API?

If it is an HTSlib-internal thing, would it not be better as _os_internal.h_ or so, in the top-level directory and in particular outside the public header directory.

jkbonfield:
Given that samtools depends on it, it has to be public.  The intention is to provide missing OS functionality so that htslib can rationalise code that expects a POSIX interface.  It's not ideal, but the alternative is to duplicate such code in both packages, which also isn't ideal.

jmarshall:
Good luck with that. It needs more `#ifdef` trickery to avoid screwing up platforms that do actually get this right.

jkbonfield:
Care to comment what then?  It would have been good to know about this in the month the PR was waiting to be merged too!

jmarshall:
I did a quick search for “don't declare standard library functions yourself, use the system headers” but didn't immediately find a good reference in e.g. the C FAQ.  In any case, (1) C99 §7.1.3/2; (2) this is a long-standing good practice that I know you're more than familiar with; (3) you just had a refresher of this in _cram/os.h_.

So for example those declarations of `[sdl]rand48()` ought to be omitted on platforms where `<stdlib.h>` has already provided them, possibly doing so as macros or with interesting annotations — in which case _hts_os.h_'s versions would conflict with the system ones.  This could be via configure but that's extremely fragile for a public header, or hopefully something like `#ifdef _POSIX_VERSION` or similar would do the trick.

Care to name the relevant samtools PR?

Presumably you have colleagues who are still in a position to review PRs as part of their paid jobs :smile:

jmarshall:
This replacing system functions malarky comes down to the same considerations that previously resulted in HTSlib introducing its own `hts_md5_init()` et al MD5-equivalent functions, rather than providing `MD5_Init()` et al with their usual names and attempting the gymnastics required to avoid colliding with system-provided functions (here “merely” declarations, mostly) with the same names.

jkbonfield:
Agreed the extern definitions of drand et al ought to be ifdef guarded and checked via autoconf for consistency and the fact they're not is simply jjust me forgetting as the others have various ifdefs around.  However they are the POSIX prototypes so haven't yet caused issues on any POSIX compliant systems.

[...]

PR #951, opened 2019-10-15
Add HTS_VERSION macro for compile-time HTSlib version checks

jmarshall:
Here is a proposal for adding a public `HTS_VERSION` macro, for use by third-party code wishing to probe for the existence of newly-added HTSlib capabilities (without going to the effort of writing configure tests to check for them!), as wished for in #794.

HTSlib already has a `hts_version()` function that reports the version of HTSlib that a program is _running against_, which is the most useful thing for third-party programs to report as it reflects the HTSlib behaviour observed and in particular what library bugs may have been fixed.

It would additionally be useful for HTSlib to have a preprocessor macro so that programs could enquire about the version of HTSlib being _compiled against_ at compile time, and emit an understandable `#error` or (if they are overachievers) adjust their code accordingly. This PR adds a macro so that user code can write tests like

```c
#if HTS_VERSION >= 101000  // Check for v1.10 or later
// …use hts_parse_region() or some other new API functions
#endif
```
    
This encodes the version number X.Y.Z into a single integer XYYYZZ so that comparisons are easy. The XYYYZZ encoding is the same as that used by [Boost](https://www.boost.org/doc/libs/1_71_0/boost/version.hpp).

(I had prepared a version of this with an associated `HTS_VERNUM` macro to hide the numeric encoding details (see jmarshall/htslib@debead4d39d7d97ac0f05ed7d91651aa27c00df1) to be used as `#if HTS_VERSION >= HTS_VERNUM(1,10,0)`, but that alas leads to errors when compiling against earlier HTSlib releases that do not define `HTS_VERNUM`. See also https://github.com/irods/irods/issues/2671#issuecomment-100267437.)

Because you don't want a public header to be dependent on configuration or building — cf https://github.com/samtools/htslib/issues/559#issuecomment-314106899 and the following conversation — this would be updated manually (or via the release scripts) during and after each release. Bumping the patch-number after release means that the value in a development Git repository header is distinct from the value in the preceding release, but it's not proposed to update it constantly with each commit to **develop**! OTOH from time to time during development, you may wish to bump the patch-number from 90 to 91 to 92 etc when truly significant facilities are merged — somewhat similarly to the current `TWO_TO_THREE_TRANSITION_COUNT` development counter.

There's some pain in this as like the version numbers in _version.sh_, _NEWS_, and the manual pages, it would need to be updated during the release process. However you already have the _mkrelease_ and _tag_release_ scripts for that so after it was added to those, it would not be onerous.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,7 @@ NUMERIC_VERSION := $(shell ./version.sh numeric)
 version.h: $(if $(wildcard version.h),$(if $(findstring "$(PACKAGE_VERSION)",$(shell cat version.h)),,force))
 
 version.h:
-	echo '#define HTS_VERSION "$(PACKAGE_VERSION)"' > $@
+	echo '#define HTS_VERSION_TEXT "$(PACKAGE_VERSION)"' > $@
 
 print-version:
 	@echo $(PACKAGE_VERSION)
--- a/NEWS
+++ b/NEWS
@@ -162,6 +162,11 @@ Noteworthy changes in release a.b
   alignment record data; see documentation in htslib/sam.h for more
   information.
 
+* In addition to the existing hts_version() function, which reflects the
+  HTSlib version being used at runtime, <htslib/hts.h> now also provides
+  HTS_VERSION, a preprocessor macro reflecting the HTSlib version that
+  a program is being compiled against.  (#794, PR #951)
+
 * Changes to hfile_s3, which provides support for the AWS S3 API.
 
   - hfile_s3 now uses version 4 signatures by default.  Attempting to write to
--- a/hfile_gcs.c
+++ b/hfile_gcs.c
@@ -124,7 +124,7 @@ int PLUGIN_GLOBAL(hfile_plugin_init,_gcs)(struct hFILE_plugin *self)
 
 #ifdef ENABLE_PLUGINS
     // Embed version string for examination via strings(1) or what(1)
-    static const char id[] = "@(#)hfile_gcs plugin (htslib)\t" HTS_VERSION;
+    static const char id[] = "@(#)hfile_gcs plugin (htslib)\t" HTS_VERSION_TEXT;
     if (hts_verbose >= 9)
         fprintf(stderr, "[M::hfile_gcs.init] version %s\n", strchr(id, '\t')+1);
 #endif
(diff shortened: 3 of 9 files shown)


jkbonfield:
Good idea and much needed.

I'm a bit wary of changing HTS_VERSION to HTS_VERSION_TEXT and replacing it with a numerical define instead.

On the plus side, this shouldn't matter because it was never exported and thus it only affects our own code, all of which you've changed.

On the minus side, htslib is embedded 100s of times in various git projects with all sorts of horrid shenanigans going on, so I wouldn't guarantee people aren't using HTS_VERSION already and this is likely to break something, somewhere.

On the plus side, this punishes such underhanded behaviour...

OK maybe I've talked myself into it after all. :-)

jkbonfield:
Also I'm assuming the correct usage is something like this:

```
#if defined(HTS_VERSION) && HTS_VERSION >= 101000                                          
char *foo = "new";                                                                         
#else                                                                                      
char *foo = "old";                                                                         
#endif                                                                                     
```

So we can handle htslib before this was added as well as those after.  We really need to start some htslib documentation, instead of shoving it all over to the samtools package, so we can document how to correctly use the library.  That's a very different PR though!

I'm not convinced on the bumping patch level to 90 straight away for develop as it doesn't leave much leeway for bumpage.  Either we'd want 3 digits there too, or pick something more midway.  The idea is sound though.

jmarshall:
The preprocessor treats undefined macros in `#if` as 0, so either with or without the `defined(HTS_VERSION)` is fine for old HTSlibs.

90 was chosen because it's familiar from [GNU project pre-releases](https://www.gnu.org/prep/maintain/html_node/Test-Releases.html). Third-party code would be unwise to depend on particular **develop** snapshots, so I would think it would be unusual to get bumped past 90 at all — such bumps would only be optional hints for the most significant API additions, so should be even rarer than `TWO_TO_THREE_TRANSITION_COUNT` bumps.

jmarshall:
> On the plus side, this punishes such underhanded behaviour...

Possibly it doesn't punish it quite enough. This PR could also rename the built _version.h_ to _htslib_version.h_ so such code failed with an unequivocal header-file-not-found, which would also add further bulletproofing to samtools/samtools#817.

PR #952, opened 2019-10-16
Make the sam parser detect out-of-range numbers

daviesrob:
Replace STRTOL64() and STRTOUL64() functions with versions that can detect when the parsed value will not fit into the destination.  The new functions are called hts_str2int() and hts_str2uint() as the
interface has changed somewhat from that of strtol().  The new functions have been put in textutils_internal.h so that they can be used by other parsers (notably vcf) in future updates.

Add tests for the new hts_str2int() and hts_str2uint() functions.

Update sam_parse1() to use these functions and report an error if a number was too large.  The B array code is pulled out into its own function.  The new function includes code to fix up B arrays where the subtype code was set incorrectly (at the cost of a couple of extra scans through the data).

This has been designed to have minimal impact on the speed of the SAM parser.  If fact it may even be faster as the compiler (notably clang) can see that the "fast" loop is bounded in size and decides to unroll it.

Credit to OSS-Fuzz
Fixes oss-fuzz 18142

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,7 @@ BUILT_TEST_PROGRAMS = \
 	test/test_kstring \
 	test/test_realn \
 	test/test-regidx \
+	test/test_str2int \
 	test/test_view \
 	test/test_index \
 	test/test-vcf-api \
@@ -384,6 +385,7 @@ maintainer-check:
 check test: $(BUILT_PROGRAMS) $(BUILT_TEST_PROGRAMS)
 	test/hts_endian
 	test/test_kstring
+	test/test_str2int
 	test/fieldarith test/fieldarith.sam
 	test/hfile
 	test/test_bgzf test/bgziptest.txt
@@ -427,6 +429,9 @@ test/test-regidx: test/test-regidx.o libhts.a
 test/test-parse-reg: test/test-parse-reg.o libhts.a
 	$(CC) $(LDFLAGS) -o $@ test/test-parse-reg.o libhts.a $(LIBS) -lpthread
 
+test/test_str2int: test/test_str2int.o
+	$(CC) $(LDFLAGS) -o $@ test/test_str2int.o
+
 test/test_view: test/test_view.o libhts.a
 	$(CC) $(LDFLAGS) -o $@ test/test_view.o libhts.a $(LIBS) -lpthread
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -456,6 +461,7 @@ test/test_kstring.o: test/test_kstring.c config.h $(htslib_kstring_h)
 test/test-parse-reg.o: test/test-parse-reg.c config.h $(htslib_hts_h) $(htslib_sam_h)
 test/test_realn.o: test/test_realn.c config.h $(htslib_hts_h) $(htslib_sam_h) $(htslib_faidx_h)
 test/test-regidx.o: test/test-regidx.c config.h $(htslib_regidx_h) $(hts_internal_h)
+test/test_str2int.o: test/test_str2int.c $(textutils_internal_h)
```
jmarshall: Also _config.h_.

test/test_str2int.c
```diff
@@ -0,0 +1,155 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.  */
+
+
+#include "config.h"
```
jmarshall: `<config.h>`

daviesrob:
Both `config.h` issues now fixed.

textutils_internal.h
```diff
@@ -171,6 +171,125 @@ static inline int isupper_c(char c) { return isupper((unsigned char) c); }
...
+The address of the first character following the converted number will
+be stored in *end.
+
+Both end and failed must be non-NULL.
+ */
+
+static inline uint64_t hts_str2uint(const char *in, char **end, int bits,
+                                      int *failed) {
+    uint64_t n = 0, limit = (bits < 64 ? (1ULL << bits) : 0) - 1;
+    uint64_t limit_d_10 = limit / 10, limit_m_10 = limit - 10 * limit_d_10;
```
valeriuo: The initialization of `limit_d_10` and `limit_m_10` could be done inside the `if(!fast)` block, as it has been done above, for `hts_str2int`.

daviesrob:
Rebased and made initialization of `limit_d_10` and `limit_m_10` is consistent between the two functions.

The point where they get set up shouldn't make much difference in practice.  The static inlines are always called with a constant value for `bits` so, as long as the function is inlined, `fast`, `limit`, `limit_d_10` and `limit_m_10` are also constant and can be calculated at compile time.

textutils_internal.h
```diff
@@ -171,6 +171,125 @@ static inline int isupper_c(char c) { return isupper((unsigned char) c); }
...
+            } else {
+                do { v++; } while (*v - ascii_zero < 10);
+                n = limit;
+                *failed = 1;
+                break;
+            }
+        }
+    }
+
+    *end = (char *)v;
```
valeriuo: `strtoll` allows the end pointer to be null, in which case it is not used. We should probably implement a similar behaviour here.

daviesrob: I didn't bother to do that as every use of this function has a non-NULL `end`.  Mostly it's passing back a pointer to the location after the number so whatever parser called it can continue.  Are there any places where we might want to use this that do pass `NULL` here?

valeriuo: If we don't see any use beyond the current ones, considering also that is not a public method, I suppose there is no need. I was thinking of a scenario where `hts_str2int` was called inside a `kstrtok` loop.

valeriuo:
Merged into develop as d123a44adb57dd3927c47d59ab45cf677e439483

PR #954, opened 2019-10-21
Return lengths from bam_cigar2rqlens as hts_pos_t and make it static

daviesrob:
A bit of left-over 64 bit work, discovered by oss-fuzz.

While it's unlikely that a single CIGAR could cover over 4 Gbases of reference, I wouldn't like to guarantee that someone won't try to make one so we may as well support it.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -450,7 +450,7 @@ static void bam_cigar2rqlens(int n_cigar, const uint32_t *cigar,
     }
 }
 
-int bam_cigar2qlen(int n_cigar, const uint32_t *cigar)
+hts_pos_t bam_cigar2qlen(int n_cigar, const uint32_t *cigar)
 {
     int k, l;
     for (k = l = 0; k < n_cigar; ++k)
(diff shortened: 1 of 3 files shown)


jkbonfield:
Given the presence of the N operator and tools that can do RNA-seq alignments, it's quite possible that we can get extreme CIGAR ops, at least in ASCII SAM terms.

valeriuo:
Merged into develop

sam.c
```diff
@@ -449,7 +450,7 @@ void bam_cigar2rqlens(int n_cigar, const uint32_t *cigar, int *rlen, int *qlen)
     }
 }
 
-int bam_cigar2qlen(int n_cigar, const uint32_t *cigar)
+hts_pos_t bam_cigar2qlen(int n_cigar, const uint32_t *cigar)
```
jmarshall: If changing the return type, the type of `l` needs to change too.

daviesrob: This shows the danger of updating code while jet-lagged.  Should now be fixed by 8ac3b5c4.

PR #955, opened 2019-10-23
Replace assert with an error mesage and return

valeriuo:
If there are duplicate entries in the `sam_hdr_t::target_name` of a BAM file, one or more of the ref stubs end up with a header line type null pointer. Although, symptoms of this issue appear earlier in the call hierarchy ([sam_hrecs_refs_from_targets_array](https://github.com/samtools/htslib/blob/5162f01532d10514a44b76af6e41f205aa8b646b/header.c#L1018) and [sam_hrecs_update_hashes](https://github.com/samtools/htslib/blob/5162f01532d10514a44b76af6e41f205aa8b646b/header.c#L216)), a check for consistency and validity of the internal header data is still necessary in `add_stub_ref_sq_lines`.

Credit to OSS-Fuzz
Fixes oss-fuzz 18148

### DIFF
--- a/header.c
+++ b/header.c
@@ -210,19 +210,17 @@ static int sam_hrecs_update_hashes(sam_hrecs_t *hrecs,
             }
 
             // If here, the name is a duplicate.
-            // Check to see if it matches the SN: tag from the earlier
-            // record.  If it does, drop this and keep the existing one.
+            // Check to see if it matches the SN: tag from the earlier record.
             if (strcmp(hrecs->ref[nref].name, name) == 0) {
-                hts_log_warning("Duplicate entry \"%s\" in sam header",
+                hts_log_error("Duplicate entry \"%s\" in sam header",
                                 name);
-                return 0;
+                return -1;
             }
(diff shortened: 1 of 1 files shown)


daviesrob:
I wonder if we should bail out earlier in this case (by changing the warning in `sam_hrecs_refs_from_targets_array` to an error)?  I think the idea was that it should be possible to recover from the duplication by making the two tids effectively aliases for each other, but obviously that doesn't actually work.

valeriuo:
> I think the idea was that it should be possible to recover from the duplication by making the two tids effectively aliases for each other, but obviously that doesn't actually work.

It could probably work if we remove the duplicate entry at the point of origin, the `sam_hdr_t::target_name` array, and adjust the `sam_hdr_t::n_targets` in the process. I'll give it a try.

daviesrob:
It looks like that would break the relationship between indexes into the reference list and tid values in the alignment records.

I think just giving up on broken headers is the easiest solution.

PR #957, opened 2019-10-23
add check for invalid tid < 0 and return null

### LINKED ISSUE #956: Incorrect result sam_hdr_tid2name with negative tid
### Description

`const char *sam_hdr_tid2name(const sam_hdr_t *h, int tid)` does not check for a negative `tid` and can return a garbage address (which lead to a crash trying to make a std::string)

### Example
```
htsFile* file = hts_open("some.bam", 'r');
bam_hdr_t* header = sam_hdr_read(file);
const char* c = sam_hdr_tid2name(header, -1); 
Expects(c == nullptr); // fails
```

Version: recentish develop: d59c364

sg-james:
closes #956


daviesrob:
Thanks for the fix.

PR #961, opened 2019-10-28
Cram unmapped reset

jkbonfield:
For transition from mapped to unmapped the types of codec to use in CRAM can change, particularly for the BA data series (used for unmapped sequence bases).

The data that is unmapped but placed along side the mapped, is often fairly random looking and entropy encoding does well.  The pure unmapped data is also usually fairly random looking, but with "samtools sort -M" can have strong similarities with previous reads making gzip or lzma a good tool.

As a demonstration on SRR1770413 (bwa aligned, sort -M).  Current htslib gives BA at 12.5Mb (larger with big slice sizes as it doesn't get time to restart the compression trials).  With this patch it's 6.6Mb for the same data (5.0Mb with lzma).

The second commit enables libdeflate usage within CRAM.  Previously we only used this for BAM and for CRCs in CRAM, but it's now available as an alternative to zlib for actual gzip compressed blocks.

One reason we didn't do this in the past was because enabling libdeflate doesn't save a great deal of time in CRAM - it mostly spends time on non-deflate things - and also because it was giving poorer compression ratios.  We've worked around this latter case so may as well take the small speed wins.  A small test set:

```
       	   Decode(s)		Encode(s)		Size(Mb)
       	   LibD	  Zlib		LibD   Zlib		LibD	Zlib
HiSeq      2.16	  2.20		5.36   5.65		59.03	59.02
HiSeqX     1.37	  1.43		3.72   3.78		17.48	17.78
NovaSeq	   1.93	  2.00		4.68   4.83		22.46   22.46
```

This paves the way to permitting higher compression ratios now as libdeflate's max compression is considerably higher than zlib's, albeit at a CPU cost:

```
       	   Decode(s)		Encode(s)		Size(Mb)
       	   LibD	  Zlib		LibD   Zlib		LibD	Zlib
HiSeq      ?	  ?		16.5   10.0		58.41	58.69
HiSeqX     ?	  ?		21.0   10.4		16.65	16.90
NovaSeq	   ?	  ?		20.2   12.1		22.01   22.26
```

Marginal I guess.

### DIFF
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -2630,13 +2630,9 @@ static cram_container *cram_next_container(cram_fd *fd, bam_seq_t *b) {
                      c->ref_seq_start + c->ref_seq_span -1);
 
         /* Encode slices */
-        if (fd->pool) {
-            if (-1 == cram_flush_container_mt(fd, c))
-                return NULL;
-        } else {
-            if (-1 == cram_flush_container(fd, c))
-                return NULL;
-
+        if (-1 == cram_flush_container_mt(fd, c))
+            return NULL;
+        if (!fd->pool) {
             // Move to sep func, as we need cram_flush_container for
             // the closing phase to flush the partial container.
             for (i = 0; i < c->max_slice; i++) {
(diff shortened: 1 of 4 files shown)


daviesrob:
Rebased with minor tidying up.

PR #962, opened 2019-10-29
Prevent overflowing of string parsed integers in VCF files

valeriuo:
The success of `strto{l,ll,d}` methods is rarely checked. However, when the out-of-range parsed value is subsequently used in arithmetic operations, it could lead to an overflow.

Credit to OSS-Fuzz
Fixes oss-fuzz 18149
Fixes oss-fuzz 18164
Fixes oss-fuzz 18193

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -2327,11 +2327,20 @@ static int vcf_parse_format(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v, char *p
                 if (z->is_gt) { // genotypes
                     int32_t is_phased = 0, *x = (int32_t*)(z->buf + z->size * m);
                     for (l = 0;; ++t) {
-                        if (*t == '.') ++t, x[l++] = is_phased;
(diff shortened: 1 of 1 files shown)


vcf.c
```diff
@@ -2321,11 +2321,19 @@ static int vcf_parse_format(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v, char *p
...
-                        else x[l++] = (strtol(t, &t, 10) + 1) << 1 | is_phased;
-#if THOROUGH_SANITY_CHECKS
-                        assert( 0 );    // success of strtol,strtod not checked
-#endif
+                        if (*t == '.') {
+                            ++t, x[l++] = is_phased;
+                        } else {
+                            char *tt = t;
+                            x[l] = strtol(t, &t, 10);
+                            if ((errno&ERANGE) == ERANGE || x[l] > (INT32_MAX>>1)-1) {
```
jmarshall: That's not how `errno` works. Reread Linux's `strtol` man page.

PR #965, opened 2019-11-06
Changed int64_t to hts_pos_t for *_parse_region API

jkbonfield:
Also fixed a few other assumptions of hts_pos_t equalling int64_t.

Updated the test-parse-reg to use hts_pos_t as well.

Added a comment to hts_parse_reg64 indicating that it is not our
preferred API.  It was added in a separate commit for ease of API
migration from hts_parse_reg, but was added only because the
hts_parse_region code hadn't been accepted.  This is now the correct
interface.

---

Note the name of hts_parse_reg64 is mildly annoying.  When Linux switched from fseek taking long to taking off_t it didn't use fseek64, but fseeko.  It's largely irrelevant though as both hts_parse_reg and hts_parse_reg64 should be considered deprecated.

### DIFF
--- a/faidx.c
+++ b/faidx.c
@@ -744,7 +744,7 @@ static int fai_get_val(const faidx_t *fai, const char *str,
     khiter_t iter;
     khash_t(s) *h;
     int id;
-    int64_t beg, end;
+    hts_pos_t beg, end;
 
     if (!fai_parse_region(fai, str, &id, &beg, &end, 0)) {
         hts_log_warning("Reference %s not found in FASTA file, returning empty sequence", str);
@@ -913,7 +913,8 @@ int faidx_has_seq(const faidx_t *fai, const char *seq)
 }
 
 const char *fai_parse_region(const faidx_t *fai, const char *s,
-                             int *tid, int64_t *beg, int64_t *end, int flags)
+                             int *tid, hts_pos_t *beg, hts_pos_t *end,
+                             int flags)
 {
     return hts_parse_region(s, tid, beg, end, (hts_name2id_f)fai_name2id, (void *)fai, flags);
 }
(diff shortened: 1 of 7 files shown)


jmarshall:
`hts_parse_reg64` wasn't present in the 1.9 release. So if samtools is updated to use `hts_parse_region` instead, then `hts_parse_reg64` could just be removed.

And `hts_parse_reg` could have a `HTS_DEPRECATED` annotation. (I think at the moment the compatibility NOTE you've added is in the wrong place if it's for `hts_parse_reg64`.)

jkbonfield:
The note was put on that function for a reason and IMO it's not in the wrong place.  Existing code will be using hts_parse_reg.   I wanted to persuade people that new code ought to be using hts_parse_region and not hts_parse_reg64.

I'm also clearly aware that `hts_parse_reg64` is new, but it's there because it is an easy upgrade from `hts_parse_reg` for those that for some reason can't keep track of headers or provide name validation lookup functions.  I did think about removing it completely and forcing people to migrate properly.  However this all arrived in the wrong order - or more to the point the work got done twice as we've been laggy in merging.

As for deprecating things.  I think originally my intention was to do that, spread over two releases with the initial code deprecated and the second release to remove it.  See the comments above hts_parse_reg64.   Although sadly it's taken so long that I've rather forgotten what my plans were!

jmarshall:
Indeed. So there is the opportunity to remove it or rename it if the name was sufficiently annoying.

> My note was put on that function for a reason and IMO it's not in the wrong place.

Yeah, at the moment it says that `hts_parse_reg` is “for compatibility with `hts_parse_reg` only”. Seems like the comment block one above was the one you were aiming for…

jkbonfield:
Duh sorry!  I see what you mean, and yes I got it wrong.  Thanks.

I'll let @daviesrob decide whether or not we want to introduce a new API function that we don't recommend.  I'm betwixt and between on it.

PR #966, opened 2019-11-06
Fix crash when loading external plug-ins into Windows executables that statically link HTSlib

### LINKED ISSUE #964: External plug-ins are broken for static builds on Windows
Due to [differences in the way shared libraries work on Windows](https://stackoverflow.com/questions/6251171), plug-in .dlls end up linking against the shared HTSlib dll even if the executable includes the library in a static link.  This results in there being two copies of the [`schemes`](https://github.com/samtools/htslib/blob/565560e/hfile.c#L865) global variable leading to a segmentation fault when the plug-in tries to register itself by calling [`hfile_add_scheme_handler`](https://github.com/samtools/htslib/blob/565560e/hfile.c#L904).

This doesn't happen on UNIX-like systems because the run-time link resolves `schemes` to the one in the executable itself.  On Windows, the executable initialises its copy, but the plug-in only sees the one in the dll, which never gets set up so the pointer is NULL.

Possible solutions:

* Only use dynamic links on Windows.  As long as everything uses the same .dll the plug-in gets the right symbol.

* If `hfile_add_scheme_handler` finds that `schemes` is NULL on entry, try using `dlopen`/`dlsym` to locate the copy in the main executable and use that instead.

daviesrob:
Works around the plug-in's insistence on linking against and using the shared hts.dll by spotting when this causes a problem and using `dlopen()` /  `dlsym()` to find `hfile_add_scheme_handler()` in the executable and call that instead.  This ensures the plug-in is registered in the correct place.  It's not elegant but it works.

Additional bonus fixes:
- A missing dependencies on hts.dll.a which could result in the plug-in dlls trying to build before the shared library was available.
- Some char-subscripts warnings due to use of `ctype` functions with `char` types.  These don't appear to trigger warnings in our CI builds (including Appveyor) but did when I made a MinGW build by hand.

Fixes #964

### DIFF
--- a/hfile.c
+++ b/hfile.c
@@ -34,6 +34,13 @@ DEALINGS IN THE SOFTWARE.  */
 
 #include <pthread.h>
 
+#ifdef ENABLE_PLUGINS
+#if defined(_WIN32) || defined(__CYGWIN__) || defined(__MSYS__)
+#define USING_WINDOWS_PLUGIN_DLLS
+#include <dlfcn.h>
+#endif
+#endif
+
 #include "htslib/hfile.h"
 #include "hfile_internal.h"
 #include "htslib/kstring.h"
@@ -906,11 +913,54 @@ static inline int priority(const struct hFILE_scheme_handler *handler)
     return handler->priority % 1000;
 }
 
+#ifdef USING_WINDOWS_PLUGIN_DLLS
+/*
+ * Work-around for Windows plug-in dlls where the plug-in could be
+ * using a different HTSlib library to the executable (for example
+ * because the latter was build against a static libhts.a).  When this
+ * happens, the plug-in can call the wrong copy of hfile_add_scheme_handler().
+ * If this is detected, it calls this function which attempts to fix the
+ * problem by redirecting to the hfile_add_scheme_handler() in the main
+ * executable.
+ */
+static int try_exe_add_scheme_handler(const char *scheme,
+                                      const struct hFILE_scheme_handler *handler)
+{
+    static void (*add_scheme_handler)(const char *scheme,
+                                      const struct hFILE_scheme_handler *handler);
+    if (!add_scheme_handler) {
+        // dlopen the main executable and resolve hfile_add_scheme_handler
+        void *exe_handle = dlopen(NULL, RTLD_LAZY);
+        if (!exe_handle) return -1;
+        *(void **) (&add_scheme_handler) = dlsym(exe_handle, "hfile_add_scheme_handler");
+        dlclose(exe_handle);
+    }
+    // Check that the symbol was obtained and isn't the one in this copy
+    // of the library (to avoid infinite recursion)
+    if (!add_scheme_handler || add_scheme_handler == hfile_add_scheme_handler)
+        return -1;
+    add_scheme_handler(scheme, handler);
+    return 0;
+}
+#else
+static int try_exe_add_scheme_handler(const char *scheme,
+                                      const struct hFILE_scheme_handler *handler)
+{
+    return -1;
+}
+#endif
+
 HTSLIB_EXPORT
 void hfile_add_scheme_handler(const char *scheme,
                               const struct hFILE_scheme_handler *handler)
 {
     int absent;
+    if (!schemes) {
+        if (try_exe_add_scheme_handler(scheme, handler) != 0) {
+            hts_log_warning("Couldn't register scheme handler for %s", scheme);
+        }
+        return;
+    }
     khint_t k = kh_put(scheme_string, schemes, scheme, &absent);
     if (absent || priority(handler) > priority(kh_value(schemes, k))) {
         kh_value(schemes, k) = handler;
--- a/plugin.c
+++ b/plugin.c
@@ -44,7 +44,7 @@ static DIR *open_nextdir(struct hts_path_itr *itr)
     DIR *dir;
 
     while (1) {
-        const char *colon = strchr(itr->pathdir, ':');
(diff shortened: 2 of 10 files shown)


jkbonfield:
I can confirm using mingw64 that schemes is still NULL when trying to add the handler, causing the crash.  However I can also confirm that this does not fix it.

Infact it makes things worse.  With PLUGIN_PATH set to an inappropriate location it can read ftp:// (falling back to knet).  With PLUGIN_PATH set correctly it can't read ftp.  It no longer just crashes as before, but it also cannot read the files.

jkbonfield:
One other thing,  I had to remove  `-rdynamic` from the configure script to get `--with-plugins` to build on Windows.  So that's also something to add on the to-fix list.

daviesrob:
libcurl with http works; https and ftp don't for some reason.

daviesrob:
Autoconf now checks that the compiler understands `-rdynamic`.

jmarshall:
Not sure why the answer isn't “tell Windows people to build their executables in a consistent way” 😛  (and am curious why it didn't work for James) but if this wart is for the benefit of Windows only maybe it could be wrapped in `#ifdef _WIN32`.

Now that the configure script checks for `-rdynamic` support (yay), the `TODO` comment can go away. Or maybe you left it to signal further scope for tweaking this stuff..?

daviesrob:
It turns out that `#ifdef _WIN32` is only defined for `mingw-w64-x86_64` builds.  The `MSYS` one doesn't define it but still needs the hack to get the right symbol.  As it should never do anything on UNIX-like systems it just seems easier to leave the guards out.

The TODO comment was indeed left to signal that we should really work out which platforms really need `-rdynamic` - but at least it's less likely to break things now.

jmarshall:
I'm sure [MSYS defines something](https://github.com/msys2/msys2/wiki/Porting#platform-checks), and it could all be encapsulated at the top of this file with its other platform warts:

```diff
--- a/hfile.c
+++ b/hfile.c
@@ -501,12 +501,16 @@ void hclose_abruptly(hFILE *fp)
 #ifndef _WIN32
 #include <sys/socket.h>
 #include <sys/stat.h>
 #define HAVE_STRUCT_STAT_ST_BLKSIZE
+#if defined __CYGWIN__ || defined __MSYS__
+#define NEED_DYNAMIC_LINKING_WORKAROUND
+#endif
 #else
 #include <winsock2.h>
 #define HAVE_CLOSESOCKET
 #define HAVE_SETMODE
+#define NEED_DYNAMIC_LINKING_WORKAROUND
 #endif
 #include <fcntl.h>
 #include <unistd.h>
 ```

and then `#if defined ENABLE_PLUGINS && defined NEED_DYNAMIC_LINKING_WORKAROUND`…
(or `HAVE_BROKEN_DLLS` if you prefer the `HAVE_…` pattern :smile:)

**Comment by daviesrob** [comment]
By popular request, I've restricted this work-around to Windows builds.

File: plugin.c [hunk 1]
```diff
@@ -44,7 +44,11 @@ static DIR *open_nextdir(struct hts_path_itr *itr)
     DIR *dir;
 
     while (1) {
+#if defined(_WIN32) || defined(__MSYS__)
+        const char *colon = strchr(itr->pathdir, ';');
+#else
```
jmarshall: Did you consider defining `PATH_SEPARATOR` as `':'` or `';'` as appropriate instead of sprinkling this guano everywhere and duplicating code? (Yes, you'll need `PATH_SEPARATOR_STRING` or other tricks as well.)

htslib/hts_defs.h
```diff
@@ -111,4 +111,12 @@ DEALINGS IN THE SOFTWARE.  */
...
 #endif
 
+#if defined(_WIN32) || defined(__MSYS__)
+#define HTS_PATH_SEPARATOR_CHAR ';'
+#define HTS_PATH_SEPARATOR_STR  ";"
+#else
+#define HTS_PATH_SEPARATOR_CHAR ':'
+#define HTS_PATH_SEPARATOR_STR  ":"
+#endif
+
```
jmarshall: This could be private to HTSlib, alongside the `hts_path_itr` stuff in _hts_internal.h_.

Or if it's decided to make it public, it would be better in _hts_os.h_ or perhaps _hts.h_ — but not really _hts_defs.h_ which is about compiler capabilities.

jmarshall: The “other tricks” BTW is the following (though if made public it's probably best to provide both of them anyway rather than making users needing a string use this trick too);

```c
static const char path_sep_string[] = { HTS_PATH_SEPARATOR_CHAR, '\0' };
```

daviesrob:
I've moved the `#define`s to `htslib/hts.h`.  I think it's useful to make them public, in case anything wants to find out easily which path separator is in use.

hfile.c
```diff
@@ -906,11 +913,54 @@ static inline int priority(const struct hFILE_scheme_handler *handler)
...
 void hfile_add_scheme_handler(const char *scheme,
                               const struct hFILE_scheme_handler *handler)
 {
     int absent;
+    if (!schemes) {
+        if (try_exe_add_scheme_handler(scheme, handler) != 0) {
+            hts_log_warning("Couldn't register scheme handler for %s", scheme);
+        }
+        return;
+    }
```
jmarshall: Putting this inside `#ifdef USING_WINDOWS_PLUGIN_DLLS` too indicates that this test is irrelevant on Unix. (And then you don't need the `#else` and stub above.)

daviesrob: I thought of doing that, but decided that it would be better that in the (very unlikely) event of getting here with `schemes == NULL` on UNIX, it prints the message and bails out gracefully.

PR #967, opened 2019-11-08
Search for a local .crai index first

valeriuo:
When doing a indexed search in a CRAM file, look first for a local CRAI index file in the current folder. If unsuccessful and the CRAM file is remote (e.g. S3, HTTP, FTP), go to the same remote location and try to fetch it from there.

Fixes https://github.com/samtools/samtools/issues/1132
Improves https://github.com/samtools/htslib/pull/870

### DIFF
--- a/cram/cram_index.c
+++ b/cram/cram_index.c
@@ -143,7 +143,8 @@ static int kget_int64(kstring_t *k, size_t *pos, int64_t *val_p) {
  *        -1 for failure
  */
 int cram_index_load(cram_fd *fd, const char *fn, const char *fn_idx) {
-    char *fn2 = NULL;
+
+    char *tfn_idx = NULL;
     char buf[65536];
     ssize_t len;
     kstring_t kstr = {0};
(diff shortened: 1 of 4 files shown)


whitwham:
Looks good and works well apart from in one case.  File names with a # in them (or a ?) fail because the file handler is treating them as a simple URL and not treating anything after these characters as part of the name. e.g.

andrew_test_file#1.cram // _fails, becomes andrew_test_file.crai#1.cram_

If I manually escape the # it works. e.g.

andrew_test_file%231.cram // _works, becomes andrew_test_file%231.cram.crai_

Extension handling code is [here](https://github.com/samtools/htslib/blob/ca891e4ba6af28cac0fb48a88cbc59ee5974f561/hfile.c#L1081)

valeriuo:
I see that Pandas have solved a similar issue by using a `allow_fragments` [flag](https://github.com/pandas-dev/pandas/pull/25992/files) in their URL parsing method. This would mean another user option that would trickle down to `haddextension` function.

A simpler solution would be a hardcoded exception for S3 URLs baked into `haddextension`. Something like:
```c
--- a/hfile.c
+++ b/hfile.c
@@ -1085,7 +1085,7 @@ char *haddextension(struct kstring_t *buffer, const char *filename,

     if (find_scheme_handler(filename)) {
         // URL, so alter extensions before any trailing query or fragment parts
-        trailing = filename + strcspn(filename, "?#");
+        trailing = filename + strncmp(filename, "s3://", 5) ? strcspn(filename, "?#") : strlen(filename);
     }
```

Opinions?

whitwham:
For the simple solution, maybe when should put the code inside **\#ifdef ENABLE_S3**.  So only handle it if the S3 plugin has been enabled.

daviesrob:
`?` can occur in s3 URLs, so we should always search for it.  I'm less bothered about `#` as it doesn't have any meaning for the file types we are interested in.

hfile_s3 accepts URLs in any of the forms `s3://`, `s3+http://` or `s3+https://`, so we would need to check for all of these if we wanted to make an exception.

PR #968, opened 2019-11-11
Fix waiting mechanism on Windows

valeriuo:
`select` on Windows is unreliable as a sleeping mechanism, as it spuriously returns -1 before timeout, without setting `errno`.

https://curl.haxx.se/libcurl/c/multi-double.html

### DIFF
--- a/hfile_libcurl.c
+++ b/hfile_libcurl.c
@@ -700,8 +700,19 @@ static int wait_perform(hFILE_libcurl *fp)
                 timeout = 10000;  // as recommended by curl_multi_timeout(3)
             }
         }
-        if (maxfd < 0 && timeout > 100)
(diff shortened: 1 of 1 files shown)


jmarshall:
The problem on Windows is only for the `maxfd == -1` case, and the curl example uses `Sleep` conditionally on that. So this would be better as (and the code made less manky by)

```c
+#ifdef _WIN32
+        // Windows's select() does not support the nfds==0 case
+        if (maxfd < 0 && timeout > 0)
+            Sleep(timeout);
+        else
+#endif
         if (timeout > 0) {
             struct timeval tval;
```

Or add `{ Sleep(timeout); timeout = 0; }` to the `Sleep` if you prefer to avoid the `else // #endif // if` trick, though that would be more obscure.

It deserves a comment, and it would be good to have a link to the curl example in the commit message.

PR #969, opened 2019-11-11
Fix htslib.mk parallel build

daviesrob:
Fixes a missing shared dependency on the HTSlib object files for the (Windows and Mac) shared and static library targets in `htslib.mk`.  This was causing random [build failures](https://ci.appveyor.com/project/samtools/bcftools/builds/28720476) for BCFtools on Appveyor because it was trying to build both the shared and static libraries at the same time, resulting in the `.o` files being built twice and one of the libraries to become corrupt because one of the object files being included got rebuilt as it was added.

The fix creates an empty file target file `hts-object-files` in HTSlib's Makefile which `htslib.mk` uses to track the shared dependency for the libraries.  While adding an empty target isn't ideal (for example it won't notice missing `.o` files if the `.c` file is unchanged), it seems to be the best solution given that `htslib.mk` doesn't keep track of HTSlib's configure options, so it can't easily work out which object files are going into the libraries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ hfile_*.cygdll
 hfile_*.dll
 hfile_*.so
 
+hts-object-files
 htslib_static.mk
 
 cyg*.dll
--- a/Makefile
+++ b/Makefile
@@ -299,6 +299,10 @@ cyghts-$(LIBHTS_SOVERSION).dll: $(LIBHTS_OBJS)
 hts-$(LIBHTS_SOVERSION).dll hts.dll.a: $(LIBHTS_OBJS)
 	$(CC) -shared -Wl,--out-implib=hts.dll.a -Wl,--enable-auto-import -Wl,--exclude-all-symbols $(LDFLAGS) -o $@ -Wl,--whole-archive $(LIBHTS_OBJS) -Wl,--no-whole-archive $(LIBS) -lpthread
 
+# Target to allow htslib.mk to build all the object files before it
+# links the shared and static libraries.
+hts-object-files: $(LIBHTS_OBJS)
+	touch $@
 
 .pico.so:
 	$(CC) -shared -Wl,-E $(LDFLAGS) -o $@ $< $(LIBS) -lpthread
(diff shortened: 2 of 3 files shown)


jmarshall:
Good debugging, but I feel like the proposed solution is treating the effect, not the cause. (And I'm not sure that there aren't still race conditions.)

The root cause of this is that the bcftools Makefile caused two submakes to be running in the htslib directory at the same time. Any time you have two makes running in the same directory at the same time (e.g., just type `make` in two different terminal windows), chaos ensues. The solution is “don't do that”.

So it's not about the libraries; rather, it's about the `$(MAKE)` recipes. There should be a way just in _htslib.mk_ to ensure those half dozen rules can't run in parallel, but off the top of my head I'm not sure what it is right now…

jkbonfield:
My suggestion to this was to replace "make a; make b" type statements with "make a b" so there's only one make and it can parallelise correctly without trampling on each other.

However it's complicated because the instructions on which things to make are in htslib.mk.

daviesrob:
The only way I've found so far that would stop the `$(MAKE)` invocations from running at the same time would be to add `.NOTPARALLEL`.  As it would affect the entire Makefile and not  just these rules, it would be a bit of a high impact solution.  There's also no way (as far as I can tell) of gathering all the targets you want together and running them in a single recursive `$(MAKE)`.

Any better solutions would be welcome, meanwhile this does solve the immediate problem.

A build of bcftools that uses this patch can be seen [here](https://ci.appveyor.com/project/daviesrob/bcftools/builds/28768659).

jmarshall:
> My suggestion to this was to replace "make a; make b" type statements with "make a b" so there's only one make and it can parallelise correctly without trampling on each other.

Hmmm… funnel them all through a single common submake. That's rather nice and could be done something like this:

1. As well as / instead of the rules in _htslib.mk_, have _htslib.mk_ provide

```make
htslib--%:
        +cd $(HTSDIR) && $(MAKE) $(subst :, ,$*)
```

2. In the bcftools Makefile, have the rules that need HTSlib products depend on `htslib-prereqs`, which is in _bcftools/Makefile_ as

```make
htslib-prereqs: htslib--lib-static:lib-shared:bgzip:tabix
```

Or it might be easier just to write the `cd && $(MAKE)` rule out directly in _bcftools/Makefile_ and deprecate _htslib.mk_'s existing convenience rules as alas, not parallelisable.

daviesrob:
@jmarshall Wouldn't `htslib-prereqs` attempt to build the HTSlib targets whether needed or not, which is a regression from what we currently have?  Also, `lib-shared` is only needed on Windows (which is why we've not seen this problem on Mac).

jmarshall:
The submake would immediately do nothing and return, but you're right that this would be a regression from the current and desirable _Nothing to be done for `all'_ behaviour.

So it's important to keep listing the conservative dependencies on `$(HTSLIB_ALL)` etc. The extrapolation of that is something like [this **single-submake** branch](https://github.com/jmarshall/htslib/commit/b9a3bc87c74a8d5bea3a78b31fea162ef0fccc9d), which is fairly similar to your solution but probably more general.

Issue #971, opened 2019-11-14
Issues with long ref and ref-skips


jkbonfield:
TODO: Fix the 28-bit limit on CIGAR length size in the internal memory structures.

While we have no immediate plans to support reads several Gb long, we do support references this long now.

However exome or tRNA style alignments (eg tophat) can produce long reference alignments by the addition of huge N ref-skip ops.   Such alignments would be rare to be over 1Gb in genuine cases, but there is always room for badly mapped data and errors, plus there is potential for this to happen with circular genomes.

An example:

```
@SQ	SN:LONG	LN:8000000000
SRR065390.14978392	16	LONG	2	1	27M7000000000N73M	*	0	0	CCTAGCCCTAACCCTAACCCTAACCCTAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAA	#############################@B?8B?BA@@DDBCDDCBC@CDCDCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC
```

Samtools currently states:

```
$ samtools view ~/tmp/long.sam
[E::sam_parse1] numeric value out of allowed range
[W::sam_read1] Parse error at line 2
[main_samview] truncated file.
```

We can try chopping it up, but even 1000000000N1000000000N1000000000N1000000000N1000000000N1000000000N1000000000N doesn't work due to the 28-bit size limit in BAM's cigar handling.  The internals of htslib shouldn't be so wedded to the BAM on disk format and ought to be using more than 28-bits.  The SAM parser at least can swallow this if we go below 28-bits. eg:

```
@SQ	SN:LONG	LN:8000000000
[...]

PR #972, opened 2019-11-14
Fix possible integer overflows in sam_parse1()

daviesrob:
bam_cigar2rlen() returns hts_pos_t, so store the result in a variable of the same type.

Add a check to ensure c->pos + reflen is within the range we support.

Credit to OSS-Fuzz
Fixes oss-fuzz 18916

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -669,10 +669,19 @@ int bam_write1(BGZF *fp, const bam1_t *b)
     } else { // with long CIGAR, insert a fake CIGAR record and move the real CIGAR to the CG:B,I tag
         uint8_t buf[8];
         uint32_t cigar_st, cigar_en, cigar[2];
+        hts_pos_t cigreflen = bam_cigar2rlen(c->n_cigar, bam_get_cigar(b));
+        if (cigreflen >= (1<<28)) {
+            // Length of reference covered is greater than the biggest
+            // CIGAR operation currently allowed.
+            hts_log_error("Record %s with %d CIGAR ops and ref length %"PRIhts_pos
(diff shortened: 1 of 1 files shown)


daviesrob:
The possible problem in `bam_write1` is an example of #971, caused by the 28 bit limit on the length of a single CIGAR operation in BAM.  You're correct, it could overflow if we have a very long read with a lot of CIGAR operations (which are very likely to go together) then this limit could be hit.  On the other hand (barring ref skips) I don't think anyone has got close to making a read that long yet.

It could be fixed by making more than one CIGAR operation at that point, but we would have to make the rest of the world agree on the correct way to interpret it.  We would also need to make other changes, e.g. in [bam_tag2cigar](https://github.com/samtools/htslib/blob/71d8683/sam.c#L493) so data round trips properly.

Meanwhile, in the hope that it's a little while before anyone makes a 268 Mbase read, possible we should just give up and report an error at this point if we detect an overflow might happen (and suggest writing a SAM or CRAM file, which wouldn't have the limitation).

jkbonfield:
> A similar fix is probably required for [bam_write1](https://github.com/samtools/htslib/blob/71d8683df6084b304a999cca639229e64e499316/sam.c#L675), otherwise we risk generating corrupt BAM files.

This is the same issue I refer to in https://github.com/samtools/htslib/issues/971

Unless we deem we want to change the ABI and actually make the exposed cigar field 64-bit (which I think was decided not to when I raised it before), then this is something we can fix in the next release.  It's possible to represent in BAM as a series of concatenated cigar-ops.  Messy, but we have no choice there anyway.

valeriuo:
Merged into develop

PR #973, opened 2019-11-14
Silence gcc warnings

daviesrob:
For some optimisation levels, gcc would report a format-truncation warning for the line that measured how much memory was needed to store the header string.  Silence it by supplying a reasonably big buffer, which has the added bonus that the malloc and second call to vsnprintf can be avoided completely in most cases.


daviesrob:
@pd3 reported another warning from gcc version 4.  It looks like a false positive, but it's trivial to fix so I've added a commit to make it go away.

PR #976, opened 2019-11-15
Fixes involving CRAM and header API for long references.

jkbonfield:
This is obviously only partial because we need CRAM4 for long reads, but it's now better at reporting errors and also now copes correctly with the cases inbetween 28-bit and 32-bit sizes for long cigar ops.

The header API has also had some fixes, which were discovered in the process of updating CRAM.  The main culprit was sam_hdr_dup which didn't honour h->sdict.


sam.c
```diff
@@ -179,6 +226,14 @@ sam_hdr_t *sam_hdr_dup(const sam_hdr_t *h0)
         h->text[h->l_text] = '\0';
     }
 
+    if (h0->sdict) {
+        // We have a reason to use new API, so build it so we
```
valeriuo: @jkbonfield As discussed, this would be better addressed inside `sam_hdr_update_target_arrays`.

jkbonfield:
Moving the code to `sam_hdr_update_target_arrays` looks to have fixed the problem in a more elegant way.  Thanks.

Issue #816, opened 2019-01-19
hts_tpool_process_flush double prototype in header


jblachly:
First with slightly better function documentation:
https://github.com/samtools/htslib/blob/a80f5fd602eb375dd4a9068ad53601a3fa51a4bc/htslib/thread_pool.h#L134-L145

Second:
https://github.com/samtools/htslib/blob/a80f5fd602eb375dd4a9068ad53601a3fa51a4bc/htslib/thread_pool.h#L231-L239

jkbonfield:
Thanks for the bug report.   I'll remove the second copy.

PR #984, opened 2019-11-20
Update the NEWS file

jkbonfield:
*Phew*!

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,187 +1,44 @@
 Noteworthy changes in release a.b
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* Incompatible changes:  Several functions and data types have been changed
-  in this release, and the shared library soversion has been bumped.
+Brief summary
+-------------
 
-  - On systems that support it, the default symbol visibility has been
-    changed to hidden and the only exported symbols are ones that form part
-    of the officially supported ABI.  This is to make clear exactly which
-    symbols are considered parts of the library interface.  It also
-    helps packagers who want to check compatibility between HTSlib versions
-    (see for example #311, #616, #695).
+There are many changes in this release, so the executive summary is:
 
-  - HTSlib now supports 64 bit reference positions.  This means several
-    structures, function parameters, and return values have been made bigger
-    to allow larger values to be stored.  While most code that uses
-    HTSlib interfaces should still build after this change, some alterations
-    may be needed - notably to printf() formats where the values of structure
-    members are being printed.
+* Addition of support for references longer than 2Gb (NB: SAM and VCF
+  formats only, not their binary counterparts).  This may need changes
+  in code using HTSlib.  See README.large_positions.md for more information.
 
-    Due to file format limitations, large positions are only supported
-    when reading and writing SAM and VCF files.
+* Added a SAM header API.
 
-    See README.large_positions.md for more information.
+* Major speed up to SAM reading and writing.  This also now supports
+  multi-threading.
 
-  - An extra field has been added to the kbitset_t struct so bitsets can
-    be made smaller (and later enlarged) without involving memory allocation.
+* We can now auto-index on-the-fly while writing a file.  This also
+  includes to bgzipped SAM.gz.
 
-  - A new field has been added to the bam_pileup1_t structure to keep track
-    of which CIGAR operator is being processed.  This is used by a new
-    bam_plp_insertion() function which can be used to return the sequence of
-    any inserted bases at a given pileup location.  If the alignment includes
-    CIGAR P operators, the returned sequence will include pads.
+* Overhaul of the S3 interface, which now supports version 4
+  signatures.  This also makes writing to S3 work.
 
-  - The hts_itr_t and hts_itr_multi_t structures have been merged and can be
-    used interchangeably.  Extra fields have been added to hts_itr_t to support
-    this.  hts_itr_multi_t is now a typedef for hts_itr_t; sam_itr_multi_next()
-    is now an alias for sam_itr_next() and hts_itr_multi_destroy() is an alias
-    for hts_itr_destroy().
+These also required some ABI changes.  See below for full details.
 
-  - An improved regidx interface has been added.  To allow this, struct
-    reg_t has been removed, regitr_t has been modified and various new
-    API functions have been added to htslib/regidx.h.  While parts of
-    the old regidx API have been retained for backwards compatibility,
-    it is recommended that all code using regidx should be changed to use
-    the new interface.
 
-  - Elements in the hts_reglist_t structure have been reordered slightly
-    so that they pack together better.
-
-  - bgzf_utell() and bgzf_useek() now use type off_t instead of long for
-    the offset.  This allows them to work correctly on files longer than
-    2G bytes on Windows and 32-bit Linux.
-
-  - A number of functions that used to return void now return int so that
-    they can report problems like memory allocation failures.  Callers
-    should take care to check the return values from these functions.
-    The affected functions are:
-       ksort.h:             ks_introsort(), ks_mergesort()
-       sam.h:               bam_mplp_init_overlaps()
-       synced_bcf_reader.h: bcf_sr_regions_flush()
-       vcf.h:               bcf_format_gt(), bcf_fmt_array(),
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -1,231 +1,444 @@
...
+   works when bgzipped.)
+
+ * HTSlib now supports the @SQ-AN header field, which lists alternative names
+   for reference sequences.  This means given "@SQ SN:1 AN:chr1", tools like
+   samtools can accept requests for "1" or "chr1" equivalently.  (PR #931)
+
+ * Completely blank files are no longer considered to be a SAM file
+   with no header and no alignments.  Thus pipelines such as `somecmd
+   | samtools ...` with `somecmd` aborting before outputting anything
+   will not propagate the error to the second command. (#261)
```
jmarshall: s/not/now/!!

But I don't really see why you've rewritten the paragraph entirely. Previous was

```
* Empty (0-length) files are no longer considered to be valid SAM files.
  Input files without any headers or records at all are usually seen in
  pipelines such as `somecmd | samtools ...` with `somecmd` aborting and
  outputting nothing, and it is beneficial for the second command to
  propagate the error.  (#261)
```

This wording was carefully chosen. “Empty (0-length)” is clearer and more precise than “Completely blank”; “blank files … a SAM file” is something of a number mismatch; the user documentation uses both “records” and “alignments” and in this context I think the former is clearer; the line break around the pipe was avoided; the thing that is doing the propagating is the second command itself not some other magic. OTOH _aborting before outputting anything_ is an improvement and “it is beneficial for” was awkward and would benefit from some rewording.

jmarshall:
Formatting nits:

1. The existing entries in this document use `*` in the first column as the top-level bullet and `-` in the third column as a second-level bullet.

   These additions have a mixture of zero, one, or two spaces indentation before the `*` bullet. And some mixture of indentation styles for the second-level `-` too. It would be good to change this to be consistent with the existing style.

2. The new entries are inconsistent as to whether they end in a fullstop or not. Previous convention has been that they do (but the final `(#816)` references don't).

NEWS
```diff
@@ -1,231 +1,444 @@
...
+ * Fixed bug where some 8 or 16-bit negative integers were stored using values
+   reserved by the BCF specification.  These numbers are now promoted to the
+   next size up, so -121 to -128 are stored using at least 16 bits, and -32761
+   to -32768 are stored using 32 bits.
+
+   Note that while BCF files affected by this bug are technically incorrect,
+   it is still possible to read them.  When converting to VCF format,
+   HTSlib (and therefore bcftools) will interpret the values as intended
+   and write out the correct negative numbers. (#766, thanks to John
+   Marshall, samtools/bcftools#874)
```
jmarshall: Please don't break people's names across lines (cf samtools/samtools#1020).

NEWS
jkbonfield: The reason I rewrote it is mainly because trying to merge changes listed in around 5000 lines of git logs with a couple thousand lines of sporadic NEWs comments, which are all reordered, was just nigh on impossible.

Instead I started again from scratch, and then searched through the old news file to see what was worth lifting over and to check I hadn't missed anything.  I found plenty of cases where I'd rephrased it, but generally prefer brevity over excessive verbiage.  People can look at the git logs if they want the fine detail.  We're not trying to write a specification.  Words like "blank" are fine IMO if it saves extra waffle.

jmarshall: I appreciate that this was a lot of effort; reviewing it is going to be quite an effort too. Should I assume that the indentation nits will be fixed up before I proofread the whole thing again, or proofread it as is?

Brevity is good of course, as is accuracy and clarity. Where effort has been made over the last 16 months to write clear accurate NEWS entries contemporaneously with the associated changes, it seems unfortunate to discard them.

daviesrob: Minor formatting problems will be fixed before this is merged.  Unfortunately it might take a bit longer than expected to get on to this as my equivalent updates to samtools just got blown away by a brief power cut. :disappointed:

NEWS
```diff
@@ -1,231 +1,444 @@
...
+
+ * Sped up decoding of CRAM by around 10% when the MD tag is being
+   generated.
+
+ * CRAM REF_PATH now supports %Ns (where N is a single digit)
+   expansion in http URLs, similar to how it already supported this
+   for directories.
+
+ * BGZF now permits indexing and seeking using virtual offsets in
+   completely uncompressed streams.
```
valeriuo: (#904, thanks to Adam Novak)

daviesrob:
Formatting has been adjusted and links to pull requests added.  Ignore white-space when comparing to make the substantive changes easier to see.

NEWS
```diff
@@ -1,214 +1,392 @@
...
+
+* New API functions bam_set_mempolicy() and bam_get_mempolicy() have
+  been added.  These allow more control over the ownership of bam1_t
+  alignment record data; see documentation in htslib/sam.h for more
+  information. (#922)
+
+* Added more HTS_RESULT_USED checks, this time for VCF I/O. (#805)
+
+* khash can now hash kstrings.  This makes it easier to hash
+  non-nul-terminated strings. (#713)
```
jmarshall:
```suggestion
  non-NUL-terminated strings. (#713)
```

NEWS
```diff
@@ -1,214 +1,392 @@
...
+* htsFile has a new 'state' member to support SAM multi-threading. (#916)
+
+* A new field has been added to the bam1_t structure, and others
+  have been rearranged to remove structure holes. (#709; #922)
+
+
+Bug fixes
+---------
+
+* Several bgzf format fixes:
```
jmarshall:
```suggestion
* Several BGZF format fixes:
```

NEWS
```diff
@@ -1,214 +1,392 @@
...
+  supplied fasta file differs to the order of the @SQ headers. (#935)
+
+* Fixed BAM and CRAM multi-threaded decoding when used in conjunction
+  with the multi-region iterator. (#830; #577, #822, #926 all reported by
+  Brent Pedersen)
+
+* Removed some unaligned memory accesses in CRAM encoder and
+  undefined behaviour in BCF reading (#867, thanks to David Seifert)
+
+* Repeated calling of bcf_empty() no longer crash. (#741)
```
jmarshall:
```suggestion
* Repeated calling of bcf_empty() no longer crashes. (#741)
```

NEWS
```diff
@@ -218,14 +396,86 @@ Noteworthy changes in release a.b
   Note that while BCF files affected by this bug are technically incorrect,
   it is still possible to read them.  When converting to VCF format,
   HTSlib (and therefore bcftools) will interpret the values as intended
-  and write out the correct negative numbers. (#766, thanks to John Marshall)
+  and write out the correct negative numbers. (#766, thanks to John Marshall;
+  samtools/bcftools#874)
+
+* Allow repeated invocations of bcf_update_info and format_*
```
jmarshall:
```suggestion
* Allow repeated invocations of bcf_update_info() and bcf_update_format_*()
```

jmarshall:
Good work all — that's looking in good shape. I didn't spot any other typos, my only other suggestion would be multiple blank lines to set this epic off from the previous _Noteworthy changes in release 1.9_ heading.

daviesrob:
Thanks, suggestions applied.

NEWS
```diff
@@ -218,14 +396,91 @@ Noteworthy changes in release a.b
   Note that while BCF files affected by this bug are technically incorrect,
   it is still possible to read them.  When converting to VCF format,
   HTSlib (and therefore bcftools) will interpret the values as intended
-  and write out the correct negative numbers. (#766, thanks to John Marshall)
+  and write out the correct negative numbers. (#766, thanks to John Marshall;
+  samtools/bcftools#874)
+
+* Allow repeated invocations of bcf_update_info() and format_*()
```
jmarshall: Suggestions almost applied :smile: (Original was 8ab9acf01b5fc95c2a6c18ace1e0a489d0089261 “Allow repeated invocations of bcf_update_info/format_*()” but with `/` replaced by `and` it probably needs the full function name prefix for clarity.

```suggestion
* Allow repeated invocations of bcf_update_info() and bcf_update_format_*()
```

daviesrob: Should be right now.

jmarshall: :+1: thanks

PR #986, opened 2019-11-22
Add missing Makefile dependencies and ignored test executable

jmarshall:
Another round of updating dependencies to correspond to `#include` updates in recent PRs.

It has occurred to me (and it seems obvious in retrospect!) that I could add a version of my dependency checking script to `make maintainer-check` so these would be caught immediately. (You could even use it as a pre-commit hook if you wanted to.) Could do this after the dust has settled on the current release activities.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ shlib-exports-*.txt
 /test/test-parse-reg
 /test/test_realn
 /test/test-regidx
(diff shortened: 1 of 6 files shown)


daviesrob:
Looks good, thanks.  A script to check this would be useful - I started writing one but didn't get very far...

PR #987, opened 2019-11-25
Bump shared library version to 3

daviesrob:
Pull request to mark the milestone.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,11 +104,8 @@ include htslib_vars.mk
 # into here.
 PACKAGE_VERSION := $(shell ./version.sh)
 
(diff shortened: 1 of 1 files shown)


jmarshall:
Previous HTSlib releases just used `LIBHTS_SOVERSION` directly in `-compatibility_version`; `$(MACH_O_COMPATIBILITY_VERSION)` was also introduced in 2a86c68b2555e20bcb99ff9d6811035216f9ecae and could be removed too. (Unless you're intentionally leaving it for next time, but the commit message didn't mention that.)

daviesrob:
It's been left deliberately.  It might be useful at some point in the future, and I won't have to rediscover it.

PR #988, opened 2019-11-25
Delay CRAM crc32 checks to block decompression step.

jkbonfield:
This means we don't validate the CRC until we have a need to use that
data.  Thus we don't pay the price when doing sub-queries such as
samtools index and samtools flagstats.

This brings htslib into line with io_lib's implementation which has
done this for some time.

Some benchmarks below on 1 million PacBio reads, which show this problem up more than most files due to the small nature of flags vs long sequence & quality blocks:

This branch:
```
time MALLOC_TOP_PAD_=1000000000 ./samtools flagstat /tmp/_1m.cram
real 0m0.523s
user 0m0.332s
sys  0m0.184s

time MALLOC_TOP_PAD_=1000000000 ./samtools index /tmp/_1m.cram
real 0m0.221s
user 0m0.024s
sys  0m0.196s

time MALLOC_TOP_PAD_=1000000000 ./samtools view -c /tmp/_1m.cram
real 0m30.824s
user 0m30.426s
sys  0m0.324s
```
Dev branch:
```
time MALLOC_TOP_PAD_=1000000000 samtools flagstat /tmp/_1m.cram
real 0m1.317s
user 0m1.048s
sys  0m0.260s

time MALLOC_TOP_PAD_=1000000000 samtools index /tmp/_1m.cram
real 0m1.029s
user 0m0.816s
sys  0m0.212s

time MALLOC_TOP_PAD_=1000000000 samtools view -c /tmp/_1m.cram
real 0m31.412s
user 0m30.894s
sys  0m0.452s
```
Note this is with libdeflate's fast CRC.  Zlib's shows an even larger
discrepancy.

PS.  For the curious, the MALLOC_TOP_PAD_=1000000000 is to work around
deficiency of GNU's libc malloc implementation.  Without it (but with
this patch) we have excessive system call time due to wasted time
spent zeroing pages of memory that we keep repeatedly giving back:
```
# This branch
time ./samtools flagstat /tmp/_1m.cram
real 0m1.328s
user 0m0.280s
sys  0m1.044s

# Dev branch
real  0m2.110s
user  0m1.104s
sys   0m0.996s
```
Also, as a demonstration of CRAM vs BAM, this is the BAM equivalent:
```
-rw-r--r-- 1 jkb team117 2160547816 Nov 25 15:30 /tmp/_1m.bam
-rw-r--r-- 1 jkb team117 1303241699 Nov 25 15:05 /tmp/_1m.cram

time MALLOC_TOP_PAD_=1000000000 ./samtools index /tmp/_1m.bam
real	0m20.564s
user	0m19.953s
sys	0m0.572s

time MALLOC_TOP_PAD_=1000000000 ./samtools view -c /tmp/_1m.bam
real	0m19.663s
user	0m19.257s
sys	0m0.372s
```

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -885,13 +885,10 @@ cram_block *cram_read_block(cram_fd *fd) {
             return NULL;
         }
 
-        crc = crc32(crc, b->data ? b->data : (uc *)"", b->alloc);
-        if (crc != b->crc32) {
-            hts_log_error("Block CRC32 failure");
-            free(b->data);
-            free(b);
-            return NULL;
-        }
+        b->crc32_checked = fd->ignore_md5;
+        b->crc_part = crc;
+    } else {
+        b->crc32_checked = 1; // CRC not present
     }
 
     b->orig_method = b->method;
@@ -991,6 +988,15 @@ void cram_free_block(cram_block *b) {
 int cram_uncompress_block(cram_block *b) {
     char *uncomp;
     size_t uncomp_size = 0;
+
+    if (b->crc32_checked == 0) {
+        uint32_t crc = crc32(b->crc_part, b->data ? b->data : (uc *)"", b->alloc);
+        b->crc32_checked = 1;
+        if (crc != b->crc32) {
+            hts_log_error("Block CRC32 failure");
+            return -1;
+        }
+    }
 
     if (b->uncomp_size == 0) {
         // blank block
--- a/cram/cram_structs.h
+++ b/cram/cram_structs.h
@@ -275,6 +275,9 @@ struct cram_block {
 
     // To aid compression
     cram_metrics *m; // used to track aux block compression only
+
+    int crc32_checked;
+    uint32_t crc_part;
 };
 
 struct cram_codec; /* defined in cram_codecs.h */


jkbonfield:
Note it's possible to go one step further, but it's a much more invasive change.

If `cram_read_slice` is amended to take a `cram_container *c` option (or we invent a new function) then we can copy the main guts of the `cram_dependent_data_series()` function to determine which blocks will be needed and which will not at the time of reading the slice.

This in turn permits a choice of calling `hread` vs `hseek` to read or skip the data section of each block.  Doing this brings index down by 2.5x and flagstat by 1.8x:

```
time MALLOC_TOP_PAD_=1000000000 samtools index /tmp/_1m.cram
real	0m0.083s
user	0m0.012s
sys	0m0.068s

time MALLOC_TOP_PAD_=1000000000 samtools flagstat /tmp/_1m.cram
real	0m0.297s
user	0m0.276s
sys	0m0.016s
```

Index is now fully 20x faster than develop branch on this pacbio data set.  On an Illumina data set it's only around 5x faster (but still that's very signifiant).

We can accept this as a second commit in the same PR if you wish to wait, or do the PR in two stages as desired.  If the latter it'll take longer to tidy it up.  I've just got a horrid hack at the moment to test it.

PR #990, opened 2019-11-26
Allow reading of .fai and .fqi files line-by-line via hts_open/hts_ge…

pd3:
…tline/hts_close.

This functionality is occasionally used by external programs and was lost in 1a1b5b9ec3.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,9 @@
 Noteworthy changes in release a.b
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Noteworthy changes in release 1.10 (6th December 2019)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 Brief summary
 -------------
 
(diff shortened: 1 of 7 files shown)


jmarshall:
Which external programs?  Why are they using `hts_open` rather than `hopen` or `fopen`?

While these files happen to come under `hts_open()`'s rubric of “possibly-compressed textual line-orientated file”, note that allowing these **index** files to be opened by `hts_open()` reduces the ability for commands to detect being given the wrong kind of file, e.g.,

    samtools view foo.fa.fai

pd3:
For example this one https://github.com/pd3/utils/blob/master/perm-test/perm-test.c#L254

As for why, it is convenient to use `hts_getline()` on arbitrary text files, plain or compressed. In this field there are many ad-hoc formats.

jmarshall:
Yes. That's what `bgzf_open()` / `bgzf_getline()` / `bgzf_close()` are for.

jkbonfield:
> Yes. That's what `bgzf_open()` / `bgzf_getline()` / `bgzf_close()` are for.

That said, is there any harm in permitting hopen to also work?  The change is minimal.

jmarshall:
I explained the harm in https://github.com/samtools/htslib/pull/990#issuecomment-558597248.

jkbonfield:
Then I don't understand what the original change was intended to do.

I thought it was to permit us to detect more types of files, including fai.  However if by doing so it transpires samtools view cannot detect the difference between a .sam and a .fai file then it implies we have other problems at hand, such as the lack of error checking in those tools.

I suspect this boils down to our love of macros and that sam_open is just hts_open in disguise.  It should probably be sam_open as a function that calls hts_open and then checks the file type is valid.

This is something for the next release though.

PR #992, opened 2019-11-26
Mention _destroy function in each init function's documentation

jmarshall:
This is a very tedious PR that just expands the API documentation comments in _htslib/*.h_. No non-comment changes.

Add a note about the corresponding `*_destroy()`/`*_free()` function to the API doxygen for each function that allocates a structure and returns it with the expectation that the caller will eventually free it.


daviesrob:
Excellent, thanks very much.

PR #996, opened 2019-12-02
Mark new internal synced_bcf_reader.c functions as static

jmarshall:
Ensuring properly-prefixed functions (whose names start with `hts_`, `sam_`, or whatever) are `static` where possible is less necessary now that visibility control has been merged. However it is still important (at least for static linking against _libhts.a_) for functions that would otherwise pollute the user's namespace. Also remove the leading underscore from the new identifiers, as that is reserved to the compiler.

(These functions appeared in db90fb58b06d6f3b27b5b873588da7c77844da22 which was not done via a pull request, so was not reviewed.)

(Also a recent addition and a function that could be `static` is `sam_hdr_dup_sdict()`, but as that is properly prefixed there's no need to address it.)

### DIFF
--- a/synced_bcf_reader.c
+++ b/synced_bcf_reader.c
@@ -817,7 +817,7 @@ static int _regions_add(bcf_sr_regions_t *reg, const char *chr, hts_pos_t start,
     return 0; // FIXME: check for errs in this function
 }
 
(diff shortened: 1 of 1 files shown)


daviesrob:
Looks good, thanks.

PR #997, opened 2019-12-03
Fixes to pass cppcheck-1.90 warnings

### LINKED ISSUE #995: Multiple issues found with cppcheck
Hi! Here is an output from cppcheck showing a range of different issues (memory leaks, resource leaks, double free). The command run was `cppcheck  --enable=warning -q -f --language=c ./` on the root directory. Output:
```
(htslib) ( develop ) ><> cppcheck  --enable=warning -q -f --language=c ./                                                                                                          17:40:31
[htslib/kbitset.h:106]: (error) Common realloc mistake: 'bs' nulled but not freed upon failure
[cram/cram_codecs.c:2232]: (error) Memory leak: t
[cram/cram_decode.c:1148]: (error) Common realloc mistake: 'cigar' nulled but not freed upon failure
[cram/cram_encode.c:552]: (error) Memory leak: buf
[cram/cram_io.c:2830]: (error) Invalid calloc() argument nr 1. The value is 0 but the valid values are '1:'.
[cram/cram_io.c:2439]: (error) Memory leak: seq
[cram/cram_io.c:3788]: (error) Memory leak: header
[cram/cram_io.c:3840]: (error) Memory leak: header
[cram/cram_io.c:4079]: (error) Memory leak: pads
[cram/mFILE.c:351]: (error) Resource leak: fp
[cram/pooled_alloc.c:68]: (error) Memory leak: p
[cram/pooled_alloc.c:202]: (error) Memory leak: items
[cram/rANS_static.c:902]: (error) Resource leak: infp
[cram/rANS_static.c:902]: (error) Resource leak: outfp
[cram/string_alloc.c:59]: (error) Memory leak: a_str
[header.c:643]: (warning) When using 'char' variables in bit operations, sign extension can generate unexpected results.
[header.c:2406]: (warning) When using 'char' variables in bit operations, sign extension can generate unexpected results.
[htslib/sam.h:639] -> [header.c:1645]: (warning) Function 'sam_hdr_remove_lines' argument order different: declaration 'h, type, id, rh' definition 'bh, type, id, h'
[header.c:531]: (warning) When using 'char' variables in bit operations, sign extension can generate unexpected results.
[hfile_s3.c:922]: (error) Memory leak: queries
[hts.c:1435]: (error) Common realloc mistake: 's' nulled but not freed upon failure
[hts.c:1452]: (error) Common realloc mistake: 's' nulled but not freed upon failure
[hts.c:1457]: (error) Common realloc mistake: 's' nulled but not freed upon failure
[hts.c:1465]: (error) Common realloc mistake: 's' nulled but not freed upon failure
[hts.c:1473]: (error) Common realloc mistake: 's' nulled but not freed upon failure
[synced_bcf_reader.c:318]: (error) Uninitialized variable: n
[test/hfile.c:183]: (error) Memory leak: original
[test/hfile.c:183]: (error) Memory leak: text
[test/test_bgzf.c:381]: (warning) %d in format string (no. 1) requires 'int' but the argument type is 'unsigned int'.
[test/test_kstring.c:122]: (warning) %u in format string (no. 1) requires 'unsigned int' but the argument type is 'signed int'.
[thread_pool.c:644] -> [thread_pool.c:620]: (warning) Either the condition '!q' is redundant or there is possible null pointer dereference: q.
[thread_pool.c:644] -> [thread_pool.c:626]: (warning) Either the condition '!q' is redundant or there is possible null pointer dereference: q.
[thread_pool.c:644] -> [thread_pool.c:640]: (warning) Either the condition '!q' is redundant or there is possible null pointer dereference: q.
[thread_pool.c:63]: (error) va_list 'args' was opened but not closed by va_end().
[vcf.c:553]: (error) Memory pointed to by 'str' is freed twice.
```

jkbonfield:
Command ran: cppcheck --inline-suppr  --enable=warning -q -f --language=c .

This is a mix of silencing false positives, rewriting code to be less confusing to cppcheck (and sometimes to us), and genuine minor issues around handling running out of memory.

In some cases I've added inline suppressions, eg

    // cppcheck-suppress objectIndex

This isn't ideal, but I personally think it's worth having a little annotation in order to make detection of future buglets easier to find.

Fixes #995

Propose merging immediately *after* 1.10 release as there are no serious bugs fixed.

Note cppcheck 1.72 still finds some false positives, but I don't propose silencing them.  1.90 is the latest version at time of PR.

### DIFF
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -1145,7 +1145,7 @@ static int cram_decode_seq(cram_fd *fd, cram_container *c, cram_slice *s,
 
         if (ncigar+2 >= cigar_alloc) {
             cigar_alloc = cigar_alloc ? cigar_alloc*2 : 1024;
-            if (!(cigar = realloc(cigar, cigar_alloc * sizeof(*cigar))))
+            if (!(cigar = realloc(s->cigar, cigar_alloc * sizeof(*cigar))))
                 return -1;
             s->cigar = cigar;
         }
@@ -1724,7 +1724,7 @@ static int cram_decode_seq(cram_fd *fd, cram_container *c, cram_slice *s,
 
         if (ncigar+1 >= cigar_alloc) {
             cigar_alloc = cigar_alloc ? cigar_alloc*2 : 1024;
-            if (!(cigar = realloc(cigar, cigar_alloc * sizeof(*cigar))))
+            if (!(cigar = realloc(s->cigar, cigar_alloc * sizeof(*cigar))))
                 return -1;
             s->cigar = cigar;
         }
@@ -1754,7 +1754,7 @@ static int cram_decode_seq(cram_fd *fd, cram_container *c, cram_slice *s,
     if (cig_len) {
         if (ncigar >= cigar_alloc) {
             cigar_alloc = cigar_alloc ? cigar_alloc*2 : 1024;
-            if (!(cigar = realloc(cigar, cigar_alloc * sizeof(*cigar))))
+            if (!(cigar = realloc(s->cigar, cigar_alloc * sizeof(*cigar))))
                 return -1;
             s->cigar = cigar;
         }
@@ -1876,9 +1876,8 @@ static int cram_decode_aux_1_0(cram_container *c, cram_slice *s,
         r |= c->comp_hdr->codecs[DS_TN]->decode(s, c->comp_hdr->codecs[DS_TN],
                                                 blk, (char *)&id, &out_sz);
         if (out_sz == 3) {
-            tag_data[0] = ((char *)&id)[0];
-            tag_data[1] = ((char *)&id)[1];
-            tag_data[2] = ((char *)&id)[2];
+            // Tag name stored as 3 chars instead of an int?
+            memcpy(tag_data, &id, 3);
         } else {
             tag_data[0] = (id>>16) & 0xff;
             tag_data[1] = (id>>8)  & 0xff;
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -1779,12 +1779,10 @@ static int cram_decode_seq(cram_fd *fd, cram_container *c, cram_slice *s,
     if ((ds & CRAM_QS) && (cf & CRAM_FLAG_PRESERVE_QUAL_SCORES)) {
         int32_t out_sz2 = cr->len;
 
-        if (ds & CRAM_QS) {
-            if (!c->comp_hdr->codecs[DS_QS]) return -1;
-            r |= c->comp_hdr->codecs[DS_QS]
-                            ->decode(s, c->comp_hdr->codecs[DS_QS], blk,
-                                     qual, &out_sz2);
-        }
+        if (!c->comp_hdr->codecs[DS_QS]) return -1;
+        r |= c->comp_hdr->codecs[DS_QS]
+            ->decode(s, c->comp_hdr->codecs[DS_QS], blk,
+                     qual, &out_sz2);
     }
 
     s->cigar = cigar;
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -547,7 +547,8 @@ cram_block *cram_encode_slice_header(cram_fd *fd, cram_slice *s) {
     if (!b)
         return NULL;
 
-    if (NULL == (cp = buf = malloc(16+5*(8+s->hdr->num_blocks)))) {
+    cp = buf = malloc(16+5*(8+s->hdr->num_blocks));
+    if (NULL == buf) {
         cram_free_block(b);
         return NULL;
     }
--- a/thread_pool.c
+++ b/thread_pool.c
@@ -617,6 +617,7 @@ static void *tpool_worker(void *arg) {
 }
 
 static void wake_next_worker(hts_tpool_process *q, int locked) {
+    if (!q) return;
     hts_tpool *p = q->p;
     if (!locked)
         pthread_mutex_lock(&p->pool_m);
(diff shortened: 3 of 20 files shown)


jmarshall:
There's lots to be quibbled about with some of these fixes — e.g., it would be better to rewrite the reallocs in `hts_readlist()` and `hts_readlines()` via `hts_resize` instead.

So it will be good to come back and discuss this after the upcoming release.

jkbonfield:
I'd forgotten about `hts_resize`.  In my mind it's still the abomination of zero error checking, but I'd conflated it with `hts_expand` that it replaced. :)

Agreed there are better solutions out there for some code and lots of it could be rewritten (especially in vcf.c), but IMO that's a second larger task than the naive quick fixes to get the code validation passing so personally I'd prefer separating the two out.

header.c
```diff
@@ -528,7 +528,7 @@ static int sam_hrecs_vadd(sam_hrecs_t *hrecs, const char *type, va_list ap, ...)
     sam_hrec_type_t *h_type;
     sam_hrec_tag_t *h_tag, *last=NULL;
     int new;
-    khint32_t type_i = (type[0]<<8) | type[1], k;
+    khint32_t type_i = ((unsigned)type[0]<<8) | type[1], k;
```
jmarshall: This code appears eight times in _header.c_. Better to abstract it into `static inline unsigned int hdrtype2int(const char *type)` and fix seven sets of latent sign extension bugs at once!

header.c
```diff
@@ -1642,6 +1642,7 @@ int sam_hdr_remove_except(sam_hdr_t *bh, const char *type, const char *ID_key, c
     return 0;
 }
 
+// cppcheck-suppress funcArgOrderDifferent
 int sam_hdr_remove_lines(sam_hdr_t *bh, const char *type, const char *id, void *h) {
```
jmarshall: Could be fixed instead by renaming the last parameter:

```suggestion
int sam_hdr_remove_lines(sam_hdr_t *bh, const char *type, const char *id, void *vrh) {
```

cram/cram_decode.c
```diff
@@ -1145,9 +1145,10 @@ static int cram_decode_seq(cram_fd *fd, cram_container *c, cram_slice *s,
 
         if (ncigar+2 >= cigar_alloc) {
             cigar_alloc = cigar_alloc ? cigar_alloc*2 : 1024;
```
jmarshall: This false alarm (and the same one in _htslib/kbitset.h_) no longer occurs with current development cppcheck 1.90pre.

cram/cram_encode.c
```diff
@@ -48,6 +48,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "htslib/hts.h"
 #include "htslib/hts_endian.h"
 
+// cppcheck-suppress unknownMacro
```
jmarshall: I think this is an artefact of not using `cppcheck -I.`

thread_pool.c
```diff
@@ -60,7 +60,8 @@ static int worker_id(hts_tpool *p) {
 int DBG_OUT(FILE *fp, char *fmt, ...) {
     va_list args;
     va_start(args, fmt);
-    return vfprintf(fp, fmt, args);
+    int ret = vfprintf(fp, fmt, args);
```
jmarshall: The `#else` dummy `DBG_OUT` is `void`, so really just discard this return value and change the function declaration to `void DBG_OUT(etc)`.

jmarshall:
IMHO it would be good to split this into the parts that fix actual defects (the lack of error handling in _hfile_s3.c_, the miscellaneous test code leaks and %d vs. %u stuff) and get that merged (with commit messages like _Fix order_query_string() error handling_ etc).

Then separately look at the best way to suppress the remaining cppcheck blather.

daviesrob:
Rebased with a few extra fixes.  It's now in three commits:

1. Fixes for cppcheck-1.9 warnings that look like genuine problems in htslib.

2. Fixes for other warnings from cppcheck 1.9 that are either false positives; real problems in test harness code; or problems in code that is not usually built (e.g. code to print out debugging messages).  These two commits should result in a clean output from cppcheck 1.9.

3. Fixes for some of the remaining warnings printed by cppcheck 1.8, which is still the default for some current Linux distributions.  These are mostly `if` statements with the side-effect of allocating something, which can be easily fixed by doing the allocation then the `if`.  A couple of more complicated cases of this haven't been fixed as they would be rather more painful to do, and there probably isn't much point as cppcheck 1.9 gets them right anyway.

jmarshall:
The recent cppcheck release is 1.90; the release with more false positives that's still in Debian & Ubuntu is 1.86. (1.08 and 1.09 were well over a decade ago!)

Your ‘genuine problems’ commit still contains a number of `realloc`-related changes that were previously cppcheck false positives and that are no longer produced by cppcheck 1.90. (Namely 3 x `cig_tmp` changes in _cram_decode.c_ and the _kbitset.h_ change.)

IMHO the original code is fine and also clearer. If you really want to suppress these false positives from an outdated version of an ad hoc linter, you could pessimise the code less by just changing the `realloc` argument instead: `if (cigar = realloc(s->cigar, …)` and `bs = (kbitset_t *) realloc(*bsp, …)`.

daviesrob:
The `realloc` related stuff has been simplified and moved into the second commit.  I've also fixed the incorrect cppcheck version number in the commit messages.

The fixes for the old version have been removed.  It started complaining about pointers allocated by `hts_resize()`, so I've given up on it.

Issue #999, opened 2019-12-08
64bit integers break VCF/BCF


pd3:
The `BCF_BT_INT64` type breaks VCF and BCF. It might be better to replace out-of-range values with missing values instead of producing files that cannot be processed.

Test 1:
````
$ cat test1.vcf
##fileformat=VCFv4.2
##INFO=<ID=MPOS,Number=A,Type=Integer,Description="dummy">
##INFO=<ID=NALOD,Number=A,Type=Float,Description="dummy">
##INFO=<ID=NLOD,Number=A,Type=Float,Description="dummy">
##INFO=<ID=POPAF,Number=A,Type=Float,Description="dummy">
##contig=<ID=chr1,length=248956422>
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
chr1	1	.	G	C	.	.	MPOS=-2147483641;NALOD=-8.279e-01;NLOD=15.45;POPAF=6.00

$ bcftools view test1.vcf 
[E::bcf_fmt_array] Unexpected type 0
````

Test 2:
````
$ cat test2.vcf
##fileformat=VCFv4.2
##INFO=<ID=MPOS,Number=A,Type=Integer,Description="dummy">
##contig=<ID=chr1,length=248956422>
#CHROM  POS ID  REF ALT QUAL    FILTER  INFO
chr1    1   .   G   C   .   .   MPOS=42949672950

$ bcftools view test2.vcf -Ou | bcftools view -o /dev/null 
[W::bcf_record_check] Bad BCF record: Invalid INFO type 4 (unknown)
Error: BCF read error
````

This was originally reported in https://github.com/samtools/bcftools/issues/1123

jkbonfield:
test2.vcf is simply impossible in BCF.  Like BAM, it simply doesn't support 64bit coordinates.  Hence you must keep all data in VCF (or likewise in SAM).  However it probably ought to complain when encoding to BCF rather than only decoding.

The first case I don't understand why it's failing.  Maybe it doesn't like negatives.  What is the permitted range of values for MPOS?  I assume it's meant to be unsigned.

pd3:
The test2 fails on the reserved integer values, 64bits are used for these.

[...]

PR #1000, opened 2019-12-09
Fixes BCF in-memory decoding with 64-bit integers.

jkbonfield:
Any 64-bit INFO field that wasn't the last in the list would cause subsequent fields to be decoded incorrectly. This commit fixes that, plus updates the tests accordingly so the bug could be triggered.

Fixes the first part of #999 (test1.vcf), but doesn't fix the second part (BCF output silently being broken).
Fixes samtools/bcftools#1123

Note attempting to fix the second part of #999 is tricky as writing BCF just dumps out the internal memory representation and this has been modified to cope with 64-bit integers so VCF round-tripping of large numbers works.

We'd need to add a flag to the bcf record that indicates out of bound data and get the bcf writing function to them reject it or to replace with legal representatives (eg "missing") according to the users wishes.  This is left for a different PR though, or at least as a subsequent commit, as it's more complex to do and I don't wish to delay a fix for the most immediate problem.


jkbonfield:
Note there are other changes we need to do if we want full 64-bit support for VCF too.

- The current implementation cannot cope with arrays.  It was only ever designed for 64-bit references and not arbitrary 64-bit INFO fields, hence it dealt with the one exception of INFO "END" and nothing more.

- Bcftools 1.9 would silently overflow integers and just deal with the bottom 32-bit.  The new one still causes broken bcf, but we cannot just revert back to our old also-broken behaviour.  We need a decision here from the spec maintainers, but I like @pd3's suggestion of filling them out with the *missing* value (plus appropriate warning).

pd3:
I propose for now we revert to the old behavior but done properly: replace with missing values for now with a warning and later enable a mechanism to control the behavior. In the current form we are unable to use the BCF writing part. I will add to this pull request and more tests

daviesrob:
Looks good to me.  Obviously I should have put END as the first item on the original test instead of the last.

Unfortunately due to time pressure I didn't get in all the checks to catch 64-bit values being written on BCF.  We should add these in a future update.  My preferred solution would be to stop if a too-large value is encountered and suggest using compressed VCF instead.  It might also be possible to scan the chromosome list and automatically switch to VCF if a long one is encountered, in the hope that whatever is reading the data will manage to handle the situation.  While this wouldn't solve the case of random large INFO fields, it would make working with large reference positions a bit easier.

PR #1001, opened 2019-12-09
Make build system more flexible

mr-c:
So that Debian can pass hardening flags

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,13 +29,13 @@ RANLIB = ranlib
 # Default libraries to link if configure is not used
 htslib_default_libs = -lz -lm -lbz2 -llzma -lcurl
 
-CPPFLAGS =
+CPPFLAGS +=
 # TODO: probably update cram code to make it compile cleanly with -Wc++-compat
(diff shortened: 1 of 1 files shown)


mp15:
Pardon me for my ignorance but aren't those supposed to be set when you run ./configure? (I recall a somewhat similar pull request before #748) Can Debian not harden that way? (This seems to be another reason we should drop non-configure builds too)

jmarshall:
Debian runs _configure_, so they don't need this change. (Nor does anyone else.)

You will find that if you export your hardening flags or otherwise inform configure of them, this will already work.

mr-c:
Ah, this was in the context of a packaged version of `vg` which includes [a fork](https://github.com/vgteam/htslib/tree/90826f8a4d8f36ecb5371a750612a3e6e09ebe89) of `htslib`; I will adjust our patch of their Makefile to ensure that all variable are passed to htslib's configure.

https://github.com/vgteam/vg/blob/master/Makefile#L479

Hopefully they will adopt the new release of htslib and we can remove their code copy.

Thanks for the quick reply!

PR #1002, opened 2019-12-10
Use srand48_deterministic() on platforms that have it

jmarshall:
This is a prerequisite for an upcoming samtools PR to make the test suite work on OpenBSD, where there are test suite failures testing `samtools merge -c -p` because OpenBSD produces a different random number sequence and so the renamed reference sequences are different from the tests' expected output.

This could be fixed in samtools, but as _hts_os.h_ collects together platform-specific `srand48()/drand48()/etc` foibles, it seems better to fix it here. Typically for HTSlib clients, cross-platform repeatability seems more important than unneeded  cryptographic secureness on one platform.

It would be possible to `#define srand48(S) hts_srand48((S))` on OpenBSD to get the standard sequence whichever of the two functions were used, but (1) this is very fragile as it is impossible to ensure that no system headers are ever #included **after** _htslib/hts_os.h_ (and such a `#define` could wreak havoc on a system header) and (2) it seems better to give client programs the choice.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,11 @@
 Noteworthy changes in release a.b
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* hts_srand48() now seeds the same POSIX-standard sequences of pseudo-random
+  numbers regardless of platform, including on OpenBSD where plain srand48()
+  produces a different cryptographically-strong non-deterministic sequence.
+
+
 Noteworthy changes in release 1.10.2 (19th December 2019)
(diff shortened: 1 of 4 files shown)


jmarshall:
The associated samtools PR is samtools/samtools#1151, and the background on `srand48_deterministic()` is <https://www.mail-archive.com/tech@openbsd.org/msg21465.html>.

jkbonfield:
Thanks.  Looks good, but will probably be back up to merging speed in the new year.

daviesrob:
This looks fine.  Could you rebase and fix the conflict that now exists on NEWS, please, or would you prefer me to do it?

jmarshall:
Rebased and deconflicted.

daviesrob:
Cheers, now merged.

PR #1011, opened 2020-01-08
Add items to `make all`; fix some warnings in header.c [minor]

jmarshall:
Another collection of minor miscellanea:

* Add a couple of minor makefile targets to `make all`. While converting lofreq to use HTSlib directly, I wanted to use _htslib_static.mk_ to get `$(HTSLIB_static_LIBS)` but didn't want to use _htslib.mk_'s dependency infrastructure. This meant that e.g. https://github.com/CSB5/lofreq/pull/86/files#diff-354f30a63fb0907d4ad57269548329e3 couldn't just use `make all` but had to list `htslib_static.mk` as an explicit target. Adding it to HTSlib's `all` would remove this need.

* Fix the semi-theoretical sign-extension bugs in `TYPEKEY()` and use it throughout _header.c_. This replaces a bunch of bit-shifting on plain maybe-signed `chars` and makes for (to my eye) clearer code. Also prevents some `-Wextra` and cppcheck warnings.

* In the same source file, trivially rename a `sam_hdr_remove_lines()` parameter to prevent another cppcheck warning.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,8 @@ BUILT_THRASH_PROGRAMS = \
 	test/thrash_threads6 \
 	test/thrash_threads7
 
-all: lib-static lib-shared $(BUILT_PROGRAMS) plugins $(BUILT_TEST_PROGRAMS)
+all: lib-static lib-shared $(BUILT_PROGRAMS) plugins $(BUILT_TEST_PROGRAMS) \
+     htslib_static.mk htslib-uninstalled.pc
 
 HTSPREFIX =
 include htslib_vars.mk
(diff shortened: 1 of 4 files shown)


jmarshall:
* I don't know whether @jblachly or others are still using htslib et al on Illumos, but this small fix tidies up one issue that has reappeared.

With this, #487 is probably fixed. I have a followup for samtools which brings samtools/samtools#663 up to date, though much of that PR has since been superseded.

daviesrob:
Thanks.  I can confirm that this does make builds on Illumos work better.

jmarshall:
Thanks for merging and confirming.

jblachly:
Thanks @jmarshall -- we do still have some machines in service

PR #1014, opened 2020-01-20
Fix missing return parameter value

janinl:
Bug visible when trying to run `bcftools view "https://url1.gvcf.gz##idx##https://url2.gvcf.gz.tbi" 1:10000-10010`
The calling function `idx_filename` was receiving ret=0, but `local_len` wasn't set. As it was never initialized it randomly failed at the `memmove` operation.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -3416,7 +3416,7 @@ static int idx_test_and_fetch(const char *fn, const char **local_fn, int *local_
         }
 
         if (download) {
-            if ((local_fp = fopen(s.s, "w")) == 0) {
+            if ((local_fp = fopen(s.s, "wb")) == 0) {
                 hts_log_error("Failed to create file %s in the working directory", p);
                 goto fail;
             }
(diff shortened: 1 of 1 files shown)


daviesrob:
Thanks for the bug fix.  Out of interest, how exactly did you get this to go wrong?  When I tried to reproduce this, htslib either found the local file and read it before reaching the buggy code or it downloaded the remote file again.  The only way I managed to actually hit the bug was by creating a race condition where the local file didn't exist before `idx_test_and_fetch()` was entered, and then putting it in place in time for the `fopen()` in that function.

janinl:
Hi @daviesrob 

Thanks for accepting this PR!

> Out of interest, how exactly did you get this to go wrong? When I tried to reproduce this, htslib either found the local file and read it before reaching the buggy code or it downloaded the remote file again.

In my case it was also finding the remote index file and reading it first, but then it was trying to locate it again when reading the vcf header, and that's where it was failing. In this 2nd attempt,  the "##idx##..." part of filename wasn't available anymore.

PR #1017, opened 2020-01-24
Add support for remote reference files

valeriuo:
1. The new method `fai_path` takes the full path to a reference file (e.g. _prot://path/to/ref.fa_) and returns the path to a corresponding index file (e.g. _prot://path/to/ref.fa.fai_). 

    - If the index file is missing and the reference file is local, the index file is generated locally, in the same location as the reference.
    - If the reference file or index file is remote, neither will be downloaded locally.
    - If the reference file is remote and an index file cannot be located, **NULL** is returned and the user will see an error message.

2. The method `refs_load_fai` can now handle remote files, via `hFILE` pointers, and only tries to infer the name of the reference file starting from the index file, if a previous reference has not been provided (e.g. via `sam_open_format`).

3. **UR** tags in `@SQ` lines can now be set to remote URIs.

4. The reference file and the index file can now be provided as a single string argument, using the format **reference.fa##idx##index.fa.fai**.

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -1764,7 +1764,7 @@ static refs_t *refs_create(void) {
 static BGZF *bgzf_open_ref(char *fn, char *mode, int is_md5) {
     BGZF *fp;
 
-    if (!is_md5) {
+    if (!is_md5 && !hisremote(fn)) {
         char fai_file[PATH_MAX];
 
         snprintf(fai_file, PATH_MAX, "%s.fai", fn);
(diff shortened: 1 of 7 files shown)


cram/cram_io.c
```diff
@@ -1811,41 +1810,58 @@ static refs_t *refs_load_fai(refs_t *r_orig, const char *fn, int is_err) {
...
 
-    if (!(r->fn = string_dup(r->pool, fn)))
-        goto err;
-
-    if (fn_l > 4 && strcmp(&fn[fn_l-4], ".fai") == 0)
-        r->fn[fn_l-4] = 0;
+    /* Look for a FASTA##idx##FAI format */
+    char *fn_delim = strstr(fn, HTS_IDX_DELIM);
+    if (fn_delim) {
+        char *fn_tmp = strdup(fn);
```
daviesrob: Instead of making a copy here, I think `string_ndup(r->pool, fn, fn_delim - fn)` should work.

cram/cram_io.c
```diff
@@ -1811,41 +1810,58 @@ static refs_t *refs_load_fai(refs_t *r_orig, const char *fn, int is_err) {
...
 
-    /* Parse .fai file and load meta-data */
-    sprintf(fai_fn, "%.*s.fai", PATH_MAX-5, r->fn);
+        fn_delim += strlen(HTS_IDX_DELIM);
+        snprintf(fai_fn, PATH_MAX, "%s", fn_delim);
+    } else {
+        /* An index file was provided, instead of the actual reference file */
+        if (fn_l > 4 && strcmp(&fn[fn_l-4], ".fai") == 0) {
+            if (!r->fn) {
+                if (!(r->fn = string_dup(r->pool, fn)))
```
daviesrob: Use `string_ndup()` here too?

htslib/faidx.h
```diff
@@ -312,6 +312,8 @@ const char *fai_parse_region(const faidx_t *fai, const char *s,
 HTSLIB_EXPORT
 void fai_set_cache_size(faidx_t *fai, int cache_size);
 
+HTSLIB_EXPORT
+char *fai_path(const char *fa);
```
daviesrob: Could you add some doxygen to say what this does, please?

faidx.c
```diff
@@ -922,3 +922,33 @@ const char *fai_parse_region(const faidx_t *fai, const char *s,
...
+char *fai_path(const char *fa) {
+    char *fai = NULL;
+    if (!fa) {
+        hts_log_error("No reference file specified");
+    } else {
+
+        char *fai_tmp = strstr(fa, HTS_IDX_DELIM);
+        if (fai_tmp) {
+            fai_tmp += strlen(HTS_IDX_DELIM);
+            fai = strdup(fai_tmp);
```
daviesrob: Maybe `hts_log_error()` if this fails (all other failures get an error message of some sort),

cram/cram_io.c
```diff
@@ -1818,28 +1818,21 @@ static refs_t *refs_load_fai(refs_t *r_orig, const char *fn, int is_err) {
...
-            hts_log_error("%s", strerror(errno));
+        char *fn_tmp = string_ndup(r->pool, fn, fn_delim - fn);
+        if (!fn_tmp)
             goto err;
-        }
-        fn_tmp[fn_delim - fn] = '\0';
 
-        if (!(r->fn = string_dup(r->pool, fn_tmp))) {
-            free(fn_tmp);
+        if (!(r->fn = string_dup(r->pool, fn_tmp)))
```
valeriuo: I just saw that I could get rid of this, as well. I will amend the commit.

htslib/faidx.h
jmarshall: Does the caller have to `free()` the string returned?

Does the answer to that depend on whether the `fa` given had `##idx##` in it? (Hope not — that'd be tedious to use!)

valeriuo: > Does the caller have to free() the string returned?

Yes, but in this respect, the method is not different in usage from the previous `samfaipath`. I'll add a note to the documentation to make it clearer.

> Does the answer to that depend on whether the fa given had ##idx## in it? (Hope not — that'd be tedious to use!)

No. In any scenario, the returned string is a dynamically allocated char array (or NULL).

jmarshall: :+1: Sounds good, just needs mentioning in the doxygen doco.

PR #1023, opened 2020-02-13
Add / fix byte swapping code for big-endian platforms

### LINKED ISSUE #119: System endiannes and unaligned memory access issues on mips

### LINKED ISSUE #355: BCF code makes no attempt to account for endianness

daviesrob:
Properly swap expanded l->n in idx_read_core()

Fix various places where bcf INFO and FORMAT values were not being accessed correctly.

There may still be issues on some platforms with bcf's use if signalling NANs, but this sorts out the endianness problems.  Samtools works out-of-the box with this; a separate PR to bcftools will be needed to get it to work on big-endian.

I haven't added the Travis s390x test here, but I may include it in a future PR as it actually worked quite well when I tried it.  This has mainly been tested on a qemu emulated sparc32 running netbsd.

Fixes #119
Fixes #355


mr-c:
I can confirm this cross-compiling for s390x and running the tests using qemu-s390x-static. However, the symbol `mfmmap@Base` seems to be missing from the resulting shared library..

LocutusOfBorg:
I confirm testsuite is all green, and the package is back on Ubuntu s390x!

hts.c
```diff
@@ -2153,7 +2153,7 @@ static int idx_read_core(hts_idx_t *idx, BGZF *fp, int fmt)
         if (fmt != HTS_FMT_CSI) { // load linear index
             int j;
             if (bgzf_read(fp, &l->n, 4) != 4) return -1;
-            if (is_be) ed_swap_4p(&l->n);
+            if (is_be) ed_swap_8p(&l->n);
```
jmarshall: I'm having a hard time convincing myself that this works — and at best since `l->n` went to 64 bits even on little endian it's subtly contingent on `l` just having been calloced.

Probably better to do the swapping on the data size that's in the file:
```
uint32_t x;
bgzf_read(fp, &x, 4);
if (is_be) ed_swap_4p(&x);
l->n = x;
```

jmarshall:
This PR doesn't touch `bcf_read1_core()`, which as noted in https://github.com/samtools/htslib/issues/355#issuecomment-579897032 reads a bunch of 32-bit ints and floats into the core fields of a `bcf1_t` but makes no attempt to byte-swap them. Similarly `bcf_write()` when writing to a BCF file just writes the in-memory values of these fields out without any byte-swapping.

So I suspect on big-endian platforms htslib is still writing out partially byte-reversed BCF files that would not be readable on a little-endian host. Because both reading and writing make the same mistake, the correct values are read back in again on a write BCF & read BCF test, so the test suite would not detect this.

We need to add a BCF file generated on a little-endian host as an input data file checked in to the test suite, or we need an extra test program that reads a BCF file in a very simpleminded way using its own code and checks that the 32-bit int values therein have been written correctly.

I suspect either such test case (which would demonstrate interoperability between endiannesses) would currently still fail, rather instructively :smile:

hts.c
daviesrob: If it doesn't work on big-endian, then it doesn't on little-endian either for the same reason that not all the bytes get touched.  I will add a fix.

daviesrob:
Oops, yes I forgot to check for a test that started with a bcf file.  I'll add one.

hts.c
jmarshall: I think actually it does work, it's just that it's super not obvious…

jmarshall: …in any case, I guess it's this single line that caused the tabix test regression (as it became broken when `l->n` changed size in 1.10) that brought all this to a head in the distros' s390x testing. Which is pretty amusing :smile:

daviesrob:
Now updated with extra swapping in `bcf_read1_core()` and `bcf_write()`.  I added a test to read a pre-built bcf file, based on `test/tabix/vcf_file.vcf` as it's got a bit more content than the one @jmarshall chose.  I also resurrected #970 to add an s390x test for Travis (slightly modified from the PR as we expect it to work now).

jmarshall:
I just grabbed the first file I saw for a quick demo :smile:

Looks good to me. This test case exercises the fixes to `bcf_read1_core()` but I guess `bcf_write()` is not directly individually exercised. However I just did a trial of breaking the `bcf_read1_core`/`bcf_write()` symmetry and there are four other tests that essentially indirectly require a working VCF→BCF→VCF roundtrip. So the combination of tests means that both are covered.

This doesn't prove that there are no endianness bugs lurking elsewhere, but it does mean that all the ones known about in #355 have been addressed. 🎉

jkbonfield:
I picked a much richer BCF file with many INFO fields including both floats and ints, made on x86_64 along side the VCF equivalent and validated VCF->BCF and BCF->VCF on s390x worked as expected.

jkbonfield:
Note to any users looking at this:  it doesn't mean bcftools is fine on big endian platforms.  I haven't checked, but it's an entirely different can of fish ;-).   It's a different repo so would be tracked in a different issue if and when we find bugs there.

jmarshall:
e.g. `bcftools concat --naive` is known to not be fine on big-endian platforms (see https://github.com/samtools/htslib/issues/355#issuecomment-553366446).

PR #1024, opened 2020-02-14
Protect against CRAM compression headers having negative ref_seq_start.

jkbonfield:
This could trigger buffer read overruns in the MD5 calculation code
(and perhaps more).

Credit to OSS-Fuzz
Fixes oss-fuzz 20694


valeriuo:
Merged as d6f15361ee63ff13dfb57f3eac119e3d50e65c56

PR #1027, opened 2020-02-19
Write index fix

jkbonfield:
Fixes samtools/samtools#1197

Fixes the bug which caused the auto-indexing (--write-index) code to produce different output when multi-threaded.  The difference being whether a BGZF virtual offset is the first byte of the next block or points to the end of the previous block.  Hence we'll no longer produce indices like this which gives data compatibility with older samtools releases.

Also fixed the bug which caused usual offsets pointing to the end of a block to be misinterpreted as hitting EOF in the bgzf file.  This gives compatibility with reading indices produced using multi-threaded auto-indexing using samtools 1.10.


jkbonfield:
Not sure what broke with my git pull of develop in order to rebase this.  I'll fix and redo it.

Issue #1013, opened 2020-01-16
HTSLIB_EXPORT pessimises clang error messages


jmarshall:
Consider the following program, which misuses `hts_open()`:

```c
#include <htslib/hts.h>

htsFile *do_open(const char *fname) {
    return hts_open(fname);
}
```

Compiling this with GCC produces an accurate error message:
```
$ gcc-9 -I. bork.c 
bork.c: In function 'do_open':
bork.c:4:12: error: too few arguments to function 'hts_open'
    4 |     return hts_open(fname);
      |            ^~~~~~~~
In file included from bork.c:1:
./htslib/hts.h:474:10: note: declared here
  474 | htsFile *hts_open(const char *fn, const char *mode);
      |          ^~~~~~~~
```

However compiling it with Clang produces the following diagnostics:
```
$ clang -I. bork.c 
bork.c:4:26: error: too few arguments to function call, expected 2, have 1
    return hts_open(fname);
           ~~~~~~~~      ^
./htslib/hts.h:473:1: note: 'hts_open' declared here
HTSLIB_EXPORT
^
./htslib/hts_defs.h:109:23: note: expanded from macro 'HTSLIB_EXPORT'
#define HTSLIB_EXPORT HTS_DLL_EXPORT
                      ^
./htslib/hts_defs.h:101:24: note: expanded from macro 'HTS_DLL_EXPORT'
#define HTS_DLL_EXPORT __attribute__((__visibility__("default")))
                       ^
1 error generated.
```
Here the error message at bork.c:4 is accurate, but the notes showing you the location of the `hts_open` declaration are useless and voluminous.

This inaccurate location for the declaration is surely a Clang bug; e.g., it was reported as [bug 23564](https://bugs.llvm.org/show_bug.cgi?id=23564) in 2015, but sadly the lack of response on that bug report also gives you a clue as to how likely this is to be fixed in Clang any time soon.

To be precise, Clang is reporting the location of the declaration as the first character (post macro expansion) that is in any way a part of the declaration (and in our headers that is now on the line before the interesting one — the point of showing the declaration is to show you the parameters it was expecting), and more egregiously if that character is part of a macro expanding to something non-empty, it shows a trace of the macro expansions (which are irrelevant in this case).

jmarshall:
This can be worked around in several ways:

```c
extern HTSLIB_EXPORT htsFile *hts_open(const char *fn, const char *mode);

[...]

PR #1029, opened 2020-02-20
Define HTSLIB_EXPORT without using a helper macro

jmarshall:
Assuming that it's not the intention for `HTS_DLL_EXPORT` to be independently useful, this PR removes it and defines `HTSLIB_EXPORT` directly instead. This improves #1013 a bit, as there's only one macro to print out spuriously rather than two — see the discussion on that issue.

You may wish to squash this when merging, but it's presented as two commits because the individual diffs are understandable but the combined diff less so.


daviesrob:
Thanks.  I think the `HTS_DLL_EXPORT` stuff was for the intricacies of exports on Windows, but it seems we can get on fine without it.

PR #1031, opened 2020-02-21
Adds linear index support

jkbonfield:
- Adds BAI linear index support for standard iterators.  This uses a bit more memory as it doesn't discard the BAI linear index, but permits it to be used for optimising range queries.  The effect is minimal on deep data, but profound on shallow.

- Also adds BAI linear index support for multi-region iterators.  This is harder and requires some rather ghastly hacks to avoid changing the ABI.  It's possible there are better methods using the off[] array, but this is rather complex and therefore liable to be error prone given the substantial disconnect between intervals and offsets, which can be many to many and also include partial overlaps - not to mention also being in potentially different orders.

- Finally fixes a pre-existing bug on multi-region iterators for CRAM where it could terminate early and omit records.

Benchmarks, from the first commit message for normal iterator:

    Indexer     Branch   Time(s) BAM I/O(Mb)  Idx I/O(Mb)
    -----------------------------------------------------
    htslib-BAI  develop  1.488   164.1        2.43
    htsjdk-BAI  develop  0.502    53.6        8.27
    htslib-BAI  this     0.498    53.5        2.43
    htsjdk-BAI  this     0.530    53.3        8.27
    htslib-CSI  this     1.532   163.3        0.49

For multi-region iterator (NB many more regions included in this test):

    Indexer     Branch   Time(s) BAM I/O(Mb)  Idx I/O(Mb)
    -----------------------------------------------------
    htslib-BAI  develop  15.184  1397.0       2.43
    htslib-BAI  this      2.991   305.2       2.43

It looks stark, but remember this is due to shallow data - around 2x coverage.  For deep data the difference is minimal.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -2746,7 +2746,7 @@ int hts_itr_multi_cram(const hts_idx_t *idx, hts_itr_t *iter)
                     if (end >= HTS_POS_MAX) {
                        e = cram_index_last(cidx->cram, tid, NULL);
                     } else {
-                       e = cram_index_query(cidx->cram, tid, end+1, NULL);
+                       e = cram_index_query_last(cidx->cram, tid, end+1);
                     }
 
                     if (e) {
--- a/hts.c
+++ b/hts.c
@@ -2165,7 +2165,7 @@ static int idx_read_core(hts_idx_t *idx, BGZF *fp, int fmt)
             if (is_be) for (j = 0; j < l->n; ++j) ed_swap_8p(&l->offset[j]);
             for (j = 1; j < l->n; ++j) // fill missing values; may happen given older samtools and tabix
                 if (l->offset[j] == 0) l->offset[j] = l->offset[j-1];
-            update_loff(idx, i, 1);
+            update_loff(idx, i, 0);
         }
     }
     if (bgzf_read(fp, &idx->n_no_coor, 8) != 8) idx->n_no_coor = 0;
@@ -2508,6 +2508,10 @@ hts_itr_t *hts_itr_query(const hts_idx_t *idx, int tid, hts_pos_t beg, hts_pos_t
             } while (bin);
             if (bin == 0) k = kh_get(bin, bidx, bin);
             min_off = k != kh_end(bidx)? kh_val(bidx, k).loff : 0;
+            if (idx->lidx[tid].offset
+                && beg>>idx->min_shift < idx->lidx[tid].n
+                && min_off < idx->lidx[tid].offset[beg>>idx->min_shift])
+                min_off = idx->lidx[tid].offset[beg>>idx->min_shift];
 
             // compute max_off: a virtual offset from a bin to the right of end
             bin = hts_bin_first(idx->n_lvls) + ((end-1) >> idx->min_shift) + 1;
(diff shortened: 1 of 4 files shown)


daviesrob:
Rebased with modified multi-region iterator code.  The new version passes the information needed to skip file regions in the `off[]` array, so it keeps within the existing ABI.  It also skips a few more locations as it's better at eliminating regions from longer indexing bins that are not really needed.

hts.c
```diff
@@ -3334,11 +3369,15 @@ int hts_itr_multi_next(htsFile *fd, hts_itr_t *iter, void *r)
                 } else {
                     ret = -1; break;
                 }
-            } else if (iter->i < iter->n_off - 1) {
-                iter->curr_off = iter->off[++iter->i].u;
-                if (iter->seek(fp, iter->curr_off, SEEK_SET) < 0) {
-                    hts_log_error("Seek at offset %" PRIu64 " failed.", iter->curr_off);
-                    return -1;
+            } else if (iter->i < iter->n_off) {
```
valeriuo: This check is now redundant, but the compiler takes care of it, when using `-O2`.

valeriuo:
Merged into develop as 059abb967d4616507917c994234231eb684e3ad0 to 6149ea6f779939451b56e6f63b28409baead2990

jmarshall:
Note that this changes behaviour for some data files. Specifically: unmapped reads are not added to the linear index, so e.g. if the first read in a query region is a placed unmapped read (i.e., it's filled in with its mate's coordinate) and that read is earlier in the file than the first mapped read in the same linear interval — then using the BAI linear index will cause that first read to be omitted from the iteration output as it skips over it to the linear interval's first mapped read.

Unmapped reads stopped being added to the linear index in samtools/samtools@cb5f44947629a68d16a48d32eef6e42a66e482fe for not entirely clear reasons.

This has made seven pysam test cases fail (when pysam is updated to use htslib-1.11) as they no longer match their expected output. The solution for pysam will probably be to adjust the test data file so the first read isn't unmapped (as this foible isn't what the test case is trying to check!), but the difference may be worth considering in general…

jkbonfield:
Eww, so this isn't so much a flaw in the idea of using the linear index for optimisation, but a bug in our handling of unmapped reads.  However that's presumably something which gets baked into the file written to disk, so fixing that bug won't magically fix all those broken BAMs.

I wonder what the reasoning was, or whether it was even intentional.  If I'm reading your comment correctly then A-mapped followed by A-placed is fine, while A-placed followed by A-mapped is not as A-placed *may* be omitted.  I'm not sure what the default behaviour is for aligners, but I'm hoping it's the former and this is just a quirk of test data rather than a real-world scenario.  Needs checking.

Do we know what htsjdk does?  Was this change in behaviour something discussed with all implementors, or just a whim?  Annoyingly there's nothing about the change in the commit message, but 2011 was rather wild-west era.

jmarshall:
Yes, I'm inclined to agree that the problem is in index generation. The read that gets omitted and the first-mapped-read-in-interval don't need to be related to each other in any way other than happening to be in the same 2<sup>14</sup> base linear interval, so it's not down to an unlucky quirk or aligner behaviour.

As you say, you get different results if the reads happen to be in the file in this-then-that order compared to morally the same data in that-then-this order. So I guess I'll raise this as an issue…

Haven't checked yet what HTSJDK does. I would be surprised if back in 2011 this wasn't just a slightly unfortunate bug fix to prevent calling `insert_offset2(…, -1, …)` in the unplaced unmapped case.

jmarshall:
Issue raised, and probably a pretty simple fix (what to do with the other use of `is_mapped` needs thinking about though). I started testing pysam with 1.11-to-be last week; pity I didn't track down the cause of the test failures a bit sooner… :smile:

PR #1040, opened 2020-03-05
Add more error checking to the tabix application

daviesrob:
Check for error returns from bcf_itr_next(), hts_getline() and tbx_itr_next() so failures can be reported properly.

Add checks in other places that could fail, mostly memory allocations and writes to the output file.

Note that bcf_itr_querys() and tbx_itr_querys() return NULL for both general failures and those caused by the requested region not being present.  As tabix silently skips missing regions, it is currently not possible to detect errors in these functions.

### DIFF
--- a/tabix.c
+++ b/tabix.c
@@ -1,7 +1,7 @@
 /*  tabix.c -- Generic indexer for TAB-delimited genome position files.
 
     Copyright (C) 2009-2011 Broad Institute.
-    Copyright (C) 2010-2012, 2014-2019 Genome Research Ltd.
+    Copyright (C) 2010-2012, 2014-2020 Genome Research Ltd.
 
     Author: Heng Li <lh3@sanger.ac.uk>
 
@@ -41,6 +41,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include "htslib/bgzf.h"
 #include "htslib/hts.h"
 #include "htslib/regidx.h"
+#include "htslib/hts_defs.h"
 
 typedef struct
 {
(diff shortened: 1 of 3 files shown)


tabix.c
```diff
@@ -138,89 +167,120 @@ static char **parse_regions(char *regions_fname, char **argv, int argc, int *nre
...
-    if ( !fp ) error("Could not read %s\n", fname);
+    if ( !fp ) error_errno("Could not open \"%s\"", fname);
     enum htsExactFormat format = hts_get_format(fp)->format;
 
     regidx_t *reg_idx = NULL;
     if ( args->targets_fname )
     {
         reg_idx = regidx_init(args->targets_fname, NULL, NULL, 0, NULL);
-        if ( !reg_idx ) error("Could not read %s\n", args->targets_fname);
+        if (!reg_idx) error_errno("Could not read \"%s\"", args->targets_fname);
```
valeriuo: There are a number of reasons for which `regidx_init` can fail. Perhaps the error message could be more indicative (e.g. "Failed to build region information from file: \\"%s\\"") so as to not make the user think there is just a problem with reading the file.

tabix.c
```diff
@@ -49,15 +50,34 @@ typedef struct
...
+{
+    va_list ap;
+    fflush(stdout);
+    if (format) {
+        va_start(ap, format);
+        vfprintf(stderr, format, ap);
+        va_end(ap);
+    }
+    if (errno) {
+        fprintf(stderr, "%s%s\n", format ? ": " : "", strerror(errno));
```
jmarshall: `fflush` and `vfprintf` might clobber `errno`, so best to sample it at the top of the function.

tabix.c
daviesrob: Text has been amended.

tabix.c
daviesrob: Good point, I now take a copy.

PR #1041, opened 2020-03-06
Added $(DEBUG) to CFLAGS and LDFLAGS usage.

jkbonfield:
This has zero effect normally, but permits easier enabling and disabling of certain debugging types, such as address and thread sanitizer.

Consider an existing htslib build where we've configured it with libdeflate, which in turn needs us to set CPPFLAGS and LDFLAGS accordingly, and maybe upping the CFLAGS to include -O3 and -march=native.  Additionally the Makefile is also automatically adding things like -fvisibility=hidden itself which weren't on our configure line.

If at some point I just want to enable the thread sanitizer, it's a bit of a rigmarole to get this done.  I can `head config.log` and cut and paste in the configure line, taking care to edit it and add all the quotes which annoyingly have been removed, etc etc, but it's slow and tedious.  Maybe I also don't want to wait for configure to do its laborious thing again when I just want to add one extra parameter to an otherwise completely identical system.

So I could do it on the command line with make, but that's also laborious and tedious.  Changing CFLAGS on make removes some other things added automatically which had nothing to do with my configure line (eg the visibility hack).  The correct make line on my current build I need to recompile everything as before, but adding address sanitizer, is this:

```
make CFLAGS="-g -O3 -march=native -fvisibility=hidden -fsanitize=address" LDFLAGS="-L/nfs/users/nfs_j/jkb/ftp/compression/libdeflate -Wl,-rpath,/nfs/users/nfs_j/jkb/ftp/compression/libdeflate -fvisibility=hidden -fsanitize=address"
```

If we add $(DEBUG) to the CFLAGS and LDFLAGS lines in the Makefile, which initially does nothing at all, then we gain the ability to chop and change these without having to regurgitate a full build line thusly:

```
make DEBUG=-fsanitize=address
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,10 @@ EXTRA_CFLAGS_PIC = -fpic
 LDFLAGS  = -fvisibility=hidden
 LIBS     = $(htslib_default_libs)
 
+# Added to CFLAGS and LDFLAGS.  Useful for adding e.g. -fsanitize=address without
+# needing to repeat all of CFLAGS and LDFLAGS.
+DEBUG    =
+
 prefix      = /usr/local
 exec_prefix = $(prefix)
 bindir      = $(exec_prefix)/bin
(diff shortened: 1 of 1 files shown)


jmarshall:
I feel your pain…

It would be a bit more general as `USER_CFLAGS`/`USER_LDFLAGS` (maybe also `USER_CPPFLAGS`; or named as `ADHOC_*`) that could be mixed into the rules via the `ALL_CFLAGS`/`ALL_LDFLAGS`-style variables that the samtools Makefile uses. Admittedly then you'd have to write out the `-fsanitize=…` twice.

What I usually do for ad-hoc compiler flags like this is open _config.mk_ in my editor and add them there.

Another trick is that `$(CC)` practically never has anything interesting in it, so is ripe for ad-hoc overriding. Thus I think you could do your correct make line on your current build as

```
make CC="gcc -fsanitize=address"
```

jkbonfield:
Oh I like the idea of changing CC.  Why didn't I think of that earlier?

Yes, it works fine.  I think given that then this PR isn't a major issue, unless we can think of other reasons why users may want to explicitly manipulate CFLAGS and LDFLAGS without munging everything else.

PR #1044, opened 2020-03-10
Fixed a raft of integer overflows in VCF land.

jkbonfield:
- Cast data into size_t before multiplication to avoid wrapping around
  int32.

- Added checks for return values to align_mem and ks_resize

- Simplified the byzantine calculation in align_mem

- Fixed kroundup_size_t and kroundup32 so they cannot wrap around to
  zero and turn the realloc into a free!

- Also added a check for ~2Gb on total length of FORMAT fields, which
  nullifies the need for some of the above.  We may wish to remove
  this at some point if we want to cope with truely mammoth
  multi-sample data, and the above fixes means doing so will not
  expose bugs.

  However for now this check adds protection against malformed data
  creating excessive memory usage and CPU requirements.

Credit to OSS-Fuzz
Fixes oss-fuzz 21139
Fixes oss-fuzz 20881

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -2347,7 +2347,7 @@ static int vcf_parse_format(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v, char *p
         // Limit the total memory to ~2Gb per VCF row.  This should mean
         // malformed VCF data is less likely to take excessive memory and/or
         // time.
-        if (v->n_sample * (size_t)f->size > INT_MAX) {
+        if (v->n_sample * (uint64_t)f->size > INT_MAX) {
             hts_log_error("Excessive memory required by FORMAT fields");
             v->errcode |= BCF_ERR_LIMITS;
             return -1;
--- a/vcf.c
+++ b/vcf.c
@@ -2193,8 +2193,7 @@ static inline int align_mem(kstring_t *s)
     int e = 0;
     if (s->l&7) {
         uint64_t zero = 0;
-        int l = ((s->l + 7)>>3<<3) - s->l;
-        e = kputsn((char*)&zero, l, s) < 0;
+        e = kputsn((char*)&zero, 8 - (s->l&7), s) < 0;
     }
     return e == 0 ? 0 : -1;
 }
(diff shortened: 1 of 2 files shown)


vcf.c
```diff
@@ -2367,9 +2383,15 @@ static int vcf_parse_format(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v, char *p
...
+            if (!z->buf) {
+                hts_log_error("Memory allocation failure for FORMAT field type %d",
+                              z->y>>4&0xf);
+                v->errcode |= BCF_ERR_LIMITS;
+                return -1;
+            }
             if ((z->y>>4&0xf) == BCF_HT_STR) {
                 if (z->is_gt) { // genotypes
-                    int32_t is_phased = 0, *x = (int32_t*)(z->buf + z->size * m);
+                    int32_t is_phased = 0, *x = (int32_t*)(z->buf + z->size * (size_t)m);
```
jmarshall: `m` is a local variable. Rather than all this casting, can it be declared more accurately as `size_t m` instead? This might trickle down to `fmt_aux_t::max_m` and/or `fmt_aux_t::size` and suggest changing them to `size_t` too, but that doesn't seem like a bad thing…

jkbonfield: The function is so vast I lose track of the scope of these variables, so I just went with the minimalist change.

I'll take a look at changing the type for local var m.  I don't wish to start changing data types though as that requires a deeper understanding of vcf which I don't currently have time to, nor the desire due to the cost of -10 sanity points!

jkbonfield: Hmm looking at it `m` is overloaded.   The bit of code you're referring to uses m as sample id.  In other places it's just listed as a vector (unspecified which).  If I change type of m, should I also change type of g? of r? of l? or j? etc.  I could create a new var `sid` say for sample id and use that throughout the specific block of code you refer to, but it's just as much code changing to read.

I think I'll keep this as is and request that @pd3 looks at changing types if he feels it is valid.  I simply don't understand all the knock on implications of changing type here given the overloading of var names.

jmarshall: Aarrgggghhhhh… fair enough

vcf.c
```diff
@@ -2338,9 +2337,26 @@ static int vcf_parse_format(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v, char *p
...
+            v->errcode |= BCF_ERR_LIMITS;
+            return -1;
+        }
         f->offset = mem->l;
-        ks_resize(mem, mem->l + v->n_sample * f->size);
+
+        // Limit the total memory to ~2Gb per VCF row.  This should mean
+        // malformed VCF data is less likely to take excessive memory and/or
+        // time.
+        if (v->n_sample * (size_t)f->size > INT_MAX) {
```
daviesrob: The cast needs to be to `uint64_t`, as with `size_t` the multiplication could still wrap around on 32-bit platforms.  For the same reason, if this limitation is ever relaxed the test would have to be kept but with `INT_MAX` changed to `SIZE_MAX`.

valeriuo:
Merged into develop as 29c294e6842a56ba3b9a24a24a5f6de1575b0961

PR #1045, opened 2020-03-10
Bt.1164

pd3:
A straightforward extension to recognise `<NON_REF>` produced by GATK the same way the symbolic star allele `<*>` is.


daviesrob:
Rebased to get rid of stray commit...

PR #1046, opened 2020-03-10
Add crypt4gh file detection

daviesrob:
Adds format detection for crypt4gh files (see https://samtools.github.io/hts-specs/crypt4gh.pdf).

If hts_hopen() finds a crypt4gh file, it will try to re-open it via the hfile_crypt4gh() plug-in.  Detection occurs at the same point as htsget, so it does both in a loop to allow for crypt4gh served via htsget.

Actual decryption of the file requires an external plug-in, available from [samtools/htslib-crypt4gh](https://github.com/samtools/htslib-crypt4gh). As this may not be present, a dummy handler is included that prints an error message indicating that it is needed.  This is overridden by the real one when it is available.

### DIFF
--- a/hfile.c
+++ b/hfile.c
@@ -1,6 +1,6 @@
 /*  hfile.c -- buffered low-level input/output streams.
 
-    Copyright (C) 2013-2019 Genome Research Ltd.
+    Copyright (C) 2013-2020 Genome Research Ltd.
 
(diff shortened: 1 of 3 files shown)


valeriuo:
Merged as b22e03d805866c04ea02a4b04c9b58a167853418

PR #1048, opened 2020-03-11
Removed gcc9 warning in kroundup.

jkbonfield:
The (size_t)-1 cap to avoid roundup overflows sometimes triggers a new gcc warning (eg in samtools bam_fastq.c).  Given getting to this point is almost always an error somewhere else, we take the alternative strategy of not rounding up instead.

This said, I believe the roundup strategy is fundamentally poor anyway.  Rather than doubling we'd be better off with a lower power, eg roundup(x) being (x+10)*1.5  or similar, with appropriate bracketing and bounds checks etc.  This is almost certainly less code, more frugal on memory, and realistically not going to be a major issue for memory pools either as the number of reallocs and memmoves still isn't vast.  (It's ~70% more, but probably more of those will be the same pointer instead of a new alloc and move.)

### DIFF
--- a/htslib/kstring.h
+++ b/htslib/kstring.h
@@ -38,18 +38,30 @@
 #include "hts_defs.h"
 
 #ifndef kroundup32
-#define kroundup32(x) (--(x), (x)|=(x)>>1, (x)|=(x)>>2, (x)|=(x)>>4, (x)|=(x)>>8, (x)|=(x)>>16, ++(x),(x)=(x)?(x):(uint32_t)-1)
+#define kroundup32(x) ((x)<INT_MAX                                      \
+  ? (--(x),                                                             \
+     (x)|=(x)>>1,                                                       \
+     (x)|=(x)>>2,                                                       \
+     (x)|=(x)>>4,                                                       \
+     (x)|=(x)>>8,                                                       \
+     (x)|=(x)>>16,                                                      \
+     ++(x))                                                             \
+  : ((x)=UINT_MAX))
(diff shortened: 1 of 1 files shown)


daviesrob:
There's still a problem on 32-bit platforms, albeit only one out:
```
In file included from bam_fastq.c:39:0:
In function ‘ks_resize.isra.4.part.5’,
    inlined from ‘ks_resize.isra.4’,
    inlined from ‘kputsn’ at ../htslib/htslib/kstring.h:252:24,
    inlined from ‘kputs’ at ../htslib/htslib/kstring.h:262:9,
    inlined from ‘insert_index_sequence_into_linebuf’ at bam_fastq.c:313:13:
../htslib/htslib/kstring.h:175:7: warning: argument 2 value ‘2147483648’ exceeds maximum object size 2147483647 [-Walloc-size-larger-than=]
   tmp = (char*)realloc(s->s, size);
   ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from bam_fastq.c:28:0:
../htslib/htslib/kstring.h: In function ‘insert_index_sequence_into_linebuf’:
/usr/i686-linux-gnu/include/stdlib.h:549:14: note: in a call to allocation function ‘realloc’ declared here
 extern void *realloc (void *__ptr, size_t __size)
              ^~~~~~~

```

Might be easier to change how `ks_resize()` rounds up instead...

jkbonfield:
Superceded by #1051

Issue #1037, opened 2020-03-03
tabix randomly and silently dropping records retrieved from GCS bucket


mwalker174:
OS Ubuntu 18.04 (within a docker container)
htslib v1.10.2 compiled from source (also when installed from the v0.7 apt package)
gcloud SDK v235.0.0-0 installed from apt package)

We have a large workflow that involves retrieving tabix-indexed records from a large file (~80GB) in a Google bucket. This workflow is run on VMs using the Google Compute Engine. Our workflows scatter across many shards. We have noticed that `tabix` seems to be dropping chunks of records randomly in multiple shards. The records are lost randomly (different records every time we retry the workflow) but consistently (it is observed in multiple shards every run). tabix does not produce any warnings or errors when this happens.

Example command:
```
GCS_OAUTH_TOKEN=`gcloud auth application-default print-access-token` \
      tabix -R region.bed $URI | bgzip -c > out.txt.gz
```
We haven't noticed any patterns other than that records seem to be dropped in contiguous groups, resulting in intervals that are only partially retrieved.

daviesrob:
I suspect the web request to get data is failing, and the error is not being propagated up correctly.  I'll see if I can find out where it's being lost.

daviesrob:
Pull request #1040 adds more error checking to tabix, which should make it more likely to fail in a prompt fashion.  Would it be possible to build that version and try it to see if the failures get reported correctly?

In my testing, the current `develop` branch did report an error from `hts_close()` on failures.  However, my test was a bit crude (I simply killed the web server providing the files) so it's possible that more subtle errors might not have been reported.  Hopefully the new version will make that less likely.

Also, if you're running `tabix` in a pipe (as in the example above), are you using the bash `set -o pipefail` option so that the pipeline will report the `tabix` failure?  (If not, the result will be the one from `bgzip`, which presumably doesn't notice anything wrong and returns success).

mwalker174:
Thank you for the quick turnaround @daviesrob. We have been using `set -o pipefail` in all our scripts. We will give your branch a try and get back to you. We may get in touch with Google to understand why the errors are occurring in the first place because if GCS streaming is this unreliable we may have to implement an alternative approach.

daviesrob:
Currently `tabix` doesn't have a way to turn up HTSlib's logging.  I'll try adding that to my pull request, as it would probably help you to diagnose the problem.

mwalker174:
[...]

PR #1049, opened 2020-03-11
Report more BGZF errors

daviesrob:
* Make `bgzf_read_block()` print more error messages, including the offset of the BGZF block where it failed.

Note that this introduces a small change of behaviour as block_address is now updated after an empty block has been read.  This may change the location stored in indexes of files with empty blocks.

* Make `hts_itr_next()` return an error on seek failure.

The return code needs to be less that -1, or the caller will incorrectly interpret it as end of data, causing it to silently finish too early.

* Fix missing newline in tabix error messages.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -958,6 +958,9 @@ int bgzf_read_block(BGZF *fp)
 
         if (j->errcode) {
             fp->errcode = j->errcode;
+            hts_log_error("BGZF decode jobs returned error %d "
+                          "for block offset %"PRId64,
+                          j->errcode, j->block_address);
             return -1;
         }
 
@@ -1025,6 +1028,9 @@ int bgzf_read_block(BGZF *fp)
         count = hread(fp->fp, fp->uncompressed_block, BGZF_MAX_BLOCK_SIZE);
         if (count < 0)  // Error
         {
+            hts_log_error("Failed to read uncompressed data "
+                          "at offset %"PRId64"%s%s",
+                          block_address, errno ? ": " : "", strerror(errno));
             fp->errcode |= BGZF_ERR_IO;
             return -1;
         }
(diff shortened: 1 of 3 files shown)


jmarshall:
There is a philosophy question here: should library functions print error messages themselves, or should they return -1 and anticipate that their callers will print a better-contextualised error message themselves?

Personally I prefer the latter, as the caller knows better what the high-level problem is, and otherwise you get a cascade of error messages from different levels pertaining to the same event. (OTOH messages from the library can give you the low-level technical details that aid debug tracing.)

TL;DR — perhaps your newly added `hts_log_error` calls would be better as `hts_log_debug` or so.

daviesrob:
I'm going towards more logging, to avoid problems like #1037 where we now know something went wrong but have no idea what caused it.  This also (at last) prints the location in the file where the problem occurs, which is not something that the caller may have available to it - especially in the multi-threaded case.

jkbonfield:
Ideally I'd favour not printing errors, but only *if* there is a proper hts_strerror and and hts_errno type model meaning the client tools can identify the actual source of error rather than a cryptic or misleading thing.

Long term I think overhauling error handling in that manner would be a win, but until it's done I favour printing them up in place - for now.

jmarshall:
Hence the suggestion to make these `hts_log_debug()` — just like the existing one already in `bgzf_read_block()`. Best of both worlds IMHO.

daviesrob:
These messages report actual errors when reading and decompressing the input files, so I think `hts_log_error()` is more appropriate than `hts_log_debug()`.  The messages will normally never be seen, but if there is a problem I want to know where it is without having to re-run a possibly lengthy process with all the debugging turned on.

If anything, we should remove some of the messages from the intermediate functions that don't add much information.  For example, the "Read block operation failed ..." message is rarely if ever useful as while it tells you something happened, it usually gives little clue as to what the problem was and where in the input it happened.

PR #1054, opened 2020-03-31
Fix memory leak on failure in cram_decode_slice()

daviesrob:
Makes the error case decrement the reference count on any used refs and free the refs array.

While this fixes the case found by the fuzzer, `cram_decode_slice()` is a complicated function so it's possible other memory or reference count leaks still exist.  It would be a good idea for others to check to see if there are any obvious ones left.

Credit to OSS-Fuzz
Fixes oss-fuzz 21393

### DIFF
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -2335,13 +2335,13 @@ int cram_decode_slice(cram_fd *fd, cram_container *c, cram_slice *s,
 
         out_sz = 1; /* decode 1 item */
         if (ds & CRAM_BF) {
-            if (!c->comp_hdr->codecs[DS_BF]) return -1;
+            if (!c->comp_hdr->codecs[DS_BF]) goto block_err;
             r |= c->comp_hdr->codecs[DS_BF]
                             ->decode(s, c->comp_hdr->codecs[DS_BF], blk,
                                      (char *)&bf, &out_sz);
             if (r || bf < 0 ||
                 bf >= sizeof(fd->bam_flag_swap)/sizeof(*fd->bam_flag_swap))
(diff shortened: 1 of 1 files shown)


cram/cram_decode.c
```diff
@@ -2708,6 +2708,7 @@ int cram_decode_slice(cram_fd *fd, cram_container *c, cram_slice *s,
                 cram_ref_decr(fd->refs, i);
         }
         free(refs);
+        refs = NULL;
     } else if (ref_id >= 0 && s->ref != fd->ref_free && !embed_ref) {
         cram_ref_decr(fd->refs, ref_id);
```
jkbonfield, line 2713: I see the original code has this else clause.  The `if (refs)` is to handle multiple reference slices, such as when we have 1000s of baits or unplaced contigs each with just 1 or 2 reads in, while the else clause is for the more normal single-ref slices.

This single-ref case is set using `cram_get_ref` near the start of the function and I see that your first change to `goto block_err` comes after that point, so it "feels" like we should have that else clause.  Did you test this?  I'm curious why you didn't copy it.

However given the severity of not freeing is low (one small block of memory leaked) while the severity of getting it wrong is high (double free and a crash), I'm still inclined to go with the safe option of what we have unless we can demonstrate an actual leak.  I'm assuming the fuzz tester will find it in time if it exists!

daviesrob: I didn't copy that bit because it's not entirely clear under what conditions it should happen, and if they're still true if you bail out early.  Also it looks like even if the ref isn't released here, it will be freed at the time the cram_fd is closed, if not earlier.

PR #1059, opened 2020-04-22
Print position of offending VCF/BCF records to help with debugging

pd3:


daviesrob:
Unfortunately, `bcf_record_check()` gets called with `hdr = NULL` from `bcf_readrec()` (i.e. when used from an iterator).  This would lead to a segfault if it tried to call `bcf_seqname()` to report an error.

The safest solution would be to improve `bcf_seqname()` and `bcf_hdr_id2name()` so that they can handle being called with invalid inputs and always return a string.  I'm not sure what the best string to return in the error case would be, but it shouldn't be something that would look like a genuine contig name.  Maybe an empty string, or possibly `"*"`?

jmarshall:
Perhaps `(unknown)` or similar — to parallel printf's common `(null)`, and the parentheses separate it from any genuine contig name.

pd3:
Good catch! This would be only needed for `bcf_seqname()`, but I think a better solution is to fix the code in `bcf_record_check()`, which I'll do now.

daviesrob:
It would be much better to put the fix in `bcf_seqname()` itself, as it will fix all callers (including ones we don't write).  Also I have visions of the fuzz tester finding all sorts of edge cases where it manages to sneak an invalid `rec->rid` into `bcf_seqnames()`, thereby turning a simple error log into a crash.  It also needs therefore to check that `rid` is valid.  Suggestion:

```
    static inline const char *bcf_hdr_id2name(const bcf_hdr_t *hdr, int rid) {
        if (hdr && rid >= 0 && rid < hdr->n[BCF_DT_CTG])
             return hdr->id[BCF_DT_CTG][rid].key;
        else
            return "(unknown)";
    }
    static inline const char *bcf_seqname(const bcf_hdr_t *hdr, bcf1_t *rec) {
        return bcf_hdr_id2name(hdr, rec ? rec->rid : -1);
    }

```

(Slightly off-topic: Someone needs to rebuild [VCFv4.3.pdf](https://samtools.github.io/hts-specs/VCFv4.3.pdf) because at the moment it still says you can have brackets.)

pd3:
I have reservations towards adding this functionality, it attempts to alleviate programming errors rather than user data. If a check needs to be added, `assert()` would be more appropriate.

jmarshall:
> (Slightly off-topic: Someone needs to rebuild VCFv4.3.pdf because at the moment it still says you can have brackets.)

Some of the SAM & CRAM ones need updating too, so I'll take care of this this afternoon.

daviesrob:
An invalid `rid` could easily come from user data, and in the case of an iterator `bcf_record_check()` is unable to filter them out.

The NULL checks will be [eliminated](https://gcc.gnu.org/news/null.html) by the compiler if they're not needed, so there's no harm in having them.

daviesrob:
Squashed and rebased.  The wrapper around `bcf_seqname()` which returns "(unknown)" is now in the header file as `bcf_seqname_safe()` as it's likely to be useful outside of HTSlib.

PR #1060, opened 2020-04-30
Minor improvements and public header cleansing

jmarshall:
A couple of stacked up trivialities:

* Use the newly-available `ks_expand()` to avoid kstring-related explicit `kroundup32+realloc` calls

* Tidy up inclusions of standard C headers from the public _htslib/*.h_ headers — in particular don't include `<assert.h>` from any public headers. (Verified that current samtools and bcftools still build after this change.)


daviesrob:
Thanks.  Good to keep things tidy.

PR #1062, opened 2020-05-07
[header API] Link a new PG line only to valid chains

valeriuo:
1. The length of PG chains is now computed and only chains with at least two PG lines are considered valid candidates for further extension.
2. PG linkage is redone before adding a new PG line.

Fixes https://github.com/samtools/samtools/issues/1235

### DIFF
--- a/header.c
+++ b/header.c
@@ -1244,8 +1244,10 @@ int sam_hdr_rebuild(sam_hdr_t *bh) {
     if (!hrecs->dirty)
         return 0;
 
-    if (hrecs->pgs_changed)
-        sam_hdr_link_pg(bh);
+    if (hrecs->pgs_changed && sam_hdr_link_pg(bh) < 0) {
+        hts_log_error("Linking @PG lines has failed");
+        return -1;
+    }
 
     kstring_t ks = KS_INITIALIZE;
     if (sam_hrecs_rebuild_text(hrecs, &ks) != 0) {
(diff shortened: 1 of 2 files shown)


header.c
```diff
@@ -2170,6 +2181,12 @@ int sam_hdr_add_pg(sam_hdr_t *bh, const char *name, ...) {
         hrecs = bh->hrecs;
     }
 
+    bh->hrecs->pgs_changed = 1;
+    if (sam_hdr_link_pg(bh) < 0) {
```
daviesrob, line 2185: This makes `sam_hdr_add_pg()` fail if an existing chain has a `@PG PP:` tag that points to a non-existent `@PG ID:`, which is a change from previous behaviour.  Is this something we actually want to fail on, or should we just ignore it (maybe by not setting `ret` to -1 when the [hash table lookup fails](https://github.com/samtools/htslib/blob/8859b09/header.c#L2081))?

I also see the `sam_hdr_link_pg()` [call](https://github.com/samtools/htslib/blob/8859b09/header.c#L1248) in `sam_hdr_rebuild()` didn't check the return value, which is why it used to work.

valeriuo: > 
> 
> This makes `sam_hdr_add_pg()` fail if an existing chain has a `@PG PP:` tag that points to a non-existent `@PG ID:`, which is a change from previous behaviour. Is this something we actually want to fail on, or should we just ignore it (maybe by not setting `ret` to -1 when the [hash table lookup fails](https://github.com/samtools/htslib/blob/8859b09/header.c#L2081))?

Print a warning and continue, perhaps.

daviesrob: Yes, it already seems to do the continue bit.

PR #1063, opened 2020-05-10
Maybe consider alignments which consume no reference bases to have length 1

jmarshall:
For the purposes of indexing, maybe for (somewhat invalid) alignment records whose CIGARs consume no reference bases htslib should consider them to have length 1 — similarly to reads that are unmapped or have no CIGAR string at all.

Thus they would be considered to cover one reference position (rather than zero), and would be returned by iterator-based queries. See for example samtools/samtools#1240, in which alignment records with CIGAR=`66P` are used to represent something somewhat different from mapping to a reference.

This may or may not be a good idea, but it would mean that htslib/samtools would do something more sensible for these records (even though they do not represent a traditional mapping).

### DIFF
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -1050,7 +1050,8 @@ hts_pos_t bam_cigar2rlen(int n_cigar, const uint32_t *cigar);
 
       @discussion For a mapped read, this is just b->core.pos + bam_cigar2rlen.
       For an unmapped read (either according to its flags or if it has no cigar
-      string), we return b->core.pos + 1 by convention.
+      string) or a read whose cigar string consumes no reference bases at all,
+      we return b->core.pos + 1 by convention.
  */
 HTSLIB_EXPORT
 hts_pos_t bam_endpos(const bam1_t *b);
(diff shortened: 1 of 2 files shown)


jmarshall:
(Oh, and I suspect the ghost reads are printing uninitialised memory — hence the single failed Travis build.)

jmarshall:
If you're happy that this doesn't have any untoward effect on the “it's harder than that!” multiple alignment pileups you discussed https://github.com/samtools/samtools/issues/1240#issuecomment-626773988 and following, then I'm happy.

You are right that the SAM spec covers this case for unmapped reads, so ought to for this case too if it's a good idea. That is now PR samtools/hts-specs#521.

The change does indeed move such reads on a bin boundary up one level to a larger bin, but they'll still be returned by the same iterator queries (and some new queries, just like those not near a bin boundary). I'm not too concerned by this change, as it affects only reads that shouldn't appear at all in aligner output and can be considered a bug fix for whatever usages there are in which these unorthodox mappings might be used…

jkbonfield:
I thought about this a bit before adding my +1 on the hts-specs issue.  It feels like it's erring on the right side.

To erroneously put a read in a higher bin means we very occasionally do more work than is strictly necessary, but the results do not change.  To erroneously put a read in a lower bin means we occasionally return the wrong result (missing data).  So I feel the change is a safe one and a reasonable recommendation to make.

daviesrob:
Thanks.  Merged in anticipation that the hts-specs PR will go through.

PR #1064, opened 2020-05-12
Fix the multi-threaded CRAM multi-region iterator regression.

### LINKED ISSUE #1061: Multi-region iterator inefficiency regression
On CRAM, multi-threaded decoding in conjunction with the multi-region iterator is very slow and inefficient, causing lots of excess work.

However it didn't used to be this way.  The changes in bfc9f0d35bb608f46f8fdbf6a284931af78c1461 interact badly with threading.  I have that commit and the commit before compiled up:

```
time /tmp/test_view.slow -@8 -M -B NA06985.final.chr1.cram `perl -lane 'print "$F[0]:". ($F[1]+1) ."-$F[2]"' 10k.bed`

real	0m15.145s
user	0m54.005s
sys	0m1.820s

time /tmp/test_view.fast -@8 -M -B NA06985.final.chr1.cram `perl -lane 'print "$F[0]:". ($F[1]+1) ."-$F[2]"' 10k.bed`

real	0m4.247s
user	0m14.079s
sys	0m0.838s
```

For reference, the unthreaded version is the comparable in both.  Here's an example:

```
time /tmp/test_view.fast -M -B NA06985.final.bam `perl -lane 'print "$F[0]:". ($F[1]+1) ."-$F[2]"' 10k.bed`

real	0m2.649s
user	0m2.604s
sys	0m0.044s
```

I was attempting to work how to speed this up by changing the seeks to CRAM_OPT_RANGE requests and noticed that 1.10 was already massively faster than current develop.

My suggestion though is indeed to use CRAM_OPT_RANGE.  Something like (but not quite) this:

```
                if (iter->curr_off < iter->off[iter->i].u) {                    
                    iter->curr_off = iter->off[iter->i].u;                      
                    // CRAM has the capability of setting an end location.      
                    // This means multi-threaded decodes can stop once they     
                    // reach that point, rather than pointlessly decoding       
                    // more slices than we'll be using.                         
                    if (iter->is_cram && iter->curr_intv == 0) {                
                        cram_range r = {                                        
                            iter->curr_tid,                                     
                            iter->reg_list[iter->curr_reg].min_beg,             
                            iter->reg_list[iter->curr_reg].max_end              
                        };                                                      
                        cram_set_option(fp, CRAM_OPT_RANGE, &r);                
                    } else   // else needed? Unsure!                                                   
                    if (iter->seek(fp, iter->curr_off, SEEK_SET) < 0) {         
                        hts_log_error("Seek at offset %" PRIu64 " failed.", iter->curr_off);                                                                   
                        return -1;                                              
                    }                                                           
                }
            }                                                                   
        }                                                                       
                                                                                
        ret = iter->readrec(fp, fd, r, &tid, &beg, &end);                       
        if (ret < 0) {                                                          
            if (iter->is_cram && cram_eof(fp)) {                                
                fprintf(stderr, "eof in this range, try next\n");               
                iter->curr_off = iter->off[iter->i].v; // skip to end of range  
                continue;                                                       
            }                                                                   
            break;                                                              
        }                                                                       
                                                              
```

The thought here is we know the start and end locations in genomic coords, and we've mapped the start over to file offsets.  When we get to the next genomic coord we do another seek to the file offset.  However with multi-threading it's decoding ahead of time.  We have no way to tell BAM to stop, but CRAM does have this mechanism.  We can tell it to cease decoding once it hits the end genomic location, so the threading won't waste lots of CPU decoding something that isn't in our range.  It's not perfect (that would be to speculatively start decoding the next range), but it's better than what we have now.  It doesn't help though due to the aforementioned issue.

I haven't tried yet to see if these changes applied to 1.10 speed it up.

jkbonfield:
Fixes #1061

This appears to have been introduced during bfc9f0d and fixed for BAM only in 6149ea6.  The effect on multi-threading decoding for CRAM was significant.

This fix takes a totally different approach, which not only fixes the regression but passes it.  CRAM can do CRAM_OPT_RANGE queries, used by the single version of the iterator, which informs the decode of both start and end range positions.  When threading this means we don't preemptively decode the next container unless it'll actually be used. This same logic is now used in the multi-region iterator, although it's complex.

The general strategy is as follows:

- Ensure we know the next container start from index.  This needs a small tweak to cram_index struct as the next container isn't quite the same as this container + slice offset + slice size (sadly).  I think it's missing the size of the container struct itself.

- When being given a start..end offset, step into the reg_list to find the corresponding chr:start-end range.

- Produce a new CRAM_OPT_RANGE_NOSEEK option that does the same job as the old CRAM_OPT_RANGE opt but without the seek.  This is necessary hts.c does the seek for us and the pseek/ptell only work with each other and not within the cram subdir code itself).

- Identify neighbouring file offsets and merge together their correponding ranges so a block of adjacent offsets becomes a single CRAM_OPT_RANGE query.  We cache the end offset (.v) in iter->end so we can avoid duplicating the seek / range request in subsequent intervals.

- Manage the EOF vs EOR (end of range) return values.  For EOR we have do the incrementing ourselves as we need to restart the loop without triggering the end-of-file or end-of-multi-iterator logic below it.

- Tweak the region sorting a bit so ties are resolved by the max value.  We want the index into reg_list to also be sorted, so we can accurately step through them.  Note this logic is CRAM specific, as the sorting wouldn't work on BAI anyway due to the R-tree.

Some benchmarks on ~37,000 regions returning ~110 million seqs across NA06985.final.cram:

```
CRAM-1.10 no threads:  real  4m37.361s   user  4m21.128s   sys  0m7.988s
CRAM-1.10 -@16:        real  3m14.371s   user 28m48.872s   sys  4m52.442s
CRAM-dev -@16:         real  5m55.670s   user 61m0.005s    sys 11m23.147s
CRAM-current -@16:     real  1m55.701s   user  5m54.234s   sys  0m51.699s
```

The increase in user time between unthreaded and 16 threads is modest. System time increase is considerable, but this appears to primarily be down to glibc malloc inefficiencies.  Using libtcmalloc.so instead gives:
```
1M CRAM-current -@16:  real  1m30.882s   user  6m9.103s    sys  0m4.960s
```
We're still nowhere near using 16 threads, but this is because the average size of each region is significantly smaller than 16 containers worth.

### DIFF
--- a/cram/cram_index.c
+++ b/cram/cram_index.c
@@ -481,6 +481,34 @@ cram_index *cram_index_query_last(cram_fd *fd, int refid, hts_pos_t end) {
             first++;
     }
 
+    // Compute the start location of next container.
+    //
+    // This is useful for stitching containers together in the multi-region
+    // iterator.  Sadly we can't compute this from the single index line.
+    //
+    // Note we can have neighbouring index entries at the same location
+    // for when we have multi-reference mode and/or multiple slices per
+    // container.
+    cram_index *next = first;
+    do {
+        if (next >= last) {
+            // Next non-empty reference
+            while (++refid+1 < fd->index_sz)
+                if (fd->index[refid+1].nslice)
+                    break;
+            if (refid+1 >= fd->index_sz) {
+                next = NULL;
+            } else {
+                next = fd->index[refid+1].e;
+                last = fd->index[refid+1].e + fd->index[refid+1].nslice;
+            }
+        } else {
+            next++;
+        }
+    } while (next && next->offset == first->offset);
+
+    first->next = next ? next->offset : 0;
+
     return first;
 }
 
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -4870,6 +4870,25 @@ int cram_set_voption(cram_fd *fd, enum hts_fmt_option opt, va_list args) {
         return r;
     }
 
+    case CRAM_OPT_RANGE_NOSEEK: {
+        // As per CRAM_OPT_RANGE, but no seeking
+        pthread_mutex_lock(&fd->range_lock);
+        cram_range *r = va_arg(args, cram_range *);
+        fd->range = *r;
+        if (r->refid == HTS_IDX_NOCOOR) {
+            fd->range.refid = -1;
+            fd->range.start = 0;
+        } else if (r->refid == HTS_IDX_START || r->refid == HTS_IDX_REST) {
+            fd->range.refid = -2; // special case in cram_next_slice
+        }
+        if (fd->range.refid != -2)
+            fd->required_fields |= SAM_POS;
+        fd->ooc = 0;
+        fd->eof = 0;
+        pthread_mutex_unlock(&fd->range_lock);
+        return 0;
+    }
+
     case CRAM_OPT_REFERENCE:
         return cram_load_reference(fd, va_arg(args, char *));
 
--- a/cram/cram_structs.h
+++ b/cram/cram_structs.h
@@ -663,6 +663,7 @@ typedef struct cram_index {
     int     slice;  // 1.0 landmark index, 1.1 landmark value
     int     len;    //                     1.1 - size of slice in bytes
     int64_t offset; // 1.0                 1.1
+    int64_t next;   // derived: offset of next container.
 } cram_index;
 
 typedef struct {
--- a/hts.c
+++ b/hts.c
@@ -1603,9 +1603,11 @@ int hts_check_EOF(htsFile *fp)
 #define META_BIN(idx) ((idx)->n_bins + 1)
 
 #define pair64_lt(a,b) ((a).u < (b).u)
+#define pair64max_lt(a,b) ((a).u < (b).u || \
+                           ((a).u == (b).u && (a).max < (b).max))
 
 KSORT_INIT_STATIC(_off, hts_pair64_t, pair64_lt)
-KSORT_INIT_STATIC(_off_max, hts_pair64_max_t, pair64_lt)
+KSORT_INIT_STATIC(_off_max, hts_pair64_max_t, pair64max_lt)
 
 typedef struct {
     int32_t m, n;
@@ -2832,7 +2834,9 @@ int hts_itr_multi_cram(const hts_idx_t *idx, hts_itr_t *iter)
                     }
 
                     if (e) {
-                        off[n_off++].v = e->offset + e->slice + e->len;
+                        off[n_off++].v = e->next
+                            ? e->next
+                            : e->offset + e->slice + e->len;
                     } else {
                         hts_log_warning("Could not set offset end for region %d:%"PRIhts_pos"-%"PRIhts_pos". Skipping", tid, beg, end);
                     }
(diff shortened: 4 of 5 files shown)


jkbonfield:
I know this is quite hairy and long-winded, but the fix is complex!  I've tried my best to test it.  Things I've looked at:

- Neighbouring offset ranges that span chromosomes; eg every end of one chr and very start of the next.

- The boundary of mapped to unmapped data (query region "*"), both with neighbouring offsets and non-neighbouring offsets.

- Multi slice containers

- Slices with multi-ref mode (RI cram data series) giving multiple references using the same offset.

- General random stress testing, around about 100,000 ranges of varying size across whole genomes.

Please do test more though! It's hard to think of all the cases and I kept finding bugs when inventing new test cases so it's really tricky to validate this is complete.  That said I did stumble across the indexing bug in 1.10 and earlier (that we'd already fixed) so my testing is apparently more complete than we'd done for several htslib releases.

jmarshall:
The documentation added for `CRAM_OPT_RANGE_NOSEEK` — “CRAM_OPT_RANGE minus the seek” — is fairly unclear for a user trying to understand which `hts_fmt_option` they want to use.

Adding documentation of `hts_itr_t` to the public header file will encourage users to access these fields directly, which they should not be doing. It would be better to put such internal documentation in _hts.c_ near the iterator code instead.

(This struct is not opaque for historical reasons only. It is used only in _hts.c_ and _sam.c_ (and also in _tabix.c_ but that should be fixed, and once in _htslib/sam.h_ but that `if` statement could easily be refactored inside a non-inlined library API function in _sam.c_). If those changes were made, `hts_itr_t` could be made opaque in a future ABI-breaking release.

A couple of fields are used directly in a couple of places in samtools, but this is also something that should be fixed.)

jkbonfield:
We really need to go through our header files adding comments above the private structures, informing people that they need to be considered as private and are public only due to use by static-inline functions.

The other change regarding -1 to HTS_IDX_NOCOOR is a potential change in behaviour.  HTS_IDX_NOCOOR is -2, not -1.   However I think it's the correct change as `cram_index_query` has code that recognises HTS_IDX_NOCOOR and replaces it with tid -1, as does `cram_seek_to_refpos`.  Giving it -1 would have worked though as it's what NOCOOR ends up being changed to.  Hence it looks OK.

Did you experimentally verify it though by using "*" as one of your multi-region chromosomes?

valeriuo:
> We really need to go through our header files adding comments above the private structures, informing people that they need to be considered as private and are public only due to use by static-inline functions.

I second that. Having the documentation on top of the struct definition helps with retrieving it faster when needed, especially by IDE doc pop-ups. And explicitly telling people not to access the struct fields directly is probably a better guarantee they would not do it.

> The other change regarding -1 to HTS_IDX_NOCOOR is a potential change in behaviour. HTS_IDX_NOCOOR is -2, not -1. However I think it's the correct change as `cram_index_query` has code that recognises HTS_IDX_NOCOOR and replaces it with tid -1, as does `cram_seek_to_refpos`. Giving it -1 would have worked though as it's what NOCOOR ends up being changed to. Hence it looks OK.

I think doing away with the hard-coded value makes the code clearer. Also the macro name corresponds exactly to the state the iterator is in.

> Did you experimentally verify it though by using "*" as one of your multi-region chromosomes?

Yes

PR #1066, opened 2020-05-13
Collected minor issues: warning fixes, buglets

jmarshall:
More collected minor items:

* Fix a buglet in the #917 check: this should be tested even if this is the first read on a new chromosome.

* Fix `-Wextra` warnings emanating from the public _htslib/*.h_ headers:

```
$ gcc-9 -c -Wall -Wextra includeall.c
In file included from includeall.c:4:
htslib/hfile.h: In function 'hwrite':
htslib/hfile.h:271:35: warning: comparison of integer expressions of different signedness: 'long int' and 'size_t' {aka 'long unsigned int'} [-Wsign-compare]
  271 |         if (fp->limit - fp->begin < nbytes){
      |                                   ^
In file included from includeall.c:7:
htslib/hts_endian.h: In function 'le_to_i32':
htslib/hts_endian.h:245:34: warning: operand of ?: changes signedness from 'int' to 'uint32_t' {aka 'unsigned int'} due to unsignedness of other operand [-Wsign-compare]
  245 |     return v < 0x80000000U ? v : -((int32_t) (0xffffffffU - v)) - 1;
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
htslib/hts_endian.h: In function 'le_to_i64':
htslib/hts_endian.h:257:19: warning: operand of ?: changes signedness from 'long long int' to 'uint64_t' {aka 'long long unsigned int'} due to unsignedness of other operand [-Wsign-compare]
  257 |             ? v : -((int64_t) (0xffffffffffffffffULL - v)) - 1);
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

(Clang only spots the _hfile.h_ one. The _hts_endian.h_ ones appear to be another instance of dc3303b6da542f74c965b3f3f50a58776f87104b and in fact there's another instance without the cast in `safe_itf8_get()` in _cram/cram_io.h_ that probably needs it too.)

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -1906,7 +1906,7 @@ int hts_idx_push(hts_idx_t *idx, int tid, hts_pos_t beg, hts_pos_t end, uint64_t
         hts_log_error("Unsorted positions on sequence #%d: %"PRIhts_pos" followed by %"PRIhts_pos, tid+1, idx->z.last_coor+1, beg+1);
         return -1;
     }
-    else if (end < beg) {
+    if (end < beg) {
         // Malformed ranges are errors. (Empty ranges (beg==end) are unusual but acceptable.)
         hts_log_error("Invalid record on sequence #%d: end %"PRId64" < begin %"PRId64, tid+1, end, beg+1);
         return -1;
--- a/htslib/hfile.h
+++ b/htslib/hfile.h
@@ -267,8 +267,9 @@ hwrite(hFILE *fp, const void *buffer, size_t nbytes)
     extern ssize_t hwrite2(hFILE *, const void *, size_t, size_t);
     extern int hfile_set_blksize(hFILE *fp, size_t bufsiz);
 
-    if(!fp->mobile){
-        if (fp->limit - fp->begin < nbytes){
+    if (!fp->mobile) {
+        size_t n = fp->limit - fp->begin;
+        if (n < nbytes) {
             hfile_set_blksize(fp, fp->limit - fp->buffer + nbytes);
             fp->end = fp->limit;
         }
--- a/htslib/hts_endian.h
+++ b/htslib/hts_endian.h
@@ -242,7 +242,7 @@ static inline int16_t le_to_i16(const uint8_t *buf) {
  */
 static inline int32_t le_to_i32(const uint8_t *buf) {
     uint32_t v = le_to_u32(buf);
-    return v < 0x80000000U ? v : -((int32_t) (0xffffffffU - v)) - 1;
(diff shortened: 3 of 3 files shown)


daviesrob:
Looks good, thanks.

jmarshall:
Cheers. HTSlib itself doesn't get built with `-Wextra` (way too many such warnings in the _*.c_ code) but people using HTSlib headers in their programs might want their own code to be `-Wextra` clean, and fair enough.

I was about to add another commit with 4 more instances of “there's another instance without the cast”, which I guess I'll do as a followup tiny PR.

jkbonfield:
> htslib/hts_endian.h: In function 'le_to_i64':
> htslib/hts_endian.h:257:19: warning: operand of ?: changes signedness from 'long long int' to 'uint64_t' {aka 'long long unsigned int'} due to unsignedness of other operand [-Wsign-compare]
>   257 |             ? v : -((int64_t) (0xffffffffffffffffULL - v)) - 1);
>       |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```

I'd still argue that this is just excessive naval gazing which is apparently causing as many issues as it solves (although maybe now fixed?).

This is simply a manual implementation of 2s-complement notation, because assuming your system uses 2s-complement is technically undefined behaviour.  However in reality it's how *everything* we support works, and furthermore I think it's becoming mandatory behaviour in a more recent C standard (although my memory is fuzzy on that so maybe it's wrong).

Eg the cram_io.c change was (in part):

```
-        *val_p = val;
+        *val_p = c[1] < 0x80 ? uval : -((int64_t) (0xffffffffffffffffULL - uval)) - 1;
```

Given lack of comment, I reckon the previous code was 10 times better! :-)

jmarshall:
> I think it's becoming mandatory behaviour in a more recent C standard (although my memory is fuzzy on that so maybe it's wrong)

You're correct — it took a while to track down our previous conversation on this, but it's here: https://github.com/samtools/htslib/issues/995#issuecomment-560550617.

¯\\\_(ツ)\_/¯

daviesrob:
I checked.  While the representation of `int16_t` etc. is well defined, exactly what happens when you cast an unsigned value to one isn't, even in N2412.  The C standards people need to add a paragraph explaining exactly how the cast works, like they have for the signed-to-unsigned case.

jmarshall:
@daviesrob: :thinking: \<interesting… emoji>

PR #1068, opened 2020-05-15
Fix buffer overflow in hts_md5_update()

valeriuo:
The already present check in `cram_decode_slice` did not account for the last byte. This has been corrected.

Credit to OSS-Fuzz
Fixes oss-fuzz 22231


jkbonfield:
You beat me to it.  I was just wrestling vcf a bit more and was going to take a peek this afternoon.

Yes the fix looks correct.  Although I wonder if it should just be `if (s->hdr->ref_seq_span > b->uncomp_size)` as after all the + and -s basically that's what it boils down to, and it becomes very obvious why it was bugged too then.

valeriuo:
> Although I wonder if it should just be `if (s->hdr->ref_seq_span > b->uncomp_size)` as after all the + and -s basically that's what it boils down to, and it becomes very obvious why it was bugged too then.

That would probably be the cleaner solution.

PR #1070, opened 2020-05-15
[manpage, minor] Mention TBI chromosome length limit

jmarshall:
A _tabix.1_ companion to PR samtools/samtools#1248. In fact the [bfx.SE question](https://bioinformatics.stackexchange.com/questions/13282/vcf-merge-fails-due-to-tabix-not-producing-tbi-files) was about tabix indexes…


daviesrob:
Looks good, thanks.

PR #1073, opened 2020-05-19
Improve SAM parser unrecognised RNAME warnings

jmarshall:
Dusting off another minor branch from a few months ago… This addresses most of the small warning improvements noted in passing on samtools/samtools#1158:

> There is room for improvement in the warning: it could give an indication of the line number in question and/or print out the problematic contig name (carefully, as this is potentially adversarial input); and there is a buglet (`s/tid/mtid/`) nearby (and a typo in the message).

To print out the contig name, the PR adds a hopefully armour-plated `hts_strprint()` utility function which formats user-supplied text truncated to a sensible length and with unprintable characters escaped. This may be helpful for printing out adversarial input elsewhere in htslib. Added some tests (which pass plain `valgrind`) to _test/test_str2int.c_ as it exercises other functions in _textutils_internal.h_ but they might be better somewhere less misleadingly named… :smile:

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1960,14 +1960,14 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
 #endif
 
 #define _get_mem(type_t, x, b, l) if (possibly_expand_bam_data((b), (l)) < 0) goto err_ret; *(x) = (type_t*)((b)->data + (b)->l_data); (b)->l_data += (l)
-#define _parse_err(cond, msg) do { if (cond) { hts_log_error(msg); goto err_ret; } } while (0)
-#define _parse_err_param(cond, msg, param) do { if (cond) { hts_log_error(msg, param); goto err_ret; } } while (0)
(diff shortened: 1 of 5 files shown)


jmarshall:
```c
const char *hts_strprint(char *buf, size_t buflen, const char *s, size_t len);
```

This might be better with an additional `char quote` parameter, which could be `\0`, `'`, or `"`. If it's non-NUL, `hts_strprint()` would wrap the output in quotes, escape any embedded quote characters, and maybe put the ellipsis outside the quotes. That'd probably be better, and would make this a drop-in replacement for _vcf.c_'s `dump_char()`.

jmarshall:
Refactored to include quote mark handling.

daviesrob:
Thanks.  I think the ultimate solution would be to build a custom printf format decoder for `hts_log()` so that it can sanitise character data itself, but this is a step in the right direction.

jmarshall:
Cheers. That's a pretty good ultimate solution!

Issue #1074, opened 2020-05-20
Compiling error: cram/cram.h


lufuhao:
I got some errors when I was compiling htslib in ubuntu 20; Please help
gcc version gcc (Ubuntu 9.3.0-10ubuntu2) 9.3.0
$ git clone https://github.com/samtools/htslib.git
$ cd htslib
$ autoheader
$ autoconf
$ ./configure
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for ranlib... ranlib
checking for grep that handles long lines and -e... /bin/grep
checking for C compiler warning flags... -Wall
checking for pkg-config... /usr/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking for special C compiler options needed for large files... no
checking for _FILE_OFFSET_BITS value needed for large files... no
checking shared library type for unknown-Linux... plain .so
checking whether the compiler accepts -fvisibility=hidden... yes
checking how to run the C preprocessor... gcc -E
checking for egrep... /bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking for stdlib.h... (cached) yes
checking for unistd.h... (cached) yes
checking for sys/param.h... yes
checking for getpagesize... yes
checking for working mmap... yes
checking for gmtime_r... yes
checking for fsync... yes
checking for drand48... yes
checking for srand48_deterministic... no
checking whether fdatasync is declared... yes
checking for fdatasync... yes
checking for library containing log... -lm
checking for zlib.h... yes
checking for inflate in -lz... yes
checking for library containing recv... none required
checking for bzlib.h... yes
checking for BZ2_bzBuffToBuffCompress in -lbz2... yes
checking for lzma.h... yes
checking for lzma_easy_buffer_encode in -llzma... yes
checking for libdeflate.h... no
checking for libdeflate_deflate_compress in -ldeflate... no
checking for curl_easy_pause in -lcurl... yes
checking for CCHmac... no
[...]

PR #1075, opened 2020-05-20
Use relative pathnames for header includes.

### LINKED ISSUE #347: HTSlib should prefer its own headers when being compiled
If you have an old version of HTSlib installed in a built-in or `$CPATH` location, or otherwise have external copies of _hts.h_ etc available, building HTSlib itself should always use the headers within itself rather than external ones.

See #117, #205, and samtools/samtools#544 where this has been encountered before but the reporters keep closing the issues when they've been resolved for themselves but before we've sorted it out once and for all! :smile:

This probably comes down to combinations of `#include <htslib/hts.h>` vs `#include "htslib/hts.h"` vs `#include "hts.h"`, but this is complicated by the various subdirectories (perhaps _cram/cram_io.c_ wants to say `#include "../htslib/hts.h"`?) and the fixed `-I.` (but cf configuring in a separate build directory) used.

(There is probably also a samtools/etc followup, along the lines of _when building samtools, ensure that it's using the headers from the desired htslib_.  There the question of `<htslib/hts.h>` vs `"htslib/hts.h"` is more fraught, as `"…"` is not appropriate when building against an installed htslib.)

jkbonfield:
This fixes #347 for users that explicitly add older versions of htslib into the C_INCLUDE_PATH environment variables.  I'm not sure where that meme comes from, but this makes the code base immune to it.  Ths fix feels a bit clunky, but I cannot think of a better way to protect against deliberately scuppering your build chances.

The problem can be demonstrated by doing an `apt-get install libhts-dev` on Ubuntu Bionic (which gets an earlier incompatible version) and compiling with `C_INCLUDE_PATH=/usr/include/: make`.  See #1074 for example of the most recent errors and gcc -H output showing where it deviates into /usr/include land.

We can fix it to get the compilation working, but to find all of them including cases where the files are compatible but in the wrong location, I did this:

```
make -n | sed 's/gcc/gcc -H/' > _
C_INCLUDE_PATH=/usr/include/: sh -x _ 2>&1 | less
```

Then search for usr/include/htslib in the output.

### DIFF
--- a/cram/cram.h
+++ b/cram/cram.h
@@ -56,6 +56,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //
 // This contains duplicated portions from cram_io.h and cram_structs.h,
 // so we want to ensure that the prototypes match.
-#include "htslib/cram.h"
+#include "../htslib/cram.h"
 
 #endif
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -50,7 +50,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "cram/cram.h"
 #include "cram/os.h"
-#include "htslib/hts.h"
+#include "../htslib/hts.h"
 
 //Whether CIGAR has just M or uses = and X to indicate match and mismatch
 //#define USE_X
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -46,8 +46,8 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "cram/cram.h"
 #include "cram/os.h"
 #include "sam_internal.h" // for nibble2base
-#include "htslib/hts.h"
-#include "htslib/hts_endian.h"
+#include "../htslib/hts.h"
+#include "../htslib/hts_endian.h"
 
 KHASH_MAP_INIT_STR(m_s2u64, uint64_t)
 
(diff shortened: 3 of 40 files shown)


jkbonfield:
Note, there is likely a comparable issue with samtools.   That however would be different to fix as we can't very easily assume where the htslib is coming from (subdir, sibling, somewhere else).

jmarshall:
On further perusal, this covers the necessary changes of `#include "htslib/*.h"` to `#include "../htslib/*.h"`.

However it makes some but not all of the similar changes to `#include "../cram/*.h"` and `#include "../*_internal.h"`, so it doesn't insulate against users who might have _cram/*.h_ (or even an entire HTSlib source directory) in C_INCLUDE_PATH. See PR #1078 for a version that makes all of these sets of changes.

jkbonfield:
I'm not sure I understand what you mean there.  Are you saying C_INCLUDE_PATH contains not only e.g. /usr/include/htslib but also the explicit subdirectories such as /usr/include/htslib/cram?  (Why would you do that!?)  There are [memes out there](https://github.com/walaj/SeqLib) that do the former which IMO are misguided, but there is no reason to ever want to add the subdirectories.

I did validate that with C_INCLUDE_PATH including /usr/include/htslib that it never included anything from that directory.

Ah, or maybe I see what you mean.  If you have C_INCLUDE_PATH to not be an installed directory but a source tree then you'll get internal headers as well. Again WTF would you do that!?   I guess never underestimate the "inventiveness" of people, and infact I see the SeqLib instructions do explicitly ask you to add their git copy of htslib to C_INCLUDE PATH.  Blergh!

daviesrob:
Debian were [shipping](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=881170) `/usr/include/htslib/cram` for quite a while, mainly due to SeqLib.

We should probably go for the comprehensive solution to ensure we don't get any more problems, and remember to do future includes with relative paths.

jkbonfield:
Happy enough to do that.

daviesrob:
Replaced by #1078

PR #1076, opened 2020-05-20
Allow multiple bcf readers without index

xbrianh:
This adds logic allowing htslib to warn, instead of throw, when using multiple bcf readers without an index. The default is to throw.

This is will allow bcftools, when lightly modified, to merge multiple VCFs without requiring indexes. This is useful for cases where `bcftools view` commands have been used to create subsampled VCFs for multiple levels of data access.

The above situation is currently present for some users in the [BioData CATALYST](https://biodatacatalyst.nhlbi.nih.gov/) project, a consortium project involve the Broad institute, the Universty of Chicago, The University of California Santa Cruz, and others. Large multisample VCFs are available in cloud buckets without indexes. Currently there is no straightforward way to use bcftools to merge these files without first indexing them, even though they are known to be aligned.

There was already an [abortive attempt to bring this functionality to samtools](https://github.com/samtools/htslib/pull/203). At that time the samtools developers seemed open to the implementation proposed in this PR, but unfortunately the work never went forward.

### DIFF
--- a/htslib/synced_bcf_reader.h
+++ b/htslib/synced_bcf_reader.h
@@ -38,7 +38,7 @@ DEALINGS IN THE SOFTWARE.  */
 
         bcf_srs_t *sr = bcf_sr_init();
         bcf_sr_set_opt(sr, BCF_SR_PAIR_LOGIC, BCF_SR_PAIR_BOTH_REF);
-        bcf_sr_set_opt(sr, BCF_SR_REQUIRE_IDX);
+        bcf_sr_set_opt(sr, BCF_SR_REQUIRE_IDX_THROW);
         for (i=0; i<nfiles; i++)
             bcf_sr_add_reader(sr,files[i]);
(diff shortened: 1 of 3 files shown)


pd3:
This would end badly when input files have chromosomes in different order. A better solution is to add support for creating indexes on the fly.

pd3:
This pull request is flawed and cannot work as is. I will check if we can do something to support this functionality.

xbrianh:
@pd3 The goal of this patch is to support merging VCF streams where it is known that the chromosomes are ordered.

Is it possible to create indexes for streams on the fly? How would you avoid processing the entire stream to create the index?

It seems samtools does not support this use case in order to prevent users from shooting themselves in the foot. However, for users who know that the chromosomes are in the same order, it seems reasonable to provide an override.

pd3:
No, this would not work reliably even when the chromosome blocks come in the same order. More changes are needed, not just letting the code run through and hope for the best.

PR #1077, opened 2020-05-21
Ensure HTSlib prefers its own headers when being compiled

### LINKED ISSUE #347: HTSlib should prefer its own headers when being compiled

jmarshall:
The full preprocessor search path for #include "..." is typically:

    1) directory of the file containing the #include directive
    2) environment variables $CPATH, $C_INCLUDE_PATH, etc
    3) command line options -I etc
    4) system include directories

It's easy to oversimplify (1) as “the current directory” but that's misleading when the compiler's PWD is always the top-level htslib source directory but the “current directory of the `#include` directive” might be one of the subdirectories.

When a top-level _*.[ch]_ file does `#include "htslib/foo.h"`, the nearby header file within the source tree is found due to (1). When a file in a subdirectory (_htslib/*.h_, _cram/*.[ch]_, or _test/*.c_) does the same `#include`, the nearby header file is intended to be found due to (3) via the Makefile's hard-coded `-I.` — `.` here being the compiler process's $PWD, i.e., the top-level HTSlib source directory.

However the latter can be subverted if there is an "htslib" directory containing foo.h in $CPATH, $C_INCLUDE_PATH, or on the command line prior to the hard-coded `-I.`. This will be the case if users have made a previous HTSlib installation accessible via $CPATH/etc or by adding a -I option to $(CFLAGS) (rather than $(CPPFLAGS)). One reason we have found this hard to reproduce in the past is that we have naturally added dummy `-I` options to CPPFLAGS, but that comes later in the search path than `-I.`. Naive users OTOH are likely to add their dummy `-I` options to CFLAGS.

Avoid this and fix #347 by adding `../` to #includes in subdirectories (and always using `""` rather than `<>`), so that includes within HTSlib (even from the source subdirectories) are found due to (1).

This PR is split into three commits: for _htslib/*.h_, for _cram/*.h_, and for all other _*.h_ files. The question is whether to apply 1, 2, or all 3 commits.

* The first commit covers the public headers, which is the likely scenario of “oops I added my previous HTSlib installation to `$C_INCLUDE_PATH`”.

* Adding the second commit extends this to the private _cram/*.h_ headers. See the commit messages for details; note that for a (long) time [Debian packaged and shipped these headers](https://twitter.com/jomarnz/status/1207233704822747139) alongside _htslib/*.h_.

* Adding the third commit as well extends this to all intra-HTSlib inclusions, so they're all similarly somewhat uglified. With this, the only `#include`s in HTSlib still depending on (3)'s hard-coded `-I.` are each _*.c_ file's `#include <config.h>`.

Probably the answer is either to apply 1 (for the minimal change) or all 3 commits (for consistency).

### DIFF
--- a/cram/cram.h
+++ b/cram/cram.h
@@ -43,7 +43,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define CRAM_ALL_H
 
 #include "cram_samtools.h"
-#include "header.h"
+#include "../header.h"
 #include "cram_structs.h"
 #include "cram_io.h"
 #include "cram_encode.h"
--- a/cram/cram.h
+++ b/cram/cram.h
@@ -56,6 +56,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //
 // This contains duplicated portions from cram_io.h and cram_structs.h,
 // so we want to ensure that the prototypes match.
-#include "htslib/cram.h"
+#include "../htslib/cram.h"
 
 #endif
--- a/cram/cram.h
+++ b/cram/cram.h
@@ -42,7 +42,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifndef CRAM_ALL_H
 #define CRAM_ALL_H
 
-#include "cram/cram_samtools.h"
+#include "cram_samtools.h"
 #include "header.h"
 #include "cram_structs.h"
 #include "cram_io.h"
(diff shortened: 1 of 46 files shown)


jmarshall:
I suppose I could have pushed these commits one at a time so they could all be exercised by travis. I'll do that in a new PR.

PR #1087, opened 2020-06-24
Check for null pointer in source string.

valeriuo:
Considering `kputs` is exposed as a public method and used extensively with data parsed directly from input files, it should check that its source string is not null.
Probably, a similar check should be implemented for `kputsn`, which has the potential of an undefined behaviour when calling `memcpy` with a null pointer.

Credit to OSS-Fuzz
Fixes oss-fuzz 23670

### DIFF
--- a/htslib/kstring.h
+++ b/htslib/kstring.h
@@ -228,6 +228,7 @@ static inline int kputsn(const char *p, size_t l, kstring_t *s)
 
 static inline int kputs(const char *p, kstring_t *s)
 {
+	if (!p) { errno = EFAULT; return -1; }
 	return kputsn(p, strlen(p), s);
 }
 


jmarshall:
If this was discovered by fuzzing, isn't the real problem (or an additional problem, at least) in the location where `kputs(NULL, …)` was called?

daviesrob:
Yes.  It was `vcf_format()` trying to print a NULL CHROM.  So there's likely more work to track down the root cause.

valeriuo:
True, the VCF code would benefit from more field validation and error checking.

daviesrob:
See #1092 for BCF fixes.

This should probably set `errno = EFAULT` in the error case (which is what, for example, `read()` sets if you give it a bad pointer).

PR #1088, opened 2020-06-24
Improve bam_aux_update_str()

daviesrob:
While recommending `bam_aux_update_str()` elsewhere, I noticed that it's interface was somewhat unfriendly as it required a NUL-terminated string, and the passed-in length had to include the NUL-terminating byte.  This PR removes that restriction in a backwards-compatible manner so that forgetting to count the NUL no longer creates an invalid tag.  As a side benefit, it becomes possible to create a tag from part of a longer string.

`bam_aux_update_str()` is made a bit more efficient, especially when replacing an existing tag, by reducing the number of `memmove()` calls used to get any following aux tags into the correct position.  It also calculates the new record size more accurately when resizing.

Includes more tests for tag growth and shrinkage, and fixes a (mostly) unrelated bug in `test_mempolicy()` which was not growing records nearly as much as expected.

### DIFF
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -1,7 +1,7 @@
 /// @file htslib/sam.h
 /// High-level SAM/BAM/CRAM sequence file operations.
 /*
-    Copyright (C) 2008, 2009, 2013-2019 Genome Research Ltd.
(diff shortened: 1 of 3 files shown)


daviesrob:
Rebased with updated copyright dates.

PR #1089, opened 2020-06-24
Support for simultaneous reading of unindexed VCF/BCF files.

pd3:
In order for this to work, the caller must ensure that the input
files have chromosomes in the same order and consistent with the
order of sequences in the header.

Replaces #1076.


daviesrob:
Well, so far I've failed to fool the out-of-order detection logic, so I think this should be safe enough.  I did notice that it isn't worried about `POS` not in order within a chromosome.  It seems to work, albeit outputting out of order records, so presumably it's no more of a problem than for the single-file case?

Would you have any objection to me renaming `BCF_SR_REQUIRE_IDX_WARN` to something like `BCF_SR_NO_INDEX_WARN`?  The current symbol seems a little wrong as it's describing the opposite of what's being asked for.  I can handle the necesary changes if you're happy with that.

daviesrob:
Rebased with added tests.  `BCF_SR_REQUIRE_IDX_WARN` has been changed to `BCF_SR_ALLOW_NO_IDX` to better reflect what the option does.  I'll adjust the bcftools part to match when this has gone in.

PR #1090, opened 2020-06-24
Permit format string ubam and ucram for uncompressed data.

jkbonfield:
Eg samtools view -O ubam in.sam | ...

Getting this to handle file suffixes would be tricky.  Currently sam_open_mode parses filename extensions to turn .bam, .cram, .sam or .sam.gz into mode string b, c, "" or z, but supporting .ubam would need "b0" which means the caller has to handle longer mode strings. That's not true without changing other code, and we've never explained how long the mode should be as it was always liimted to 1 char in the past.

Addresses comments raised in samtools/samtools#1265

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -601,6 +601,12 @@ htsFile *hts_open_format(const char *fn, const char *mode, const htsFormat *fmt)
     }
     mode_c = cp2;
     *cp2++ = fmt_code;
+
+    // Permit ubam and ucram formats
+    if (fmt && fmt->compression_level >= 0 && fmt->compression_level <= 9
+        && (fmt->format == bam || fmt->format == cram))
+        *cp2++ = '0' + fmt->compression_level;
+
     *cp2++ = 0;
 
     // Set or reset the format code if opts->format is used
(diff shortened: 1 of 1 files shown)


dkj:
Not a blocker, but note that I think some people use `uBAM` to mean _unaligned_ BAM.

daviesrob:
@dkj just mentioned that the term `ubam` is used in GATK for [unmapped bam files](https://gatk.broadinstitute.org/hc/en-us/articles/360035532132-uBAM-Unmapped-BAM-Format).

jkbonfield:
Duh, well in that case it's going to cause more problems than it solves.  Maybe just stick with the -u options that we've already implemented.

jkbonfield:
I was also pondering about bam0 to bam9 and setting the compression level accordingly, as a synomym for the existing bam,level=0 to bam,level=9.  But I'm not sure we really *need* synonyms when we have a single letter option to do the same thing anyway.  I guess it's more about other tools linking against htslib and whether they want this.  Although it didn't come as a request from there.

jkbonfield:
Closing as it's likely to cause confusion and isn't hugely beneficial.

PR #1091, opened 2020-06-25
 Fix dirty default build by including latest pkg.m4 instead of using aclocal.m4

zamaudio:
~~The aclocal.m4 file can be regenerated with `autoreconf -fi` and is required to be run anyway to get a `configure` script.~~

With this change, I can build a non-dirty git build of htslib directly from the tree.


jmarshall:
Looks like a good rendition of option 2. It would be good to rebase this against the **develop** branch, and also to remove `aclocal` from _.appveyor.yml_ (as that will catch other items sneaking in in future requiring aclocal).

jkbonfield:
Hallelujah!  This has been bugging me for years. :)

Thanks.

zamaudio:
@jmarshall I have made the requested changes.

daviesrob:
It looks like it keeps `aclocal.m4` from coming back.  Thanks for the update.

PR #1092, opened 2020-06-25
Add more checks for broken BCF files

daviesrob:
It is possible to make a sparse set of ids in the VCF header using IDX= annotations, which can cause problems
if a BCF record refers to one of the indexes that does not exist.  This adds checks to `bcf_record_check()` and `vcf_format()` to catch these cases and return gracefully instead of crashing.

Ideally the checks in `bcf_record_check()` would be enough, but it isn't able to access the header when called from an iterator so has to skip these tests in that case.  `vcf_format()` does have access to the header, so can at least stop itself from behaving badly.  Any code looking up these dictionary values between the two functions still needs to take similar precautions to ensure values in BCF records are valid, at least until we can work out how to make `bcf_record_check()` work better in the iterator case.

A future PR will add some more error checking to `vcf_format()`, this one has deliberately been limited to just catching these invalid index values.

Credit to OSS-Fuzz
Fixes oss-fuzz 23670

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -3022,8 +3022,16 @@ int bcf_unpack(bcf1_t *b, int which)
 int vcf_format(const bcf_hdr_t *h, const bcf1_t *v, kstring_t *s)
 {
     int i;
+    int32_t max_dt_id = h->n[BCF_DT_ID];
+    const char *chrom = bcf_seqname(h, v);
+    if (!chrom) {
+        hts_log_error("Invalid BCF, CONTIG id=%d not present in the header",
+                      v->rid);
+        errno = EINVAL;
+        return -1;
+    }
     bcf_unpack((bcf1_t*)v, BCF_UN_ALL);
-    kputs(h->id[BCF_DT_CTG][v->rid].key, s); // CHROM
+    kputs(chrom, s); // CHROM
     kputc('\t', s); kputll(v->pos + 1, s); // POS
     kputc('\t', s); kputs(v->d.id ? v->d.id : ".", s); // ID
     kputc('\t', s); // REF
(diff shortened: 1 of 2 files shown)


daviesrob:
Test files attached: [test_bcf_files.zip](https://github.com/samtools/htslib/files/4832514/test_bcf_files.zip)

`sparse_good.bcf` is the original file.  All the `*_bad.bcf` files have had `IDX=` values modified in the header to make the alignment record invalid.  To test without the index, use `./test/test_view sparse_chrom_bad.bcf`, etc.  To force use of the iterator interface, use `./test/test_view sparse_chrom_bad.bcf .` (Note the `.` at the end).

vcf.c
```diff
@@ -3016,8 +3022,16 @@ int bcf_unpack(bcf1_t *b, int which)
 int vcf_format(const bcf_hdr_t *h, const bcf1_t *v, kstring_t *s)
 {
     int i;
+    int32_t max_dt_id = h->n[BCF_DT_ID];
```
valeriuo, line 3025: In `bcf_record_check`, when initializing `max_id`, you checked first the header pointer for NULL value. Is there no need for that here?

daviesrob: If `bcf_record_check()` gets called via an iterator read there's no header available so it is passed in as NULL and the function needs to deal with that.  `vcf_format()` won't work at all without a header so it's a coding error if `h == NULL`.  I guess you could argue that it should either return an error code or abort in that case (and the same would be true for `v` and `s`).

valeriuo:
Merged into develop as 1d6b68263a58786b6ab88f579e2866ad9e13c17b to 2c49d1913b05e754e1f180fc64d187ba0baba1a8

PR #1093, opened 2020-06-25
[CRAM] incomplete decode to encode conversion

valeriuo:
Because of the missing `else`, `cram_codec_decoder2encoder` was erroneously returning **-1**, causing its caller method `cram_block_compression_hdr_decoder2encoder` to abort converting the other codecs from decoding to encoding. This resulted in uninitialized codec fields, which were subsequently used.

Fixes https://github.com/samtools/samtools/issues/1276


jkbonfield:
Looks like a good catch for a trivial error.  That code was added with the 64-bit position support, which explains why it broke between 1.9 and 1.10.  I'll merge once travis has finished testing.

PR #1096, opened 2020-06-30
Implement vcf_open_mode method

### LINKED ISSUE #1095: Add vcf_open_mode
Similar to `sam_open_mode`.

valeriuo:
Similar to `sam_open_mode` for variant call files.

Fixes #1095

### DIFF
--- a/hts_internal.h
+++ b/hts_internal.h
@@ -27,8 +27,9 @@ DEALINGS IN THE SOFTWARE.  */
 #include <ctype.h>
 
 #include "htslib/hts.h"
-
 #include "textutils_internal.h"
+
+#define HTS_MAX_EXT_LEN 8
 
 #ifdef __cplusplus
 extern "C" {
@@ -120,17 +121,19 @@ int bgzf_idx_push(BGZF *fp, hts_idx_t *hidx, int tid, hts_pos_t beg, hts_pos_t e
  */
 void bgzf_idx_amend_last(BGZF *fp, hts_idx_t *hidx, uint64_t offset);
 
-static inline int find_file_extension(const char *fn, char ext_out[5])
+static inline int find_file_extension(const char *fn, char ext_out[static HTS_MAX_EXT_LEN])
 {
     const char *delim = fn ? strstr(fn, HTS_IDX_DELIM) : NULL, *ext;
     if (!fn) return -1;
     if (!delim) delim = fn + strlen(fn);
     for (ext = delim; ext > fn && *ext != '.' && *ext != '/'; --ext) {}
-    if (*ext == '.' && delim - ext == 3 && ext[1] == 'g' && ext[2] == 'z') {
-        // permit .sam.gz as a valid file extension
+    if (*ext == '.' &&
+        ((delim - ext == 3 && ext[1] == 'g' && ext[2] == 'z') || // permit .sam.gz as a valid file extension
+        (delim - ext == 4 && ext[1] == 'b' && ext[2] == 'g' && ext[3] == 'z'))) // permit .vcf.bgz as a valid file extension
+    {
         for (ext--; ext > fn && *ext != '.' && *ext != '/'; --ext) {}
     }
-    if (*ext != '.' || delim - ext > 7 || delim - ext < 4) return -1;
+    if (*ext != '.' || delim - ext > HTS_MAX_EXT_LEN || delim - ext < 4) return -1;
     memcpy(ext_out, ext + 1, delim - ext - 1);
     ext_out[delim - ext - 1] = '\0';
     return 0;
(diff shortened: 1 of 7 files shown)


vcf.c
```diff
@@ -2901,6 +2901,22 @@ int vcf_parse(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v)
...
+{
+    if (format == NULL) {
+        // Try to pick a format based on the filename extension
+        char extension[7];
+        if (find_file_extension(fn, extension) < 0) return -1;
+        return vcf_open_mode(mode, fn, extension);
+    }
+    else if (strcmp(format, "bcf") == 0) strcpy(mode, "b");
+    else if (strcmp(format, "vcf") == 0) strcpy(mode, "");
+    else if (strcmp(format, "vcf.gz") == 0) strcpy(mode, "z");
```
pd3, line 2914: Another extension often used is  `vcf.bgz` and `strcasecmp` to make the comparison case insensitive.

hts_internal.h
```diff
@@ -120,6 +120,22 @@ int bgzf_idx_push(BGZF *fp, hts_idx_t *hidx, int tid, hts_pos_t beg, hts_pos_t e
  */
 void bgzf_idx_amend_last(BGZF *fp, hts_idx_t *hidx, uint64_t offset);
 
+static inline int find_file_extension(const char *fn, char ext_out[5])
```
daviesrob, line 123: The array size here is incorrect.  It doesn't strictly matter because C treats `ext_out` as a `char *` and the array passed by the caller is big enough.  It is misleading though so should be corrected.  We could also use a C99 feature and turn it into `ext_out[static 7]` which means it's treated more like you'd expect (a `char` array with at least 7 elements).  `gcc` seems to ignore the declared size here, but `clang` will complain if it's bigger than the array passed in by the caller.

hts_internal.h
```diff
@@ -120,6 +120,22 @@ int bgzf_idx_push(BGZF *fp, hts_idx_t *hidx, int tid, hts_pos_t beg, hts_pos_t e
...
+{
+    const char *delim = fn ? strstr(fn, HTS_IDX_DELIM) : NULL, *ext;
+    if (!fn) return -1;
+    if (!delim) delim = fn + strlen(fn);
+    for (ext = delim; ext > fn && *ext != '.' && *ext != '/'; --ext) {}
+    if (*ext == '.' && delim - ext == 3 && ext[1] == 'g' && ext[2] == 'z') {
+        // permit .sam.gz as a valid file extension
```
daviesrob, line 130: There's no specific check for `.bgz` here, which has issues for `.vcf.bgz` detection.

If we add it as a specific case, `ext_out` would need to become `ext_out[static 8]` (or `.bgz` could be converted to `.gz` when filling `ext_out`).

vcf.c
```diff
@@ -2901,6 +2901,22 @@ int vcf_parse(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v)
...
+{
+    if (format == NULL) {
+        // Try to pick a format based on the filename extension
+        char extension[7];
+        if (find_file_extension(fn, extension) < 0) return -1;
+        return vcf_open_mode(mode, fn, extension);
+    }
+    else if (strcasecmp(format, "bcf") == 0) strcpy(mode, "b");
+    else if (strcasecmp(format, "vcf") == 0) strcpy(mode, "");
+    else if (strcasecmp(format, "vcf.gz") == 0 || strcasecmp(format, "vcf.bgz")) strcpy(mode, "z");
```
daviesrob, line 2914: Missing `== 0` for `vcf.bgz` here, which means the corresponding test in `test_open_format()` succeeds when it shouldn't due to missing `.bgz` detection in `find_file_extension()`.

test/test-vcf-api.c
```diff
@@ -527,6 +527,26 @@ void test_invalid_end_tag(void)
...
+        error("Mode '%s' does not match the expected value '%s'", mode, "r");
+    mode[1] = 0;
+    vcf_open_mode(mode+1, "mode1.vcf.gz", NULL);
+    if (strncmp(mode, "rz", 2))
+        error("Mode '%s' does not match the expected value '%s'", mode, "rz");
+    mode[1] = 0;
+    vcf_open_mode(mode+1, "mode1.vcf.bgz", NULL);
+    if (strncmp(mode, "rz", 2))
+        error("Mode '%s' does not match the expected value '%s'", mode, "rz");
+}
```
daviesrob, line 548: This could do with a test with an invalid (for vcf/bcf) extension to ensure `vcf_open_mode()` returns -1 in such a case.  See comments in `vcf_open_mode()`.

valeriuo:
Mended. Thanks for reviewing.

PR #1100, opened 2020-07-08
typo "Number=R" instead of "Number=G" in error msg

VorontsovIE:


daviesrob:
Typo is now fixed.  Thanks very much.

Issue #528, opened 2017-05-05
Memory issues in bcf_get_format_values (and other places that use realloc)


reedacartwright:
## 1. Memory leaks caused by failed `realloc` function.

`bcf_get_format_values` uses `realloc` internally to resize as needed the memory block pointed to by `*dst`.  E.g.

```C
        if ( *ndst < n )
        {
            *dst  = realloc(*dst, n);
            if ( !*dst ) return -4;     // could not alloc
            *ndst = n;
        }
```
https://github.com/samtools/htslib/blob/develop/vcf.c#L3765-L3767

The problem with this block is that if realloc fails, it returns a null pointer and leaves the original block untouched. See http://en.cppreference.com/w/c/memory/realloc. However, by assigning the return value of `realloc` to `*dst` the library and the caller of the function loses the original block, making it impossible to free unless a copy of the pointer was made before hand. The above block (and ones like it) should be replaced by

```C
        if ( *ndst < n )
        {
            void *newdst  = realloc(*dst, n);
            if ( !newdst ) return -4;     // could not alloc
            *dst = newdst;
            *ndst = n;
        }
```
This way, the original value of `*dst` is preserved if `realloc` fails allowing it to be freed by the caller.

## 2. Typo

I also spotted a typo at https://github.com/samtools/htslib/blob/develop/vcf.c#L3780.  This line should be `if ( !*dst ) return -4;`.

jkbonfield:
Good spot on that typo.

The realloc issue sadly turns up in a lot of places.  Eg a grep of realloc in htslib/cram/*.c I see 39 reallocs of which only 12 appear to assign to a new variable.  I'm guilty of those issues.  In htslib/*.c it's 51 reallocs with 11 assigning to new variables.  So it's definitely a common idiom.

If I recall ```cppcheck``` spots these and that's one of the things we want to work our way through these and other issues spotted.

reedacartwright:
Yeah, I figured that the typo would be the easiest to fix and the rest involves a massive update.

PR #1102, opened 2020-07-10
Fix minor issue 994: avoid bcf_update_alleles() crash

### LINKED ISSUE #994: very minor issue: bcf_update_alleles segmentation faults if nals<=0
In htslib/vcf.h the function bcf_update_alleles is defined as:
```
    /**
     *  bcf_update_alleles() and bcf_update_alleles_str() - update REF and ALLT column
     *  @alleles:           Array of alleles
     *  @nals:              Number of alleles
     *  @alleles_string:    Comma-separated alleles, starting with the REF allele
     */
    HTSLIB_EXPORT
    int bcf_update_alleles(const bcf_hdr_t *hdr, bcf1_t *line, const char **alleles, int nals);

    HTSLIB_EXPORT
    int bcf_update_alleles_str(const bcf_hdr_t *hdr, bcf1_t *line, const char *alleles_string);
```
It does not say that nals must be positive. If the function is run with nals=0 on a new bcf1_t record that was just cleared, segmentation fault ensues. I would clarify that nals has to be positive and add a sanity check in the vcf.c code. Also, ALT is misspelled as ALLT in the header file.

jmarshall:
Avoid crash for `bcf_update_alleles(hdr, rec, NULL, 0)`.

Updating a `bcf1_t` to have 0 alleles is not particularly meaningful, but it is a valid state as that's what bcf_init() and bcf_clear() produce. Hence it is a valid thing for bcf_update_alleles() to do too, so calling it with `nals==0` should not crash. Instead set `rlen` the same way that bcf_init() and bcf_clear() leave it. Fixes #994.

### DIFF
--- a/htslib/vcf.h
+++ b/htslib/vcf.h
@@ -765,7 +765,7 @@ set to one of BCF_ERR* codes and must be checked before calling bcf_write().
     HTSLIB_EXPORT
     int bcf_has_filter(const bcf_hdr_t *hdr, bcf1_t *line, char *filter);
     /**
-     *  bcf_update_alleles() and bcf_update_alleles_str() - update REF and ALLT column
+     *  bcf_update_alleles() and bcf_update_alleles_str() - update REF and ALT column
      *  @alleles:           Array of alleles
      *  @nals:              Number of alleles
      *  @alleles_string:    Comma-separated alleles, starting with the REF allele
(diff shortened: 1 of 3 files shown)


jmarshall:
Updated to also fix the `bcf_get_format_values()` error checking typo noted in #528.

PR #1104, opened 2020-07-13
Fix check for VCF record size

daviesrob:
The check for excessive record size in vcf_parse_format() only looked at individual fields.  It was therefore possible to exceed the limit and overflow fmt_aux_t::offset by having multiple fields with a combined size that went over INT_MAX.
Fix by including the amount of memory used so far in the check.

Credit to OSS-Fuzz
Fixes oss-fuzz 24097


valeriuo:
Merged into develop as dcd4b7304941a8832fba2d0fc4c1e716e7a4e72c

PR #1105, opened 2020-07-14
Fix hfile_libcurl breakage when using libcurl 7.69.1 or later

daviesrob:
Curl update curl/curl#5050 (pause: bail out on bad input) made curl_easy_pause() return an error if passed
a disconnected CURL handle.  In hfile_libcurl this could happen when libcurl_read() or libcurl_close() was called after all
bytes had been read; or all the time in restart_from_position().

Fix by checking for fp->finished in libcurl_read() and libcurl_close(); and by removing the curl_easy_pause() in restart_from_position().

Thanks to @jmarshall for tracking down the exact libcurl change that caused the incompatibility.

Fixes samtools/samtools#1254 (Trouble reading SRA CRAM files in Google Cloud buckets)
Fixes samtools/samtools#1284 (samtools view crashes reading from amazon s3 http url?)

### DIFF
--- a/hfile_libcurl.c
+++ b/hfile_libcurl.c
@@ -221,6 +221,8 @@ static int easy_errno(CURL *easy, CURLcode err)
         return EEXIST;
 
     default:
+        hts_log_error("Libcurl reported error %d (%s)", (int) err,
+                      curl_easy_strerror(err));
         return EIO;
     }
 }
@@ -241,6 +243,8 @@ static int multi_errno(CURLMcode errm)
         return ENOMEM;
 
     default:
+        hts_log_error("Libcurl reported error %d (%s)", (int) errm,
+                      curl_multi_strerror(errm));
         return EIO;
     }
 }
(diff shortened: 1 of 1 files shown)


hfile_libcurl.c
```diff
@@ -818,8 +822,13 @@ static ssize_t libcurl_read(hFILE *fpv, void *bufferv, size_t nbytes)
         fp->buffer.ptr.rd = buffer;
         fp->buffer.len = nbytes;
         fp->paused = 0;
-        err = curl_easy_pause(fp->easy, CURLPAUSE_CONT);
-        if (err != CURLE_OK) { errno = easy_errno(fp->easy, err); return -1; }
+        if (!fp->finished) {
```
jmarshall, line 825: Local style is `if (! fp->finished)`

valeriuo:
It fixes https://github.com/samtools/samtools/issues/1284 on a Windows machine with libcurl 7.71.0 installed.

jmarshall:
LGTM (though I didn't look at the `restart_from_position()` part in any detail).

Thought about suggesting adding a parameter to `easy_errno()` so its verbose logging can say which libcurl function reported the error, but it doesn't matter much as it's mostly always `curl_easy_pause()` anyway.

PR #1108, opened 2020-07-16
[tabix] Add --separate-regions option

### LINKED ISSUE #1107: tabix multipule regions with delimited between output regions #30
I was not sure where to post my question.. so the question is [here](https://github.com/samtools/tabix/issues/30)

Hi,

Say I have a bed file, `file.bed`, and I want to retrieve multipule regions.
file.bed:

```
chr1    10468   10469   
chr1    10470   10471   
chr1    10483   10484   
chr1    10488   10489   
chr1    10492   10493   
chr1    10496   10497 
```
running `tabix file.bed chr1:10468-10490 chr1:10478-10490 chr1:10496-10497` gives me all the rows that matches region in the file.bed:

```
chr1    10468   10469  
chr1    10470   10471  
chr1    10483   10484  
chr1    10488   10489  
chr1    10483   10484  
chr1    10488   10489 
chr1    10496   10497 
```
Can I have an option to separate between the results, i.e. to define which rows came from each region? Something like:

```
chr1    10468   10469  
chr1    10470   10471  
chr1    10483   10484  
chr1    10488   10489
  
chr1    10483   10484  
chr1    10488   10489 

chr1    10496   10497 
```

valeriuo:
Adds `tabix --separate-regions` option, which lets the user split the output by
the number of target regions supplied in the command line.

Fixes #1107


jmarshall:
_Split_ tends to imply writing the output to multiple separate output files (cf `samtools split`), so I'd suggest calling the option `--separate-regions` or similar (that's supposed to be ‘separate’ as a verb, but it's hard to convey…).

No need to put `[off]` for a flag option, and it would be good to shorten the usage description to fit in 80 columns.

Just looking at the code, it looks like this'll put an extra blank line right at the end too?

valeriuo:
I went with `--separate-regs`, as we already use `--list-chroms`.

> Just looking at the code, it looks like this'll put an extra blank line right at the end too?

Fixed. Thanks for reviewing.

jmarshall:
There's also already `--regions`, so abbreviating to `regs` in an option name would be a departure from existing tabix practice… ¯\\\_(ツ)\_/¯

PR #1112, opened 2020-07-27
Download index files atomically in idx_test_and_fetch()

jmarshall:
In https://github.com/samtools/samtools/issues/1242#issuecomment-627386086, @valeriuo wrote:

> A footnote: when running parallel jobs on a S3 file, it is advisable to download the index file in the current folder a priori, otherwise each job might try to do that independently, resulting in a corrupt index file.

This is a pretty tragic bug in the index downloading — instead it should use the same technique as the CRAM reference code does to download into a unique temporary filename and atomically `rename()` the file into place at the end.

This PR extracts filename uniquification into a new `hts_open_tmpfile()` function declared in _hts_internal.h_ and uses it in the two places where files are currently downloaded and saved as a byproduct of htslib operations. (Are there any others?)

No code in _hts.c_ currently uses pthreads directly, so instead of mixing `pthread_self()` into `tmpname`, mix in the pointer value of one of the arguments — which will vary across threads. Trick inspired by musl's `mkstemp()` implementation.

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -2433,16 +2433,13 @@ static int cram_populate_ref(cram_fd *fd, int id, ref_entry *r) {
             return -1;
         }
 
-        if (hwrite(fp, r->seq, r->length) != r->length) {
-            perror(path);
-        }
-        if (hclose(fp) < 0) {
+        ssize_t length_written = hwrite(fp, r->seq, r->length);
+        if (hclose(fp) < 0 || length_written != r->length ||
+            chmod(path_tmp.s, 0444) < 0 ||
+            rename(path_tmp.s, path) < 0) {
+            hts_log_error("Creating reference at %s failed: %s",
+                          path, strerror(errno));
             unlink(path_tmp.s);
-        } else {
-            if (0 == chmod(path_tmp.s, 0444))
-                rename(path_tmp.s, path);
-            else
-                unlink(path_tmp.s);
         }
     }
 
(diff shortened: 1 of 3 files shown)


cram/cram_io.c
```diff
@@ -2449,23 +2428,25 @@ static int cram_populate_ref(cram_fd *fd, int id, ref_entry *r) {
...
             perror(path);
         }
         if (hclose(fp) < 0) {
-            unlink(path_tmp);
+            unlink(path_tmp.s);
         } else {
-            if (0 == chmod(path_tmp, 0444))
-                rename(path_tmp, path);
+            if (0 == chmod(path_tmp.s, 0444))
+                rename(path_tmp.s, path);
```
valeriuo, line 2443: Can you also check for `rename()` error here and unlink the temp file in that case? Probably all three conditions `hclose(fp) < 0 || chmod(path_tmp.s, 0444) != 0 || rename(path_tmp.s, path) != 0` could be grouped into a single check resulting in `unlink(path_tmp.s)`.

jmarshall: It's a pre-existing buglet, but sure — since I touched the code I guess I get to fix it :smile:

hts.c
```diff
@@ -1394,6 +1395,25 @@ int hts_set_fai_filename(htsFile *fp, const char *fn_aux)
...
+    unsigned ptr = (uintptr_t) tmpname;
+    int n = 0;
+    hFILE *fp;
+
+    do {
+        // Attempt to further uniquify the temporary filename
+        unsigned t = ((unsigned) time(NULL)) ^ ((unsigned) clock()) ^ ptr;
+        n++;
+
+        ksprintf(ks_clear(tmpname), "%s.tmp_%d_%d_%u", fname, pid, n, t);
```
daviesrob, line 1410: A check on `ksprintf()` would be useful here, to prevent the possibility of passing NULL to `hopen()`.

daviesrob:
Thanks, hopefully we'll have fewer problems with index downloads from now on.

PR #1115, opened 2020-07-31
Add struct tags to htslib/*.h public typedefs

### LINKED ISSUE #1106: Enable forward declarations
Hi there,

I am working in C++, and want to avoid including htslib headers in my headers, so that the global namespace of my library doesn't get polluted with other library's symbols. However, to be able to use htslib data in my classes, I then would at least need to forward declare them. Unfortunately, the htslib structs are all declare as typdefs of anonymous structs, such as 

    typedef struct {
        // ...
    } htsFile;

These cannot be forward declared (see [here](https://stackoverflow.com/questions/10249548/forward-declaring-a-typedef-of-an-unnamed-struct) and [here](https://stackoverflow.com/questions/38874235/forward-declaring-anonymous-structure-from-c-in-c) for example). Hence, I suggest to add names to these structs, such as

    typedef struct htsFile_s {
        // ...
    } htsFile;

As far as I can tell, this would not change anything else, and would not destroy any compatibilities. But it would make integration into C++ way nicer. (Even better would be to somehow offer namespaces if compiled with C++, but let's go one step at a time).

Is that something that would be possible? I could write a PR if needed.

Best
Lucas

jmarshall:
This PR adds struct tags so it's possible to write e.g. `struct htsFile` to forward declare htsFile as an incomplete struct in circumstances when you'd rather not `#include <htslib/hts.h>`, thus fixes #1106.

Presented as separate commits in vaguely increasing order of theoretical API compatibility issues:

* Adds tags where there weren't any before — no issues.

* Removes `aux_key_t` which was accidentally added and never used in HTSlib.

* Renames reserved-to-the-compiler tags such as `typedef struct __bgzidx_t bgzidx_t` — third party code could conceivably have been using these tags but as they start with underscores users should have …uh… been expecting undefined results.

* Renames `variant_t` to `bcf_variant_t`. This struct is used only as the type of a field within `bcf_dec_t`, which itself is an internal field of `bcf1_t` and is unlikely to be used directly. So user code is very unlikely to be using this identifier directly.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -167,7 +167,7 @@ typedef struct
 }
 bgzidx1_t;
 
-struct __bgzidx_t
+struct bgzidx_t
 {
     int noffs, moffs;       // the size of the index, n:used, m:allocated
     bgzidx1_t *offs;        // offsets
--- a/faidx.c
+++ b/faidx.c
@@ -52,7 +52,7 @@ typedef struct {
 } faidx1_t;
 KHASH_MAP_INIT_STR(s, faidx1_t)
 
-struct __faidx_t {
+struct faidx_t {
     BGZF *bgzf;
     int n, m;
     char **name;
--- a/hts.c
+++ b/hts.c
@@ -1645,7 +1645,7 @@ typedef struct {
     uint64_t *offset;
 } lidx_t;
 
-struct __hts_idx_t {
+struct hts_idx_t {
     int fmt, min_shift, n_lvls, n_bins;
     uint32_t l_meta;
     int32_t n, m;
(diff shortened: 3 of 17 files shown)


jmarshall:
While in the area, clarify that `htsFile` should be considered opaque by end-user code — there has long been a note here, but unfortunately it was aimed at maintainers and not explicit enough for HTSlib API users.

daviesrob:
Thanks for this.  I've hunted around for packages that reference the changed symbols.  While a few third-party projects (mostly language bindings) mention them, they seem to build happily enough with the modified names.  The only package I've found that failed to build is [MethylDackel](https://github.com/dpryan79/MethylDackel), and as that isn't compatible with HTSlib 1.10 yet (it builds but then segfaults due to differing struct layouts) I don't think it's a problem.  So it looks like we can safely get away with merging all of the changes.

PR #1116, opened 2020-08-04
Add missing Makefile dependency [minor]

jmarshall:
This was missed during PR #1035.

It's still a back burner project to update my script for checking these to deal with the `../` added in #1078, but for now: also verified that all necessary `#include "../etc"` relative paths are correct in htslib source code; and also that there's no missing dependencies in current **develop** on samtools or bcftools.


daviesrob:
Thanks.  Is there any chance of getting your script into, for example, `test/maintainer`?

jmarshall:
I have a draft with it in `make maintainer-check` as _test/maintainer/check_dependencies.pl_ but there's some updates needed to make it usable… I was planning to leave it on the back burner until after the upcoming release…?

daviesrob:
I don't think there's any hurry, but it would certainly be useful to have.

PR #1117, opened 2020-08-04
Sam threading improvements

jkbonfield:
May or may not solve the problems in https://github.com/samtools/samtools/issues/1293.  We can't reproduce that exact scenario, but have found others and it's unclear if we're fixing the same problem or not.

The issues fixed here are mainly for error recovery; what happens if we fail parsing due to malformed SAM or receive and external source of error (I/O, memory, etc).  Various deadlocks and/or race conditions could occur, sometimes leading to crashes too.

### DIFF
--- a/hfile.c
+++ b/hfile.c
@@ -1005,6 +1005,11 @@ void hfile_add_scheme_handler(const char *scheme,
         return;
     }
     khint_t k = kh_put(scheme_string, schemes, scheme, &absent);
+    if (absent < 0) {
+        hts_log_warning("Couldn't register scheme handler for %s : %s",
+                        scheme, strerror(errno));
+        return;
+    }
     if (absent || priority(handler) > priority(kh_value(schemes, k))) {
         kh_value(schemes, k) = handler;
     }
@@ -1014,7 +1019,10 @@ static int init_add_plugin(void *obj, int (*init)(struct hFILE_plugin *),
                            const char *pluginname)
 {
     struct hFILE_plugin_list *p = malloc (sizeof (struct hFILE_plugin_list));
-    if (p == NULL) abort();
+    if (p == NULL) {
+        hts_log_debug("Failed to allocate memory for plugin \"%s\"", pluginname);
+        return -1;
+    }
 
     p->plugin.api_version = 1;
     p->plugin.obj = obj;
@@ -1035,7 +1043,11 @@ static int init_add_plugin(void *obj, int (*init)(struct hFILE_plugin *),
     return 0;
 }
 
-static void load_hfile_plugins()
+/*
+ * Returns 0 on success,
+ *        <0 on failure
+ */
+static int load_hfile_plugins()
 {
     static const struct hFILE_scheme_handler
         data = { hopen_mem, hfile_always_local, "built-in", 80 },
@@ -1043,7 +1055,8 @@ static void load_hfile_plugins()
         preload = { hopen_preload, is_preload_url_remote, "built-in", 80 };
 
     schemes = kh_init(scheme_string);
-    if (schemes == NULL) abort();
+    if (schemes == NULL)
+        return -1;
 
     hfile_add_scheme_handler("data", &data);
     hfile_add_scheme_handler("file", &file);
@@ -1085,6 +1098,8 @@ static void load_hfile_plugins()
     // carry on; then eventually when the program exits, we'll merely close
     // down the plugins uncleanly, as if we had aborted.
     (void) atexit(hfile_exit);
+
+    return 0;
 }
 
 /* A filename like "foo:bar" in which we don't recognise the scheme is
@@ -1118,7 +1133,10 @@ static const struct hFILE_scheme_handler *find_scheme_handler(const char *s)
     scheme[i] = '\0';
 
     pthread_mutex_lock(&plugins_lock);
-    if (! schemes) load_hfile_plugins();
+    if (!schemes && load_hfile_plugins() < 0) {
+        pthread_mutex_unlock(&plugins_lock);
+        return NULL;
+    }
     pthread_mutex_unlock(&plugins_lock);
 
     khint_t k = kh_get(scheme_string, schemes, scheme);
--- a/htslib/hts.h
+++ b/htslib/hts.h
@@ -537,8 +537,17 @@ const char *hts_format_file_extension(const htsFormat *format);
 HTSLIB_EXPORT
 int hts_set_opt(htsFile *fp, enum hts_fmt_option opt, ...);
 
+/*!
+  @abstract         Read a line (and its \n or \r\n terminator) from a file
+  @param fp         The file handle
+  @param delimiter  Unused, but must be '\n' (or KS_SEP_LINE)
+  @param str        The line (not including the terminator) is written here
+  @return           Length of the string read;
+                    -1 on end-of-file; <= -2 on error
+*/
 HTSLIB_EXPORT
 int hts_getline(htsFile *fp, int delimiter, kstring_t *str);
+
 HTSLIB_EXPORT
 char **hts_readlines(const char *fn, int *_n);
 /*!
--- a/sam.c
+++ b/sam.c
@@ -2239,6 +2239,7 @@ typedef struct sp_lines {
 enum sam_cmd {
     SAM_NONE = 0,
     SAM_CLOSE,
+    SAM_CLOSE_DONE,
 };
 
 typedef struct SAM_state {
@@ -2327,17 +2328,31 @@ int sam_state_destroy(htsFile *fp) {
         if (fd->h) {
             // Notify sam_dispatcher we're closing
             pthread_mutex_lock(&fd->command_m);
-            fd->command = SAM_CLOSE;
+            if (fd->command != SAM_CLOSE_DONE)
+                fd->command = SAM_CLOSE;
             pthread_cond_signal(&fd->command_c);
             ret = -fd->errcode;
-            if (!ret) hts_tpool_wake_dispatch(fd->q); // unstick the reader
+            if (fd->q)
+                hts_tpool_wake_dispatch(fd->q); // unstick the reader
+
+            if (!fp->is_write && fd->q && fd->dispatcher) {
+                for (;;) {
+                    // Avoid deadlocks with dispatcher
+                    if (fd->command == SAM_CLOSE_DONE)
+                        break;
+                    hts_tpool_wake_dispatch(fd->q);
+                    pthread_mutex_unlock(&fd->command_m);
+                    usleep(10000);
+                    pthread_mutex_lock(&fd->command_m);
+                }
+            }
             pthread_mutex_unlock(&fd->command_m);
 
             if (fp->is_write) {
                 // Dispatch the last partial block.
                 sp_bams *gb = fd->curr_bam;
                 if (!ret && gb && gb->nbams > 0)
-                  ret = hts_tpool_dispatch(fd->p, fd->q, sam_format_worker, gb);
+                    ret = hts_tpool_dispatch(fd->p, fd->q, sam_format_worker, gb);
 
                 // Flush and drain output
                 hts_tpool_process_flush(fd->q);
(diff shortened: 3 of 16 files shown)


hfile.c
```diff
@@ -1035,15 +1035,20 @@ static int init_add_plugin(void *obj, int (*init)(struct hFILE_plugin *),
...
+static int load_hfile_plugins()
 {
     static const struct hFILE_scheme_handler
         data = { hopen_mem, hfile_always_local, "built-in", 80 },
         file = { hopen_fd_fileuri, hfile_always_local, "built-in", 80 },
         preload = { hopen_preload, is_preload_url_remote, "built-in", 80 };
 
     schemes = kh_init(scheme_string);
-    if (schemes == NULL) abort();
+    if (schemes == NULL)
```
jmarshall, line 1050: (At the time, I judged there was little point in extreme malloc error handling here.) This change is incomplete without also checking the `kh_put` in `hfile_add_scheme_handler()`.

htslib/hts.h
```diff
@@ -538,6 +538,15 @@ HTSLIB_EXPORT
 int hts_set_opt(htsFile *fp, enum hts_fmt_option opt, ...);
 
 HTSLIB_EXPORT
+/*!
+    @abstract        Reads a line of text
+    @param fp        htsFile pointer
+    @param delimiter Line separator char, or KS_SEP_* values (see kseq.h)
```
jmarshall, line 544: Since 71c03b88edf892a93eae7bb7932a6da84858480a `hts_getline` only works on actual lines, so `delimiter` must be literally `'\n'` or `KS_SEP_LINE` — see the TODO in that commit message.

(`HTSLIB_EXPORT` goes after the comment.)

hfile.c
jkbonfield: I'm aware there are many more things to fix, such as all the khash and kstring checks.  However my strategy is as follows (ignoring threading for this argument), using a malloc fuzzer that fails on the Nth allocation.

1. Count up from 1 to X to find the value of X where the exit code is not 1.  It should be zero, but some also crashed.
2. Count from X upwards to validate that all subsequent exist codes should be 0.  Again, I found cases where this failed as we erroneously returned success when we'd infact failed.

Now surprisingly for 2, this doesn't happen often, even on VCF which is riddled with kstring calls without checks.  The reason being that for all the missed errors, it eventually does something which does check and so blows up a bit later.  Hence the number of places to fix was small.

This was one example which made my testing logic problematic, so I fixed it so I could continue with the other tests.  Step in the right direction and all that.

FWIW this PR makes things considerably more stable in memory tight situations, but so far only for reading (both single and multi-threaded).  I've just started on the writing side.  Why do I care?  Because we very occasionally get baffling bugs reported from people who claim to have dozens of GB of memory available, only to discover they're using a batch submission system which limited the memory.  Ideally the tools should be reporting out of memory too, and they do now more often than before, but that's still not universally true.  (It also turns out the bug we were hunting this time wasn't due to that, but it's still helpful to discount that.)

htslib/hts.h
jkbonfield: I struggled writing the documentation for this as the operation of it is unclear and, obviously, there's no documentation!  Even the return values were pretty vague and confusing.

Please make a suggestion for revised text as I expect you know what it's meant to be doing far more than I do.

jmarshall: It was almost four years ago :smile: but here you go:

```
  @abstract         Read a line (and its \n or \r\n terminator) from a file
  @param fp         The file handle
  @param delimiter  Unused, but must be '\n' (or KS_SEP_LINE)
  @param str        The line (not including the terminator) is written here
  @return           Length of the string read; -1 on end-of-file; <= -2 on error
```

jkbonfield: Well it was worth asking - I didn't know it's actually unused.  Thanks will update soon.  Working on some write side checks atm.

jkbonfield:
Rebased, squashed and shuffled it a bit.

The bug believed to trigger samtools/samtools#1293 is now the first commit.  This is followed by a bunch of SAM threading fixes, then generic (all-format) decoding bug fixes found through the malloc fuzzer, and a all-format encoding bug fixes via a similar route.

I can't say there aren't more lurking bugs and I know there are still some minor memory leaks during the error recovery stage, but I haven't focused on those as tidy exit (no possible crashes or deadlocks) is the primary concern and neatness second; for now.

sam.c
```diff
@@ -2321,31 +2326,50 @@ int sam_state_destroy(htsFile *fp) {
...
                 pthread_mutex_unlock(&fd->command_m);
 
-                while (!ret && !hts_tpool_process_empty(fd->q)) {
+                while (!ret && fd->q && !hts_tpool_process_empty(fd->q)) {
                     usleep(10000);
                     pthread_mutex_lock(&fd->command_m);
-                    if (!ret) ret = -fd->errcode;
+                    ret = -fd->errcode;
+                    // not empty but shutdown implies error
+                    if (hts_tpool_process_is_shutdown(fd->q) && !ret)
```
daviesrob, line 2367: I'm not sure if I understand this bit.  `sam_dispatcher_write()` calls `hts_tpool_process_shutdown()` [on success](https://github.com/samtools/htslib/blob/d99bf404/sam.c#L2792).  That suggests there may be a race between the queue becoming empty, `sam_dispatcher_write()` shutting the queue down and the shutdown check here.  Given that there's also a `usleep()`, I'm surprised that this doesn't always lose.  I think another `!hts_tpool_empty(fd->q)` check  after finding the shutdown state would be needed to ensure that it really was caused by an error.

thread_pool.c
```diff
@@ -874,13 +905,35 @@ int hts_tpool_process_flush(hts_tpool_process *q) {
...
+    // Wait for n_input and n_processing to hit zero.
+    while (!q->shutdown && (q->n_input || q->n_processing)) {
+        struct timeval now;
+        struct timespec timeout;
+
+        while (q->n_input && !q->shutdown) {
+            gettimeofday(&now, NULL);
+            timeout.tv_sec = now.tv_sec + 1;
+            timeout.tv_nsec = now.tv_usec * 1000;
+            pthread_cond_timedwait(&q->input_empty_c, &p->pool_m, &timeout);
```
daviesrob, line 924: Is this using timedwait to avoid a lock-up found in testing, or is this just a belt-and-braces approach?

jkbonfield: If I recall it was something I found while testing.  However given multiple rounds of testing and fixing, that doesn't necessarily mean it is still needed.  Probably, but with several subsequent changes it's hard to be sure without removing and testing again.

sam.c
daviesrob: The loop is there to wait for the IO thread (in this case `sam_dispatcher_write()`) to finish.  The flush should have ensured there's no new input and no remaining jobs running, but there may be some output waiting to be written.

Looking at it again, this probably is OK.  The flush will stop new data being added, but the queue is still intact and the writer will patiently wait for more in `hts_tpool_next_result_wait()` until this loop finishes and `hts_tpool_process_shutdown()` [gets called](https://github.com/samtools/htslib/blob/d99bf404/sam.c#L2372).  I suspect the [`hts_tpool_process_shutdown()` call in `sam_dispatcher_write()`](https://github.com/samtools/htslib/blob/d99bf404/sam.c#L2792) isn't actually needed because the queue will already be shut down by the time it gets there.

vcf.c
```diff
@@ -818,6 +819,14 @@ int bcf_hdr_parse(bcf_hdr_t *hdr, char *htxt)
             }
             p += len;
         }
+        if (hrec == NULL && len != 0) {
+            char *eol = strchr(p, '\n');
+            if (*p != '\0') {
+                hts_log_warning("Could not parse header line: %.*s",
+                                eol ? (int)(eol - p) : INT_MAX, p);
+            }
+            return -1;
```
daviesrob, line 828: I just noticed that this changes the behaviour described in the [comment](https://github.com/samtools/htslib/blob/d99bf404/vcf.c#L831) below.  Does this mean the future has arrived?  I fear we may get complaints about suddenly unreadable files again.

Would it be better to only give up on the out-of-memory case (now indicated by `len == -1`) and just skip the bad lines as before?

jkbonfield: Ugh, so it's deliberate that it spams a million errors and ignores all of them?

*splutter*

I guess we could revert that bit then if needed.  I think it was in its own commit anyway.

daviesrob: Apparently so.  I will make the necessary adjustments.

jkbonfield: Thanks

daviesrob:
I've added some small adjustments.  Hopefully there are no regressions from these updates.  Comments noted so far should be fixed.  VCF should now only complain once about each malformed header line.  I've strengthened the error clean-up code in some cases where it was inconsistent or incomplete.  I also fixed a couple of new minor problems I found when testing.

There are still problems with CRAM returning success when it should have failed because it incorrectly thinks it's at EOF.  I suggest we fix these in a new pull request as this one is already rather complicated and as a result has taken a long time to review.

I will squash some of the smaller commits added on merge.

jkbonfield:
If we really care about twos-complement being implementation defined behaviour, then note this is definitely not the only place we make that assumption or indeed use that idiom.  Eg see `sam_format1_append`.

daviesrob:
Yes, the code in `sam_format1_append` is currently implementation-defined.  It also isn't altered by this pull request, so I'm not going to do anything about it here.

vcf.c
```diff
@@ -819,13 +829,17 @@ int bcf_hdr_parse(bcf_hdr_t *hdr, char *htxt)
...
-        if (hrec == NULL && len != 0) {
-            char *eol = strchr(p, '\n');
-            if (*p != '\0') {
-                hts_log_warning("Could not parse header line: %.*s",
-                                eol ? (int)(eol - p) : INT_MAX, p);
-            }
+        assert(hrec == NULL);
+        if (len < 0) {
+            // len < 0 indicates out-of-memory, or similar error
+            hts_log_error("Could not parse header line: %s", strerror(errno));
```
jkbonfield, line 835: Do we no longer wish to warn on the contents of the line that we couldn't parse?  Admittedly the `eol ? (int)(eol - p) : INT_MAX` code is a bit of a bizarrity in style, but that came from b8c6f7dd5581f4b7fcf0a997bfa80763a7ee9026.

I think it's still useful to be able to see the offending line and don't see why we'd want to hide that.  Admittedly it'll now only ever be the first line instead of every single line, but that's fair enough and usually the first one is sufficient.

daviesrob: In this case (`len < 0`) the failure would be due to something like out-of-memory, so my reasoning was that the contents of the header line aren't likely to be relevant - unless it happens to be incredibly long.  All other types of dodgy line should now be printed out once only, with unprintable characters escaped and a reasonable cap on the length.

jkbonfield: Ah yes, I see they're still printed elsewhere, and if this is just malloc error then fair enough.  strerror will be more useful (provided errno has indeed been set, which it will if malloc failure, and isn't a left-over remnant of something else).

I'm not sure 320 chars is long enough for some valid VCF header lines though (but INT_MAX was a tad over the top!).

However checking a few local files the longest pre-CHROM line I see is 311 (so pretty close), and the longest CHROM is obviously unlimited, with ~10Kb being the longest in the files I have locally.

jkbonfield:
Most of the new commits here are simply further continuation of the work to check for errors, leaks, etc and very little to do with threading per-se.

The most significant change I see is to `hts_tpool_init` and eyeballing it the cleanup looks correct.  I validated it by deliberately failing after 10 threads and confirm that it cleans up the 10 it's created only, as expected.  So thumbs up from me on the changes.

vcf.c
daviesrob: Most of the VCF header lines I've seen are fairly short.  What was in the 311 character long one?

The `#CHROM` line only gets a very basic test in `bcf_hdr_parse()` itself, with most checks delegated to `bcf_hdr_parse_sample_line()` later.  If the basic test fails, you've likely either started a header with `#` instead of `##`; spelled `CHROM` or `POS` incorrectly; or forgotten the `#CHROM` line completely.  Either way a small amount of context should be enough to see where the problem is.

jkbonfield: It was a `##INFO` tag with a length comment:

```
##INFO=<ID=CIGAR,Number=A,Type=String,Description="The extended CIGAR representation of each alternate allele, with the exception that '=' is replaced by 'M' to ease VCF parsing.  Note that INDEL alleles do not have the first matched base (which is provided by default, per the spec) referred to by the CIGAR.">
```

Because descriptions can be any length, it's hard to know what's a sensible limit.

daviesrob: Yes, I think it might be considered a feature that the warning messages get a length cap.

daviesrob:
Rebased and somewhat squashed.

PR #1118, opened 2020-08-05
Don't trash errno in bam_tag2cigar.

jkbonfield:
bam_aux_get sets errno=ENOENT when we look for a tag and don't find it.  Personally I think this is a misfeature, but it's part of the
documented API so we can't change that. Instead we undo the damage as not finding the tag is not an error.

The consequence of samtools/samtools#1296 is that now all almost errors found while decoding a BAM file get converted to "No such file or directory" prior to this commit.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -522,7 +522,13 @@ static int bam_tag2cigar(bam1_t *b, int recal_bin, int give_warning) // return 0
     cigar0 = bam_get_cigar(b);
     if (bam_cigar_op(cigar0[0]) != BAM_CSOFT_CLIP || bam_cigar_oplen(cigar0[0]) != c->l_qseq) return 0;
     fake_bytes = c->n_cigar * 4;
-    if ((CG = bam_aux_get(b, "CG")) == 0) return 0; // no CG tag
+    int saved_errno = errno;
+    CG = bam_aux_get(b, "CG");
+    if (!CG) {
+        if (errno != ENOENT) return -1;  // Bad aux data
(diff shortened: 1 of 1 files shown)


sam.c
```diff
@@ -522,7 +522,10 @@ static int bam_tag2cigar(bam1_t *b, int recal_bin, int give_warning) // return 0
     cigar0 = bam_get_cigar(b);
     if (bam_cigar_op(cigar0[0]) != BAM_CSOFT_CLIP || bam_cigar_oplen(cigar0[0]) != c->l_qseq) return 0;
     fake_bytes = c->n_cigar * 4;
-    if ((CG = bam_aux_get(b, "CG")) == 0) return 0; // no CG tag
+    int saved_errno = errno;
+    CG = bam_aux_get(b, "CG");
+    errno = saved_errno;
```
daviesrob, line 527: This should really test for and propagate the `errno == EINVAL` case when `GC == NULL`, instead of glossing over it.

jkbonfield: Ok, something like

```
if (errno == ENOENT) errno = saved_errno;
```

I'll amend when I'm finished with my current other PR, or make the trivial change during merge if you wish.

daviesrob:
```
if (!CG) {
    if (errno == EINVAL) return -1;  // Bad aux data
    errno = saved_errno;
    return 0; // no CG tag
}
```

daviesrob:
Rebased with fixed return value in the bad aux data error case.

PR #1119, opened 2020-08-14
Compile ks_free() with HTSlib

### LINKED ISSUE #1114: kstring.h ks_free() use of `free()` crashes alternative compilers
### Descriptions

On windows, in an msvc application, with htslib (v1.10-ish?) compiled via mingw.

`ks_free()` is defined in the header `kstring.h` which uses `free()`.

This uses the msvc's free, while the alloc call inside htslib uses mingw's alloc, resulting in a crash

related #807

### Example:
```
sam_hdr_t* hdr = ...
bam1_t* rec = ...
kstring kstr;
ks_initialize(&kstr);

const int nbytes = sam_format1(hdr, rec, &kstr);
std::string s(kstr.s, kstr.l); // do something with kstr

ks_free(&kstr); // <-- crashes here 
```

### Workaround

```
#include <htslib/hts.h>

// ks_free(&kstr)
hts_free(kstr.s); // use the hts_free wrapper, in this case to mingw's `free()`
```

### Possible solution(s)

- moving ks_free definition into `kstring.c` (ABI issue?)
- including `hts.h` and using `hts_free()` (more header dependencies)
- moving `hts_free()` to some independent header file (preventing dragging large header)
- do nothing - it's easy enough to work around (once you know about it)

valeriuo:
Since the memory allocation of **kstring**s is managed by the C runtime HTSlib is built with, the deallocation should be managed by the same runtime. For this purpose, `ks_free()` is now compiled together with HTSlib and not inlined in the calling application.

Fixes #1114

### DIFF
--- a/htslib/kstring.h
+++ b/htslib/kstring.h
@@ -122,6 +122,10 @@ extern "C" {
     HTSLIB_EXPORT
 	int kgetline2(kstring_t *s, kgets_func2 *fgets, void *fp);
 
+    // Safely free the underlying buffer in a kstring.
+    HTSLIB_EXPORT
+    void ks_free(kstring_t *s);
+
 #ifdef __cplusplus
 }
 #endif
@@ -204,15 +208,6 @@ static inline char *ks_release(kstring_t *s)
 	s->l = s->m = 0;
 	s->s = NULL;
 	return ss;
-}
-
-/// Safely free the underlying buffer in a kstring.
-static inline void ks_free(kstring_t *s)
-{
-    if (s) {
-        free(s->s);
-        ks_initialize(s);
-    }
 }
 
 static inline int kputsn(const char *p, size_t l, kstring_t *s)
--- a/kstring.c
+++ b/kstring.c
@@ -48,6 +48,15 @@ int ks_resize2(kstring_t *s, size_t size)
 	return 0;
 }
 
+HTSLIB_EXPORT
+void ks_free(kstring_t *s)
+{
+    if (s) {
+        free(s->s);
+        ks_initialize(s);
+    }
+}
+
 int kputd(double d, kstring_t *s) {
 	int len = 0;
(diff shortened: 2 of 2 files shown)


htslib/kstring.h
```diff
@@ -144,7 +148,7 @@ static inline void ks_initialize(kstring_t *s)
 /// Resize a kstring to a given capacity
 static inline int ks_resize(kstring_t *s, size_t size)
 {
-	extern HTSLIB_EXPORT int ks_resize2(kstring_t *s, size_t size);
+	HTSLIB_EXPORT int ks_resize2(kstring_t *s, size_t size);
```
jmarshall, line 151: Did `extern` here need to be removed because it caused a problem for one of the various flavours of `HTSLIB_EXPORT`?

jmarshall:
What happened to not pessimising the inline function for non-Windows platforms where this doesn't matter?

To be sure it's not a major problem, but wouldn't something like the following suffice in _htslib/kstring.h_?

```c
static inline void ks_free(kstring_t *s)
{
    if (s) {
#ifdef _WIN32 /* or similar */
        extern HTSLIB_EXPORT void hts_free(void *ptr);
        hts_free(s->s);
#else
        free(s->s);
#endif
        ks_initialize(s);
    }
}
```

valeriuo:
I have another branch with [similar code](https://github.com/valeriuo/htslib/commits/free_lib), but it didn't feel right.
First of all, it's not just about Windows. If one links the application with the same C library HTSlib uses on Windows, the problem doesn't occur. Also, if we start on this path, we might have to cater to other combinations of compilers and libraries.

htslib/kstring.h
valeriuo: I see now [why](https://docs.microsoft.com/en-us/cpp/cpp/definitions-and-declarations-cpp?view=vs-2019) it was needed. I'll add it back.

PR #1123, opened 2020-08-19
Fix bgzf_close return code

valeriuo:
Since https://github.com/samtools/htslib/commit/3855184b58ff51735e9a70282a286702a24ae7f3, `bgzf_close()` returns an error code (**-1**) if the file pointer it is trying to close has an attached non-null error code, even if the closing procedure is successful.
This PR attempts to fix this behaviour by only considering the error codes generated inside `bgzf_close()` when setting the final return code.

Credit to OSS-Fuzz
Fixes oss-fuzz 25037

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -2000,9 +2000,12 @@ int bgzf_close(BGZF* fp)
 {
     int ret, block_length;
     if (fp == 0) return -1;
+    uint16_t save_fp_errcode = fp->errcode;
+    fp->errcode = 0;
     if (fp->is_write && fp->is_compressed) {
         if (bgzf_flush(fp) != 0) {
             bgzf_close_mt(fp);
+            fp->errcode |= save_fp_errcode;
             return -1;
         }
         fp->compress_level = -1;
@@ -2010,12 +2013,14 @@ int bgzf_close(BGZF* fp)
         if (block_length < 0) {
             hts_log_debug("Deflate block operation failed: %s", bgzf_zerr(block_length, NULL));
             bgzf_close_mt(fp);
+            fp->errcode |= save_fp_errcode;
             return -1;
         }
         if (hwrite(fp->fp, fp->compressed_block, block_length) < 0
             || hflush(fp->fp) != 0) {
             hts_log_error("File write failed");
             fp->errcode |= BGZF_ERR_IO;
+            fp->errcode |= save_fp_errcode;
             return -1;
         }
     }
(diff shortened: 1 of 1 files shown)


jkbonfield:
I'm not convinced this is the correct behaviour.

`fclose` does work as per your change above, demonstrated via this:

```

jmarshall:
`bgzf_close()`'s error handling is fundamentally ill-considered and needs to be rethought. At present it can e.g. lead to leaks from `hts_close()`.

As `bgzf_close()` is calling code's only option for properly disposing of a `BGZF` pointer, it should destroy the `struct BGZF` in all cases, whether an error occurred or not, and then return an error indication. This is how `fclose()`, `hclose()`, and `hts_close()` (except when it leaks its `BGZF*`…) all work.

So there's not much point setting `fp->errcode`, as it ought instead to be goneburger.

For `hFILE` I followed the same thinking as in James's comment: `hclose()` intentionally returns failure if an error was ever seen on the file handle — on the basis that this is convenient for less robust code that hasn't exhaustively checked every prior `hwrite` or completely remembered the results of such prior checking. `fclose()` returns failure if it needed to do some flushing itself and that failed, and this can be thought of as a logical extension of that.

It looks a bit like `bgzf_close()` follows similar thinking, although perhaps not as consciously…

(For `hclose()` or `bgzf_close()` the user wouldn't ever need to distinguish errors from EOF as these closing functions don't signal EOF in their return codes. Non-zero is always an error for these functions.)

jkbonfield:
I see what you mean about leaks.

I looked and it's been there since 1.0 or before:

https://github.com/samtools/htslib/blob/1.0/bgzf.c#L762-L792

We (and by that I mean *I*) just obfuscated it by adding much more code and complexity regarding threading, but the core of the broken return values and memory leaks has always been there. :(

We should obviously fix these, but I think we agree on the fundamentals here which is that errors tagged on the fd should lead to bgzf_close errors too, in which case the fix for the fuzzing issue we received (via Google's OSS fuzz) is probably in the test harness itself rather than this code.

daviesrob:
The problem in the fuzzer appeared in 3855184b where `bgzf_close()` started setting the return value from `fp->errcode` prior to destroying `fp`.  I agree, we should make the fuzzer ignore errors when closing its input file and keep the behaviour of propagating the error through close.  And we certainly need improved error clean-up in `bgzf_close()`.

daviesrob:
See #1124 for the fuzzing fix.

PR #1126, opened 2020-08-24
Account for underflow in kt_fisher_exact()

jmarshall:
As reported in arq5x/bedtools2#855 (bedtools contains a lightly-modified copy of htslib's _kfunc.c_). This backports PR arq5x/bedtools2#856 and adds test cases.

When some of `kt_fisher_exact()`'s input arguments are very large, `hypergeo(…)` is dominated by the subtracted `lbinom()` so returns `exp(≤ -750)`, which underflows, hence `hypergeo()` returns 0.0.

When `q` is 0.0, `kt_fisher_exact()`'s loops never execute and which of `i` and `j` is closer to `n11` is indeterminate. Hence we handle the adjustment of `left`/`right` separately in this case.

The `n11 + n22 < n12 + n21` test in that additional handling is motivated by [this description of left/right](https://www.langsrud.com/fisher.htm#INTRO) and agrees with the implementations in R and SciPy for various test cases, but this function's code is rather opaque so who really knows…

In any case, the patch only makes a difference when `q` is (exactly) 0.0, so does not affect the result for the “usual” range of inputs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ shlib-exports-*.txt
 /test/test-bcf-translate
 /test/test_bgzf
 /test/test_index
+/test/test_kfunc
 /test/test_kstring
 /test/test-parse-reg
 /test/test_realn
--- a/Makefile
+++ b/Makefile
@@ -77,6 +77,7 @@ BUILT_TEST_PROGRAMS = \
 	test/plugins-dlhts \
 	test/sam \
 	test/test_bgzf \
+	test/test_kfunc \
 	test/test_kstring \
 	test/test_realn \
 	test/test-regidx \
@@ -393,6 +394,7 @@ maintainer-check:
 #    MSYS2_ARG_CONV_EXCL="*" make check
 check test: $(BUILT_PROGRAMS) $(BUILT_TEST_PROGRAMS) $(BUILT_PLUGINS)
 	test/hts_endian
+	test/test_kfunc
 	test/test_kstring
 	test/test_str2int
 	test/fieldarith test/fieldarith.sam
@@ -430,6 +432,9 @@ test/sam: test/sam.o libhts.a
 
 test/test_bgzf: test/test_bgzf.o libhts.a
 	$(CC) $(LDFLAGS) -o $@ test/test_bgzf.o libhts.a -lz $(LIBS) -lpthread
+
+test/test_kfunc: test/test_kfunc.o libhts.a
+	$(CC) $(LDFLAGS) -o $@ test/test_kfunc.o libhts.a -lz $(LIBS) -lpthread
 
 test/test_kstring: test/test_kstring.o libhts.a
 	$(CC) $(LDFLAGS) -o $@ test/test_kstring.o libhts.a -lz $(LIBS) -lpthread
@@ -472,6 +477,7 @@ test/pileup.o: test/pileup.c config.h $(htslib_sam_h) $(htslib_kstring_h)
 test/plugins-dlhts.o: test/plugins-dlhts.c config.h
 test/sam.o: test/sam.c config.h $(htslib_hts_defs_h) $(htslib_sam_h) $(htslib_faidx_h) $(htslib_khash_h) $(htslib_hts_log_h)
 test/test_bgzf.o: test/test_bgzf.c config.h $(htslib_bgzf_h) $(htslib_hfile_h) $(hfile_internal_h)
+test/test_kfunc.o: test/test_kfunc.c config.h $(htslib_kfunc_h)
 test/test_kstring.o: test/test_kstring.c config.h $(htslib_kstring_h)
 test/test-parse-reg.o: test/test-parse-reg.c config.h $(htslib_hts_h) $(htslib_sam_h)
 test/test_realn.o: test/test_realn.c config.h $(htslib_hts_h) $(htslib_sam_h) $(htslib_faidx_h)
--- a/kfunc.c
+++ b/kfunc.c
@@ -29,6 +29,7 @@
 
 #include <math.h>
 #include <stdlib.h>
+#include <stdint.h>
 #include "htslib/kfunc.h"
 
 /* Log gamma function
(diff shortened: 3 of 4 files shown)


daviesrob:
Unfortunately, the fix to `kt_fisher_exact()` can be broken with this example:
```
(gdb) call test_fisher(50001, 9999, 89999, 50001, 1.0, 0.0, 0.0, 0.0)
(gdb) call test_fisher(49999, 10001, 90001, 49999, 1.0, 0.0, 0.0, 0.0)
[49999 10001 | 90001 49999] LEFT: 0 (expected 1)
[49999 10001 | 90001 49999] RIGHT: 1 (expected 0)
```

Both the linked web site, R and SciPy suggest both should be LEFT 1 RIGHT 0 TWO 0 for both:
```
Web site:
       Fisher's Exact Test 
https://www.langsrud.com/fisher.htm#INTRO
------------------------------------------
 TABLE = [ 50001 , 9999 , 89999 , 50001 ]
Left   : p-value = 1
Right  : p-value = 0
2-Tail : p-value = 0
------------------------------------------
 TABLE = [ 49999 , 10001 , 90001 , 49999 ]
Left   : p-value = 1
Right  : p-value = 0
2-Tail : p-value = 0

SciPy:
>>> fisher_exact([[50001,9999],[89999,50001]],alternative="less")
(2.7781975755326176, 0.9999999996761233)
>>> fisher_exact([[50001,9999],[89999,50001]],alternative="greater")
(2.7781975755326176, 0.0)
>>> fisher_exact([[50001,9999],[89999,50001]],alternative="two-sided")
(2.7781975755326176, 0.0)
>>> fisher_exact([[49999,10001],[90001,49999]],alternative="less")
(2.7773580693507856, 0.9999999996761233)
>>> fisher_exact([[49999,10001],[90001,49999]],alternative="greater")
(2.7773580693507856, 0.0)
>>> fisher_exact([[49999,10001],[90001,49999]],alternative="two-sided")
(2.7773580693507856, 0.0)

R:
> test1 <- matrix(c(50001, 9999, 89999, 50001), nrow = 2, dimnames = list(c("a", "b"), c("c", "d")))
> fisher.test(test1, alternative="l")$p.value
[1] 1
> fisher.test(test1, alternative="g")$p.value
[1] 0
> fisher.test(test1, alternative="t")$p.value
[1] 0
> test2 <- matrix(c(49999, 10001, 90001, 49999), nrow = 2, dimnames = list(c("a", "b"), c("c", "d")))
> fisher.test(test2, alternative="l")$p.value
[1] 1
> fisher.test(test2, alternative="g")$p.value
[1] 0
> fisher.test(test2, alternative="t")$p.value
[1] 0

```

It's possible to make a version of the algorithm that doesn't have the underflow problem by keeping `p` and `q` in logarithmic space as much as possible.  This does come with a speed penalty though as the trick used to speed the algorithm up doesn't really work any more, and the loops to calculate `i` and `j` actually run:
```c
static double lhypergeo(int n11, int n1_, int n_1, int n)
{
    return lbinom(n1_, n11) + lbinom(n-n1_, n_1-n11) - lbinom(n, n_1);
}

double kt_fisher_exact(int n11, int n12, int n21, int n22, double *_left, double *_right, double *two)
{
    int i, j, max, min;
    double p, q, left, right;
    int n1_, n_1, n;

    n1_ = n11 + n12; n_1 = n11 + n21; n = n11 + n12 + n21 + n22; // calculate n1_, n_1 and n
    max = (n_1 < n1_) ? n_1 : n1_; // max n11, for right tail
    min = n1_ + n_1 - n;    // not sure why n11-n22 is used instead of min(n_1,n1_)
    if (min < 0) min = 0; // min n11, for left tail
    *two = *_left = *_right = 1.;
    if (min == max) return 1.; // no need to do test
    q = lhypergeo(n11, n1_, n_1, n); // the probability of the current table
    // left tail
    p = lhypergeo(min, n1_, n_1, n);
    for (left = 0., i = min + 1; p < q - .00000001 && i<=max; ++i) // loop until underflow
        left += exp(p), p = lhypergeo(i, n1_, n_1, n);
    --i;
    if (p < q + .00000001) left += exp(p);
    else --i;
    // right tail
    p = lhypergeo(max, n1_, n_1, n);
    for (right = 0., j = max - 1; p < q - .00000001 && j>=0; --j) // loop until underflow
        right += exp(p), p = lhypergeo(j, n1_, n_1, n);
    ++j;
    if (p < q + .00000001) right += exp(p);
    else ++j;
    // two-tail
    *two = left + right;
    if (*two > 1.) *two = 1.;
    q = exp(q);
    // adjust left and right
    if (abs(i - n11) < abs(j - n11)) right = 1. - left + q;
    else left = 1.0 - right + q;
    *_left = left; *_right = right;
    return q;
}
```

This version passes all the tests in the PR and gets my extra test case right, but I don't know if it's completely correct.  Due to the likely slow down it's probably also not the best solution.  As we know the problem manifests when `q == 0.0` in the original version, switching to an alternate on discovering that would make sense especially as we know the answer is either (1, 0, 0) or (0, 1, 0).  If we can work out a more efficient way of finding where each tail reaches the target probability in such cases we should be able to get a better fix.

jmarshall:
Thanks for the analysis. Yeah, I thought about keeping it in log space for longer too but didn't work through what effect that would have.

We could either do that in the `q == 0.0` case or, as it seems to indeed be either (1,0,0) or (0,1,0) in these cases there should be a simple(ish) classifier something like the proposed extremely naive `n11 + n22 < n12 + n21` that does the trick.

Okay, back to the drawing board (and the bottom of my to-do list :smile:) on that classifier…

daviesrob:
I think I've managed to come up with a solution for the classifier.  You need to compare `n11` with the location of the peak of the distribution.  The peak is at the mode of the [hypergeometric distribution](https://en.wikipedia.org/wiki/Hypergeometric_distribution) which is fairly easy to work out.  Plugging in the appropriate variables gives the test `n11 < (n_1 + 1) * (n1_ + 1) / (n + 2)`.

A revision of this PR using this test can be found in my [pr1126v2](https://github.com/daviesrob/htslib/tree/pr1126v2) branch.  In the implementation I multiplied through by `(n + 2)` in the classifier to get rid of the division, and cast everything to `int64_t` to avoid overflow on the multiplications.  It also includes comments to explain how it works.

jmarshall:
Nice work, Rob.

I concur with your calculations of the mode, and will believe you on the resulting `n11` classifier :smile:

I've updated the PR from your **pr1126v2** branch, adding in the two test cases from your first comment that demonstrate the difference in results between my naive classifier and your good one.

daviesrob:
Thanks for the update.  Will you port this over to [bedtools2](https://github.com/arq5x/bedtools2) as well?

jmarshall:
Yep, I've got the bedtools equivalent drafted — I was just waiting for this one to be merged before I open it over there.

PR #1129, opened 2020-08-26
Have realloc and free in the same layer of kstring

### LINKED ISSUE #1114: kstring.h ks_free() use of `free()` crashes alternative compilers

valeriuo:
This PR is an alternative solution to https://github.com/samtools/htslib/pull/1119. Instead of pushing the `free` call inside the HTSlib compiled code, this solution brings `realloc` back to the inline function `ks_resize`, so that it can be directly called by the upper layer.
To minimize the number of CPU instructions generated for `ks_resize`, the round-up to a power of two has been replaced with a simpler approach to increase the string size with 1.5 X requested size. It's the same strategy used by the MS C++ runtime when increasing the capacity of a `vector`.

Fixes #1114

### DIFF
--- a/htslib/kroundup.h
+++ b/htslib/kroundup.h
@@ -32,7 +32,7 @@
 /*
   Macro with value 1 if the highest bit in x is set for any integer type
 
-  This is written avoiding conditionals (?: operator) to reduce the liklihood
+  This is written avoiding conditionals (?: operator) to reduce the likelihood
   of gcc attempting jump thread optimisations for code paths where (x) is
   large.  These optimisations can cause gcc to issue warnings about excessively
   large memory allocations when the kroundup64() macro below is used with
--- a/htslib/kstring.h
+++ b/htslib/kstring.h
@@ -144,9 +144,15 @@ static inline void ks_initialize(kstring_t *s)
 /// Resize a kstring to a given capacity
 static inline int ks_resize(kstring_t *s, size_t size)
 {
-	extern HTSLIB_EXPORT int ks_resize2(kstring_t *s, size_t size);
-
-	if (s->m < size) return ks_resize2(s, size);
+	if (s->m < size) {
+	    char *tmp;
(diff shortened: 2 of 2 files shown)


jmarshall:
Also need to remove `ks_resize2()` (from _kstring.c_) which is now unused.

jkbonfield:
I thought about that, but it needs to be kept.

It's use is via static inline functions in header files in earlier releases, so if we remove it then we're breaking the ABI and things compiled against earlier versions will no longer work.

jmarshall:
As mentioned in https://github.com/samtools/htslib/issues/1114#issuecomment-674076003, `ks_resize2()` first appeared on **develop** in March and fortunately has not been present in any releases.

jkbonfield:
Good point.  I keep forgetting what's in dev vs last release as it's been so long now.

Yes, we should remove that function too then.

PR #1130, opened 2020-08-26
Sped up bcf_check_record.

jkbonfield:
This is mainly by speeding up bcf_dec_typed_int1_safe, but also by
some other code in there to streamline the function and keep the
loops tighter.

The effect on clang is slight.  On gcc9 reading an uncompressed BCF
file can be a bit faster (up to 15% on one file), bringing the speed
of gcc and clang to similar speeds when using -O3.  (Gcc -O2 is still
noticably poorer than clang -O2 on this test though.)

Examples on CHM1_CHM13_2.1.gatk_raw.gvcf.ubcf and gcc9 -O3.

Before:
           1549.79 msec task-clock                #    0.999 CPUs util
        5549055377      cycles                    #    3.581 GHz
       16764915993      instructions              #    3.02  insn per

After:
           1345.83 msec task-clock                #    0.999 CPUs util
        4806341286      cycles                    #    3.571 GHz
       13942053336      instructions              #    2.90  insn per

Record checking is still around 50% of the entire decode CPU cost though.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -1270,27 +1270,27 @@ static int bcf_dec_typed_int1_safe(uint8_t *p, uint8_t *end, uint8_t **q,
     /* Use if .. else if ... else instead of switch to force order.  Assumption
        is that small integers are more frequent than big ones. */
     if (t == BCF_BT_INT8) {
-        *q = p + 1;
-        *val = *(int8_t *) p;
-    } else if (t == BCF_BT_INT16) {
-        if (end - p < 2) return -1;
-        *q = p + 2;
-        *val = le_to_i16(p);
-    } else if (t == BCF_BT_INT32) {
-        if (end - p < 4) return -1;
-        *q = p + 4;
-        *val = le_to_i32(p);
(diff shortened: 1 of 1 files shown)


vcf.c
```diff
@@ -1318,6 +1318,13 @@ static const char *get_type_name(int type) {
     return types[t];
 }
 
+static void bcf_record_check_err(const bcf_hdr_t *hdr, bcf1_t *rec, char *msg,
+                                 uint32_t *reports, int i, const char *s) {
+    if (*reports == 0 || hts_verbose >= HTS_LOG_DEBUG)
+        hts_log_warning(msg, msg, bcf_seqname_safe(hdr,rec), rec->pos+1, i, s);
```
daviesrob, line 1324: Too many `msg` here.  It would also be better to move the format string from the caller into here given that it's only used for "Invalid FORMAT (id|type)".  For the invalid type case, `get_type_name()` will only return "null" or "unknown" as anything else is valid, so removing that part of the message would not be a great loss.

PR #1135, opened 2020-09-10
[fuzz] Check against the actual type lenght

valeriuo:
Verify that the BCF input is within the length of its assigned type.

Credit to OSS-Fuzz
Fixes oss-fuzz 25533


jkbonfield:
Oops!  My fault and credit to Google for spotting the impact so quickly. :-)

Thanks for the prompt fix Valeriuo.

PR #1136, opened 2020-09-14
News update Sept 2020.

whitwham:
Ready for proofreading.


NEWS
```diff
@@ -1,16 +1,274 @@
...
+
+* Improved the speed of range queries when using BAI indexes by
+  making better use of the linear index data included in the file.
+  The best improvement is on low-coverage data. (#1031)
+
+* Alignments which consume no reference bases are now considered to have
+  length 1.  This would make such alignments cover 1 reference position in the
+  same manner as alignments that are unmapped or have no CIGAR strings.  These
+  alignments can now be returned by iterator based queries.  Thanks to John
+  Marshall. (#1063; fixed samtools/#1240, see also hts-specs#521).
```
valeriuo, line 71: There is a mix of styles when referring to issues from other repos. GitHub uses `<organization>/<repo>#<no>`. If all the repos are under `samtools` organization, we should go with `<repo>#<no>`.

NEWS
```diff
@@ -1,16 +1,274 @@
...
+  has been updated to avoid generating them.  (#1028; fixed samtools#1197)
+
+* In sam_hdr_create(), free newly allocated SN strings when encountering an
+  error. [fuzz] (#1034)
+
+* Prevent double free in case of idx_test_and_fetch() failure. Thanks to
+  @fanwayne for the bug report. (#1047; fixed #1033)
+
+* In the header, link a new PG line only to valid chains. Prevents an
+  explosive growth of PG tags on headers where PG tags are present
```
valeriuo, line 210: **lines** instead of **tags**

valeriuo (changes requested):

NEWS
```diff
@@ -1,16 +1,274 @@
...
+
 Noteworthy changes in release a.b
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Features and Updates
+--------------------
+
+* Support added for remote reference files.  fai_path() can take a remote
+  reference file and will return the corresponding index file.  Remote indexes
+  can be handle by refs_load_fai().  UR tags in @SQ lines can now be set to
```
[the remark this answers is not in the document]
daviesrob: Should be fixed.

NEWS
daviesrob: Opps.  I'd noticed this but forgot to change them.  Now done.

NEWS
daviesrob: Now fixed.

NEWS
jmarshall: Previously in NEWS we've written them as the full `samtools/samtools#1240` style and hence they automatically become links in the Releases text.

I don't know if the abbreviated `repo#NN` style will be automatically linkified (you sometimes see that as a link on GH web pages, but it's usually just being displayed abbreviated from its original `org/repo#NN` when it's seen on a page from the same org).

valeriuo:
Merged into develop as d50a3a531c597a24644c6878fc0068d0d1cc0a0d and 1748a2dd34e5ff4b4800fcf04dbd5860d8c30f22

PR #1137, opened 2020-09-15
Spelling corrections [minor]

jmarshall:
First commit has typo and spelling corrections in NEWS, manual pages, and public header files (in comments only).

Second commit extends that to source files (mostly in comments). You might prefer to leave this to the start of the next development cycle to avoid this churn in all these files. Or contrarily, to squash them together.


daviesrob:
Thanks very much.  I went for squashing, as it's easier to do them all in one go.

PR #1140, opened 2020-09-18
Add back assert.h to vcf.h.

jkbonfield:
This undoes a change in #1060.

It's a philosphical point.  On one hand we have an inclusion of a
spurious file that we do not utilise ourselves.  On the other hand we
have an API breakage where old software (in this case bcftools 1.10)
attemting to compile against current htslib no longer succeeds because
we've changed which symbols we declare.

This is unclean, but I think backwards compatibity trumps tidyness in
this case.

### DIFF
--- a/htslib/vcf.h
+++ b/htslib/vcf.h
@@ -41,6 +41,9 @@ DEALINGS IN THE SOFTWARE.  */
 #include "hts_defs.h"
 #include "hts_endian.h"
 
+/* Included only for backwards compatibility with e.g. bcftools 1.10 */
+#include <assert.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif


jmarshall:
Getting `<assert.h>` as a byproduct of HTSlib headers is a particularly unfortunate case, as third-party code wants visibility of when that header is included so that it is obvious how it is controlled by the third-party author's `NDEBUG` setting. (They can just `#include` it again, but it is better if HTSlib doesn't futz with it behind the author's back.)

However fair enough that bcftools-1.10 otherwise builds with htslib-current, including passing its own test suite. I'll propose the change again in the next development cycle: by the time of the following release, wanting to compile bcftools-1.10.x against htslib-1.12 and retain (some of) its old bugs will be less reasonable.

jkbonfield:
It's rock and hard place, or putting it another way I wish we didn't have to start from here.

NDEBUG I don't think matters here as including assert.h doesn't invalidate their current setting (or otherwise) of NDEBUG.  It probably only matters if people are explicitly doing something weird like:

```
#include <htslib/vcf.h>
#define NDEBUG
#include "my_local_header_that_uses_assert.h"
```

I think I'd rather deal with the observed compilation failure than a hypothetical one.  I don't like my solution, but I dislike it less than breaking the API.

PR #1143, opened 2020-09-24
Add unistd.h include to hts.c now that it used R_OK

andrewpatto:
Include unistd to guarantee access to R_OK definition.

On a mac (Xcode 11) - **when compiling via rust-htslib** - I get a compile error about missing R_OK. As far as I can tell R_OK is defined officially in unistd.h, but probably comes in via other includes.

htslib by itself compiles fine on my mac - so I realise this is not specifically a htslib problem - presumably papered over a bit by the autoconf system.

The only other spot in htslib that used R_OK does include unistd (cram/cram_io.c)

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -31,6 +31,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <string.h>
 #include <strings.h>
 #include <stdlib.h>
+#include <unistd.h>
 #include <inttypes.h>
 #include <limits.h>
 #include <stdint.h>


andrewpatto:
I should add that the the rust-htslib compile bound to the previous release of htslib did compile so this is something introduced in the latest release

jkbonfield:
On linux it looks to come via zlib.h -> zconf.h -> unistd.h.

Yes, we should be including it explicitly. Thanks.

jmarshall:
POSIX does indeed specify that `R_OK` comes from _unistd.h_, and _hts.c_ itself directly including what it itself needs is definitely the way to go.

It happens to come the same way on macOS; probably when compiling via Rust a different subset of the platform predefines that control whether _zconf.h_ activates its `#include <unistd.h>` are defined, leading to that not being activated.

PR #1146, opened 2020-09-28
Use the sover-bumping version number for -compatibility_version

### LINKED ISSUE #1144: Strange library version numbers
Hi, I'm the maintainer of htslib in MacPorts. I'm sure I've noticed this before but I guess I never reported it before so I'm reporting it now.

Your project's library has very strange version numbers. Here's what I see on my Mac after installing htslib 1.11:

```
$ ls -l /opt/local/lib/libhts*
-rwxr-xr-x  1 root  admin  1362172 Sep 25 03:39 /opt/local/lib/libhts.1.11.dylib
lrwxr-xr-x  1 root  admin       17 Sep 25 03:39 /opt/local/lib/libhts.3.dylib -> libhts.1.11.dylib
-rw-r--r--  1 root  admin  1837824 Sep 25 03:39 /opt/local/lib/libhts.a
lrwxr-xr-x  1 root  admin       17 Sep 25 03:39 /opt/local/lib/libhts.dylib -> libhts.1.11.dylib
```

That number between "libhts." and ".dylib" is the library's major version number. You're claiming that your library's major version number is both 3 and 1.11. That's strange. Usually a library's major version number is just an integer, and you increase it when you make a huge backward-incompatible change and want to force everyone who uses your library to have to relink. (So, ideally, you never do that.)

Now let's look at the minor version numbers:

```
$ otool -L /opt/local/lib/libhts.1.11.dylib
/opt/local/lib/libhts.1.11.dylib:
	/opt/local/lib/libhts.3.dylib (compatibility version 3.0.0, current version 1.11.0)
	/opt/local/lib/liblzma.5.dylib (compatibility version 8.0.0, current version 8.5.0)
	/opt/local/lib/libbz2.1.0.dylib (compatibility version 1.0.0, current version 1.0.8)
	/opt/local/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/opt/local/lib/libcurl.4.dylib (compatibility version 11.0.0, current version 11.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.50.4)
```

You're claiming that your library's minor version is currently 1.11.0, and that it's compatible with minor version 3.0.0. That's strange. Your library can't claim to be compatible with a minor version of itself that hasn't been released yet.

It's also unusual that one of your library's major version numbers and its current minor version number appear to be the same as your project's release version number. Normally these are separate things that are incremented on their own separate timetables.

jmarshall:
See #1144 for background. Closes #1144.

It seems that `-compatibility_version` checking has to date been ineffective on macOS, which could explain why it seemed that compatibility_version was not intended to be compared to current_version. Switch to a more orthodox scheme of using the version number of the HTSlib release that bumped the soversion as the compatibility_version.

The Makefile now hard-codes `LIBHTS_SOVERSION` and `LIBHTS_SOBUMP_VERSION`, both of which should be updated when the soversion is bumped. As well as being embedded in the macOS _libhts.dylib_, `LIBHTS_SOBUMP_VERSION` is a useful record in the Makefile.

Ideally perhaps this would be merged during an ABI-breaking release concurrent with a soversion bump. However as it seems that this macOS check has to date not worked anyway, this PR can probably be merged at any time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,8 +110,10 @@ include htslib_vars.mk
 # into here.
 PACKAGE_VERSION := $(shell ./version.sh)
 
+# The current library soversion, and the package release in which
+# this soversion first appeared.
 LIBHTS_SOVERSION = 3
-MACH_O_COMPATIBILITY_VERSION = $(LIBHTS_SOVERSION)
+LIBHTS_SOBUMP_VERSION = 1.10
 
 # $(NUMERIC_VERSION) is for items that must have a numeric X.Y.Z string
 # even if this is a dirty or untagged Git working tree.
@@ -293,7 +295,7 @@ libhts.so: $(LIBHTS_OBJS:.o=.pico)
 # includes this project's build directory).
 
 libhts.dylib: $(LIBHTS_OBJS)
-	$(CC) -dynamiclib -install_name $(libdir)/libhts.$(LIBHTS_SOVERSION).dylib -current_version $(NUMERIC_VERSION) -compatibility_version $(MACH_O_COMPATIBILITY_VERSION) $(LDFLAGS) -o $@ $(LIBHTS_OBJS) $(LIBS)
+	$(CC) -dynamiclib -install_name $(libdir)/libhts.$(LIBHTS_SOVERSION).dylib -current_version $(NUMERIC_VERSION) -compatibility_version $(LIBHTS_SOBUMP_VERSION) $(LDFLAGS) -o $@ $(LIBHTS_OBJS) $(LIBS)
 	ln -sf $@ libhts.$(LIBHTS_SOVERSION).dylib
 
 cyghts-$(LIBHTS_SOVERSION).dll libhts.dll.a: $(LIBHTS_OBJS)


jkbonfield:
+1 on it fixing the issue.  Thanks. :-)  I don't think we need to be overly prescriptive and state current version must be reset to 1.0.0 on changes.  Just as long as current and compatibility work together we're OK (and IMO it's useful meta-data).

However the naming could be better.  This is a pure Mac thing, but "SOBUMP_VERSION" to me immediately conjures up  a linux style `-soname` parameter so naively I'd be assuming it's used there too.  Maybe an explicit "MAC_COMPATIBILITY_VERSION" to indicate the purpose more precisely.

It could also do with some comments explaining how this version compares to the package release; specifically instructions on what to change.   The key thing here is starting from say 1.11.0 with new APIs but no ABI changing, then I'd expect both current and compatibility to be 1.11.0.

However if we find an algorithmic bug and release 1.11.1 then we'd bump current version while leaving the "LIBHTS_SOBUMP_VERSION" parameter untouched.  It's possible (albeit highly unlikely) we may even have a period of a lot of code optimisation and tidying up that doesn't add any new APIs, so release 1.12.0 is also 100% compatible with 1.11.0, in which case we wouldn't change it either.  In that regard it mustn't be viewed simply as major, minor, patchlevel and hence why having COMPATIBILITY in the name emphasises the purpose.  (This is where the libtool instructions help.  I copied a bunch of those into io_lib's configure.ac as a reminder to myself for what variables to edit whenever I do a release.)

daviesrob:
I think this isn't quite correct given how we've used `-compatibility_version` in the past.  Unless we rename the library, the first digit probably needs to be greater-than or equal to 3.  We should though be incrementing either the first or second digit whenever the ABI is changed in a compatible manner (the easiest way might be to increment the second digit).

jmarshall:
It's very intentional that the name `LIBHTS_SOBUMP_VERSION` is **not** macOS-specific but rather reflects that this **is indeed** tied to the platform-agnostic soversion/soname machinery:

```c
# The current library soversion, and the package release in which
# this soversion first appeared.
```

I can expand that comment if you wish.

The macOS compatibility_version/current_version check (as it was presumably intended :smile:) can be simply considered to be a more intricate approach to sonames. My (slightly reduced from before…) claim is that making them package versions means that compatibility_version needs to record exactly the same point in time as bumping soversion does.

@daviesrob: You're making the same argument as my last paragraph: for complete safety this should happen during a soversion bump so that the library's filename is different. However because in practice the compatibity_version field is currently **ignored** by macOS's dynamic loader, it really doesn't matter how it's been used in the past — it can be changed arbitrarily.

jkbonfield:
> 
> 
> It's very intentional that the name `LIBHTS_SOBUMP_VERSION` is **not** macOS-specific but rather reflects that this **is indeed** tied to the platform-agnostic soversion/soname machinery:

And yet it's only used by MacOS...

> ```c
> # The current library soversion, and the package release in which
> # this soversion first appeared.
> ```
> 
> I can expand that comment if you wish.

Maybe I'm misunderstanding this then.

There are ABI breaking changes such as removing or renaming functions, changing the layout of data structures, and so on.  They result in the SONAME being bumped.

Then there are API additions; adding new functions, structures etc.

Consider a tool that uses `vcf_open_mode` added to 1.11.   There hasn't been an ABI change between 1.10 and 1.11, but we do need to know we're compatible with libraries matching SONAME 3 and version 1.11 onwards.  Knowing that SONAME 3 arrived in 1.10 doesn't help us one iota here, and I don't think this is the right mindset.  Hence why I disagree with the naming of this variable.

My understanding from the mac man page is that the ABI breaking changes are simply handled by the filename matching.  We have a hard baked in dependency on libhts.so.3 or libhts.3.dylib so that's easily dealt with.  The latter case of a minimum version of that ABI is handled by current-version vs compatible-version checking.  As you mention elsewhere, probably best done by simply comparing the baked in version in the compiled executable and the compatible-version recorded in the library found on disk.

What it really needs is some hands-on testing to validate it I guess.  I'd do it, but sadly the server mac we were supplied for the office turns off networking unless someone physically prods the machine.  Sigh. :/  Even less use than a chocolate fireguard (you could eat that).

jmarshall:
> And yet it's only used by MacOS...

It's **currently** only used on macOS. If someone were sufficiently motivated, they could e.g. embed it in an ELF note section for documentation purposes on Unix. The point is, the concept recorded is not macOS-specific.

> Consider a tool that uses vcf_open_mode added to 1.11. There hasn't been an ABI change between 1.10 and 1.11, but we do need to know we're compatible with libraries matching SONAME 3 and version 1.11 onwards.

Why do we need to know this? This is not a test that's commonly done, and there's a good reason for that. Suppose I have a program that does not use `vcf_open_mode()` (or anything else that's new in 1.11) and I compile and link it against 1.11. Nonetheless that program should not be prevented from running against 1.10.

jkbonfield:
> > Consider a tool that uses vcf_open_mode added to 1.11. There hasn't been an ABI change between 1.10 and 1.11, but we do need to know we're compatible with libraries matching SONAME 3 and version 1.11 onwards.
> 
> Why do we need to know this? On the surface that seems useful, but this is not a test that's commonly done and there's a good reason for that. Suppose I have a program that does not use `vcf_open_mode()` (or anything else that's new in 1.11) and I compile and link it against 1.11's _libhts.dylib_. Nonetheless that program should not be prevented from running against 1.10.

Why you need to know it is in order to prevent run-time crashes basically.  Say you get a binary compiled against libhts.so.3 produced by htslib-1.11.  You have htslib-1.10 installed.  You do a samtools view.  You get:

```
./samtools: symbol lookup error: ./samtools: undefined symbol: fai_path
```

This is real btw.  I built samtools-1.10 from tarball.  I also built and installed htslib 1.11 and built samtools 1.11 using --with-htslib=/tmp/inst.   Then `LD_LIBRARY_PATH=/tmp/inst/lib ./samtools` works, but `LD_LIBRARY_PATH=../samtools-1.10/htslib-1.10` does not.  Both provide libhts.so.3, but one of them has an incompatible version.

Linux does have a concept of doing the same thing that the Mac does here, but it's via a more complex route.  It also answers your other query.  Try objdump on glibc and you'll see it has a whole bunch of GLIBC version numbers embedded in there.  I think these come from ld version scripts where each function gets tied to a specific symbolic version.  If you compile against 1.11 but don't use any symbols labelled with 1.11 then your dependency will be against a 1.10 (or earlier) implementation of the libhts.so.3 ABI.

We don't currently support that fine grain however.

daviesrob:
I think that while the check on compatibility version can't work for `dlopen()` (there's no way to supply it), it does happen for programs that link against HTSlib.  So the value in the library really needs to march forwards, at least until the filename gets changed.  As we've baked the ABI version into digit X, it needs to stay there until we have a major version bump and can reset the counter.  Setting digit Y to our current minor API version would do the trick, although it is slightly less convenient compared to just putting in the full API version as sort-of done here.

(Although I don't quite understand why you used `1.10` here, John?)

jkbonfield:
What would be useful is to know if the Mac actually solves the above example.  I believe it's what current-version vs compatible-version is intended to resolve.   It knows upfront that the major version exists on disk, but then it can also check that the local library we have available isn't compatible with the version our binary requires.

It makes perfect sense to me as a feature, but it sounds like the checks may not be there.

jmarshall:
> Why you need to know it is in order to prevent run-time crashes basically. Say you get a binary compiled against libhts.so.3 produced by htslib-1.11. You have htslib-1.10 installed. You do a samtools view. You get:
>
>     ./samtools: symbol lookup error: ./samtools: undefined symbol: fai_path

That's not a run-time crash. It's an error message from the dynamic loader, just like “incompatible shared library” but with a more detailed error message.

To be sure, “symbol lookup error” is fairly obscure for bfx users. However any message from ld.so / dyld would be equally obscure and also lead to biostars/etc postings…

jkbonfield:
I should also point out that the behavior of checking library versions rather than ABI releases was one of the most annoying behaviours in early linux!  There were numerous times when I copied a binary from one system to another and had to edit the binary to change the GLIBC version number it required, mainly because GNU had a horrible knack of tinkering with GNU specific things every release and your program using standard c89 would refuse to work because it *may* be incompatible with the glib on your other system.  IMO that however was a mistake of GNU by mixing slow moving standard and fast moving non-standard things in the same library.

Regardless, it's a thing of the past as the per-function versioning, while a total PITA to do, completely fixed it.  I'm not suggesting the pain of going down that road however unless someone feels sufficiently exercised.

jkbonfield:
> 
> 
> > Why you need to know it is in order to prevent run-time crashes basically. Say you get a binary compiled against libhts.so.3 produced by htslib-1.11. You have htslib-1.10 installed. You do a samtools view. You get:
> > ```
> > ./samtools: symbol lookup error: ./samtools: undefined symbol: fai_path
> > ```
> 
> That's not a run-time crash. It's an error message from the dynamic loader, just like “incompatible shared library” but with a more detailed error message.
> 
> To be sure, “symbol lookup error” is fairly obscure for bfx users. However any message from ld.so / dyld would be equally obscure and also lead to biostars/etc postings…

Ok fair enough, and if you hadn't actually used anything new then yes it'd work fine.  I'll grant you that.

I'm just trying to illustrate the *concept* which I believe Apple are addressing here.  I don't think it's wanting to know the earliest package release matching the first ABI version.  At least that's not my reading of current vs compatible in that man page.

jmarshall:
> I'm just trying to illustrate the concept which I believe Apple are addressing here.

We all understand the concept. The questions are around whether it's worth it for htslib, whether it's particularly possible, and whether Apple's (possibly currently buggy) implementation actually does it (and how).

jkbonfield:
> 
> 
> > I'm just trying to illustrate the concept which I believe Apple are addressing here.
> 
> We all understand the concept. The questions are around whether it's worth it for htslib, whether it's particularly possible, and whether Apple's (possibly currently buggy) implementation actually does it (and how).

Well, you either agree or disagree with the concept of including the package version in run-time checking rather than just ABI version.

If you disagree, we should use SONAME for both current_version and compatible_version so it's entirely package release agnostic.   If you agree then we should use the correct package version so it tracks the API properly.   I don't see the point otherwise.

jmarshall:
@daviesrob:

>I think that while the check on compatibility version can't work for dlopen() (there's no way to supply it), it does happen for programs that link against HTSlib.

Indeed, there is documentation somewhere saying this happens (or is supposed to happen) at startup but not for `dlopen()`.

> So the value in the library really needs to march forwards, at least until the filename gets changed.

However I have tested a program linked against a libhts.dylib with compatibility_version=3 then running against a this-PR-built libhts.3.dylib with compatibility_version=1.10 — and that program loads and runs just fine. So it doesn't need to march forwards, as (at present) this stuff is just not implemented in dyld (modulo the investigations into the dyld source code, which I haven't looked at yet).

> (Although I don't quite understand why you used 1.10 here, John?)

As the value of `LIBHTS_SOBUMP_VERSION` in the PR diff? Because that is the HTSlib release in which soversion was bumped to its current value, 3.

i.e., this PR is proposing using compatibility_version at the same level of granularity as the existing soversion. So that compatibility_version enforces ABI incompatibilities but not new additions, exactly as soversion traditionally does.

One can agree with the concept of including the package version in run-time checking checking — but the next question is what policy to use to update the various version numbers.

This PR proposes making both compatibility_version and current_version be comparable X.Y.Z numbers (as @ryandesign desires), making them both be the package version number for documentation and parsimoniousness (no extra numbers to invent!) reasons, and making the current_version be the current package version (duh) and making compatibility_version change in lockstep with soversion changes (because that adds no extra burden). The claim is that this is both convenient and compatible with the rules (as well as they can be discerned). It doesn't actually buy you anything over the installname/soname-based library filename search. It does not attempt to model the “new additions to API” problem, on the basis that the answer to that is either symbol versioning or who cares :smile:

jkbonfield:
> > (Although I don't quite understand why you used 1.10 here, John?)
> 
> As the value of `LIBHTS_SOBUMP_VERSION` in the PR diff? Because that is the HTSlib release in which soversion was bumped to its current value, 3.
> 
> i.e., this PR is proposing using compatibility_version at the same level of granularity as the existing soversion. So that compatibility_version enforces ABI incompatibilities but not new additions, exactly as soversion traditionally does.

Isn't this basically the same as simply deleting the current_version and compatibility_version arguments entirely, in terms of the actual affect?  The ABI version is checked by dint of it being part of the filename.  Beyond that we do no checks, in which case we may as well not fill out those fields.

Or is there a way on the Mac that it can end up at runtime looking for hts.dylib rather than hts.3.dylib?  I don't think that would be possible on a linux box unless we broke the install process.

jmarshall:
> Isn't this basically the same as simply deleting the current_version and compatibility_version arguments entirely, in terms of the actual effect? The ABI version is checked by dint of it being part of the filename.

Yes, hence “It doesn't actually buy you anything over the installname/soname-based library filename search“.

> Beyond that we do no checks, in which case we may as well not fill out those fields.

These fields do however provide interesting diagnostic information in `otool -L` output (which is essentially macOS's `ldd` equivalent).

jkbonfield:
> > Beyond that we do no checks, in which case we may as well not fill out those fields.
> 
> These fields do however provide interesting diagnostic information in `otool -L` output (which is essentially macOS's `ldd` equivalent).

True, and infact it's what I was thinking of when I wrote _and IMO it's useful meta-data_ above.

daviesrob:
I've been looking into this again as it would be useful to fix this problem before our next release.  Currently, if you build and install HTSlib from before this PR, make a samtools that links to the DLL, and then update HTSlib to include this PR, trying to run samtools gives:

```
$ ./samtools 
dyld: Library not loaded: /Users/rmd/opt/htslib/lib/libhts.3.dylib
  Referenced from: /Users/rmd/git/samtools/./samtools
  Reason: Incompatible library version: samtools requires version 3.0.0 or later, but libhts.3.dylib provides version 1.10.0
Abort trap: 6
```

which is not so good as it should really work.

I think until we have an ABI bump, we should set `compatibility_version` to either 3.1.11, or 3.11, or 11; and set `current_version` to the same thing.  The 11 can then be incremented as we make releases (in theory we could leave `compatibility_version` alone if we don't add new interfaces, but that rarely happens between releases).  After the next ABI bump we can switch to using the version number directly, as currently in this PR.

jmarshall:
> Currently, if you build and install HTSlib from before this PR, make a samtools that links to the DLL, and then update HTSlib to include this PR, [it fails to load]

Interesting. Running such a samtools against a _lilbhts.3.dylib_ built from this PR works fine for me on macOS 10.14.6 (i.e., current Mojave). What version of macOS are you testing this on? (Has this test perhaps finally been implemented in the OS?)

daviesrob:
Mine pre-dates Mojave.  It's working as I would expect from reading the documentation, so I'm not sure why yours differs.

jmarshall:
The current setup was working when it was set up in 2013 (on Mountain Lion) and is working on current Mojave (at least) but switching to this PR's improved scheme without changing the soversion, i.e. dylib filename, either also works (my machine) or doesn't work (yours, of unspecified version). 🤷
Gatekeeper settings? \</straw-grasping>

If the documentation in question is the man page, IMHO it's quite ambiguous (https://github.com/samtools/htslib/issues/1144#issuecomment-699920361) and could be read as supporting a number of interpretations.

Anyway, there's pretty clearly a problem to solve, and when the soversion is bumped it can be solved as per this PR. I guess the decision is whether to leave well enough alone until then (AIUI the status quo isn't causing any actual failures), or to bump to `3.1.11` or similar until then.

daviesrob:
I think I'll go with `3.1.11` for this, and probably hard-code it with an update as necessary on release.  I'll make a new pull request for this, as the changes will be a bit different to the ones here.

Issue #1148, opened 2020-09-29
CRLF line termination not handled properly in sam_parse_worker


anderskaplan:
To reproduce: `unix2dos test/index.sam`, then run the test suite.

Actual: The test test_index fails, but only in the multi-threaded case. The single-threaded case passes.
This is because the line parser in sam_parse_worker does not handle CRLF line endings properly. The CR is passed on to sam_parse1 where it is included in the aux field data. The resulting BAM differs from the reference.

Expected: The test suite should pass regardless of line endings.

whitwham:
Fixed in #1149.

PR #1149, opened 2020-09-29
Bug fix: handling of CRLF line terminators in sam_parse_worker()

anderskaplan:
Fixes issue #1148 

Also adds a few test cases for kstring which were written while troubleshooting.


whitwham:
Looks good, though I will have to go through the tests more carefully.

whitwham:
Added a test for a CRLF sam file.

daviesrob:
Thanks for the fix.

PR #1150, opened 2020-09-30
Variable name used by dlsym in load_plugin corrected.

whitwham:
The hfile irods plugin depends on a constructed symbol name which was no longer being used.  This is a small fix to changes made in #1072.

### DIFF
--- a/plugin.c
+++ b/plugin.c
@@ -153,7 +153,7 @@ plugin_void_func *load_plugin(void **pluginp, const char *filename, const char *
         const char *basename = slash? slash+1 : filename;
         kputsn(basename, strcspn(basename, ".-+"), &symbolg);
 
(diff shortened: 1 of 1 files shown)


jmarshall:
Alas, this was a cut'n'pasto in 00c61d2c8d83324494d4edab42c452c2d3cc26fc. /me hangs head in shame. Feel free to be more explicit in the commit message: `Fixes typo introduced in 00c61d2c8d83324494d4edab42c452c2d3cc26fc`.

Fortunately this only affects the minority of plugins needing `RTLD_GLOBAL`. If it affected all of them, there would be the start of a case for htslib-1.11.1 here.

Sigh… all that conversation and neither I nor anyone else reviewed this line carefully enough to catch this inadvertent change… :smile:

PR #1151, opened 2020-09-30
Replace the indexing range check error message.

jkbonfield:
This code detects when the end positioni s too large for the current
index.  It then computes an adjusted n_lvls parameter and tells the
user to switch to CSI with this parameter.

That has numerous failings though.

1. The only user-adjustable parameter is min-shift, not levels, so
the advice is impossible to follow.

2. CSI auto-scales based on SQ headers, so we don't need to be
explicit anyway (assuming we currently have a BAI index that doesn't
fit).

3. If we already were using CSI index and if we did have the
configuration options to do what it asks, it wouldn't fix the problem.
Given CSI auto-scales, if we see this message it's because of the
alignments don't match the headers; either the header has been
replaced by something inappropriate or there are bugs in the aligner
that has emittede rogue POS fields.  There is no fix to the index that
actually fixes the data; it's simply brushing the problem under the carpet.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -1865,26 +1865,16 @@ int hts_idx_check_range(hts_idx_t *idx, int tid, hts_pos_t beg, hts_pos_t end)
     int64_t maxpos = (int64_t) 1 << (idx->min_shift + idx->n_lvls * 3);
     if (tid < 0 || (beg <= maxpos && end <= maxpos))
         return 0;
-    int64_t max = end > beg ? end : beg, s = 1 << 14;
-    int n_lvls = 0;
-    while (max > s) {
-        n_lvls++;
-        s <<= 3;
-    }
 
     if (idx->fmt == HTS_FMT_CSI) {
-        hts_log_error("Region %"PRIhts_pos"..%"PRIhts_pos" cannot be stored in a csi index "
-                      "with min_shift = %d, n_lvls = %d. Try using "
-                      "min_shift = 14, n_lvls >= %d",
-                      beg, end,
-                      idx->min_shift, idx->n_lvls,
-                      n_lvls);
+        hts_log_error("Region %"PRIhts_pos"..%"PRIhts_pos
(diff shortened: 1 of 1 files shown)


daviesrob:
Tabix can index a few formats (notably bed) with no headers.  When doing that `n_lvls` gets [set to a default size](https://github.com/samtools/htslib/blob/30d82d6b/tbx.c#L302) - if you hunt around for the definition of `TBX_MAX_SHIFT` you eventually discover the default is 6, which means it can go up to 4Gbases so mostly people won't have noticed.  It does mean it fails on a bed file like this:
```
chr1	1	2
chr1	2	3
chr1	1234567890	1234567891
chr1	5000000000	5000000001
```

And the error message, while better, still doesn't help much:
```
$ ./tabix -C /tmp/test.bed.gz 
[E::hts_idx_check_range] Region 5000000000..5000000001 cannot be stored in a csi index. Please check headers match the data
tbx_index_build failed: /tmp/test.bed.gz
```

I wonder how hard it would be to solve this in a more user-friendly way by attempting a fix-up on the index by increasing `n_lvls` and then renumbering the existing bins?  I think renumbering should be as simple as adding a constant, but you would need to put them into a new hash.  There may be some other state that might need fixing, but I think it should be do-able.  This could even extend to upgrading a .bai/.tbi index to .csi, although we would certainly want to print a warning if that happened due to the output not being quite what was asked for.

jkbonfield:
Perhaps, but meanwhile this is a tiny trivial fix and the alternative is a considerably larger project to develop and test.  IMO do this first and then if and when we have a better system it can be replaced again.

Wanting perfection is the enemy of progress!

daviesrob:
Merging for now, but I think it would be good to see if we could get an automatic fix-up to work, along with automatic selection of CSI when needed.  It would fix a frequent pain-point for people working on larger references.

PR #1152, opened 2020-10-02
Correct the linear index to include unmapped placed reads

### LINKED ISSUE #1142: Unmapped reads are not added to the BAI linear index

valeriuo:
Unmapped placed reads are not added to the linear index, but are covered by the metabin of the bin index. To include the unmapped reads in the iterator run, correct the starting offset of the first linear index entry that can cover these reads.

Fixes #1142

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -1759,12 +1759,21 @@ static void update_loff(hts_idx_t *idx, int i, int free_lidx)
     khint_t k;
     int l;
     uint64_t offset0 = 0;
+    uint32_t unmapped = 0;
     if (bidx) {
         k = kh_get(bin, bidx, META_BIN(idx));
-        if (k != kh_end(bidx))
+        if (k != kh_end(bidx)) {
             offset0 = kh_val(bidx, k).list[0].u;
-        for (l = 0; l < lidx->n && lidx->offset[l] == (uint64_t)-1; ++l)
-            lidx->offset[l] = offset0;
+            unmapped = 1;
+        }
+        for (l = 0; l < lidx->n; ++l) {
+            if (lidx->offset[l] == (uint64_t)-1) {
+                lidx->offset[l] = offset0;
+            } else if (unmapped && offset0 < lidx->offset[l]) {
+                unmapped = 0;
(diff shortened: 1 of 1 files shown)


jmarshall:
When **generating** the index, the reason these reads are not added to the linear index is the use of `is_mapped` within `hts_idx_push()`. `is_mapped` is documented in _htslib/hts.h_ as controlling which of `n_mapped`/`n_unmapped` in the meta-bin this read will contribute to. Hence for generating the index, why not just amend the `if (is_mapped)` condition that controls whether `insert_to_l()` is called in `hts_idx_push()`?

This approach also aims to correct the in-memory data when a previously-generated index is **read in**, as discussed in https://github.com/samtools/htslib/issues/1142#issuecomment-698415867. So that's good, and worth mentioning in the commit message and PR description. But maybe it would be better to fix `hts_idx_push()` as above and do this correction separately just for the reading-in case?

This change changes `update_loff()`'s behaviour if there are `lidx->offset[l]` entries that are -1 which come after a non-`-1` entry — previously they were not updated to `offset0`. What effect does this have?

jkbonfield:
Unfortunately this doesn't seem to fix the bug.

```
@ deskpro[samtools.../htslib]; samtools view -c ~jkb/scratch/data/um.bam 20:24613254-24613799
1
@ deskpro[samtools.../htslib]; /tmp/samtools.dev view -c ~jkb/scratch/data/um.bam 20:24613254-24613799
1
@ deskpro[samtools.../htslib]; samtools110 view -c ~jkb/scratch/data/um.bam 20:24613254-24613799
11
```

Samtools110 is 1.10 release.  The .dev and this PR give the same results for this test.  @valeriuo - the test file should be visible to you internally.  There are also a mu.bam and uu.bam files in the same directory.  um.bam and mu.bam contain the same reads, but for half-mapped data um has unmapped followed by mapped while mu is the other way around, neatly demonstrating the bug.

Also in all cases I multiplied up the unmapped reads by 10 copies in order to make it significantly more likely they'd span a bgzf block.  uu.bam is entirely unmapped, as an alternative stress test.

PR #1154, opened 2020-10-07
Fix the invariant that fd->dispatcher thread exists if fd->h is set

jmarshall:
[An alternative take on the pthread `fd->dispatcher` problem identified in #1153.]

Remove `fd->dispatcher` test as it makes unwarranted assumptions about the `pthread_t` type. In fact this code block assumes that `fd->dispatcher` is set up whenever `fd->h` has been set (see the unguarded `pthread_join()` further down the function).

Fix `pthread_create()` error handling accordingly. (Also use the canonical name for `sam_hdr_destroy()`.)

This may be incomplete: it may be worth making the same change in the other nearby error handling; more significantly, the relationship between the thread pool in `fd->q` and the dispatcher thread is unclear, so resetting `fd->h` may not be quite right in some cases… maybe some of the `fd->q` handling needs to change too.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -2340,7 +2340,7 @@ int sam_state_destroy(htsFile *fp) {
             if (fd->q)
                 hts_tpool_wake_dispatch(fd->q); // unstick the reader
 
-            if (!fp->is_write && fd->q && fd->dispatcher) {
+            if (!fp->is_write && fd->q) {
                 for (;;) {
                     // Avoid deadlocks with dispatcher
                     if (fd->command == SAM_CLOSE_DONE)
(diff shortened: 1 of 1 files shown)


jmarshall:
(Note that in the present code, if a `pthread_create(&fd->dispatcher, …)` fails, then the subsequent `sam_close()` calls `sam_state_destroy()` which calls `pthread_join()` on an indeterminate pthread ⇒ :boom: on an unfriendly pthread implementation.)

jkbonfield:
It's been hammered on linux, but agreed there are probably poor bits of code that don't work well on other platforms.

Getting the SAM threading to work with valid data and no errors is fine.  The problems all come when you have a failure - malformed data, running out of memory, and so on.  If I recall the line was added after a huge round of fuzzing where we died after the Nth malloc call, to test as many possible failure code paths as possible.  The number of ways it could die (obviously) was huge.   Things like failing to launch the thread would have been one of those that lead to deadlocks.

I don't want to simply undo that check, but yes it needs to cope with the thread state being a struct.   Given SAM_state is internal it could just have a status integer to indicate if it's been initialised and use that instead of a boolean comparison on a pthread_t.

PR #1155, opened 2020-10-08
Adjust the starting offset of the iterator for unmapped placed reads

### LINKED ISSUE #1142: Unmapped reads are not added to the BAI linear index

valeriuo:
Unmapped placed reads can be skipped by an iterator read, when they fall at the beginning of a BGZF block, since they are not added to the linear index. However, the meta-bin from the bin index records the presence of unmapped reads, and can indicate whether the iterator should start reading from the beginning of a bin chunk, which includes the unmapped reads, or from the linear index offset.

Also, the code that sets the `bins_t::loff` to the `lidx_t::offset` was removed, as this check is already performed inside `update_loff` [method](https://github.com/samtools/htslib/blob/2264113e5df1946210828e45d29c605915bd3733/hts.c#L1773-L1784). 

Fixes #1142

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -2412,7 +2412,7 @@ static inline int reg2bins(int64_t beg, int64_t end, hts_itr_t *itr, int min_shi
     return itr->bins.n;
 }
 
-static inline int reg2intervals(hts_itr_t *iter, const hts_idx_t *idx, int tid, int64_t beg, int64_t end, uint32_t interval, uint64_t min_off, uint64_t max_off, int min_shift, int n_lvls)
+static inline int reg2intervals(hts_itr_t *iter, const hts_idx_t *idx, int tid, int64_t beg, int64_t end, uint32_t interval, uint64_t min_off, uint64_t max_off, int min_shift, int n_lvls, uint32_t unmapped)
 {
     int l, t, s;
     int i, j;
@@ -2444,7 +2444,7 @@ static inline int reg2intervals(hts_itr_t *iter, const hts_idx_t *idx, int tid, 
                     iter->off = off;
                     for (j = 0; j < p->n; ++j) {
                         if (p->list[j].v > min_off && p->list[j].u < max_off) {
-                            iter->off[iter->n_off].u = min_off > p->list[j].u
+                            iter->off[iter->n_off].u = !unmapped && min_off > p->list[j].u
                                 ? min_off : p->list[j].u;
                             iter->off[iter->n_off].v = max_off < p->list[j].v
                                 ? max_off : p->list[j].v;
@@ -2552,6 +2552,7 @@ hts_itr_t *hts_itr_query(const hts_idx_t *idx, int tid, hts_pos_t beg, hts_pos_t
     bidx_t *bidx;
     uint64_t min_off, max_off;
     hts_itr_t *iter;
+    uint32_t unmapped = 0;
 
     // It's possible to call this function with NULL idx iff
     // tid is one of the special values HTS_IDX_REST or HTS_IDX_NONE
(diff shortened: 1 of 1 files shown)


jkbonfield:
This seems to more or less just be the same as undoing 45715e421e1ce9d15d8880a24454ec7dc42bf4bc plus some added CPU expense to boot.

Some benchmarks on Samtools 1.10 (before the linear index commit), Samtools dev branch without this commit, and "samtools" which is a checkout of this PR.

```
@ deskpro107386[samtools.../samtools]; R=`perl -e 'srand(0);for ($i=0;$i<50000;$i++) {my $p = int(rand(60000000));$e=$p+int(rand(1000));print "20:$p-$e "}'`

@ deskpro107386[samtools.../samtools]; time samtools110 view -c ~/scratch/data/9827_2#49.bam $R
500590
real	0m45.141s
user	0m44.148s
sys	0m0.987s


@ deskpro107386[samtools.../samtools]; time /tmp/samtools.dev view -c ~/scratch/data/9827_2#49.bam $R
500590
real	0m14.350s
user	0m14.069s
sys	0m0.280s


@ deskpro107386[samtools.../samtools]; time samtools view -c ~/scratch/data/9827_2#49.bam $R
500590
real	1m30.408s
user	1m29.365s
sys	0m1.036s
```

I/O wise we get the following from io_trace:

```
@ deskpro107386[samtools.../samtools]; io_trace -r 'bam$' -S -- samtools110 view -c ~/scratch/data/9827_2#49.bam $R
...
    Num. seeks   	51656
    Num. read    	294728
    Bytes read   	8084384944

@ deskpro107386[samtools.../samtools]; io_trace -r 'bam$' -S -- /tmp/samtools.dev view -c ~/scratch/data/9827_2#49.bam $R
...
    Num. seeks   	47493
    Num. read    	80083
    Bytes read   	2268340052

@ deskpro107386[samtools.../samtools]; io_trace -r 'bam$' -S -- samtools view -c ~/scratch/data/9827_2#49.bam $R
...
    Num. seeks   	51656
    Num. read    	294728
    Bytes read   	8084384944
```

It does however help when using the multi-region iterator.  I'll do more research and post those results separately.


So sorry, I don't think this is the answer either. :(   It looks like it's basically a CPU expensive way of avoiding using the linear index.  However I'm sure there's a way because htsjdk manages this fine.  I haven't looked at how it does it yet though.

daviesrob (commented):
Unfortunately this does erase much of the benefit from #1031 for alignments with placed, unmapped reads.  It is possible to get some, but not all, of the speed back by adjusting the `min_off` value in `hts_itr_query()` like this:
```
            if (idx->lidx[tid].offset
                && beg>>idx->min_shift < idx->lidx[tid].n
                && min_off < idx->lidx[tid].offset[beg>>idx->min_shift]) {
		hts_pos_t bidx = beg>>idx->min_shift;
                min_off = idx->lidx[tid].offset[bidx];
                if (unmapped) {
                    while (bidx > 0) {
                        if (idx->lidx[tid].offset[bidx] < min_off) break;
                        bidx--;
                    }
                    min_off = idx->lidx[tid].offset[bidx];
                }
            }
```
It searches backwards for the previous linear offset position, on the grounds that there must be at least one mapped read between there and the one we started at.

There may well be better ways to do it, and the above code may not be completely correct.

Comments of course apply to the `hts_itr_multi_bam()` version as well.

jkbonfield:
With the multi-region iterator things are better, as indicated above.  Some timings (in less gory detail):

Sparse: **Averaging** 1 region every 110Kbp of reference.
```
        1.10       dev       This PR
Time	10.423s    2.944s    6.730s
Seeks	1546	   7230      4463
Bytes	1456MB	   398MB     1003MB
```

Dense: **Averaging** 1 region every 1.2Kbp of reference.
```
        1.10       dev       This PR
Time	6.511s	   1.184s    0.919s
Seeks	4423       5468      4527
Bytes	119MB      147MB     122MB
```

So it's fastest of all on the dense querying.  However note that bytes read vs cpu time aren't well correlated.  1.10 vs dev branch show more bytes read in dev but much less CPU.  I'm guessing this is in part a difference between read and decompress as not using linear index vs read-decompress; read-discard (skip); read-decompress.  We'd need to profile it to count function calls to know what work load it's putting on the decompressor.

Either way, the difference between this vs dev is small compared to either of them vs 1.10.  So in this case the PR works well *if* we have a densely populated set of regions.

valeriuo:
Replaced by #1158

PR #1156, opened 2020-10-12
Fastq

jkbonfield:
This is just fastq - no fasta at present and the hts code bizarrely already stated fastq was sequencing data while fasta was uncategorised.  I decided to go with the flow there.

The basics of this are pretty straight forward.  The intention is simply to be able to transparently treat FASTQ as unmapped data as if it was SAM, BAM or CRAM.


jkbonfield:
Bizarrely I found factoring out the aux parser sped things up.  I guess it's to do with the complexity of that function, especially all the gotos for error handling.  Maybe the compiler just has an easier job. :)  So win-win there.

As for handling the format options, I've been using the newer state field (wasn't there in 2016) which I added for the multi-threaded SAM parser.  It's a convenient way of caching format specific things in, in this case a copy of the kseq instance and the various options.  Lots of talks & meetings today, but I hope to have writing side done in the next day or two.

jmarshall:
>  I did initially start with my own fastq parser, mainly because getting kseq to work was such a huge uphill struggle! I did eventually beat it into submission though so went via that route.

I would recommend **not** using kseq. Unlike other klib #defined types, neither `kseq_t` nor `kstream_t` (which it uses internally) do any name mangling based on the “template argument” type, so using them inside libhts interferes with their use in third-party or application code (e.g. _samtools/dict.c_) — at least in theory: these days all the functions are `static` but the type names are still an ODR violation. It was a glorious day when we eradicated the use of `kstream_t` inside the library (see 71c03b88edf892a93eae7bb7932a6da84858480a) and it may cause problems to reverse that.

Also kstream does no error checking.

(Several _*.c_ files that go into libhts do `#include "htslib/kseq.h"` already, but they do so only to get a definition of `KS_SEP_LINE` and do not instantiate any kstreams or kseqs.)

jkbonfield:
> I would recommend **not** using kseq. Unlike other klib #defined types, neither `kseq_t` nor `kstream_t` (which it uses internally) do any name mangling based on the “template argument” type, so using them inside libhts interferes with their use in third-party or application code (e.g. _samtools/dict.c_) — at least in theory: these days all the functions are `static` but the type names are still an ODR violation. It was a glorious day when we eradicated the use of `kstream_t` inside the library (see [71c03b8](https://github.com/samtools/htslib/commit/71c03b88edf892a93eae7bb7932a6da84858480a)) and it may cause problems to reverse that.

Ugh!  Now I see why we've avoided using it in htslib (bcftools has extensive usage of it plus some minor bits of samtools).  Maybe however we should put some warnings there so others don't naively, like me, assume it's the supported way of reading fastq files.

I admit it was a total royal PITA to use too.  A truely hideous API and virtually impossible to debug.  Thanks for the tip.   Culling it again will be my next joyous task then. :-)

jkbonfield:
Updating this PR again then.  I'll post when I'm happy with it once more.

jkbonfield:
Squashed to fewer commits.

I think this is pretty much complete.  The only possible adjustment would be the addition of fasta, but I'm not particularly convinced with the benefit of that.  The only use I can see would be to simplify "samtools fasta" as right now fasta and fastq subcommands are linked together with a minor tweak at time of writing.  That makes it tricky to rewrite using the newer htslib functionality.

Fixes #1157 (by dint of superceding it).

kevinjacobs-progenity:
Does `sam_open_mode` need explicit checks for `fastq.gz` and `fq.gz`?

daviesrob:
Rebased...

daviesrob:
Rebased again with some minor fixes and a couple of commits adding some extra error checking.

I notice there's not much documentation on the HTSlib side for how to use this.  What do you think might be the best way to add this?  Expand the doxygen comments in `htslib/sam.h`, or maybe add an `htslib-fastq(5)` man page with details on how the library supports these formats?

jkbonfield:
Acknowledged we always need new documentation, but we have no man pages at all on API.  We have some command line tool ones (bgzip.1) or file format ones (sam.5, vcf.5), but no .3 sections.  I'm not looking to document the fasta and fastq formats.  That's someone elses task IMO.

I'm not sure what we'd add to htslib/sam.h.  It doesn't have format specific doxygen content at the moment.  The main API functions are in hts.h, and they have indeed been updated to contain the various format specifiers, so I'm not sure what you're asking for here.

jkbonfield:
2nd commit message has a typo: "Catch more errors in sam_parse1()" should be "Catch more errors in fastq_parse1()".

daviesrob:
Errant commit message has been fixed.

There's a lot of detail here about how HTSlib will interpret fasta and fastq files which is currently only discoverable by inspecting the code.  Examples would be that adding `/1` or `/2` to the end of the name will make the read paired; and what happens it you use the `FASTQ_OPT_*` options.  HTSlib already has (fairly minimal) man pages for [sam](https://github.com/samtools/htslib/blob/develop/sam.5) and [vcf](https://github.com/samtools/htslib/blob/develop/vcf.5).  A similar one describing its interpretation of fasta and fastq data, and how the options affect it, would I think be helpful.

jkbonfield:
Most of this is already in the https://github.com/samtools/samtools/pull/1323 samtools-import.1 man page.  Why there?  Because htslib simply doesn't have anywhere to put it!   Htslib has never had this sort of documentation before.  It's always left a lot of the documentation to samtools man pages instead.  It's not something I like, but I went with the flow.

We have a major reorganisation of documentation coming at some point, so this PR probably isn't the time to start redefining documentation policy knowing it'll probably get rewritten again soon anyway.

jkbonfield:
Added some docs to the hts.h `enum hts_fmt_option`.

This is the right place as we need to document things like `FASTQ_OPT_AUX` being a comma-separated list of aux tag types ad `FASTQ_OPT_BARCODE` holding the name of the barcode tag ("BC") for CASAVA processing.  None of that is related to the FASTQ format itself, so I don't think a .5 man page section is appropriate.  See `man 5 intro` for information on what those man pages are meant to contain.

sam.c
```diff
@@ -3488,6 +3559,250 @@ int sam_set_threads(htsFile *fp, int nthreads) {
...
+
+        // Barcode, maybe numeric in which case we skip it
+        if (!isdigit_c(endptr[1])) {
+            barcode = endptr+1;
+            for (i = barcode - kc->s; i < kc->l; i++)
+                if (isspace_c(kc->s[i]))
+                    break;
+
+            kc->s[i] = 0;
+            barcode_len = i+1-6;
```
daviesrob, line 3788: Fuzzing found a bug here, as it assumes a fixed length for the part before the barcode.  If `control_bits` includes lots of digits (or padding) this assumption is broken leading to a read beyond the end of the comment string when it gets added to the aux data.  A simple reproducer is:

```
>r1 1:Y:0000000000001:CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC
ACGT
```

I think you can also pad with spaces if you want to as `strtol()` will eat them up.

daviesrob:
Thanks for adding the extra documentation, although I'm afraid I noticed a typo ("Bolean") in the part for `FASTQ_OPT_RNUM`.

sam.c
jkbonfield: Thanks.

I'm not so worried about `strtol()` eating spaces.  It means we accept some potentially invalid data (assuming they're not permitted), but as long as it doesn't lead to bugs elsewhere then it's OK.

I'm guessing we can just change `barcode_len` to subtract `barcode - kc->s` instead of a fixed 6, or similar.  I'll give it a try out.

daviesrob:
I've added a commit to fix a fuzzer-detected endless loop in `aux_parse()`.  For the record, the input that caused this was:

```
> 2`:i:833333333333333333333333333333333333s333333166666666666-2T
ACGT
```

I also added a couple of lines to reset `b->l_data` if the parser detects an error in lenient mode, so that any semi-formed aux record will be cleaned up before it attempts to read the next tag.

daviesrob:
Tidied up ready for merge, including removal of an accidental htscodecs rewind.  Please be very careful not to accidentally change submodules when adding commits (in particular if using `git commit -a`).

jkbonfield:
Looks good.  Apologies for the accidental submodule problem.

Although it does beg the question how it happened.  Yes I likely did a `commit -a`, but I hadn't changed my submodules.  I'm guessing this is some brokenness with `git pull` not updating things properly?  Is there a way to get it to work including submodules?  If so, is there a way this can be part of the gitconfig for htslib directly so we don't have to get users all modifying their own configs?

daviesrob:
My submodule keeps itself up to date by having `fetchRecurseSubmodules = true` in [`.gitmodules`](https://github.com/samtools/htslib/blob/c3ba3026/.gitmodules#L4).  Yours should have it too, but you might have to do something extra to get it to work if it wasn't there from the start.  Maybe `git submodule sync` might fix it?

jmarshall:
@daviesrob: Does that fetch the new commits, or does it also for you actually alter which commit is checked out in the submodule?

@jkbonfield: I suspect you want to set `submodule.recurse` in your global or htslib git-config. There doesn't appear to be a way to set this in _.gitmodules_, presumably as the correct behaviour is in the eye of the beholder. (Maybe Rob has this setting set already?)

daviesrob:
@jmarshall I'm not sure exactly what it does, but I do see `Fetching submodule htscodecs` when I do a `git pull`, which means it should have the right commits available for the branch I'm updating.  I do not have `submodule.recurse` set in git-config.

Issue #245, opened 2015-07-20
`hts_hopen` should have an exclusive mode


SamStudio8:
`hts_hopen` should provide functionality akin to `O_EXCL` in `open` or `x` in `fopen` to allow applications to open handlers to files with exclusivity to avoid race conditions like those demonstrated in samtools/samtools#432.

jmarshall:
You can already get this effect via `open(O_EXCL) / hdopen() / hts_hopen()`, but it would indeed probably be good if `hts_open()` and `hts_hopen()` followed glibc's `fopen()` mode extensions, such as `x`.

jmarshall:
If I'd realised this would take just one line (+doco), I would have implemented it back in July—sorry!

Added because it'll be handy for _bgzip.c_ overhauling too.

PR #1157, opened 2020-10-12
[Previous draft] Read/write FASTQ as unmapped SAM

jmarshall:
For entertainment value and comparison to PR #1156 (i.e. to see if there are any aspects here that would be useful in the other PR), here is the branch that has been gathering dust since [this conversation](https://twitter.com/jomarnz/status/788228002270285824).

### DIFF
--- a/htslib/hts.h
+++ b/htslib/hts.h
@@ -457,7 +457,7 @@ char *hts_format_description(const htsFormat *format);
   @param fn       The file name or "-" for stdin/stdout. For indexed files
                   with a non-standard naming, the file name can include the
                   name of the index file delimited with HTS_IDX_DELIM
-  @param mode     Mode matching / [rwa][bceguxz0-9]* /
+  @param mode     Mode matching / [rwa][bcefguxz0-9]* /
   @discussion
       With 'r' opens for reading; any further format mode letters are ignored
       as the format is detected by checking the first few bytes or BGZF blocks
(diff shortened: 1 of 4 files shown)


jkbonfield:
It maybe would have been useful if you could have put this in as a PR in 2016!  What other useful snippets have you been sitting on? ;-)

There are some useful things in there though that could be merged over.  I did initially start with my own fastq parser, mainly because getting kseq to work was such a huge uphill struggle!  I did eventually beat it into submission though so went via that route.

Writing is the bit I haven't got yet (and arguably samtools fastq covers this from the application side, although not the library side), but I was dissuaded from this course of action by others.  Do people think writing from the library is useful?  My instincts are yes.

htslib/hts.h
```diff
@@ -457,14 +457,15 @@ char *hts_format_description(const htsFormat *format);
   @param fn       The file name or "-" for stdin/stdout. For indexed files
                   with a non-standard naming, the file name can include the
                   name of the index file delimited with HTS_IDX_DELIM
-  @param mode     Mode matching / [rwa][bceguxz0-9]* /
+  @param mode     Mode matching / [rwa][bcefguxz0-9]* /
```
jkbonfield, line 460: What's the x and e in this string?  I can't see any support for these in the code, and they're undocumented below.  Some legacy thing?

jmarshall: They are passed through to `hopen()` and interpreted there (see #245), and actually are documented in the `hts_open()` blurb — look further below.

jkbonfield: I should have opened the file to look.  :)

I saw them documented in alphabetical order and github trimmed the surrounding text at just the wrong spot.  Duh!  Sorry and thanks for the explanation.

I'll be adding bits of this as I now have the go ahead for doing the writing support too, which will start from this.  Cheers.

jkbonfield:
Obsoleted by #1156

Issue #789, opened 2018-10-30
bam1_t build function


jkbonfield:
See Will Stokes' request at https://sourceforge.net/p/samtools/mailman/message/36454348/

We desire a way to construct a BAM record from the component pieces that isn't just a manual mess of bit/byte shuffling.  Most of it is easy, but some things such as creation of the sequence is complex.

CRAM has an internal function for this (bam_construct_seq), so looking at that and surrounding functions may be a suitable starting point.

anderskaplan:
Is this still relevant? If yes, I might give it a try.

daviesrob:
Yes, it would still be useful, thanks.  As James mentioned, [`bam_construct_seq`](https://github.com/samtools/htslib/blob/8016127/cram/cram_samtools.c#L45) could be used as a starting point, although you'd probably want to pass in a `bam1_t` pointer directly and the extra_len bit really should be automatic.

A separate interface for bulk adding aux tags would also be useful, although I've no idea what it should look like.

jkbonfield:
As Rob points out it probably needs something sorting regarding auxiliary data.  The way this is used in CRAM is the aux data is stitched together separately and has some special handling for the read-group (which is stored as an integer field in CRAM), but as it does know the length of the incoming aux it passes over the length in order to avoid a second realloc.
See the call here:  https://github.com/samtools/htslib/blob/80161279b755d83df6e0dbc52942f625df414452/cram/cram_decode.c#L2899

I think it probably should take a uint8_t *preformatted_aux and size_t aux_len fields which may be passed in as NULL/0 if you don't wish to use them.  Constructing aux is a different set of APIs maybe?  Either that or we additionally have a varargs style interface for specifying key/value pairs.   There are maybe some ideas around that in the SAM header API, eg: https://github.com/samtools/htslib/blob/develop/header.c#L1316

anderskaplan:
All right, thanks for the pointers! I'll dig into it and probably come back with more questions.

anderskaplan:
Hi again, I've done some research and come up with two alternatives and a few questions. I'd appreciate your input
to be able to move forward!

The function we want has two non-trivial jobs to do. The first is to manage the bam1_t data buffer, making
sure that it gets reallocated as necessary and writing the dynamic length fields into it. The second is to convert
the sequence string from a char array to a packed integer format.

The bam_construct_seq() function in cram_samtools.c seems to do what is needed: it sets all the fields in the bam1_t
[...]

PR #1159, opened 2020-10-18
Construct a BAM record from the component pieces

anderskaplan:
Added a bam_construct() function as requested and discussed in #789 .

I am a bit confused about what is and what isn't considered to be in the public API. Therefore I kept the existing bam_construct_seq() function with its current behavior, but replaced the implementation with a call to bam_construct(). If possible I'd like to remove that function altogether, as it's largely redundant.

This pull requests also fixes an issue with the return value of cram_to_bam() where the length of the aux data was counted twice.

### DIFF
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -2842,7 +2842,7 @@ static int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
     int bam_idx, rg_len;
     char name_a[1024], *name;
     int name_len;
-    char *aux;
+    char *aux, *aux_orig;
     char *seq, *qual;
     sam_hrecs_t *bfd = sh->hrecs;
 
@@ -2887,6 +2887,7 @@ static int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
         cr->len = 0;
     }
 
+
     if (fd->required_fields & SAM_QUAL) {
         if (!BLOCK_DATA(s->qual_blk))
             return -1;
@@ -2912,7 +2913,7 @@ static int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
     if (bam_idx == -1)
         return -1;
 
-    aux = (char *)bam_aux(*bam);
+    aux = aux_orig = (char *)bam_aux(*bam);
 
     /* Auxiliary strings */
     if (cr->aux_size != 0) {
@@ -2929,7 +2930,7 @@ static int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
         *aux++ = 0;
     }
 
-    return bam_idx;
+    return bam_idx + (aux - aux_orig);
 }
 
 /*
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -2842,7 +2842,7 @@ static int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
     int bam_idx, rg_len;
     char name_a[1024], *name;
     int name_len;
-    char *aux, *aux_orig;
+    char *aux;
     char *seq, *qual;
     sam_hrecs_t *bfd = sh->hrecs;
 
@@ -2887,7 +2887,6 @@ static int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
         cr->len = 0;
     }
 
-
     if (fd->required_fields & SAM_QUAL) {
         if (!BLOCK_DATA(s->qual_blk))
             return -1;
@@ -2913,7 +2912,7 @@ static int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
     if (bam_idx == -1)
         return -1;
 
-    aux = aux_orig = (char *)bam_aux(*bam);
+    aux = (char *)bam_aux(*bam);
 
     /* Auxiliary strings */
     if (cr->aux_size != 0) {
@@ -2930,7 +2929,7 @@ static int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
         *aux++ = 0;
     }
 
-    return bam_idx + (aux - aux_orig);
+    return bam_idx;
 }
 
 /*
--- a/cram/cram_samtools.c
+++ b/cram/cram_samtools.c
@@ -55,69 +55,20 @@ int bam_construct_seq(bam_seq_t **bp, size_t extra_len,
                       int64_t isize,
                       int len,
                       const char *seq,
-                      const char *qual) {
-    static const char L[256] = {
-        15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,
-        15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,
-        15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,
-        15,15,15,15,15,15,15,15,15,15,15,15,15, 0,15,15,
-        15, 1,14, 2,13,15,15, 4,11,15,15,12,15, 3,15,15,
-        15,15, 5, 6, 8,15, 7, 9,15,10,15,15,15,15,15,15,
-        15, 1,14, 2,13,15,15, 4,11,15,15,12,15, 3,15,15,
-        15,15, 5, 6, 8,15, 7, 9,15,10,15,15,15,15,15,15,
-        15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,
-        15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,
-        15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,
-        15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,
-        15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,
-        15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,
-        15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,
-        15,15,15,15,15,15,15,15,15,15,15,15,15,15,15,15
-    };
-    bam1_t *b = (bam1_t *)*bp;
-    uint8_t *cp;
-    int i, qname_nuls, bam_len;
-
-    //b->l_aux = extra_len; // we fill this out later
-
-    qname_nuls = 4 - qname_len%4;
-    bam_len = qname_len + qname_nuls + ncigar*4 + (len+1)/2 + len + extra_len;
-    if (realloc_bam_data(b, bam_len) < 0)
-        return -1;
-    b->l_data = bam_len;
-
-    b->core.tid     = rname;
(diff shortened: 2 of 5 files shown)


daviesrob:
Thanks for this, it generally looks good.

I think one improvement would be to add a check for `data_len` going over `INT32_MAX` which is the maximum possible value that can be stored in [`bam1_t::l_data`](https://github.com/samtools/htslib/blob/2264113e/htslib/sam.h#L248).  Unfortunately this isn't completely trivial because there's a possibility of overflow, especially on 32-bit platforms.  The easiest solution might be to keep track of the amount of space left after accounting for each item.  You could also have a more restrictive limit on `l_seq` (which I think should be `(INT32_MAX-4)*2/3 = 1431655762`, the 4 is a minimal `qname`) which would help ensure it doesn't overflow when working out how long it is.

On a more pedantic note, `qname_nulls` should really be `qname_nuls` as it's referring to the ASCII NUL character `\0` rather than the NULL pointer (which is why `l_extranul` ends in a single "l").

anderskaplan:
Nice, I'll add the validation for `data_len`.

The NUL vs NULL issue, well, I don't think many people make that distinction. According to [Internet RFC 20](https://tools.ietf.org/html/rfc20) from 1969, NUL is an abbreviation for NULL just like BEL is an abbreviation for ASCII 7, BELL, which makes the teletype go bing. 😄

jmarshall:
What Rob is saying is that the HTSlib source code does make that distinction.

anderskaplan:
Ok, I've changed the null to nul. I can't help that it looks like a spelling error to me, but it's more important to stick with the conventions of the code base.

daviesrob:
Thanks for the update.  It might be possible to defeat the test `subtract_check_underflow(n_cigar * 4, &limit)` on 32-bit platforms if `n_cigar * 4` overflows, however as the associated `cigar` array would have to be bigger than the available address space I expect it would cause a problem before you get to this point - so probably not worth worrying about too much.

Would you like to squash this down before I merge it?  Or if you prefer I can do it, but I'll either need to force-push on to your branch so Github records the merge correctly; or squash it into a single commit.

jmarshall:
Needs tests with both even and odd lengths of `seq`.

Some error returns (the one due to `realloc_bam_data()`) set `errno`. IMHO it would be good if they all did.

`sam_parse1()` doesn't validate that QNAME contains only valid characters, and HTSlib generally does not validate such things. Is it really useful for this function to do so?

A bit late to the party :smile: but as this function takes an already-set-up `bam1_t*` (BTW other functions in _htslib/sam.h_ just call this parameter `b`, not `bam`), calling it `bam_construct()` is not a great name especially for C++ programmers — _construct_ means allocating a new object when none existed before.

Better names would be `bam_assign()` (à la C++) or `bam_set1()` (à la HTSlib).

jkbonfield:
Coincidentally I was just testing the performance impact of this.  Specifically what's the impact on CRAM decoding.
I found that this PR is around 4% slower to decode CRAM than develop.  Looking deeper into it, `bam_construct` is 50% slower than the original `bam_construct_seq`, and the huge bulk of this is in the qname validation.

4% doesn't sound like a lot, but CRAM's got a lot going on so that's probably the best it could be.  On something tighter it may be closer to that 50% performance hit.

We *could* optimise the validation function.  I tried this which for me worked better, but it's still at best maybe only half the slow down recovered.

```
diff --git a/sam.c b/sam.c
index 7d04713..1397355 100644
--- a/sam.c
+++ b/sam.c
@@ -488,18 +488,48 @@ static void bam_cigar2rqlens(int n_cigar, const uint32_t *cigar,
 static int validate_qname(size_t l_qname, const char *qname)
 {
     // SAM specification: the query name must be on the format [!-?A-~]{1,254}
-    if (l_qname > 254) {
+    static char vchar[256] = {
+        // 00
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+        0, 1, 1, 1,  1, 1, 1, 1,   1, 1, 1, 1,   1, 1, 1, 1,
+        1, 1, 1, 1,  1, 1, 1, 1,   1, 1, 1, 1,   1, 1, 1, 1,
+        // 40
+        0, 1, 1, 1,  1, 1, 1, 1,   1, 1, 1, 1,   1, 1, 1, 1,
diff --git a/sam.c b/sam.c
index 7d04713..1397355 100644
--- a/sam.c
+++ b/sam.c
@@ -488,18 +488,48 @@ static void bam_cigar2rqlens(int n_cigar, const uint32_t *
cigar,
 static int validate_qname(size_t l_qname, const char *qname)
 {
     // SAM specification: the query name must be on the format [!-?A-~]{1,254}
-    if (l_qname > 254) {
+    static char vchar[256] = {
+        // 00
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+        0, 1, 1, 1,  1, 1, 1, 1,   1, 1, 1, 1,   1, 1, 1, 1,
+        1, 1, 1, 1,  1, 1, 1, 1,   1, 1, 1, 1,   1, 1, 1, 1,
+        // 40
+        0, 1, 1, 1,  1, 1, 1, 1,   1, 1, 1, 1,   1, 1, 1, 1,
+        1, 1, 1, 1,  1, 1, 1, 1,   1, 1, 1, 1,   1, 1, 1, 1,
+        1, 1, 1, 1,  1, 1, 1, 1,   1, 1, 1, 1,   1, 1, 1, 1,
+        1, 1, 1, 1,  1, 1, 1, 1,   1, 1, 1, 1,   1, 1, 1, 0,
+        // 80
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+        // C0
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+        0, 0, 0, 0,  0, 0, 0, 0,   0, 0, 0, 0,   0, 0, 0, 0,
+    };
+
+    if (l_qname > 254)
         return -1;
-    }
 
-    int i;
-    for (i = 0; i < l_qname; i++) {
-        if (qname[i] < '!' || qname[i] > '~' || qname[i] == '@') {
-            return -1;
-        }
+    int i, v[4] = {1,1,1,1};
+    const unsigned char *uq = (const unsigned char *)qname;
+    for (i = 0; i < (l_qname & ~3); i+=4) {
+        v[0] &= vchar[uq[i+0]];
+        v[1] &= vchar[uq[i+1]];
+        v[2] &= vchar[uq[i+2]];
+        v[3] &= vchar[uq[i+3]];
     }
+    for (; i < l_qname; i++)
+        v[0] &= vchar[uq[i]];
 
-    return 0;
+    v[0] &= v[1];
+    v[0] &= v[2];
+    v[0] &= v[3];
+
+    return v[0] ? 0 : -1;
 }
 
 static int subtract_check_underflow(size_t length, size_t *limit)
```

If I just commented out the call to validate_qname entirely my CRAM decode was only 0.7% slower, which is only a minor change (although still oddly slower; I'd expect no change really).

daviesrob:
`sam_parse1()` arguably should validate QNAMEs, and only doesn't because it prefers speed to strict specification conformance.  As this function generates reads de-novo, it's probably a good thing that it doesn't let you create incorrect names.

jkbonfield:
In that case we should have a second implementation for CRAM, because right now if there is a CRAM file with an illegal char in the name we can no longer decode it, which in turn means we can't even pipe it into a sed script to fix it either.

jkbonfield:
If we really care about this, my preference would be a validation function, which can optionally be turned on.  It's applied after decoding a BAM object so we can have a custom "samtools validate" or on-the-fly validate via one of the --input-fmt-option parameters.

jmarshall:
As for `bam_construct_seq()`: it's not in _htslib/*.h_ and it doesn't have `HTSLIB_EXPORT`, so it's not in HTSlib's public API. It **is** in the [io_lib](https://github.com/jkbonfield/io_lib) compatibility layer, but the two _cram_decode.c_ files have diverged enough that it probably doesn't matter.

I'd suggest dropping the _cram/cram_decode.c_ and _cram/cram_samtools.c_ changes from this PR, as they're not related to the primary purpose of adding a function. When this is merged, make a followup PR that changes `cram_to_bam()` to use your new function directly instead of `bam_construct_seq()`, fixes up `cram_to_bam()`'s return value, and nukes _cram/cram_samtools.c_ entirely.

Edited to add: also maybe split the function added in this PR into an internal function and a public function that does the validation and then calls the internal one; then the later PR can have `cram_to_bam()` calling the internal function.

daviesrob:
I doubt many cram files exist with invalid names, if any, given that most names are generated by sequencing software that generally gets it right.  I expect the most likely problem might be someone using an `@` somewhere in the middle (the first character can never be `@` thanks to its use in SAM headers).  In case there are any such files around the best solution may be, as @jmarshall suggests, have an internal function that lets you choose if you want to validate the names, then make the public API call it with name validation enabled and the cram one with it disabled.

anderskaplan:
Updated according to comments:
- Changed the name to bam_assign. All the examples similar to bam_set1 that I could find in the code base were on the form bam_set_<something>, so it seems assign is the safer choice.
- Removed the qname validation altogether, both because of its surprisingly high performance hit but also because it might not be the job of this function.
- Removed the changes to cram_to_bam() and cram_construct_seq(). Will post another PR with the updates to those once this is merged.
- Added test cases, setting errno.

jmarshall:
Thanks for that.

It seems to me that `bam_set1()` would fit in with `bam_init1`/`bam_destroy1`/`bam_copy1`/ `bam_dup1` and `sam_parse1`/`sam_format1`/`sam_read1`/`sam_write1` very nicely. (And `bam_` rather than `sam_` as like those other existing `bam_*1` functions it's purely about manipulating a `bam1_t`.) And `_set` paralleling all those `bam_set_somefield` functions (btw there are no HTSlib functions currently with “assign” in the name).

daviesrob:
Thanks for the updates.  It might be useful to keep the `l_qname <= 254` check, as it's quick and could avoid some bizarre problems if the `l_qname` field wraps.

anderskaplan:
The naming is always the hardest part 😄

daviesrob:
Thanks very much.  Now squashed and merged before anyone else asks to have the name changed.

jmarshall:
I took the liberty of pushing a followup commit to tidy up the test's temporary filename.

PR #1163, opened 2020-10-23
Prevent switch/case statement "fall through" warnings [trivial]

jmarshall:
[A trivial one that's been in a holding pattern since before the 1.11 release…]

Add comments indicating intentional fall-throughs, so that `gcc -Wextra` builds don't need `-Wno-implicit-fallthrough` as well.

(Clang understands only not-yet-standard attributes and annotations, not comment text, so there's little point in catering to its `-Wimplicit-fallthrough` warning, which isn't included in `‑Wall`/ `‑Wextra` in Clang.)


jmarshall:
Thanks Andrew.

PR #1164, opened 2020-10-24
Replace bam_construct_seq() with bam_set1()

anderskaplan:
This is the continuation of #1159. It changes cram_to_bam() to call bam_set1() instead of bam_construct_seq(), fixes up cram_to_bam()'s return value (aux data length counted twice), and removes the now redundant cram/cram_samtools.c.

Note: there are other declarations in cram_samtools.h which maybe should be removed as well, such as the bam_seq_t type alias. Please advise!


anderskaplan:
The failed appveyor build does not seem related to the code changes, and unfortunately I don't see any way to retrigger the build.

jkbonfield:
The origin of cram_samtools.h is it's a wrapper file making io_lib's CRAM code work with htslib/samtools APIs and types.  This cram code originated here: https://github.com/jkbonfield/io_lib/tree/master/io_lib

Initially we wanted to keep the code between the two libraries as consistent as possible, while development and bug fixes got worked on, but it's just diverged too much for that to really be a concern any more.  Certainly once I went through replacing all use of my hash table with Heng's it was already looking radically different.  Now it's just not worth worry about.

So the different data types and API names probably needs fixing at some point, but I'd argue it's a different issue.  At least its own commit(s), but possibly a new PR too.

I'll take a look at the changes here, but I'm sure it's pretty straight forward.

Thanks for your work on this.

cram/cram_decode.c
```diff
@@ -2896,24 +2895,22 @@ static int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
...
+                       cr->ncigar, &s->cigar[cr->cigar],
+                       cr->mate_ref_id, cr->mate_pos - 1, cr->tlen,
+                       cr->len, seq, qual,
+                       cr->aux_size + rg_len);
+    if (bam_idx < 0) {
+        return bam_idx;
+    }
+
+    // mark the buffer space allocated for aux data as containing actual aux data
+    (*bam)->l_data += cr->aux_size + rg_len;
```
jmarshall, line 2910: Better to increment `…->l_data` individually in the two `if` blocks below so they're self-contained and if a third item is added later the changes needed are more self-contained.

On successful return from `bam_set1()`, `bam_idx` is the same as `…->l_data` and the incrementing maintains that. So IMHO you should change the `return` below to `return …->l_data` so there's no need to update `bam_idx`. (And then you can rename `bam_idx` to `ret` and just use it to check the return value from `bam_set1()` — I have no idea what the name `bam_idx` in the original code was trying to imply!).

anderskaplan: ok!

jkbonfield:
Thanks for this.  I took the liberty of amending your commit to include removing cram_samtools.c from htslib.mk too.

Rebased, squashed and merged as 44787d9800bba41e463e4db11c2280fac3cdbb7f

PR #1166, opened 2020-10-28
Fix AppVeyor CI

jkbonfield:
The keys used to sign MSYS package have changed, so we have to update the key store first.

The next problem is that newer MSYS packages are compressed with zstd, but the installed pacman cannot handle these.  We  eed to explicitly update pacman first.

Then finally we can install our build system.

Note none of this is strictly necessary.  The alternative solution is to comment out all pacman updates as the base image has sufficient compilers and libraries to build and pass the test harness.

The main difference is the lack of libcurl, which then means we don't test compilation of the components that depend on it.  We don't test these anyway, but it's useful to know they build at least.  However if we hit problems in the future due to the explicit path of those key files then that's a viable temporary workaround.

(It may be that AppVeyor also updates their base image at some point, in which case the key shenanigans can be deleted.)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -18,13 +18,25 @@ skip_tags: true
 #    - docs/*
 #    - '**/*.html'
 
-# We use Mingw/Msys, so use pacman for installs
+# We use Mingw/Msys, so use pacman for installs.
+# Note if we want faster builds, all of the pacman bits can be removed.
+# The base image has sufficient support to compile and pass, although it's
+# not testing building with libcurl as that is absent.
 install:
   - set HOME=.
   - set MSYSTEM=MINGW64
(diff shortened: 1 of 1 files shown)


daviesrob:
appveyor/ci#3513 [suggests](https://github.com/appveyor/ci/issues/3513#issuecomment-680128709) switching to the `Visual Studio 2019` image to get an up-to-date pacman.  Yes, I know we're not using Visual Studio, see https://www.appveyor.com/docs/build-environment/ for the available Windows images - we were using `Visual Studio 2015`.  I think that solution may have more longevity than bootstrapping a working pacman ourselves.

I've tested it in my branch https://github.com/daviesrob/htslib/tree/appveyor_img, and managed to get a [successful build](https://ci.appveyor.com/project/daviesrob/htslib/builds/36095626).

daviesrob:
I turned my branch into #1172

PR #1170, opened 2020-11-02
Added some simple introspection functions.

jkbonfield:
These permit testing of explicit features, e.g.

    if (htslib_test_feature(HTS_FEATURE_PLUGINS)) {
        ...
    }

or capturing the entire feature bit-field via htslib_features().
(This function may be redundant so perhaps is a candidate for culling?).

This also permits querying compilation details: CC, CFLAGS, LDFLAGS and CPPFLAGS.

Finally there is a htslib_feature_string() function which is mainly for verbose feature printing rather than programmatic parsing, but it conveniently also permits us to get features from a library simply by running "strings" on the binary.

If accepted, I can work on an equivalent samtools command to report its own data along with the htslib it links against.  How should we access this? `samtools version` maybe for more detail than the minimal version data in the usage statement?


jkbonfield:
Apologies, but I'll need someone with a Mac to do the final fixing of the Makefile for this.  I have no remote login permission for our work macs (as they stupidly all require you to be sat infront of it, duh!).

For reference, line 337 of Makefile is currently:

```
hts_os.o hts_os.pico: TMP_CPPFLAGS := $(CPPFLAGS)                               
```

:= is a GNU extension.  The POSIX syntax is ::= (https://www.gnu.org/software/make/manual/html_node/Flavors.html#Flavors) which is what I started with, but that doesn't work either on Macs.  Maybe this just isn't possible on whatever make system Apple provides?

jmarshall:
> Unknown why, but the POSIX syntax fails on MacOS X.

It's because the system make installed by Apple is based on GNU Make 3.81 — for their usual anti-GPLv3-jihad reasons — and that pre-dates the invention of the POSIX form of `::=`.

There is precedent for this sort of thing with _plugin.o_ in _config.mk.in_ so you could follow what that does. However in _Makefile_ I'd probably suggest generating it to a C source file that you then `#include` (à la _version.h_) instead.

This would be trivial if configure is run, so this somewhat requires a decision on whether to finally remove non-configure builds…

jmarshall:
Without expressing an opinion on whether all this is desirable: _hts_os.c/.h_ are for papering over platform-specific deficiencies, so are the wrong place for this. (Essentially no third party code that doesn't care about being portable to weird platforms should ever need to `#include <htslib/hts_os.h>`.)

You should probably just commit the trivial `hts_lrand48()` typo fix independently IMHO.

daviesrob:
I don't think the `hts_lrand48()` fix is completely trivial.  I suspect it may have ABI implications (although on platforms where it does, it probably doesn't work correctly).  I would agree it might be best in a different PR.

jkbonfield:
Oops yes I forgot about my need to hack hts_lrand48.  I was going to put that in a separate commit!  It'll have no ABI issues IMO as it's only changing on broken systems where it's already broken anyway.

I did look at how plugin.o was defined, and indeed copied that syntax from there.  The difference though is the requirement for := instead of = to avoid a recursively defined variable.  I can't include `$CPPFLAGS` in the definition for `CPPFLAGS` without explicitly utilising a non-recursive temporary variable.  It's questionable over whether we want such things to be quizzed, but my hope is with a proper introspection method we can add it to the list of questions for when people raise issues.  Ie not "which version are you using" but rather "please paste the output from `samtools version`" or similar.

I did also think where best to put this.  Initially I was going to put them in hts.c, but then saw the statement at the top that it is "format-neutral I/O, indexing, and iterator API functions" which this introspection work clearly is not.  No other file matched the bill bar the OS specific stuff, which arguably this is - it's build specific anyway.  IMO it's a good home.   I'd agree that maybe we should put the function declarations somewhere else though.

jkbonfield:
> This would be trivial if configure is run, so this somewhat requires a decision on whether to finally remove non-configure builds…

A lot would be trivial if we could cull the split-brain personility of "make or no-make".  I've been arguing this for years so +1 from me. :-)

jkbonfield:
Removed the lrand fix, squashed and rebased.

NB this won't compile until after #1171 is merged, after which we can rebase off develop again.

jmarshall:
I had a number of further comments about the proposed API here. Also the PR had not been revised to take advantage of #1187 which was applied so that one of the _Makefile_ hacks in this could be simplified.

I realise that I should have made those comments over the New Year break. Still it would be appreciated if some warning could be given when long-untouched PRs are about to be merged.

PR #1171, opened 2020-11-02
Fix a long-standing typo in hts_lrand48.

jkbonfield:
Note: this function is never used.  It was only added for reasons of completeness when we were putting in hts_drand48, used by htslib/ksort.h.

### DIFF
--- a/hts_os.c
+++ b/hts_os.c
@@ -49,7 +49,7 @@ HTSLIB_EXPORT
 double hts_drand48(void) { return drand48(); }
 
 HTSLIB_EXPORT
-double hts_lrand48(void) { return lrand48(); }
+long hts_lrand48(void) { return lrand48(); }
 #endif
 
 // // On Windows when using the MSYS or Cygwin terminals, isatty fails


jmarshall:
Okay, this was not completely trivial as it turns out :smile:

Fortunately no true ABI implications either as it's not the header file that was wrong. At worst it's a bug fix in the results of a mostly unused function. At least on the common platforms, by luck on return from `hts_lrand48()` the return value will have been in both the int and float return registers — so it has been returning the right result by luck anyway, which will have contributed to no-one noticing for three years.

> Note: this function is never used. 

Never used in htslib/samtools/bcftools. It may have been used by other third-party code.

jkbonfield:
I'd already done some basic github searches and it turns up absolutely nowhere.

I suspect it really is 100% unused.

jkbonfield:
> At worst it's a bug fix in the results of a mostly unused function. At least on the common platforms, by luck on return from `hts_lrand48()` the return value will have been in both the int and float return registers — so it has been returning the right result by luck anyway, which will have contributed to no-one noticing for three years.

Oh I see what you mean.  The usual implementation would be call lrand48 (result in rax) and then migrate that to double via something like "cvtsi2sdq       %rax, %xmm0".  The calling function with the proper declaration assumes it's long and just reads %rax anyway.

That works, *but* it's compiler specific and fails with clang optimisations.  Eg:

```
@ deskpro[/tmp]; cat dr.c
#include <stdlib.h>
double hts_lrand48(void) { return lrand48(); }

@ deskpro[/tmp]; cat r.c
#include <stdio.h>
#include <stdlib.h>

long hts_lrand48(void);

int main(void) {
    srand48(10);

    int i;
    long l;
    for (i = 0; i < 10; i++) {
	l = hts_lrand48();
	printf("%ld\n", l);
    }
}

@ deskpro[/tmp]; clang70 -c dr.c; clang70 r.c dr.o; ./a.out
1887318415
1708981709
1032568723
1128874726
986043905
2030393807
1387591810
1257232211
1374034427
2054885217

@ deskpro[/tmp]; clang70 -O2 -c dr.c; clang70 r.c dr.o; ./a.out
0
1
2
3
4
5
6
7
8
9
```

Comically non-random!

I still just view this as a bug fix and not an ABI change.  The ABI is 100% the same, but our random number generator is now more random.

daviesrob:
Agreed, given the mismatch between header and definition can result in undefined behaviour this is really a bug fix.

PR #1175, opened 2020-11-04
Add Cirrus-CI integration

jkbonfield:
This is a replacement for Travis, but it doesn't remove it (yet) so we can compare side by side for a while.  Compared to Travis is lacks the s390x test.

However we have gcc:latest (Debian),  clang (via ubuntu:devel) and MacOS X (clang only).  The mix of maintainer-checks, with/without config, libdeflate, pedantic and address sanitizer are sprinkled across the various environments so we don't have a huge matrix.  We also test libdeflate integration from local source (Debian, MacOS) and from OS installed package (Ubuntu).

Everything launches in parallel bar the 2nd MacOS X instance as it appears we're limited to just 1.

Other possibilities could include pretty much anything on docker-hub, so we could consider 32-bit builds if we feel it's worth it.
That also means we have the potential for custom containers with the prerequisites already installed so we have a leaner instance startup.  The ubuntu ones spend maybe 25-30% of their time installing compilers and libraries.

Example builds can be seen here:

https://cirrus-ci.com/build/5071567057059840

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,166 @@
+# Note we have a maximum of 16 CPUs available, so adjust our
+# builds so we can start all concurrently without needing to schedule.
+
+# Sadly though there is still a finite limit to macOS of one instance.
+# Can we cull our Mac test to just one instance?
+
+timeout_in: 10m
+
+#--------------------------------------------------
+# Template: build libdeflate dependency
+
+libdeflate_template: &LIBDEFLATE
+  libdeflate_script: |
+    if test "x$USE_LIBDEFLATE" == "xyes"; then
+      pushd "$HOME"
+      git clone --depth 1 https://github.com/ebiggers/libdeflate.git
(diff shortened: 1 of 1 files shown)


daviesrob:
Looking at the CPU and memory graphs on the results, it looks like we rarely go above 2 cpus and only use a tiny amount of the available memory.  Given that, it might be a good idea to use:
```
cpu: 2
memory: 1G
```

I don't think it will change the run time too much, but it might ensure our jobs get scheduled faster at times of high demand.

jkbonfield:
Agreed.  I've already tweaked the Samtools version to do this and plan to go back and modify the htslib one.

I don't see how to tweak these for the Mac instances though.  It may not be possible there.

jkbonfield:
A few minor tweaks squashed into one commit

- MacOS X vs macOS
- The typo Rob found above
- CPU / mem limits
- Centos build
- Improved formatting

PR #1180, opened 2020-11-10
Fix the data pointer of hts_itr_next.

### LINKED ISSUE #1179: NULL pointer on using bam_itr_next
After updating from htslib v1.9 to v1.11 we are getting regular crashes on usage of `bam_itr_next`. We are always using indexed BAM files:
```c
htsFile *fp_in = hts_open(bam_file,"rb");
hts_idx_t *idx_in = bam_index_load(bam_file);
hts_itr_t *iter = sam_itr_queryi(idx_in, tid_t, beg_t, end_t);
bam_itr_next(fp_in, iter, b);
```

This gives us now a crash due to NULL of `fp` in sam.c:958 in the call stack:

sam.h:1083
```c
#define bam_itr_next(htsfp, itr, r) hts_itr_next((htsfp)->fp.bgzf, (itr), (r), 0)
```

hts.c:3320-3335
```c
int hts_itr_next(BGZF *fp, hts_itr_t *iter, void *r, void *data)
{
    int ret, tid;
    hts_pos_t beg, end;
    if (iter == NULL || iter->finished) return -1;
    if (iter->read_rest) {
        if (iter->curr_off) { // seek to the start
            if (bgzf_seek(fp, iter->curr_off, SEEK_SET) < 0) {
                hts_log_error("Failed to seek to offset %"PRIu64"%s%s",
                              iter->curr_off,
                              errno ? ": " : "", strerror(errno));
                return -2;
            }
            iter->curr_off = 0; // only seek once
        }
        ret = iter->readrec(fp, data, r, &tid, &beg, &end);
```

Via

sam.c:1158-1166
```c
hts_itr_t *sam_itr_queryi(const hts_idx_t *idx, int tid, hts_pos_t beg, hts_pos_t end)
{
    const hts_cram_idx_t *cidx = (const hts_cram_idx_t *) idx;
    if (idx == NULL)
        return hts_itr_query(NULL, tid, beg, end, sam_readrec_rest);
    else if (cidx->fmt == HTS_FMT_CRAI)
        return cram_itr_query(idx, tid, beg, end, sam_readrec);
    else
        return hts_itr_query(idx, tid, beg, end, sam_readrec);
```

`iter->hts_readrec_func` is set to `sam_readrec`:

sam.c:954-958
```c
static int sam_readrec(BGZF *ignored, void *fpv, void *bv, int *tid, hts_pos_t *beg, hts_pos_t *end)
{
    htsFile *fp = (htsFile *)fpv;
    bam1_t *b = bv;
    fp->line.l = 0;
```

Thus, we suspect that the macro of `bam_itr_next` is wrongly defined by setting the `data` parameter of `hts_itr_next` to `0`.

Question to the maintainers: Are we (read: the users of htslib) supposed to use the generic `sam_*` API from now on - and in favour of the old-ish `bam_*` API?

valeriuo:
`bam_itr_next` now sets the `data` argument of `hts_itr_next` to the BAM file pointer (of `htsFile *` type).

Fixes #1179

### DIFF
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -1112,7 +1112,7 @@ int bam_set_qname(bam1_t *b, const char *qname);
 #define bam_itr_destroy(iter) hts_itr_destroy(iter)
 #define bam_itr_queryi(idx, tid, beg, end) sam_itr_queryi(idx, tid, beg, end)
 #define bam_itr_querys(idx, hdr, region) sam_itr_querys(idx, hdr, region)
-#define bam_itr_next(htsfp, itr, r) hts_itr_next((htsfp)->fp.bgzf, (itr), (r), 0)
+#define bam_itr_next(htsfp, itr, r) sam_itr_next((htsfp), (itr), (r))
 
 // Load/build .csi or .bai BAM index file.  Does not work with CRAM.
 // It is recommended to use the sam_index_* functions below instead.


jmarshall:
Historically this code was like this because for BAM files `hts_itr_next()` used its `BGZF *fp` argument (and ignored its `void *data` argument) and called `bam_readrec()` which used `bam_read1()` with the `BGZF *fp` argument. Meanwhile for CRAM files `hts_itr_next()` used the `void *data` argument but ignored the `BGZF *fp` argument.

This changed — and this bug was introduced — when iterating over SAM files was added and `bam_readrec()` (using `bam_read1()` and `BGZF *fp`) was replaced by `sam_readrec()` (using `sam_read1()` and `void *data` aka `htsFile *`).

Since then for BAM files, `hts_itr_next()` has mixed up seeking on the `BGZF *` with reading from the `htsFile *`. This works more by luck than design!

So IMHO this bug report is an indication that this code needs a careful review and, if moving to `htsFile *` everywhere in `hts_itr_next()`, streamlining and removal of the legacy `BGZF *` parameter (ABI issues notwithstanding). And/or more heavily deprecating the legacy `bam_itr_*` functions.

daviesrob:
Well, the `bam_itr_*` interface has now effectively gone as it just calls the `sam_itr_*` equivalents.  You're right, though, the iterator code is quite hairy and would benefit from some simplification.

PR #1182, opened 2020-11-25
Adjust parser for undefined CIGARs

valeriuo:
The newly added `sam_parse_cigar` and `bam_parse_cigar` methods can now handle unavailable/undefined CIGAR strings (`*`). The returned value (number of CIGAR operators) in this case is still 0. However, the return type of both methods was changed to signed integer, so that errors can be reported with `-1` code.
Tests were added for `sam_parse_cigar`.

Related to #1169

### DIFF
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -29,6 +29,7 @@ DEALINGS IN THE SOFTWARE.  */
 
 #include <errno.h>
 #include <stdint.h>
+#include <sys/types.h>
 #include "hts.h"
 #include "hts_endian.h"
 
@@ -1108,12 +1109,12 @@ int bam_set_qname(bam1_t *b, const char *qname);
  @param  in      [in]  pointer to the source string
  @param  end     [out] address of the pointer to the new end of the input string
                        can be NULL
- @param  a_cigar [out]  address of the destination uint32_t buffer
+ @param  a_cigar [in/out]  address of the destination uint32_t buffer
  @param  a_mem   [in/out]  address of the allocated number of buffer elements
  @return         number of processed CIGAR operators; -1 on error
  */
 HTSLIB_EXPORT
-ssize_t sam_parse_cigar(const char *in, char **end, uint32_t **a_cigar, uint32_t *a_mem);
+ssize_t sam_parse_cigar(const char *in, char **end, uint32_t **a_cigar, size_t *a_mem);
 
 /*! @function
  @abstract  Parse a CIGAR string into a bam1_t struct
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -1110,10 +1110,10 @@ int bam_set_qname(bam1_t *b, const char *qname);
                        can be NULL
  @param  a_cigar [out]  address of the destination uint32_t buffer
  @param  a_mem   [in/out]  address of the allocated number of buffer elements
- @return         number of processed CIGAR operators; 0 if error
+ @return         number of processed CIGAR operators; -1 on error
  */
 HTSLIB_EXPORT
-size_t sam_parse_cigar(const char *in, char **end, uint32_t **a_cigar, uint32_t *a_mem);
+ssize_t sam_parse_cigar(const char *in, char **end, uint32_t **a_cigar, uint32_t *a_mem);
 
 /*! @function
  @abstract  Parse a CIGAR string into a bam1_t struct
@@ -1121,10 +1121,10 @@ size_t sam_parse_cigar(const char *in, char **end, uint32_t **a_cigar, uint32_t 
  @param  end     [out] address of the pointer to the new end of the input string
                        can be NULL
  @param  b       [in/out]  address of the destination bam1_t struct
- @return         number of processed CIGAR operators; 0 if error
+ @return         number of processed CIGAR operators; -1 on error
  */
 HTSLIB_EXPORT
-size_t bam_parse_cigar(const char *in, char **end, bam1_t *b);
+ssize_t bam_parse_cigar(const char *in, char **end, bam1_t *b);
 
 /*************************
  *** BAM/CRAM indexing ***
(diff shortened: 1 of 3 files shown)


sam.c
```diff
@@ -2150,8 +2150,8 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
     if (*p != '*') {
         uint32_t *cigar = NULL;
         int old_l_data = b->l_data;
-        uint32_t n_cigar = bam_parse_cigar(p, &p, b);
-        if (!n_cigar || *p++ != '\t') goto err_ret;
+        int32_t n_cigar = bam_parse_cigar(p, &p, b);
```
jkbonfield, line 2153: Being nit-picky this should be ssize_t, but frankly the world will have broken long before we get to 2 billion cigar elements!  Infact I'm not sure it's possible in BAM as the bam record would overflow in length.

I'd just go with an unsized "int" or a specific "ssize_t" if you feel inclined to be pedantic.

test/sam.c
```diff
@@ -2149,6 +2149,27 @@ static void test_bam_set1_write_and_read_back()
...
+    char *cig = "*";
+    char *end;
+    uint32_t m = 0;
+    int32_t n;
+    n = sam_parse_cigar(cig, &end, &buf, &m);
+    VERIFY(n == 0 && m == 0 && (end-cig) == 1, "failed to parse undefined CIGAR");
+    cig = "2M3X1I10M5D";
+    n = sam_parse_cigar(cig, &end, &buf, &m);
+    VERIFY(n == 5 && m > 0 && (end-cig) == 11, "failed to parse CIGAR string: 2M3X1I10M5D");
+    n = sam_parse_cigar("722M15D187217376188323783284M67I", NULL, &buf, &m);
```
jkbonfield, line 2164: I'm not sure what the point is of this test.  187217376188323783284 is 68 bits long. 

Cigars are held inside htslib as uint32_t with 4 bits being the type, so the maximum length of cigar op we currently handle is 268435455 (2^28-1).   CRAM could handle longer lengths (if #976 got merged), but I think we rejected that idea as the in-memory htslib struct would need fixing first.

I'm sure we discussed changing the cigar type when we were doing all of our sweeping ABI changes and it was rejected, so I think we're committed to having excessively long cigar lengths as a series of concatenated smaller lengths using the same op type.  (I don't see a reason to support that though until we get reads long enough and accurate enough to generate such data.  We're considerably far off that still.)

jkbonfield: This does raise the question though why the test isn't:

```
VERIFY(n == -1, "failed to flag CIGAR string with long op length: 722M15D187217376188323783284M67I");
```

The documentation for `sam_parse_cigar` says -1 on error and "number of processed CIGAR operators" otherwise.  If the CIGAR has been rejected (even if syntacically valid) then it should probably return -1 in this case.

htslib/sam.h
```diff
@@ -1110,21 +1110,21 @@ int bam_set_qname(bam1_t *b, const char *qname);
                        can be NULL
  @param  a_cigar [out]  address of the destination uint32_t buffer
  @param  a_mem   [in/out]  address of the allocated number of buffer elements
- @return         number of processed CIGAR operators; 0 if error
+ @return         number of processed CIGAR operators; -1 on error
  */
 HTSLIB_EXPORT
-size_t sam_parse_cigar(const char *in, char **end, uint32_t **a_cigar, uint32_t *a_mem);
+ssize_t sam_parse_cigar(const char *in, char **end, uint32_t **a_cigar, uint32_t *a_mem);
```
jmarshall, line 1116: By luck _sam.h_ currently gets it via _kstring.h_, but in case that changes: need to add `#include <sys/types.h>` at the top of _sam.h_ to ensure `ssize_t` is declared.

htslib/sam.h
```diff
@@ -1110,21 +1110,21 @@ int bam_set_qname(bam1_t *b, const char *qname);
                        can be NULL
  @param  a_cigar [out]  address of the destination uint32_t buffer
```
jmarshall, line 1111: a_cigar is [in/out] presumably?

jmarshall:
While we're revisiting the signatures of these new functions,

```diff
-size_t sam_parse_cigar(const char *in, char **end, uint32_t **a_cigar, uint32_t *a_mem);
+ssize_t sam_parse_cigar(const char *in, char **end, uint32_t **a_cigar, uint32_t *a_mem);
```

using such a low-level type for the user's buffer length is not ideal: the proper type for this in C would be `size_t *a_mem`.

Also the two utility functions should not cast away `const` so soon, as this invites bugs. `end` needs to be a non-const `char *` for the same reasons as in `strtol()` etc, but it would be better to use a `(char *)` cast only when actually writing to `end`. So `read_ncigar()` should use only a `const char *` and `parse_cigar()` should primarily use a `const char *p` (see e.g. jmarshall/htslib@a12d4d447cd3ca089db05dcc35144818339a4911).

htslib/sam.h
valeriuo: Initially, I thought of just being `[out]`, as the information contained by the buffer is generated inside the method, but the pointer itself is being read. So, I guess a case can be made for `[in]`.

htslib/sam.h
valeriuo: Thanks for the catch.

htslib/sam.h
jmarshall: It's `in` because the pointed-to variable needs to already have a meaningful value before the call (this is how it differs from `end`, whose pointed-to variable may even be uninitialised). It's `out` because the pointed-to value might be different after the call.

Therefore it's `[in/out]`, the same as `a_mem`.

jkbonfield:
Rebased & squashed, merged as 2056490488b81169bf69beb5ac835cd22dc74a42

PR #1196, opened 2020-12-23
fix a serious bug when get next overlapped matched base in the case t…

wulj2:
…he match base is in just the first one in the stretch of match and this stretch of match is after some other stretch of non-match sequences(such as ins, del, .etc)
fixed#1195


valeriuo:
Thank your for the fix. However, I've made [one](https://github.com/valeriuo/htslib/commit/2d034837bff3ebd9cba6318a9ace585d2a0bbdb9) which is simpler. Instead of keeping track of a previous matching segment, it resets `icig` to -1 every time, so that the first base of each matching segment starts from 0.
I rebased and squashed your commits and added my changes on top of yours, to preserve your authorship.

wulj2:
wow, that is a really clever fix, I love it so much.

wulj2:
[Elegantly fixed](https://github.com/valeriuo/htslib/commit/2d034837bff3ebd9cba6318a9ace585d2a0bbdb9) by @valeriuo

PR #1200, opened 2021-01-05
Removed knet.

jkbonfield:
Recent bug reports [personal communication] in knet is the straw that broke the camels back.  I've been wanting to do this for years though!

There's little benefit to keeping it (basically FTP and unsecured http which has largely gone out of fashion), while curl is already required for htsget, refget (CRAM ref cache), AWS S3 and GCS protocols.  Curl is available on all platforms we support.


INSTALL
```diff
@@ -41,9 +41,10 @@ a development ('-dev' or '-devel') package separate from the main library.
 Disabling libbzip2 and liblzma will make some CRAM files unreadable, so
 is not recommended.
 
-Using libcurl provides HTSlib with better network protocol support, for
-example it enables the use of https:// URLs.  It is also required if
-direct access to Amazon S3 or Google Cloud Storage is enabled.
+Using libcurl provides HTSlib withnetwork protocol support, for
+example it enables the use of ftp and https:// URLs.  It is also
+required if direct access to Amazon S3 or Google Cloud Storage is
+enabled.
```
jmarshall, line 47: Fix typo (missing space), and suggested spelling out all the important URL schemes:

```suggestion
Using libcurl provides HTSlib with network protocol support, for example it
enables the use of ftp://, http://, and https:// URLs.  It is also required
if direct access to Amazon S3 or Google Cloud Storage is enabled.
```

Makefile
```diff
@@ -537,8 +533,8 @@ test-shlib-exports: header-exports.txt shlib-exports-$(SHLIB_FLAVOUR).txt
 	( echo "Error: Found unexported symbols (listed above)" ; false )
 
 # Extract symbols that should be exported from public headers using ctags
-# Filter out macros in htslib/hts_defs.h, and knet_win32_ functions that
-# aren't needed on non-Windows platforms.
+# Filter out macros in htslib/hts_defs.h that aren't needed on non-Windows
+# platforms.
```
jmarshall, line 537: Suspect “that aren't needed on non-Windows platforms” pertained only to `knet_win32_` and can disappear too.

hfile.c
```diff
@@ -1061,7 +1061,6 @@ static int load_hfile_plugins()
     hfile_add_scheme_handler("data", &data);
     hfile_add_scheme_handler("file", &file);
     hfile_add_scheme_handler("preload", &preload);
-    init_add_plugin(NULL, hfile_plugin_init_net, "knetfile");
```
jmarshall, line 1064: Also nuke the `hfile_plugin_init_net` declaration from _hfile_internal.h_.

INSTALL
jkbonfield: Odd.  I checked the original and the missing space is actually an entire missing " better " word.  Must have been some odd finger slippage.

Anyway thanks for the updates.  I'll amend.

Makefile
jkbonfield: I was unsure of that and ummed and ahhed over it, but happy to remove.  I did start to edit the associated perl script, but figured it's harmless for it to check for functions that don't exist and I'm more likely to break it in doing so.  It's perhaps still something that should be changed though.

jmarshall:
Huzzah. This also deserves a _NEWS_ entry, and there are some _knetfile/knet_win32_ references in _test/header_syms.pl_ and _test/maintainer/check_spaces.pl_ that could go.

I guess some will consider the removal of _htslib/knetfile.h_ and the functions declared there to be a removal of vital public API functions. Pedantically correct, but… sigh… noone should be using these functions directly.

So either this needs a soversion bump, or perhaps as a transition before removal during a later soversion bump in a future release, replace the implementations of `knet_open()` etc with dummy versions that just call through to `hopen()` (and hence newly fail if libcurl was not configured — tough!) and have them maintain `knetFile::offset` and maybe `knetFile::fd` I guess (for the sake of the `knet_tell()`, `knet_fileno()` macros)…

INSTALL
jmarshall: Removing “better“ here was clearly the right thing to do :smile: Now I'm off to get my third coffee of the morning, to fend off my own befuddlement…

jkbonfield:
That's a good point - knetfile.h was a public header.

Maybe it should have stub functions (static inlines?).

jmarshall:
Annoyingly, in the absence of a soversion bump they would need to be real symbols in _libhts.so_.

Which would mean keeping this PR as a draft to be applied in the eventual happy future, and making a new smaller one that basically keeps the infrastructure but replaces all the function bodies within _knetfile.c_. Alas…

jkbonfield:
It appears to be trivial to get *most* of the ABI to still exist, eg:

```
typedef struct knetFile_s knetFile;
knetFile *knet_open(const char *fn, const char *mode) {
    return (knetFile *)hopen(fn, mode);
}

knetFile *knet_dopen(int fd, const char *mode) {
    return (knetFile *)hdopen(fd, mode);
}

ssize_t knet_read(knetFile *fp, void *buf, size_t len) {
    return hread((hFILE *)fp, buf, len);
}

off_t knet_seek(knetFile *fp, off_t off, int whence) {
    return hseek((hFILE *)fp, off, whence);
}

int knet_close(knetFile *fp) {
    return hclose((hFILE *)fp);
}
```

API though is a bit less clear due to the `hfile_fileno` macro.  Frankly it looks to be never used outside of ancient copies of bcftools.  The macro for `knet_tell` would need changing too, which is the fly in the ointment as far as ABI goes.  It's trivial to retain API compatibility by simply changing it to `htell`.  I could keep track of offset internally of course, but that means allocating new structures etc.   Pondering...

jkbonfield:
Ok, so this second commit undoes some things in the first, so it needs squashing, but is left as-is for now so reviewers can see the history.

- htslib/knetfile.h now exists again for API compatibility.  I'm still open to removing this (it's not needed by any of this code), which will remove knet from the API without affecting ABI.  It depends on whether we wish to give warning of impending demise or not.  (It has `HTS_DEPRECATED` markers).
- hfile.c now has a dummy set of knet_* functions which are simply wrappers around `hopen`/`hread` etc.  They manually track offset so the `knet_tell` macro still works (ABI compat).  NB: it doesn't include knetfile.h so it's just a hacked duplicated.  Perhaps I could move the `hf` structure member to in-situ of another `char*` so the structure size doesn't change incase people are using arrays of knetFiles, but I suspect that's unnecessary.
- The windows knet functions have gone, but they weren't listed using `HTS_EXPORT` so I assume they were not publically callable anyway.
- The `knet_fileno` macro is hideous to implement for curl (not sure even if possible), but the only uses I could find of it were immediately followed by `fstat` so I made it work for local files only and it returns -1 otherwise.

I've validated this using this tiny program:

```
#include <stdio.h>
#include "htslib/knetfile.h"

    char dat[100];

    knetFile *fp = knet_open(argv[1], "r");
    if (!fp) {
	perror(argv[1]);
	return 1;
    }
    knet_seek(fp, 15, SEEK_SET);
    int x = knet_read(fp, dat, 50);
    printf("FD %d: Read '%.*s'\n", knet_fileno(fp), x, dat);
    printf("Now at pos %ld\n", (long)knet_tell(fp));
    return knet_close(fp);
}
```

I built this against htslib 1.11 headers, and have switched dynamic library by use of `LD_LIBRARY_PATH`. I can confirm it works as expected.

hfile.c
```diff
@@ -1213,3 +1212,82 @@ char *haddextension(struct kstring_t *buffer, const char *filename,
...
+    int64_t file_size;
+    char *path, *http_host;
+
+    // Our local addition
+    hFILE *hf;
+} knetFile;
+
+HTSLIB_EXPORT
+knetFile *knet_open(const char *fn, const char *mode) {
+    knetFile *fp = malloc(sizeof(*fp));
```
daviesrob, line 1246: It would be safer to use `calloc` here.

hfile.c
```diff
@@ -1213,3 +1212,82 @@ char *haddextension(struct kstring_t *buffer, const char *filename,
...
+ssize_t knet_read(knetFile *fp, void *buf, size_t len) {
+    ssize_t r = hread(fp->hf, buf, len);
+    fp->offset += r>0?r:0;
+    return r;
+}
+
+HTSLIB_EXPORT
+off_t knet_seek(knetFile *fp, off_t off, int whence) {
+    off_t r = hseek(fp->hf, off, whence);
+    if (r > 0)
```
daviesrob, line 1283: `>= 0`  ?

htslib/knetfile.h
```diff
@@ -72,35 +73,30 @@ typedef struct knetFile_s {
...
 #endif
 
-#ifdef _WIN32
-	int knet_win32_init();
-	void knet_win32_destroy();
-#endif
-
     HTSLIB_EXPORT
-	knetFile *knet_open(const char *fn, const char *mode);
+        knetFile *knet_open(const char *fn, const char *mode) lHTS_DEPRECATED("Please use hopen instead");
```
daviesrob, line 77: A stray `l` has appeared here, so this header currently doesn't compile.

daviesrob (commented):
Generally looks good, apart from a few minor issues.

The only stray reference to knet I could find was in a [comment in hfile_libcurl.c](https://github.com/samtools/htslib/blob/44787d98/hfile_libcurl.c#L1224).  It might be a good idea to remove that as well; the comment will still make sense without it.

Could you add a NEWS item, please?

hfile.c
jkbonfield: Yes I think so.

jkbonfield:
Thanks.  Don't know what went on with that stray "l".  I previously had commented out the deprecated lines as it made my testing a pain.  I guess I finger slipped removing the `;//`.

Comment in libcurl.c has been nuked.  I left it there before because I wasn't sure which bit to delete (is the RFC comment for knet or for the whole idea of full pathnames in general?), but I'm sure curl must be pretty compliant and if not I doubt we need to be spelling it out.

daviesrob:
The RFC comment was about using full pathnames.  RFC1738 says you should CWD through each directory, curl did by default but knet did not, which made curl quite a bit slower in comparison.

Git blame suggests you added the original comment in commit 5bd5b9538.

jkbonfield:
Fair enough.  I reinstated that bit of comment again then.

PR #1201, opened 2021-01-13
Fix incorrect setting of end position in overlap_push()

daviesrob:
Commit a2fdf3b changed `bam_endpos()` so that it would consider alignments which consume no bases (e.g. fully soft clipped) as one base long.  This had the side effect of making [`overlap_push()` change the position stored in `a->end`](https://github.com/samtools/htslib/blob/410f2b2e4/sam.c#L4948), causing reads that would previously have been ignored to be added to the pileup.  As `resolve_cigar2()` was not designed to handle such reads it resulted in nonsensical values being set in the pileup structures for these reads.

The update to `a->end` in `overlap_push()` is not necessary anyway as it will have already have been set correctly in [`bam_plp_push()`](https://github.com/samtools/htslib/blob/410f2b2e4/sam.c#L5064), so the simplest solution is to remove the call to `bam_endpos()` in `overlap_push()`.

Pointers to the cigar data in `tweak_overlap_quality()` have been changed to const to ensure it makes no changes that would change the alignment end position.

Fixes samtools/bcftools#1362 (bcftools mpileup seg fault for reads with only 1 match).

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -4772,7 +4772,11 @@ void bam_plp_destructor(bam_plp_t plp,
  *  Returns BAM_CMATCH, -1 when there is no more cigar to process or the requested position is not covered,
  *  or -2 on error.
  */
-static inline int cigar_iref2iseq_set(uint32_t **cigar, uint32_t *cigar_max, hts_pos_t *icig, hts_pos_t *iseq, hts_pos_t *iref)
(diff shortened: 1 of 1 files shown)


valeriuo (approved):

valeriuo:
Merged as f39dae751dae8c1abf8477074c5fbaea66d229df

PR #1202, opened 2021-01-14
Fix the iteration over the CIGAR segments

### LINKED ISSUE #1195: during tweak quality of overlapped paired of reads, the quality of some bases in the end of overlap region are not tweaked
Buggy results:
```
reference
>ref
GCATGGCATACAAATTATTTCATTCCCATTGAGAAATAAAATCCAATTCTCCATCACCAAGAGAGCCTTCCGAAAGAGGCCCCCCTGGGCAAACGGCCACCGATGGAGAGGTCTGCCAGTCCTCTTCTACCCCACCCACGCCCCCACCCTAATCAGAGGCCAAACCCTTCCTGGAGCCTGTGATAAAAGCAACTGTTAGCTTGCACTAGACTAGCTTCAAAGTTGTATTGACCCTGGTGTGTTATGTCTAAGAGTAGATGCCATATCTCTTTTCTGG

read pairs
read1
@rq_182_1:0:0_2:0:0_0
CACAGGCTCCAGGAAGGGTTTGGCCTCTGAATAGGGTGGGGGCGTGGGTGGGGTAGAAGAGGACTGGCAGACCTCTCCATCGGTGGCCGTTTGCCCAGGGGGGCCTCTTTCGGAAGGCTCTCTTGTT
+
2222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222

read2
@rq_182_1:0:0_2:0:0_0
GAAATAAAATCCAATTCTCCAACACCAAGCCGAAAGAGGCCCCCCTGGGCAAACGAAACACATCGGCCACCGATGGAGAGGTCTGCCAGTCCTCTTCTACCCCACCCACGCCCCCACC
+
2222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222

alignment result
rq_182_1:0:0_2:0:0_0	83	ref	56	60	127M	=	33	-150	AACAAGAGAGCCTTCCGAAAGAGGCCCCCCTGGGCAAACGGCCACCGATGGAGAGGTCTGCCAGTCCTCTTCTACCCCACCCACGCCCCCACCCTATTCAGAGGCCAAACCCTTCCTGGAGCCTGTG	2222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222	NM:i:2	MD:Z:1C94A30	MC:Z:29M8D24M10I55M	AS:i:120	XS:i:0
rq_182_1:0:0_2:0:0_0	163	ref	33	60	29M8D24M10I55M	=	56	150	GAAATAAAATCCAATTCTCCAACACCAAGCCGAAAGAGGCCCCCCTGGGCAAACGAAACACATCGGCCACCGATGGAGAGGTCTGCCAGTCCTCTTCTACCCCACCCACGCCCCCACC	2222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222	NM:i:19	MD:Z:21T7^AGAGCCTT79	MC:Z:127M	AS:i:73	XS:i:0

however samtools depth -q 10 -s  generate depth not right at two position
ref	147	2
ref	148	2
```

these two positions are overlapped of course.

I checked the code in sam.c which tweak the quality of overlapped regions and found the bug as below:
each time a new cigar M or X or = is met, if there is another cigar which is none of "MX=", the cigar will be set to 0 first, and then enter the situation to process "MX=", in this case, it increase the positions (*iseq)++; (*icig)++; (*iref)++(https://github.com/samtools/htslib/blob/develop/sam.c#L4459), thus the first matched base in this kind of "MX=" cigar will never be processed in a right way which leads the variable "iref" in https://github.com/samtools/htslib/blob/develop/sam.c#4510 will not increase to the right amount, so finally the tweak loop in https://github.com/samtools/htslib/blob/develop/sam.c#L4500 will not iterate all the overlapped positions and left some end overlapped bases unprocessed, the number of unprocessed bases equals to number of those kinds of "MX=" cigar which appears after other kinds of cigar in overlapped region.

I have fixed this bug by #1196 , please consider merge it if I have done it properly.

valeriuo:
When encountering a new CIGAR segment, the index `icig`, which indicates the position of a base inside the current segment, has to be reset. This fix changes the reset value to **-1** (instead of **0**), so that the next first matching base starts from `icig=0` in the calling method (`tweak_overlap_quality`).

Replaces #1196
Fixes #1195

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -4816,14 +4816,14 @@ static inline int cigar_iref2iseq_next(uint32_t **cigar, uint32_t *cigar_max, ht
 
         if ( cig==BAM_CMATCH || cig==BAM_CEQUAL || cig==BAM_CDIFF )
         {
-            if ( *icig >= ncig - 1 ) { *icig = 0;  (*cigar)++; continue; }
+            if ( *icig >= ncig - 1 ) { *icig = -1;  (*cigar)++; continue; }
             (*iseq)++; (*icig)++; (*iref)++;
             return BAM_CMATCH;
         }
-        if ( cig==BAM_CDEL || cig==BAM_CREF_SKIP ) { (*cigar)++; (*iref) += ncig; *icig = 0; continue; }
-        if ( cig==BAM_CINS ) { (*cigar)++; *iseq += ncig; *icig = 0; continue; }
-        if ( cig==BAM_CSOFT_CLIP ) { (*cigar)++; *iseq += ncig; *icig = 0; continue; }
-        if ( cig==BAM_CHARD_CLIP || cig==BAM_CPAD ) { (*cigar)++; *icig = 0; continue; }
+        if ( cig==BAM_CDEL || cig==BAM_CREF_SKIP ) { (*cigar)++; (*iref) += ncig; *icig = -1; continue; }
+        if ( cig==BAM_CINS ) { (*cigar)++; *iseq += ncig; *icig = -1; continue; }
+        if ( cig==BAM_CSOFT_CLIP ) { (*cigar)++; *iseq += ncig; *icig = -1; continue; }
+        if ( cig==BAM_CHARD_CLIP || cig==BAM_CPAD ) { (*cigar)++; *icig = -1; continue; }
         hts_log_error("Unexpected cigar %d", cig);
         return -2;
     }


daviesrob:
Looks good, although the operation of `tweak_overlap_quality()` is certainly somewhat obscure.

PR #1205, opened 2021-01-20
Reject VCF records where POS == 0

daviesrob:
Which prevents problems caused by v->pos becoming negative when converting from 1-based VCF positions to 0-based internal/BCF.

Catches the bad input that gave rise to issue samtools/bcftools#1366 (bcftools: END (15195) smaller than POS) and caused a third-party tool using HTSlib to crash.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -2901,6 +2901,10 @@ int vcf_parse(kstring_t *s, const bcf_hdr_t *h, bcf1_t *v)
                 hts_log_error("Position value '%s' is too large", p);
                 goto err;
             } else {
+                if (v->pos < 1) {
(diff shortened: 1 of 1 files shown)


jmarshall:
The VCF specification says POS=0 is a valid representation of an event within a telemere… 🤔

pd3:
Uh, I was not aware

daviesrob:
Hmm, that's annoying.  Presumably this should be stored as -1 in BCF (POS is `int32_t`, so that is possible).

I can only presume that it isn't used much, as HTSlib currently handles it rather badly.

daviesrob:
Looks like this one is going to be a bit more complicated...

PR #1206, opened 2021-01-21
Cram3.1 fuzzing fixes

jkbonfield:
A few minor issues thanks to Google's OSS Fuzz.

Also spotted a missing dependency in the Makefile, so fixed that while editing (in a separate commit).


jkbonfield:
Delay merge for now as more incoming to this PR.

jkbonfield:
Turns out the other fuzzing issues were in the submodule, so can merge that first and then retest.  I don't intend adding any more to this (unless more fuzzing reports arrive).

daviesrob:
Rearranged to pull all the htscodecs changes into one commit, which now exists on htscodecs master.

PR #1208, opened 2021-01-25
More OSS-Fuzz detected fixes

jkbonfield:
Credit to OSS-Fuzz

- Fixed buffer overflow in XPACK codec (security)
  Now protects against overflow of rmap[i].

  Fixes oss-fuzz 29827

- Add "cp_end" arg to varint_get* funcs in cram_varint_decode_init so
  reading beyond the buffer can be spotted.

  Fixes oss-fuzz 29837

- Set b->uncomp_size for FQZ decoder.  Already done in other codecs,
  just accidentally missed here.  Without it, it invalidates the
  boundary check in int32_get_blk.

  Fixes oss-fuzz 29855
  Fixes oss-fuzz 29908

- Fixed memory leak in XDELTA when recovering from malformed data stream.

  Fixes oss-fuzz 29858

### DIFF
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -745,7 +745,7 @@ cram_codec *cram_varint_decode_init(cram_block_compression_hdr *hdr,
                                     enum cram_external_type option,
                                     int version, varint_vec *vv) {
     cram_codec *c;
-    char *cp = data;
+    char *cp = data, *cp_end = data+size;
 
     if (!(c = malloc(sizeof(*c))))
         return NULL;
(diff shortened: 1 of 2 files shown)


jkbonfield:
NB: depending on order of merging, you may want to add the htscodecs revision update into this when merging, or add it as a new commit after.

The two PRs are independent, but I didn't include the htscodecs revision change in this so that the tests work.

cram/cram_codecs.c
```diff
@@ -1400,18 +1400,21 @@ cram_codec *cram_xpack_decode_init(cram_block_compression_hdr *hdr,
...
-        return NULL;
+        goto malformed;
     }
     c->free = cram_xpack_decode_free;
     c->size = cram_xpack_decode_size;
     c->get_block = cram_xpack_get_block;
 
     c->u.xpack.nbits = vv->varint_get32(&cp, endp, NULL);
     c->u.xpack.nval  = vv->varint_get32(&cp, endp, NULL);
+    if (c->u.xpack.nval >= 256)
```
daviesrob, line 1411: 40afaaa already added a similar check to this, but used `>` instead of `>=`.  Looking at the loop a few lines below, a value of 256 ought to be OK, so would it be best to resolve the conflict in favour of the version in 40afaaa?

jkbonfield: I thought it felt familier!  However at the time I mistakenly believed my other PR had been merged - losing track as so many outstanding things being juggled.

Agreed a value of 256 would be valid, so I'll fix it (or you can if you wish?).  It shouldn't happen though as there's no point to any value > 16 really as you're just packing 1 byte to 1 byte.  It's valid, just pointless.

daviesrob: I have a fixed version now, so I'll push it up.

daviesrob:
I have rebased and deconflicted.

PR #1210, opened 2021-01-26
remove a strange gcc10 warning.

jkbonfield:
Gcc 10 when using -O and -fsanitize=address produces a warning about
mallocing some insanely large amount of memory.

This is an impossibility given the limits on the sizes of some data
values, but gcc doesn't know this and is assuimng variables may wrap
around and go negative.  Using an unsigned type stops it from
exploring the inaccessible data ranges.


daviesrob:
Rebased to pull htscodecs updates into their own commit (and include the latest one).

Please take care when making commits to only include the intended changes.

PR #1220, opened 2021-02-03
Undefined behaviour fixes

daviesrob:
Fixes a number of warnings from running the tests on HTSlib built with clang 10.0.1 and -fsanitize=address,undefined

These fix "runtime error: applying zero offset to null pointer" warnings:
- Make `kvsprintf()` allocate some memory before trying to write a string into it
- In CRAM, always allocate memory for CIGAR data in slices
- Rearrange `cram_byte_array_stop_decode_block()` so it won't try to do maths on a NULL pointer
- Fix check on `bcf_get_format_int32()` return value in test-vcf-sweep so it doesn't try to work on a non-existent tag
- Fix dubious pointer comparisons in `bcf_update_alleles()`
- Ensure `bcf_read1_core()` always allocates memory for `rec->indiv.s` and `rec->shared.s`

Fixes for integer overflows:
- Prevent overflow of some elements in the `cram_metrics::sz` array
- Fix truncation bug in filtering expression bitwise operators, which could cause `is_true` to have the wrong value for some results

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -2080,7 +2080,7 @@ int cram_compress_block2(cram_fd *fd, cram_slice *s,
                     1.04, // 6  arithpr (O0)
                     1.05, // 7  fqz
                     1.05, // 8  tok3 (rans)
-                    9, 9, // 9,10 reserved
+                    1.00, 1.00, // 9,10 reserved
 
                     // Paramterised versions of above
                     1.01, // gzip rle
@@ -2123,6 +2123,9 @@ int cram_compress_block2(cram_fd *fd, cram_slice *s,
                     for (m = 0; m < CRAM_MAX_METHOD; m++)
                         metrics->sz[m] *= 1+(meth_cost[m]-1)/3;
                 } // else cost is ignored
+
+                // Ensure these are never used
+                metrics->sz[9] = metrics->sz[10] = INT_MAX;
 
                 for (m = 0; m < CRAM_MAX_METHOD; m++) {
                     if ((!metrics->sz[m]) || (!(method & (1u<<m))))
@@ -4395,8 +4398,8 @@ cram_slice *cram_new_slice(enum cram_content_type type, int nrecs) {
     s->block_by_id = NULL;
     s->last_apos = 0;
     if (!(s->crecs = malloc(nrecs * sizeof(cram_record))))  goto err;
-    s->cigar = NULL;
-    s->cigar_alloc = 0;
+    s->cigar_alloc = 1024;
+    if (!(s->cigar = malloc(s->cigar_alloc * sizeof(*s->cigar)))) goto err;
     s->ncigar = 0;
 
     if (!(s->seqs_blk = cram_new_block(EXTERNAL, 0)))       goto err;
@@ -4499,8 +4502,8 @@ cram_slice *cram_read_slice(cram_fd *fd) {
     }
 
     /* Initialise encoding/decoding tables */
-    s->cigar = NULL;
-    s->cigar_alloc = 0;
+    s->cigar_alloc = 1024;
+    if (!(s->cigar = malloc(s->cigar_alloc * sizeof(*s->cigar)))) goto err;
     s->ncigar = 0;
 
     if (!(s->seqs_blk = cram_new_block(EXTERNAL, 0)))      goto err;
--- a/kstring.c
+++ b/kstring.c
@@ -1,7 +1,7 @@
 /* The MIT License
 
    Copyright (C) 2011 by Attractive Chaos <attractor@live.co.uk>
-   Copyright (C) 2013-2018, 2020 Genome Research Ltd.
+   Copyright (C) 2013-2018, 2020-2021 Genome Research Ltd.
 
    Permission is hereby granted, free of charge, to any person obtaining
    a copy of this software and associated documentation files (the
@@ -151,6 +151,15 @@ int kvsprintf(kstring_t *s, const char *fmt, va_list ap)
 		l = kputd(d, s);
 		va_end(args);
 		return l;
+	}
+
+	if (!s->s) {
+		const size_t sz = 64;
+		s->s = malloc(sz);
+		if (!s->s)
+			return -1;
+		s->m = sz;
+		s->l = 0;
 	}
 
 	l = vsnprintf(s->s + s->l, s->m - s->l, fmt, args); // This line does not work with glibc 2.0. See `man snprintf'.
--- a/vcf.c
+++ b/vcf.c
@@ -4443,28 +4443,64 @@ static inline int _bcf1_sync_alleles(const bcf_hdr_t *hdr, bcf1_t *line, int nal
 int bcf_update_alleles(const bcf_hdr_t *hdr, bcf1_t *line, const char **alleles, int nals)
 {
     if ( !(line->unpacked & BCF_UN_STR) ) bcf_unpack(line, BCF_UN_STR);
-    kstring_t tmp = {0,0,0};
     char *free_old = NULL;
+    char buffer[256];
+    size_t used = 0;
 
-    // If the supplied alleles are not pointers to line->d.als, the existing block can be reused.
+    // The pointers in alleles may point into the existing line->d.als memory,
+    // so care needs to be taken not to clobber them while updating.  Usually
+    // they will be short so we can copy through an intermediate buffer.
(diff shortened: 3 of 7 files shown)


daviesrob:
Rebased with a couple of extra copyright date changes.  (I ignored test-vcf-sweep as the changes there were trivial).

vcf.c
```diff
@@ -1236,9 +1236,9 @@ static inline int bcf_read1_core(BGZF *fp, bcf1_t *v)
     shared_len = le_to_u32(x);
     if (shared_len < 24) return -2;
     shared_len -= 24; // to exclude six 32-bit integers
-    if (ks_resize(&v->shared, shared_len) != 0) return -2;
+    if (ks_resize(&v->shared, shared_len ? shared_len : 1) != 0) return -2;
     indiv_len = le_to_u32(x + 4);
-    if (ks_resize(&v->indiv, indiv_len) != 0) return -2;
+    if (ks_resize(&v->indiv, indiv_len ? indiv_len : 1) != 0) return -2;
```
jkbonfield, line 1241: Isn't this a generic issue of ks_resize with size 0, rather than a specific VCF issue?

If so it'd be better fixed in kstring.h

daviesrob: Everything `ks_resize()` does is wrapped in `if (s->m < size)` which is never true for `size == 0`.  I think it would be more intrusive to try to change this behaviour instead of just fixing the one or two places where it may cause a problem.

vcf.c
```diff
@@ -4448,8 +4448,10 @@ int bcf_update_alleles(const bcf_hdr_t *hdr, bcf1_t *line, const char **alleles,
 
     // If the supplied alleles are not pointers to line->d.als, the existing block can be reused.
     int i;
+    uintptr_t als_start = (uintptr_t) line->d.als;
+    uintptr_t als_end = line->d.als ? als_start + line->d.m_als : als_start;
     for (i=0; i<nals; i++)
-        if ( alleles[i]>=line->d.als && alleles[i]<line->d.als+line->d.m_als ) break;
+        if ( (uintptr_t) alleles[i] >= als_start && (uintptr_t) alleles[i] < als_end ) break;
```
jkbonfield, line 4454: Do you have an example of this?

I tried clang 10 with -fsanitize=undefined and this line as-is in develop, on the entire VCF test corpus (hts-specs#491) and cannot trigger this.

I find the rewrite harder to understand, so wonder if there is an issue whether it's simpler fixed.  Eg:

`if (line->d.als && alleles[i]>=line->d.als && alleles[i]<line->d.als+line->d.m_als ) break;`

Replacing the one line with 3 lines using `uintptr_t` really doesn't add to the clarity of what it's trying to do.

cram/cram_io.c
```diff
@@ -2080,7 +2080,7 @@ int cram_compress_block2(cram_fd *fd, cram_slice *s,
                     1.04, // 6  arithpr (O0)
                     1.05, // 7  fqz
                     1.05, // 8  tok3 (rans)
-                    9, 9, // 9,10 reserved
+                    1.00, 1.00, // 9,10 reserved
```
jkbonfield, line 2083: If we're redefining them elsewhere anyway, then probably something like 0 would make more sense as an explicit nul value instead of 1.00.

kstring.c
```diff
@@ -153,6 +153,15 @@ int kvsprintf(kstring_t *s, const char *fmt, va_list ap)
 		return l;
 	}
 
+    if (!s->s) {
```
jkbonfield, line 156: This has a different indentation style to the rest of the file.  (kstring.c differs to the rest of htslib as it is imported from elsewhere)

jkbonfield:
It looks like these are all genuine fixes which is great (although I see Mozilla gave up and just added a suppression for adding 0 to NULL, possibly under the reasonable grounds that it's a nop and it's explicitly permitted in C++ too so is likely to be considered for C in a future version at some point).

I would add though that I was unable to reproduce many of these.  How are you finding these problems?  If you have specific crafted VCF and CRAM files that are exercising new and interesting code paths, then maybe consider adding them to the hts-specs VCF and CRAM test files PRs as the intention for those PRs are to provide validation corpuses.

daviesrob:
You need to run the samtools and bcftools tests to get some of these to happen.  Apart from that, there were no specially crafted files.

cram/cram_io.c
daviesrob: I didn't use 0 because there's some [scaling code](https://github.com/samtools/htslib/blob/2d1e9549/cram/cram_io.c#L2115) that subtracts one from the values here before multiplying them, and I didn't want to end up with a negative value.

vcf.c
daviesrob: The existing code relies on undefined behaviour, which puts it at risk of being [optimised in an unintended way](https://devblogs.microsoft.com/oldnewthing/20170927-00/?p=97095).  Using `uintptr_t` is also strictly undefined, but I suspect it's less likely that half the test would disappear because the optimiser thinks it shouldn't happen.

I do have an alternative fix that removes the undefined behaviour completely, but it's a far bigger change and I thought it best to go for a simple one for now.

kstring.c
daviesrob: That shouldn't matter if you've set your tab size correctly.

jkbonfield: The entire rest of the file is tab indented, while this change is 4 space indented.  If we're to believe tabs vs space don't matter, then why do we have a very strict regime of absolutely zero tabs (not to mention invisible things like trailing white space) ?

I think it's also a bit much to require people to use text editors with configurable tab settings to adjust tabs to be 4 spaces.  Infact I don't think it's even possible with straight "less".

vcf.c
jkbonfield: That still doesn't explain why the far simpler fix isn't sufficient, as suggested.  It avoids the use of uintptr_t too.

I did manage to reproduce this in the end with `./test/test-vcf-api test/test-vcf-api.out` and can confirm adding `line->d.als &&` to the if statement is all that is needed.  It seems far clearer too as it's obvious what it's doing: only do the bounds check if d.als is non-NULL.

Looking at the test code, alleles may be NULL, but nals is also 0 then so the loop isn't entered.  The other scenarios are (with alleles as non-NULL) that alleles points to within `line->d.als` in which case the `&&` claus above ensures we're valid (barring strict behaviour on pointer comparisons), or `line->d.als` is NULL (the bug we're trying to fix) in which case we break without pointer comparison as an indication that alleles is not an internal part of the existing line.

daviesrob: I'm trying to fix both the bit of undefined behaviour that UBSAN complained about (adding zero to NULL) and the bit it didn't (comparing pointers to memory regions that are not part of the same array).

vcf.c
```diff
@@ -4448,8 +4448,10 @@ int bcf_update_alleles(const bcf_hdr_t *hdr, bcf1_t *line, const char **alleles,
 
     // If the supplied alleles are not pointers to line->d.als, the existing block can be reused.
     int i;
+    uintptr_t als_start = (uintptr_t) line->d.als;
+    uintptr_t als_end = line->d.als ? als_start + line->d.m_als : als_start;
```
jkbonfield, line 4452: So just to clarify what I was trying to say earlier but obviously not explaining it well, the above ?: is pointless and just obfuscates the code.

I get what it's trying to achieve: adding offset to pointer is invalid is pointer is NULL, even if offset is zero.  However you've also converted these to uintptr_t already.  `als_start` isn't a pointer, it's an integer.  Adding 0 to 0 is 100% legal and defined behaviour, so we don't need to first check whether the value is zero with a ?: operator.  It could just be `als_end = als_start + line->d.m_als`.

Given that, it's clear to see that both local variables are also unnecessary and could simply be replaced by a cast in the if statement as that's now all they're doing.

kstring.c
daviesrob: I have replaced the spaces with tabs.

vcf.c
daviesrob: This has now been replaced with a more comprehensive solution that completely eliminates the undefined behaviour.  One advantage with the new version is that it works better with the way bcftools calls the function (e.g. in [norm](https://github.com/samtools/bcftools/blob/b626e984/vcfnorm.c#L162)) and makes it much more likely to reuse the existing buffer.  When I tested, it halved the amount of time taken in `bcf_update_alleles()` (although it doesn't affect the run time too much as most time is spent elsewhere).

PR #1223, opened 2021-02-04
Fix a minor memory leak in XRLE encoding.

jkbonfield:
Credit to OSS-Fuzz
Fixes oss-fuzz 30260


valeriuo:
Merged as 8aa9bcd10c782c8a03adde978f9ef169ebaf4573

PR #1225, opened 2021-02-07
construction fix.

ihsinme:
good day.
I apologize in advance if I'm wrong.
it's just that the expressions look pretty strange, the use of `!res-> is_str` suggests this. which in the given presentation is always true.
so I suggest a possible fix.


valeriuo:
It is a chained assignment. If the result is a string, it is always true, even if empty, and `d` is not used.

PR #1230, opened 2021-02-12
Assorted build updates

daviesrob:
Small updates for build / release:

- Update `.gitattributes` to keep `.cirrus.yml` out of release tar files
- Add targets to `htslib.mk` so HTSlib's `distclean` and `check / test` can be run from packages that embed it (e.g. SAMtools / BCFtools).
- Clarify that the htscodecs submodule is part of htslib in messages printed when it's missing, so it's easier to see what needs to be updated when building SAMtools / BCFtools with HTSlib embedded.  Fixes samtools/samtools#1364.

### DIFF
--- a/htslib.mk
+++ b/htslib.mk
@@ -188,7 +188,8 @@ $(HTSDIR)/htslib.pc.tmp:
 #
 #	clean: clean-htslib
 
-all-htslib clean-htslib install-htslib plugins-htslib:
+all-htslib check-htslib clean-htslib distclean-htslib install-htslib plugins-htslib test-htslib:
 	+cd $(HTSDIR) && $(MAKE) $(@:-htslib=)
 
-.PHONY: all-htslib clean-htslib install-htslib plugins-htslib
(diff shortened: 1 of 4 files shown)


htslib.mk
```diff
@@ -188,7 +188,8 @@ $(HTSDIR)/htslib.pc.tmp:
 #
 #	clean: clean-htslib
 
-all-htslib clean-htslib install-htslib plugins-htslib:
+all-htslib check-htslib clean-htslib distclean-htslib install-htslib plugins-htslib test-htslib:
```
jkbonfield, line 191: Should we also have check-all test-all and distclean-all targets, in a similar vein to clean-all?
"distclean-all" is the one I miss the most when testing releases.

daviesrob: Those would need to be added to the samtools and bcftools Makefiles.  The parts here are the bits that they reference and need to be in place first.

jkbonfield: Ah yes my mistake.

PR #1235, opened 2021-02-16
Change CI scripts to use "autoreconf -i" and update docs

daviesrob:
autoconf 2.7 requires use of `autoreconf -i` for the AC_CANONICAL_HOST macro, which is pulled in by AC_FUNC_MMAP.

Update CI configuration to use this, and change the instructions in the INSTALL and README.md files to match.

Fixes AppVeyor builds, which have stopped working since they upgraded to autoconf 2.71

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -34,7 +34,7 @@ build_script:
   - set MSYSTEM=MINGW64
   - set PATH=C:/msys64/usr/bin;C:/msys64/mingw64/bin;%PATH%
   - git submodule update --init --recursive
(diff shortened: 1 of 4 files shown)


jmarshall:
It's autoconf 2.70 ("seventy" (!)), not 2.7.

(Also cf #1199, but this is better while autoconf 2.70+ is not very widespread. BTW there's a 2.71 release now too, a bug fix release from a few weeks ago.)

daviesrob:
Oops.  I have amended the commit message.

PR #1242, opened 2021-02-26
Update to htscodecs release 1.0

daviesrob:
Updates the htscodecs submodule and updates dependencies for the new header files.

Adds Makefile infrastructure to build and clean up the htscodecs/htscodecs/version.h file, for git checkouts only. Release tarballs will ship with a pre-built copy of this file.

Fixes issue where the Makefile rules to test if the submodule is missing would cause missing source files to be silently ignored. They now run a recipe to check for the target, and suggest updating the submodule if it's missing.

Makes hts_version() report the submodule version along with its own.

Adds documentation for the '--with-external-htscodecs' configure option to INSTALL.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -358,7 +358,7 @@ hfile_gcs.o hfile_gcs.pico: hfile_gcs.c config.h $(htslib_hts_h) $(htslib_kstrin
 hfile_libcurl.o hfile_libcurl.pico: hfile_libcurl.c config.h $(hfile_internal_h) $(htslib_hts_h) $(htslib_kstring_h) $(htslib_khash_h)
 hfile_s3_write.o hfile_s3_write.pico: hfile_s3_write.c config.h $(hfile_internal_h) $(htslib_hts_h) $(htslib_kstring_h) $(htslib_khash_h)
 hfile_s3.o hfile_s3.pico: hfile_s3.c config.h $(hfile_internal_h) $(htslib_hts_h) $(htslib_kstring_h)
-hts.o hts.pico: hts.c config.h $(htslib_hts_expr_h) $(htslib_hts_h) $(htslib_bgzf_h) $(cram_h) $(htslib_hfile_h) $(htslib_hts_endian_h) version.h config_vars.h $(hts_internal_h) $(hfile_internal_h) $(sam_internal_h) $(htslib_hts_os_h) $(htslib_khash_h) $(htslib_kseq_h) $(htslib_ksort_h) $(htslib_tbx_h)
+hts.o hts.pico: hts.c config.h $(htslib_hts_expr_h) $(htslib_hts_h) $(htslib_bgzf_h) $(cram_h) $(htslib_hfile_h) $(htslib_hts_endian_h) version.h config_vars.h $(hts_internal_h) $(hfile_internal_h) $(sam_internal_h) $(htslib_hts_os_h) $(htslib_khash_h) $(htslib_kseq_h) $(htslib_ksort_h) $(htslib_tbx_h) $(htscodecs_htscodecs_h)
 hts_expr.o hts_expr.pico: hts_expr.c config.h $(htslib_hts_expr_h) $(textutils_internal_h)
 hts_os.o hts_os.pico: hts_os.c config.h $(htslib_hts_defs_h) os/rand.c
 vcf.o vcf.pico: vcf.c config.h $(htslib_vcf_h) $(htslib_bgzf_h) $(htslib_tbx_h) $(htslib_hfile_h) $(hts_internal_h) $(htslib_khash_str2int_h) $(htslib_kstring_h) $(htslib_sam_h) $(htslib_khash_h) $(htslib_kseq_h) $(htslib_hts_endian_h)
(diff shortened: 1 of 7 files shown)


hts.c
```diff
@@ -70,7 +75,10 @@ int hts_verbose = HTS_LOG_WARNING;
 
 const char *hts_version()
 {
-    return HTS_VERSION_TEXT;
+    static char version[256];
+    snprintf(version, sizeof(version), "%s (with htscodecs %s)",
+             HTS_VERSION_TEXT, htscodecs_version());
+    return version;
```
jmarshall, line 81: `hts_version()` is documented as returning literally `N.N[.N]` or `N.N[.N]-N-gCOMMITHASH[-dirty]` and e.g. this is used to construct a one-word string in a fixed format in `samtools --version-only`.

I'd suggest leaving `hts_version()` as is and adding a function (e.g. `hts_htscodecs_version()`) declared in _htslib/hts.h_ alongside `hts_version()` that just forwards the result of htscodecs's function. Then samtools could use that explicitly to produce either human-readable or machine-readable information e.g.

```
$ samtools --version
samtools 1.11-30-g63f345d
Using htslib 1.11-104-ga117153-dirty
Using htscodecs 1.0
Copyright (C) 2021 Genome Research Ltd.
[…]

$ samtools --version-only
1.11-30-g63f345d+htslib-1.11-104-ga117153-dirty+htscodecs-1.0
```

(See also [GNU's note about the Using lines](https://www.gnu.org/prep/standards/html_node/_002d_002dversion.html); I think putting the word `Using` was my small addition to distinguish these from the main version number.)

daviesrob: Hmm, yes it does break `samtools --version-only` a bit.  I was hoping I could get away without adding another function.  Another alternative would be to add it into the `hts_feature_string()` output, although I'm not sure if it really fits well there either.

jmarshall: > … add it into the `hts_feature_string()` output …

And you could get the raw codecs version number — if you really wanted to — with something like `hts_test_feature(HTS_FEATURE_HTSCODECS)`. I like it!

daviesrob: Yes, that's a good plan.  I'll make it so.

jmarshall: I agree with James that once this settles down, it mostly won't change much. So recording it in the `hts_feature_string()` that Samtools is already printing out displays it as useful background info, and avoids needing extra code to call it out in the version number part.

It's win/win and a vindication of the introspection stuff! One thing I hesitate to mention… It does feel like `hts_test_feature()` is a wee bit misnamed… any appetite for simplifying that to `hts_feature(unsigned int id)`?

daviesrob: It isn't ideally named, but we already have `hts_features()`, so `hts_feature()` is maybe a bit too similar?

jmarshall: Gimme all `hts_features()` vs gimme one `hts_feature(HTS_FEATURE_THIS_ONE)` I think is okay :shrug:

I mostly want to fend off people saying “is `hts_test_feature()` just about hslib's testing functionality” but that could be done via documentation comments if it ever became an issue…

daviesrob: I've added a commit which moves the htscodecs version to the introspection API.  It currently has no opinion on what `hts_test_feature()` is called, as any change to that really ought to be done separately.

hts.c
```diff
@@ -204,18 +205,19 @@ const char *hts_feature_string(void) {
...
-    "bzip2=yes ";
+    "bzip2=yes "
 #else
-    "bzip2=no ";
+    "bzip2=no "
 #endif
 
+    "htscodecs=%.40s";
+
+    snprintf(config, sizeof(config), fmt,
```
jmarshall, line 215: Maybe stdlib's don't really define functions as macros any more so this doesn't really matter (and maybe there's already stuff like this in htslib, so this one doesn't matter either), but: `snprintf()` might be defined as a macro in which case you can't do `#ifdef` inside its argument list (C99 §6.10.3/11).

Yes, I really did actually get burnt by this once :smile:

daviesrob: That's somewhat annoying....

Makefile
```diff
@@ -432,6 +443,21 @@ htscodecs/htscodecs:
 	  official releases from https://www.htslib.org/\\n" ; \
 	fi
 	@false
+
+# Build the htscodecs/htscodecs/version.h file if necessary
+htscodecs/htscodecs/version.h: force
+	@if test -e htscodecs/.git && test -e htscodecs/configure.ac ; then \
```
jkbonfield, line 449: I see no else clause here.  I tried hiding my .git directory and sure enough it fails.

We need this to be working for the tarball.

hts.c
```diff
@@ -58,6 +58,11 @@ DEALINGS IN THE SOFTWARE.  */
 #include "htslib/kseq.h"
 #include "htslib/ksort.h"
 #include "htslib/tbx.h"
+#if defined(HAVE_EXTERNAL_LIBHTSCODECS)
```
jkbonfield, line 61: This works fine, but I wonder if it's the wrong method.

The traditional way would be to amend the include search path to include the htscodecs sub-module dir, so the include works using local source.  I guess the tricky bit is needing -I- or -iquote rules in order to include this dir before any system paths.  I'm not sure how portable those methods are.

Hence the above is at least robust and portable, but I wondered if you thought about the other approaches?

hts.c
```diff
@@ -74,7 +79,7 @@ const char *hts_version()
 }
 
 unsigned int hts_features(void) {
-    unsigned int feat = 0;
+    unsigned int feat = HTS_FEATURE_HTSCODECS; // Always present
```
jkbonfield, line 82: Having amended this, it should also get an amended test/test_introspection.c file to utilse the new versioning.

htscodecs_bundled.mk
```diff
@@ -25,6 +25,7 @@
 
 HTSCODECS_SOURCES = $(HTSPREFIX)htscodecs/htscodecs/arith_dynamic.c \
         $(HTSPREFIX)htscodecs/htscodecs/fqzcomp_qual.c \
+	$(HTSPREFIX)htscodecs/htscodecs/htscodecs.c \
```
jkbonfield, line 28: This is the only line using a tab in this file.  All others are using spaces.

hts.c
jkbonfield: > It does feel like `hts_test_feature()` is a wee bit misnamed… any appetite for simplifying that to `hts_feature(unsigned int id)`?

My thought process is if you're calling it on a specific feature then you're most likely testing whether that feature exists, otherwise you're just asking for the whole list of them.  However I'm ambivalent and could live with hts_feature too if others prefer it.

Makefile
daviesrob: The idea is that non-git distributions (i.e. the tarball) will supply a version of this file that has the correct version number in it.  If it isn't there then the distribution is incomplete.

hts.c
daviesrob: It's deliberately done this way to avoid issue #347 (HTSlib should prefer its own headers when being compiled).  I doubt other solutions are very portable, and this solution is already used in source files in the `cram` subdirectory.

Makefile
jkbonfield: Ah fair enough and makes sense.  I'll delay my assessment then until the tarballs are ready.

hts.c
daviesrob: Done, test_introspection now exercises this.

htscodecs_bundled.mk
daviesrob: Oops, tab has now gone.

PR #1246, opened 2021-03-03
When not using configure, define _XOPEN_SOURCE

jmarshall:
Non-configure `-std=c99` builds (as opposed to `-std=gnu99` builds), e.g., `make CC='gcc -std=c99'`, fail as glibc (and maybe others) suppresses non-Standard-C functions etc in standard headers in this mode:

```
$ make distclean
$ make CC='gcc -std=c99'  # On Linux i.e. glibc
…
gcc -std=c99 -g -Wall -O2 -fvisibility=hidden -I.  -c -o kfunc.o kfunc.c
kfunc.c: In function ‘kf_erfc’:
kfunc.c:76:16: error: ‘M_SQRT2’ undeclared (first use in this function)
  z = fabs(x) * M_SQRT2;
                ^~~~~~~

…This is the first error of many — e.g., missing strdup() declarations, etc
```

In particular: Rhtslib, Rsamtools, and other R-based builds do not use configure and don't supply their own _config.h_, and may specify `-std=c99`. See for example rwdavies/STITCH#45, where HTSlib is built without running configure as part of an R build process — and that R build process supplies `-std=c99`.

This is part of something I've been working on to tidy up the setting of these `_POSIX/_XOPEN…SOURCE` defines in _configure.ac_, but this is not the time to be proposing that! This part by itself will prevent some build failures for the R people and others though.


jmarshall:
Well, well… so that's why we need 600! At first glance, this appears to be a bug in macOS's headers…

jkbonfield:
> 
> 
> [`macosx + clang USE_CONFIG:no` CI build fails]
> Well, well… so that's why we need 600! At first glance, this appears to be a bug in macOS's headers…

Interesting.  I see it failed on `vsnprintf`, which the linux man page claims needs `_XOPEN_SOURCE >= 500 || _ISOC99_SOURCE`.  So it ought to pass on both fronts (although it doesn't appear the compile used -stdc=c99 and I don't know what level of compliance it defaults to).

From what I can tell from searching, macos man pages for vsnprintf mentioned c99 but are silent regarding XOpen.

jmarshall:
Mojave's `<stdio.h>` conditions `[v]snprintf` on `__DARWIN_C_LEVEL`, `_C99_SOURCE`, and/or `__cplusplus`, and you'd think `_C99_SOURCE ` was supposed to be true in C99 mode. I dunno what macOS clang's default language level is, but even with `clang -std=c99` that one doesn't fire. Nothing in the headers defines `_C99_SOURCE` and clang doesn't seem to predefine it. So it seems like this vestigial `_C99_SOURCE` macro is left over from when it got removed from _sys/cdefs.h_ (their equivalent of glibc's _features.h_) or something.

So setting 600 means it gets activated via `__DARWIN_C_LEVEL` instead.

I've spent a while looking through #423 and the rest of your old `PTHREAD_MUTEX_RECURSIVE` stuff trying to figure out why we settled on 600 instead of 500, but maybe this is the answer to that one. :smile:

jkbonfield:
I can't recall why I chose 600, but I expect it was pragmatically what worked.

daviesrob:
According to the [feature_test_macros](https://linux.die.net/man/7/feature_test_macros) man page, `_XOPEN_SOURCE = 600` is the one that gets all of the C99 stuff.

It certainly helps to be more resistant to random compiler options, thanks.

jmarshall:
That is an interesting point that 600 implies the C99 functions, and I think we've explained the pragmaticness of 600, and thanks for merging.

Still… other things, notably `-std=c99`, imply the C99 functions too. It'd be an obtuse C implementation that interpreted `cc -D_XOPEN_SOURCE=500 -std=c99` as “give me C99 language constructs but take the C99 library functions away“ rather than as a statement about what level of POSIX+SUS the code wants to have available in addition to the C99 basics. And in fact on Mojave `_XOPEN_SOURCE=500` does not remove other C99 additions like `labs()`, `lldiv()`, the C99 maths functions, or `isblank()`. So I'll still quixotically file a bug with Apple…

PR #1249, opened 2021-03-04
Add options to force S3 address style.

### LINKED ISSUE #1377: S3 file access fails with long paths
#### Are you using the latest version of [samtools](https://github.com/samtools/samtools/releases/latest) and [HTSlib](https://github.com/samtools/htslib/releases/latest)? If not, please specify.
(run `samtools --version`)

`samtools 1.10`

 #### Please describe your environment.
- OS (run `uname -sr` on Linux/Mac OS or ` wmic os get Caption, Version` on Windows)
- machine architecture (run `uname -m` on Linux/Mac OS or ` wmic os get OSArchitecture` on Windows)
- compiler (run `gcc --version` or `clang --version`)

`Linux 3.10.0-957.27.2.el7.x86_64`

#### Please specify the steps taken to generate the issue, the command you are running and the relevant output.

When accessing a file on CEPH storage using the S3 protocol, samtools fails on longer paths. I haven't determined yet whether the root cause is the length of the path string, or the depth of nesting. I figured I would check here if this is a known issue before diving too deep into debugging, as I can solve the problem in the short term by moving the data to a location with shorter paths. I'll need to resolve this eventually though.

whitwham:
There are now three ways to force the URL style in the S3 plugin.  By setting addressing_style in the aws credentials file, by setting host_bucket in .s3cfg to anything not containing %(bucket).s (this is how s3cmd works) or by setting the  HTS_S3_ADDRESS_STYLE env variable to **path**.

Should fix samtools/samtools#1377.

### DIFF
--- a/hfile_s3.c
+++ b/hfile_s3.c
@@ -29,6 +29,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <strings.h>
 #include <time.h>
 
 #include <errno.h>
@@ -503,6 +504,7 @@ static s3_auth_data * setup_auth_data(const char *s3url, const char *mode,
     ptrdiff_t bucket_len;
     int is_https = 1, dns_compliant;
     char *query_start;
+    enum {s3_auto, s3_virtual, s3_path} address_style = s3_auto;
 
     if (!ad)
         return NULL;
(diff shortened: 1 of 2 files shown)


hfile_s3.c
```diff
@@ -555,29 +557,73 @@ static s3_auth_data * setup_auth_data(const char *s3url, const char *mode,
...
     if (ad->id.l == 0) {
+        kstring_t url_style = KS_INITIALIZE;
         const char *v = getenv("AWS_SHARED_CREDENTIALS_FILE");
         parse_ini(v? v : "~/.aws/credentials", profile.s,
                   "aws_access_key_id", &ad->id,
                   "aws_secret_access_key", &ad->secret,
                   "aws_session_token", &ad->token,
-                  "region", &ad->region, NULL);
+                  "region", &ad->region,
+                  "addressing_style", &url_style, NULL);
```
jmarshall, line 578: You or I should have put this `NULL` on the next line — so extra entries can be added more clearly — ages ago :smile: (and similarly in the other `parse_ini` call below).

```suggestion
                  "addressing_style", &url_style,
                  NULL);
```

daviesrob: Adjusted (will squash on merge).

daviesrob:
Thanks, I've tested the various configuration options and they all behave as expected.

Will merge shortly...

PR #1258, opened 2021-03-23
Remove compressBound assertions.

### LINKED ISSUE #1257: Support for zlib-ng.
Add support for `zlib-ng`.
https://github.com/zlib-ng/zlib-ng

zlib-ng has its own API, but can also be compiled with zlib compatibility support.
https://github.com/zlib-ng/zlib-ng/blob/develop/PORTING.md

For now it seems to fail when running `samtools view`:
```
samtools: bgzf.c:540: bgzf_hopen: Assertion `compressBound(0xff00) < 0x10000' failed.
```
as zlib-ng might require a bigger buffer than zlib.

jkbonfield:
These trip on with zlib-ng as the worse case expansion is 9-bits per byte (at level 1 only).  The test is done on opening a file, which seems strange.

We could do the check on opening for write only, but it seems more productive to error only when we actually can't fit the data rather than just incase it may if given truely random input.  Our real data doesn't seem to trigger this.

TODO: if we ever actually find genuine data that trips this up, then consider resetting level to above 1 and trying again, so the error is handled rather than exiting.

The worst case I could find on any of our data files was an ONT BAM file with high randomness in quality values, leading to 57k blocks at level 1.

Fixes #1257

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -491,7 +491,6 @@ fail:
 BGZF *bgzf_open(const char *path, const char *mode)
 {
     BGZF *fp = 0;
-    assert(compressBound(BGZF_BLOCK_SIZE) < BGZF_MAX_BLOCK_SIZE);
     if (strchr(mode, 'r')) {
         hFILE *fpr;
         if ((fpr = hopen(path, mode)) == 0) return 0;
@@ -514,7 +513,6 @@ BGZF *bgzf_open(const char *path, const char *mode)
 BGZF *bgzf_dopen(int fd, const char *mode)
 {
     BGZF *fp = 0;
-    assert(compressBound(BGZF_BLOCK_SIZE) < BGZF_MAX_BLOCK_SIZE);
     if (strchr(mode, 'r')) {
         hFILE *fpr;
         if ((fpr = hdopen(fd, mode)) == 0) return 0;
@@ -537,7 +535,6 @@ BGZF *bgzf_dopen(int fd, const char *mode)
 BGZF *bgzf_hopen(hFILE *hfp, const char *mode)
 {
     BGZF *fp = NULL;
-    assert(compressBound(BGZF_BLOCK_SIZE) < BGZF_MAX_BLOCK_SIZE);
     if (strchr(mode, 'r')) {
         fp = bgzf_read_init(hfp, NULL);
         if (fp == NULL) return NULL;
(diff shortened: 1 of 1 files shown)


jkbonfield:
Meh, my error message seems pointless.  I can trigger it on grossly small sizes (which is how I initially tested it) as there is still avail input left.  On very close cases of not-quite-enough, both zlib and zlib-ng set avail_in to 0, avail_out to 0, and ret to 0.  Why ret of 0?  It didn't succeed, and it definitely didn't compress all of the data either.

So we just get:

```
[E::bgzf_compress] Deflate operation failed: [0] unknown
```

Very unhelpful!  That was testing using /dev/urandom on zlib-ng, but also on normal zlib by manually hacking avail_out.  I'll expand on the error reporting.

(For what it's worth, on 100 random blocks of data, the biggest expansion was 65280 to 68914 for zlib-ng, so I think it's compressBound is rather pessimistic unless there are non-random edge cases that trigger even poorer behaviour.)

jkbonfield:
Updated to store uncompressed data for the grossly-maladjusted output size case too instead of err message (although I don't expect this case should ever occur).

daviesrob:
Unfortunately I've spotted a memory leak if the fall-back to storing uncompressed data is triggered.  It skips the call to `deflateEnd()`, which means the data allocated for the `z_stream` never gets freed.  For example:
```
$ dd if=/dev/urandom bs=64k count=10 | valgrind ./bgzip -l 1 > /dev/null
[ ... uninteresting bit snipped ... ]
==27023== HEAP SUMMARY:
==27023==     in use at exit: 2,354,320 bytes in 50 blocks
==27023==   total heap usage: 65 allocs, 15 frees, 3,124,368 bytes allocated
==27023== 
==27023== LEAK SUMMARY:
==27023==    definitely lost: 48,320 bytes in 8 blocks
==27023==    indirectly lost: 2,306,000 bytes in 42 blocks
```

Calling `deflateEnd()` (without checking the return code, as I expect it will be unhappy) before the `goto uncomp;` seems to work as a fix.  It would need to be done in both places.

I also noticed that `bgzf_gzip_compress()` could also fail due to running out of space.  I'm inclined to ignore that for now though as it would be harder to fix, and you would have to work quite hard to get it to happen - you'd need to link zlib-ng, set gzip mode and a very low compression level.  It would have to be in third-party code, as the only thing we use gzip mode for is cram indexes, and they always use the default level.

jkbonfield:
Good spot.  I did think briefly about that but naively assumed it didn't allocate as it was already doing a return -1 without freeing it anyway.  So this is an old memory leak, but we ought to fix it while tweaking this.

Issue #1261, opened 2021-03-30
libcurl enabled applied by default?


mkirsche:
Hi,

I just tried to install htslib on a system without libcurl with the following commands: 

```
./configure --prefix=/path/to/htslib --without-libdeflate --disable-bz2
make
make install
```

However I got this error:  `hfile_libcurl.c:47:10: fatal error: curl/curl.h: No such file or directory`

I was eventually able to install it by adding the command line flag `--disable-libcurl`, but this option is not listed in the documentation as far as I could see (and the INSTALL document implies that libcurl is disabled by default since it is listed as an optional dependency), so I had to guess that it existed based on similar command line flags for other dependencies.  Could you please update either the documentation or the default behavior?

Thanks!
Melanie

jkbonfield:
This is also a bug of the configure script apparently.  It's meant to spot if it doesn't have curl and build without it automatically.  It looks like it only tests against run-time install rather than a compile-time one so it won't spot missing headers.

Note your best solution though is to install the recommended dependencies. Eg from INSTALL:

```
Debian / Ubuntu
---------------

sudo apt-get update  # Ensure the package list is up to date
sudo apt-get install autoconf automake make gcc perl zlib1g-dev libbz2-dev liblzma-dev libcurl4-gnutls-dev libssl-dev

Note: libcurl4-openssl-dev can be used as an alternative to libcurl4-gnutls-dev.
```

jmarshall:
When I tried to reproduce this, it **did** spot that it didn't have curl and built without it automatically; see also #1177.

So it would be good to know what version you were trying to build (were you starting from an _htslib-X.Y.tar.bz2_ tarball?), what operating system you're on, what output from `./configure …` you got, and perhaps to attach the resulting _config.log_.

jmarshall:
> It looks like it only tests against run-time install rather than a compile-time one

For libcurl it checks `AC_CHECK_LIB([curl], [curl_easy_pause], …)`, which is the standard autoconf way to check for a compile-time-available library function. This tries to link against `-lcurl`, which will succeed if _libcurl.so_ is present (which is supplied by the libcurl-devel package) but not if only _libcurl.so.4_ is present (which is supplied by the libcurl runtime package). So this test too is checking the compile-time capabilities.

[...]

Issue #1177, opened 2020-11-04
configure: WARNING: libcurl not enabled: library not found


JonEilers:
Hi, 
when running ./configure, it can't find curl

```bash
./configure
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for ranlib... ranlib
checking for grep that handles long lines and -e... /usr/bin/grep
checking for C compiler warning flags... -Wall
checking for pkg-config... /usr/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking for special C compiler options needed for large files... no
checking for _FILE_OFFSET_BITS value needed for large files... no
checking shared library type for unknown-Linux... plain .so
checking whether the compiler accepts -fvisibility=hidden... yes
checking how to run the C preprocessor... gcc -E
checking for egrep... /usr/bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking for stdlib.h... (cached) yes
checking for unistd.h... (cached) yes
checking for sys/param.h... yes
checking for getpagesize... yes
checking for working mmap... yes
checking for gmtime_r... yes
checking for fsync... yes
checking for drand48... yes
checking for srand48_deterministic... no
checking whether fdatasync is declared... yes
checking for fdatasync... yes
checking for library containing log... -lm
checking for zlib.h... yes
checking for inflate in -lz... yes
checking for library containing recv... none required
checking for bzlib.h... yes
checking for BZ2_bzBuffToBuffCompress in -lbz2... yes
checking for lzma.h... yes
checking for lzma_easy_buffer_encode in -llzma... yes
checking for libdeflate.h... no
checking for libdeflate_deflate_compress in -ldeflate... no
checking for curl_easy_pause in -lcurl... no
checking for curl_easy_init in -lcurl... no
configure: WARNING: libcurl not enabled: library not found
configure: WARNING: GCS support not enabled: requires libcurl support
configure: WARNING: S3 support not enabled: requires libcurl support
[...]

PR #1265, opened 2021-04-06
Add curl/curl.h configure check and improve INSTALL

jmarshall:
This addresses a _configure_ issue and configure option documentation omissions from _INSTALL_ uncovered by #1261:

* Unlike the checks for other libraries, the libcurl check validated only that the link library _libcurl.so/.dylib/etc_ is available, but did not validate that the headers are available. In the #1261 case, inconsistent LDFLAGS and CPPFLAGS settings meant that the link library was available to the linker but the headers were not available to the compiler, which should have been detected at configure time.

   (It's not recorded in the commit messages, but I think the other `AC_CHECK_HEADER`/`AC_CHECK_LIB` checks may have been written thus precisely so that such environment variable problems could be detected.)

* The default behaviour of `--enable-libcurl` was not documented in _INSTALL_, and it was not clear that
`--enable`/`--disable` and `--with`/`--without` options come in pairs.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -150,7 +150,10 @@ various features and specify further optional external requirements:
 
 --enable-libcurl
     Use libcurl (<http://curl.se/>) to implement network access to
-    remote files via FTP, HTTP, HTTPS, etc.
+    remote files via FTP, HTTP, HTTPS, etc.  By default or with
+    --enable-libcurl=check, configure will probe for libcurl and include
+    this functionality if libcurl is available.  Use --disable-libcurl
+    to prevent this.
 
 --enable-gcs
     Implement network access to Google Cloud Storage.  By default or with
@@ -175,6 +178,12 @@ various features and specify further optional external requirements:
     and decompression.  It also includes a fast crc32 implementation.
     By default, ./configure will probe for libdeflate and use it if
     available.  To prevent this, use --without-libdeflate.
+
+Each --enable-FEATURE/--disable-FEATURE/--with-PACKAGE/--without-PACKAGE
+option listed also has an opposite, e.g., --without-external-htscodecs
+or --disable-plugins.  However, apart from those options for which the
+default is to probe for related facilities, using these opposite options
+is mostly unnecessary as they just select the default configure behaviour.
 
 The configure script also accepts the usual options and environment variables
 for tuning installation locations and compilers: type './configure --help'
--- a/configure.ac
+++ b/configure.ac
@@ -351,15 +351,20 @@ HTSlib.])])])])
 
 libcurl=disabled
 if test "$enable_libcurl" != no; then
-  AC_CHECK_LIB([curl], [curl_easy_pause],
-    [AC_DEFINE([HAVE_LIBCURL], 1, [Define if libcurl file access is enabled.])
-     libcurl=enabled],
+  libcurl_devel=ok
+  AC_CHECK_HEADER([curl/curl.h], [], [libcurl_devel="headers not found"], [;])
+  AC_CHECK_LIB([curl], [curl_easy_pause], [:],
     [AC_CHECK_LIB([curl], [curl_easy_init],
-       [message="library is too old (7.18+ required)"],
-       [message="library not found"])
-     case "$enable_libcurl" in
-       check) AC_MSG_WARN([libcurl not enabled: $message]) ;;
-       *) MSG_ERROR([libcurl $message
+       [libcurl_devel="library is too old (7.18+ required)"],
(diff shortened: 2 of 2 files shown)


jmarshall:
(This is the first `elif` in _configure.ac_, but there are already lots in the autoconf-generated parts of _configure_ so there are no portability issues in using it.)

jmarshall:
Non-configure builds have required libcurl since 1.10 (d5a0887f13f734a90b6e00009bc7c5d022f5c0c9); as noted in https://github.com/samtools/htslib/issues/1177#issuecomment-722010747, the maintainers may consider that it's time to promote libcurl to being required at build-time by default in configure-based builds too.

This would need a one-word change in _configure.ac_ and the `--enable-libcurl` documentation to be inverted, something like this:

```diff
diff --git a/configure.ac b/configure.ac
index c6b09db..4391e92 100644
--- a/configure.ac
+++ b/configure.ac
@@ -95,7 +95,7 @@ AC_SYS_LARGEFILE
 AC_ARG_ENABLE([libcurl],
   [AS_HELP_STRING([--enable-libcurl],
                   [enable libcurl-based support for http/https/etc URLs])],
-  [], [enable_libcurl=check])
+  [], [enable_libcurl=yes])
 diff --git a/INSTALL b/INSTALL
index e8d0bea..1d77f7c 100644
--- a/INSTALL
+++ b/INSTALL
@@ -148,12 +148,13 @@ various features and specify further optional external requirements:
---enable-libcurl
-    Use libcurl (<http://curl.se/>) to implement network access to
-    remote files via FTP, HTTP, HTTPS, etc.  By default or with
-    --enable-libcurl=check, configure will probe for libcurl and include
-    this functionality if libcurl is available.  Use --disable-libcurl
-    to prevent this.
+--enable-libcurl=check
+--disable-libcurl
+    By default or with --enable-libcurl, HTSlib includes code that uses
+    libcurl (<http://curl.se/>) to implement network access to remote files
+    via FTP, HTTP, HTTPS, etc.  Use --enable-libcurl=check to probe for
+    libcurl and include this functionality only if libcurl is available.
+    Use --disable-libcurl to omit this functionality.
```

daviesrob:
This doesn't quite work where the header can't be found, as it still defines `HAVE_LIBCURL`.  On a test machine where I removed `curl.h` I got:

```
checking for curl/curl.h... no
checking for curl_easy_pause in -lcurl... yes
configure: WARNING: libcurl not enabled: headers not found

[ ... ]

gcc -fvisibility=hidden  -o test/test_bgzf test/test_bgzf.o libhts.a -lz -lcurl -llzma -lbz2 -lz -lm   -lpthread
/usr/bin/ld: libhts.a(hfile.o): in function `load_hfile_plugins':
/home/debian/htslib/hfile.c:1084: undefined reference to `hfile_plugin_init_libcurl'
collect2: error: ld returned 1 exit status

[ ... ]

$ grep CURL config.h
#define HAVE_LIBCURL 1
```

It looks like this is because the default behaviour of `AC_CHECK_LIB()` isn't being overridden when looking for `curl_easy_pause` so it defines `HAVE_LIBCURL` when it finds the library.  Luckily the fix is easy:

```diff
diff --git a/configure.ac b/configure.ac
index c6b09db..53d03c1 100644
--- a/configure.ac
+++ b/configure.ac
@@ -353,7 +353,7 @@ libcurl=disabled
 if test "$enable_libcurl" != no; then
   libcurl_devel=ok
   AC_CHECK_HEADER([curl/curl.h], [], [libcurl_devel="headers not found"], [;])
-  AC_CHECK_LIB([curl], [curl_easy_pause], [],
+  AC_CHECK_LIB([curl], [curl_easy_pause], [:],
     [AC_CHECK_LIB([curl], [curl_easy_init],
        [libcurl_devel="library is too old (7.18+ required)"],
        [libcurl_devel="library not found"])])
```

Could you update the PR with this change, please, if you agree?

jmarshall:
D'oh, you're absolutely right, thanks.

I usually test these by changing the checks to `AC_CHECK_HEADER([curl/curlbork.h], …)` etc, which is easier than breaking my development machine. Unfortunately my testing focussed on enable/disable and not so much on check… thanks for catching this!

daviesrob:
Thanks for the update.

I can recommend throw-away virtual machines as very useful for tests that involve messing around with system files.

jmarshall:
Good point again — I have a stable of VMs running different platforms, but I'm mostly unpractised with (the UI for) making short-term throwaway clones of them for hacking on like this. Which is probably half the point for developers of using VMs…

PR #1270, opened 2021-04-15
Prevent assert() in bcf_sr_set_regions

kdm9:
It seems `bcf_sr_set_regions` ought be called before `bcf_sr_add_reader`. However, if a user unaware of this calls `bcf_sr_set_regions` **after** `bcf_sr_add_reader`, they are met with a failing `assert()`. This is particularly odd, as `bcf_sr_set_regions` already contains a useful error message about exactly this usage error.

This patch reorders the check for incorrect usage and the assert(), so that if an API user stuffs up the order of the calls, they get the nice message that already exists and not an assert().

I'm not sure if the assert() is even needed here, and probably is best replaced with a nice `hts_log_error()` in any case. I think
`readers->regions != NULL` only when either `bcf_sr_add_reader` or `bcf_sr_set_regions` has already been called, but I'll leave this in the maintainer's hands.

### DIFF
--- a/htslib/synced_bcf_reader.h
+++ b/htslib/synced_bcf_reader.h
@@ -302,6 +302,9 @@ int bcf_sr_set_samples(bcf_srs_t *readers, const char *samples, int is_file);
  *  file are currently not supported.
  *  Targets (but not regions) can be prefixed with "^" to request logical complement,
  *  for example "^X,Y,MT" indicates that sequences X, Y and MT should be skipped.
+ *
+ *  API note: bcf_sr_set_regions/bcf_sr_set_targets MUST be called before the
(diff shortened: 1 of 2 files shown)


valeriuo:
```c
    if ( readers->nreaders || readers->regions )
    {
        hts_log_error("Must call bcf_sr_set_regions() before bcf_sr_add_reader()");
        return -1;
    }
```
would probably be the cleanest solution; together with documenting on top of the `bcf_sr_set_regions` declaration, in `synced_bcf_reader.h`, that this method must be called before `bcf_sr_add_reader`.

kdm9:
@valeriuo I've implemented your suggestion for both set_regions and set_targets, and added a short note in the header file.

PR #1275, opened 2021-04-22
Permit platform specific BAQ parameters.

jkbonfield:
The flag passed to sam_prob_realn can now contain platform data, encoded via the htsRealnPlatform enum.

It also automatically selects long-read parameters if the read length is bigger than 1kb, but otherwise it defaults to the same parameters as before.

This helps bcftools mpileup call SNPs on PacBio CCS reads.

### DIFF
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -2031,6 +2031,21 @@ typedef struct bam_mplp_s *bam_mplp_t;
 HTSLIB_EXPORT
 int sam_cap_mapq(bam1_t *b, const char *ref, hts_pos_t ref_len, int thres);
 
+// Used as flag parameter in sam_prob_realn.
+enum htsRealnFlags {
+    BAQ_APPLY = 1,
+    BAQ_EXTEND = 2,
+    BAQ_REDO = 4,
+
+    // Platform subfield, in bit position 3 onwards
+    BAQ_AUTO = 0<<3,
+    BAQ_ILLUMINA = 1<<3,
+    BAQ_PACBIOCCS = 2<<3,
+    BAQ_PACBIO = 3<<3,
+    BAQ_ONT = 4<<3,
+    BAQ_GENAPSYS = 5<<3
+};
+
 /// Calculate BAQ scores
 /** @param b   BAM record
     @param ref     Reference sequence
@@ -2045,9 +2060,15 @@ This function calculates base alignment quality (BAQ) values using the method
 described in "Improving SNP discovery by base alignment quality", Heng Li,
 Bioinformatics, Volume 27, Issue 8 (https://doi.org/10.1093/bioinformatics/btr076).
 
+The @param flag value can be generated using the htsRealnFlags enum, but for
+backwards compatibilty reasons is retained as an "int".  An example usage
+of the enum could be this, equivalent to flag 19:
+
+    sam_prob_realn(b, ref, len, BAQ_APPLY | BAQ_EXTEND | BAQ_PACBIOCCS);
+
 The following @param flag bits can be used:
 
-Bit 0: Adjust the quality values using the BAQ values
+Bit 0 (BAQ_APPLY): Adjust the quality values using the BAQ values
 
  If set, the data in the BQ:Z tag is used to adjust the quality values, and
  the BQ:Z tag is renamed to ZQ:Z.
@@ -2055,22 +2076,28 @@ Bit 0: Adjust the quality values using the BAQ values
  If clear, and a ZQ:Z tag is present, the quality values are reverted using
  the data in the tag, and the tag is renamed to BQ:Z.
 
-Bit 1: Use "extended" BAQ.
+Bit 1 (BAQ_EXTEND): Use "extended" BAQ.
 
  Changes the BAQ calculation to increase sensitivity at the expense of
  reduced specificity.
 
-Bit 2: Recalculate BAQ, even if a BQ tag is present.
+Bit 2 (BAQ_REDO): Recalculate BAQ, even if a BQ tag is present.
 
  Force BAQ to be recalculated.  Note that a ZQ:Z tag will always disable
  recalculation.
+
+Bits 3-10: Choose parameters tailored to a specific instrument type.
+
+ One of BAQ_AUTO, BAQ_ILLUMINA, BAQ_PACBIOCCS, BAQ_PACBIO, BAQ_ONT and
+ BAQ_GENAPSYS.  The BAQ parameter tuning are still a work in progress and
+ at the time of writing mainly consist of Illumina vs long-read technology
+ adjustments.
 
 @bug
 If the input read has both BQ:Z and ZQ:Z tags, the ZQ:Z one will be removed.
 Depending on what previous processing happened, this may or may not be the
 correct thing to do.  It would be wise to avoid this situation if possible.
 */
-
 HTSLIB_EXPORT
 int sam_prob_realn(bam1_t *b, const char *ref, hts_pos_t ref_len, int flag);
 
--- a/realn.c
+++ b/realn.c
@@ -103,13 +103,32 @@ static int realn_check_tag(const uint8_t *tg, enum htsLogLevel severity,
     return 0;
 }
 
-int sam_prob_realn(bam1_t *b, const char *ref, hts_pos_t ref_len, int flag)
-{
-    int k, bw, y, yb, ye, xb, xe, apply_baq = flag&1, extend_baq = flag>>1&1, redo_baq = flag&4, fix_bq = 0;
+int sam_prob_realn(bam1_t *b, const char *ref, hts_pos_t ref_len, int flag) {
+    int k, bw, y, yb, ye, xb, xe, fix_bq = 0, apply_baq = flag & BAQ_APPLY,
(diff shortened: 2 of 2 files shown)


realn.c
```diff
@@ -106,10 +106,29 @@ static int realn_check_tag(const uint8_t *tg, enum htsLogLevel severity,
...
+
+    // d(I) e(M) band
+    probaln_par_t conf = { 0.001, 0.1, 10 }; // Illumina
+
+    if (b->core.l_qseq > 1000 || system > BAQ_illumina) {
+        // Params that work well on PacBio CCS 15k.  Unknown if they
+        // help other long-read platforms yet, but likely better than
+        // the short-read tuned ones.
+        //
+        // This function has no access to thinks like the SAM header.
```
jmarshall, line 122:
```suggestion
        // This function has no access to things like the SAM header.
```

I guess in the fullness of time it might be worth having an HTSlib function that took an `@RG` header and returned a `sam_prob_realn()` flags value…

realn.c
```diff
@@ -106,10 +106,29 @@ static int realn_check_tag(const uint8_t *tg, enum htsLogLevel severity,
...
+        // Ideally the calling function would check for e.g.
+        // @RG PL = "PACBIO" and FS contains "READTYPE=CCS".
```
jmarshall, line 124: DS?

realn.c
```diff
@@ -106,10 +106,29 @@ static int realn_check_tag(const uint8_t *tg, enum htsLogLevel severity,
 int sam_prob_realn(bam1_t *b, const char *ref, hts_pos_t ref_len, int flag)
 {
     int k, bw, y, yb, ye, xb, xe, apply_baq = flag&1, extend_baq = flag>>1&1, redo_baq = flag&4, fix_bq = 0;
+    enum htsRealnPlatform system = (flag>>3)<<3;
```
daviesrob, line 109: I think this might be better as `system = flag & (0x1f << 3)` to select the exact bits mentioned in the header.

jkbonfield: I thought about that.  My comment hypothecated it being bits 3-7, but we don't use anything higher so it makes zero difference so I just use whatever is remaining to provide freedom for changing things.

I question whether this amendment of enum values really matters anyway.  We're trading having to do shifting in the caller (which admittedly doesn't even exist at the moment), vs having to do shifting in the htslib side.  Given we already have a bit field where the user is expected to shift-or, shift-or, shift-or, it may even be counterintuitive to require one and only one element to not be shifted.

jmarshall:
> The per platform values aren't separate bits, it's just been obfuscated a bit.

Oops, so they are. D'oh! (Maybe don't call them “bits 3-7” then :smile:)

> The reason I chose a platform field rather than breaking it down to specific bits is because in time I may adjust the `probaln_par_t` values further.

Yes, hence the expectation for further specific flag bits, and the RG-based function that would also be an HTSlib API function — so programs using it would also benefit from the improved capabilities when linking against future HTSlibs.

> Therefore the alternative to a platform bit would be an entirely new function with the probaln_par_t parameter being passed in instead. I felt I'd try the easier solution first though.

Yes, the alternative would be `sam_prob_realn2()` or so that also allowed the caller to specify the `probaln_par_t` to be used (and I don't what what other knobs it might be useful to be able to adjust, but maybe flags would suffice there). Combined perhaps with an RG-based function that would return a suitable default `probaln_par_t` for your platform.

The difference is between specifying a platform in a fairly inflexible way and having all the magic happen on the inside of `sam_prob_realn()`, versus exposing the magic on the outside so that anyone can tune the parameters easily if they want to.

Anyway, that's my 2¢…

> I like the idea of moving the other bit fields into the enum though (with a type rename). That would then genuinely mean we can just OR together enum terms to construct the bit field, in addition to literal values like 3 and 7 for compatibility.

Yep, tres cool.

jkbonfield:
> > The per platform values aren't separate bits, it's just been obfuscated a bit.
> 
> Oops, so they are. D'oh! (Maybe don't call them “bits 3-7” then 😄)
> 

Well it was correct when the PR was made.

Shifting them didn't really fit with me as it felt muddled, but with hindsight the right approach would have been to embrace it further as per your suggestion of putting other bit values in there too.  I'll update the PR.  Thanks.

htslib/sam.h
```diff
@@ -2065,14 +2081,20 @@ Bit 2: Recalculate BAQ, even if a BQ tag is present.
...
 @bug
 If the input read has both BQ:Z and ZQ:Z tags, the ZQ:Z one will be removed.
 Depending on what previous processing happened, this may or may not be the
 correct thing to do.  It would be wise to avoid this situation if possible.
 */
-
 HTSLIB_EXPORT
-int sam_prob_realn(bam1_t *b, const char *ref, hts_pos_t ref_len, int flag);
+int sam_prob_realn(bam1_t *b, const char *ref, hts_pos_t ref_len,
+                   enum htsRealnFlags flag);
```
jmarshall, line 2097: In theory this is an ABI breakage because the underlying type of an enumeration is implementation-defined and might not be the same size as `int`. In practice on modern platforms for parameter passing, there is probably not a problem.

Where this change certainly **does** cause a problem is in C++: any existing C++ code that calls `sam_prob_realn()` with an `int` for a flag will become a compile-time error.

jmarshall: BTW ensuring the enum sizes are future-proofed even with `-fshort-enums` is why _hts.h_'s enums (`htsFormatCategory` and friends) have a rather goofy `foo_maximum = 32767` entry that's otherwise useless.

jkbonfield: Meh!  I did do a very cursory search which said it was int (but I suspect it was just comparable to int, rather than always represented as int). Anyway C++ as usual barges in to spoil the party!  I guess we'll just have to keep it as int and document the enum in comments then.

Thanks.

jkbonfield:
I've removed the enum from the public function protoype, and rebased / squashed.

The commit message mentions the possible future TODOs regarding header usage, either selecting the bit field automatically or more flexibly selecting the parameters and passing in to a new realignment function.  However that's ongoing work and can be subject of a follow up commit.

jmarshall:
Nice detection of line 1903's potential pointer arithmetic on NULL by the debian-pedantic CI there — but how come that line hasn't caused a CI failure before?!

daviesrob:
A new gcc recently came out.  Maybe it got upgraded on cirrus-ci?

jkbonfield:
I'm struggling to work out how it could fail.  `cp_end` is initialised, and it can never be NULL as `dat` malloc is checked too - *unless* `in_size` is zero.  In that case both pointers are NULL, but the code is unreachable.

We should tidy up the code to silence the false warning though.

jmarshall:
`dat` isn't checked until _after_ the arithmetic occurs, so the compiler deduces that `cp_end` is the result of undefined behaviour. [It's a classic!](https://lwn.net/Articles/575563/) (the second example):smile:

jkbonfield:
> `dat` isn't checked until _after_ the arithmetic occurs, so the compiler deduces that `cp_end` is the result of undefined behaviour. [It's a classic!](https://lwn.net/Articles/575563/) (the second example)😄

Interesting.  It's a valid concern that the check should be before the addition, although in this case it's not what the compiler is complaining about.  It claims it may be uninitialised.  I guess this is a result of some curious optimisation that's been applied due to the potential for an undefined behaviour code path.

jkbonfield:
Anyway,  it needs fixing somewhere so I may as well add a new commit so we can test it here. :-)

daviesrob:
Please fix it in a separate pull request, not here.

jkbonfield:
Does it really matter?  I meant a separate commit of course, but if we want the extra red tape then fine I'll do it that way.  However it will mean you have to juggle the order of merging if you also want to merge this only when passing all tests.

daviesrob:
I prefer not to mix things up too much.  Also, this PR still needs work - mainly to comments which now don't match the code they describe.  Some typos have not been fixed (see suggestions above). Also, I don't think it's a good idea to use every single remaining bit in `flags` for the platform as we may want to define other flags in the future.  Masking out a reasonable number of bits for the platform type would seem a good idea.

jkbonfield:
Updated with review comments.

- Fixed some more comment typos that I missed before, sorry.
- Be explicit with exactly 8 bits permitted for system type (ie reserve a maximum of 255 different configurations).

daviesrob:
Rebased to fix tests...

realn.c
```diff
@@ -103,13 +103,32 @@ static int realn_check_tag(const uint8_t *tg, enum htsLogLevel severity,
...
+
+    // d(I) e(M) band
+    probaln_par_t conf = { 0.001, 0.1, 10 }; // Illumina
+
+    if (b->core.l_qseq > 1000 || system > BAQ_ILLUMINA) {
+        // Params that work well on PacBio CCS 15k.  Unknown if they
+        // help other long-read platforms yet, but likely better than
+        // the short-read tuned ones.
+        //
+        // This function has no access to thinkg like the SAM header.
```
jmarshall, line 122:
```suggestion
        // This function has no access to things like the SAM header.
```

Close, but no CIGAR :smile:

jkbonfield: Lol it's probably a good thing I called it a day when I did!  Clearly head not screwed on right this afternoon.

PR #1277, opened 2021-04-29
Support building in a separate build directory

### LINKED ISSUE #231: out of tree builds
Nice to have a configure script - slightly frustrating that it does not allow building out of the source tree e.g.:
```
dj3@sf2-farm-srv1:~/repos/htslib/build$ ../configure && make
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for ranlib... ranlib
checking for zlib.h... yes
checking for inflate in -lz... yes
configure: creating ./config.status
config.status: creating config.mk
make: *** No targets specified and no makefile found. Stop.
```

jmarshall:
Unlike the traditional recursive _Makefile.in_ autoconf approach, HTSlib uses a single non-recursive non-generated Makefile that includes fragments generated by _configure_. Hence _configure.ac_ needs to detect out-of-tree configuration as a special case and explicitly create _Makefile_ and the tree of subdirectories in the build directory.

Contrary to some naive expectations, using autoconf and/or automake does not mean that out-of-tree builds are automatically easily supported: you need to ensure that your makefile rules are written with `$(srcdir)/…` in appropriate places as well. For HTSlib's `make test` rule in particular, builddir and srcdir are intricately intertwined; moreover some of the commands `cd` into a subdirectory and then expect data files in srcdir and executables in builddir to both be in the current directory. This first commit does not attempt to disentangle the `make test` recipe commands.

Fixes #231, at least for `make && make install` — though currently not for `make test`.

Further details from the commit message:

Add _builddir_vars.mk.in_, which for an out-of-tree build (only) is generated into _builddir/htslib_vars.mk_. It sets `VPATH` and some other additional variables and then includes the real _srcdir/htslib_vars.mk_. This is used to override `srcdir` etc for out-of-tree builds, and notably adds an `-I` option for _htscodecs/version.h_ which in an in-tree build is found by being in the same directory as the C file that includes it.

HTSlib also provides _htslib.mk_ for use by external projects. Split its variables and rules into `$(HTSDIR)` and `$(HTSSRCDIR)` as appropriate. These are the same by default, but _builddir_vars.mk.in_ overrides `$(HTSSRCDIR)` to be `$(HTSDIR)/$(`[HTSlib's ]`srcdir)` instead, taking care to handle the case of `$(srcdir)` being an absolute path.

The `include` in _htscodecs.mk.in_ would need to be `$(srcdir)/selected.mk` for _Makefile_ but `$(HTSPREFIX)selected.mk` for _htslib.mk_. Rather than arrange this, it's easier to generate _builddir/htscodecs.mk_ as a symlink to _srcdir/selected.mk_ instead, using `AC_CONFIG_LINKS`.

Add `$(srcdir)/` or `$(srcprefix)` (for tidiness in in-tree build logs) to commands in recipes containing explicit filenames in source directories. This has been done for `make all install` but not yet for `make test`.


jmarshall:
I have a followup commit that fixes a few of the commands in `make test`, but it is nontrivial especially for the commands that change the working directory. I am fairly sure this is where the draft of separate builddir support got bogged down years ago when the _configure_ script was first introduced…

I'd suggest merging this now, and adding separate builddir support to the `make test` recipes incrementally later.

PR #1278, opened 2021-04-30
Detect incorrect number of AC fields

pd3:
Incorrect number of AC fields (values greater than nALT) could cause writes beyond
allocated memory, resulting in core dumps and cryptic messages such as
"corrupted size vs. prev_size" or "munmap_chunk(): invalid pointer"


valeriuo:
Merged into develop as bf32d5624d9732824af039fbae157a9c3e41bd28

PR #1280, opened 2021-05-04
Fix undefined behaviour in cram_xdelta_encode_char.

jkbonfield:
Assumption: this is a gcc-11 warning, but I cannot be sure as I don't
have a local install.

    cram/cram_codecs.c:1916:19: error: 'cp_end' may be used uninitialized [-Werror=maybe-uninitialized]
     1916 |             cp += c->vv->varint_put32(cp, cp_end, zigzag16(c->u.e_xdelta.last));
          |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

cp_end is initialised, but due to the malloc potentially returning
NULL, the initialisation of cp_end will have undefined behaviour if
the malloc fails.  The code does then check for NULL malloc and exits,
but the compiler generated code appears to have at least one path
which doesn't initialise cp_end due to this undefined behaviour.
(Unknown if a reachable path, but it's unlikely.)

### DIFF
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -1898,16 +1898,20 @@ int cram_xdelta_encode_int(cram_slice *slice, cram_codec *c,
     return -1;
 }
 
+#if __GNUC__ == 11 && __GNUC_MINOR__ == 1
+// See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100417
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
 int cram_xdelta_encode_char(cram_slice *slice, cram_codec *c,
-                           char *in, int in_size) {
+                            char *in, int in_size) {
     char *dat = malloc(in_size*5);
     if (!dat)
         return -1;
     char *cp = dat, *cp_end = dat + in_size*5;
 
     c->u.e_xdelta.last = 0; // reset for each new array
-    switch(c->u.e_xdelta.word_size) {
-    case 2: {
+    if (c->u.e_xdelta.word_size == 2) {
         int i, part;
 
         part = in_size%2;
@@ -1923,9 +1927,6 @@ int cram_xdelta_encode_char(cram_slice *slice, cram_codec *c,
             c->u.e_xdelta.last = le_int2(in16[i]);
             cp += c->vv->varint_put32(cp, cp_end, zigzag16(d));
         }
-
-        break;
-    }
     }
     if (c->u.e_xdelta.sub_codec->encode(slice, c->u.e_xdelta.sub_codec,
                                       (char *)dat, cp-dat)) {
@@ -1936,6 +1937,9 @@ int cram_xdelta_encode_char(cram_slice *slice, cram_codec *c,
     free(dat);
     return 0;
 }
+#if __GNUC__ == 11 && __GNUC_MINOR__ == 1
+#pragma GCC diagnostic pop
+#endif
 
 void cram_xdelta_encode_free(cram_codec *c) {
     if (!c) return;
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -1900,9 +1900,10 @@ int cram_xdelta_encode_int(cram_slice *slice, cram_codec *c,
 
 int cram_xdelta_encode_char(cram_slice *slice, cram_codec *c,
                            char *in, int in_size) {
-    char *dat = malloc(in_size*5), *cp = dat, *cp_end = dat + in_size*5;
+    char *dat = malloc(in_size*5);
     if (!dat)
         return -1;
+    char *cp = dat, *cp_end = dat + in_size*5;
 
     c->u.e_xdelta.last = 0; // reset for each new array
     switch(c->u.e_xdelta.word_size) {


jkbonfield:
Or not...  I'll build gcc-11 locally so I can test prioperly.  Mutter!

jmarshall:
LOL wut? I guess the next question is what's up with that single-item `switch`…

daviesrob:
It might be worth temporarily making the build script do `gcc -v` so we know which version it's actually using.

jkbonfield:
If my gcc 11 tests don't reveal a problem I'll do that, but it seems unnecessary work for now.

daviesrob:
Yes, I see our cirrus-ci configuration uses `gcc:latest`, which at this point in time is [`11.1.0`](https://github.com/docker-library/gcc/blob/bb63178b4a0781555d904c3ac3897c9b81d922c8/11/Dockerfile)

jkbonfield:
Wow, this is really broken!

```
#include <stdlib.h>                                                             
                                                                                
void ext_func(const char *endp);                                                
                                                                                
void func(char *in) {                                                           
    char *dat;                                                                  
    if (!(dat = malloc(10)))                                                    
        return;                                                                 
                                                                                
    ext_func(dat);                                                              
                                                                                
    free(dat);                                                                  
}                                                                               
```

Nice and simple, but:

```
_.c: In function 'func':
_.c:10:5: warning: 'dat' may be used uninitialized [-Wmaybe-uninitialized]
   10 |     ext_func(dat);
      |     ^~~~~~~~~~~~~
_.c:3:6: note: by argument 1 of type 'const char *' to 'ext_func' declared here
    3 | void ext_func(const char *endp);
      |      ^~~~~~~~
```

Removing the `const` from the `ext_func` prototype removes the warning.  I suspect it's a new gcc bug.  Maybe malloc has some attributes which is triggering an issue somewhere with undefined behaviour.

If I declare `extern void *alloc(size_t size);` and call that instead of `malloc` then it works.

jkbonfield:
Oh.... I get it now!

It's claiming the memory pointed to by the variable may be uninitialised, not that the variable itself is uninitialised.  The presence of `const` informs the compiler it's only being read from and not written to, so the lack of any chance for writing to it also means the data cannot be initialised in that unknown external function.

jkbonfield:
So this minimal example demonstrates it better in the context that the gcc authors intended probably.

```
#include <stdio.h>                                                              
#include <stdlib.h>                                                             
                                                                                
void fill(char *start, const char *end);                                        
                                                                                
char *func(char *in) {                                                          
    char *dat;                                                                  
    //if (!(dat = calloc(1, 10))) // works                                      
    if (!(dat = malloc(10)))                                                    
        return NULL;                                                            
                                                                                
    fill(dat, dat+10);                                                          
    return dat;                                                                 
}                                                                               
```

The calloc call instead fixes it, as it knows that dat is ininitialised.

jmarshall:
Annoying. I didn't immediately find a report of this by searching [GCC Bugzilla](https://gcc.gnu.org/bugzilla/) for “may be used uninitialized”. It would have been clearer if the message was “data pointed to by ‘cp_end’ may be used uninitialized” or so…

This could also be suppressed by marking `varint_put32()` in _cram_structs.h_ as `__attribute__((access(none, 2)))` but that causes other spurious warnings.

jkbonfield:
I filed a bug report anyway: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100417

We'll see what they say.  I can see why a naive implementation would be reporting this, but too many false positives undermines the benefit of the warnings.

daviesrob:
It's a bit ugly, but I guess it will do until we think of something better.  I'll be interested to see what the response to the gcc bug report is.

jkbonfield:
I wasn't convinced this is the best way to solve the problem, but it's merged now so if we come up with an alternative we can change it once again obviously.

There are some comments to the bug report now, with acknowledgement that the check probably shouldn't happen on a zero sized object (which cp_end is given it points to the end of the malloc buffer).  It's possible that there may be attributes to disable checking too for this function, which could be a better solution, eg `__attribute__ ((access (none, 2)))`, but this gives other errors complaining about size 0 (odd given we've stated we're not accessing it).

jmarshall:
Attributes would help in future, but as long as 11.1 is a reasonably current version out there with this behaviour (and especially as long as it's in CI with `-Werror`!) the current workaround would be necessary.

The alternative would be to drop `const` from the `int (*varint_put32) (char *cp, const char *endp, int32_t val_p)` signature and the other similar functions. To be honest, I think this would be the right thing to do: `cp` is already a `char *` pointer into the same array (so that array already has to be non-const), and `endp` is mostly just used in arithmetic with `cp`. However this would be an htscodecs API change and changing this would imply changes sprinkled across htslib, htscodecs, and probably io_lib as well…

jkbonfield:
Agreed on the const change.  For now we needed a temporary hack so we don't get CI breakage.  The alternative would have been to modify cirrus-CI to explicitly request gcc 10 instead of "latest".  This may have been better as I suspect this is the tip of the iceburg here.  I've seen similar new errors in samtools already from gcc 11 and almost certainly will in bcftools too.

Still, to their credit, I'm impressed with how rapidly latest migrated over.  It took about a week from release.  It's already shown its worth too as we spotted an issue before the users did. :-)

jmarshall:
It turns out that the `const` change doesn't percolate all the way through htscodecs and io_lib after all; see PR #1285 for a alternative without needing to disable diagnostics.

PR #1281, opened 2021-05-04
[index] Fix the update of the minimum offset, when unmapped reads are present

### LINKED ISSUE #1279: BAM subsetting broken in 1.12
Subsetting appears to be broken in htslib 1.12. In particular, reads are being missed when the target region is too large:

```shell
$ samtools version | head -2
samtools 1.12
Using htslib 1.12
$ samtools view mybam.bam 13:23907893-23907932 | wc -l
87
$ samtools view mybam.bam 13:16907893-23907932 | wc -l
1750673
$ samtools view mybam.bam 13:15907893-23907932 | wc -l
0
```

The problem does not occur with 1.11:

```shell
$ samtools 2>&1 >/dev/null | head -3

Program: samtools (Tools for alignments in the SAM format)
Version: 1.11 (using htslib 1.11)
$ samtools view mybam.bam 13:15907893-23907932 | wc -l
2124456
```

This is with Conda installed samtools/htslib on Linux but I see the same issue on MacOS with Homebrew installed samtools/htslib. I'm confident this is a htslib issue as I also see the same behaviour in my tool [octopus](https://github.com/luntergroup/octopus) that directly calls htslib.

valeriuo:
The linear index can be used to fine tune the minimum file offset from where a targeted read starts. However, the linear index does not cover unmapped reads, so a solution was constructed to look backwards for a valid offset which is guaranteed to cover the unmapped reads. 
The problem arose form a heuristic that was employed when no other offset value was found in the backwards search. The current code uses the starting offset of the first chunk of the first matching bin from the bin index, in the assumption that this chunk will cover the starting coordinate of interest. However, this assumption was not always met, so the current fix reverts to setting the minimum offset to the corresponding bin `loff` value, which was the initial value of the minimum offset, before it was updated by the linear index.

Fixes #1279

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -2966,6 +2966,8 @@ hts_itr_t *hts_itr_query(const hts_idx_t *idx, int tid, hts_pos_t beg, hts_pos_t
                 if (min_off < idx->lidx[tid].offset[rel_off])
                     min_off = idx->lidx[tid].offset[rel_off];
                 if (unmapped) {
+                    // unmapped reads are not covered by the linear index,
+                    // so search backwards for a smaller offset
                     int tmp_off;
                     for (tmp_off = rel_off-1; tmp_off >= 0; tmp_off--) {
                         if (idx->lidx[tid].offset[tmp_off] < min_off) {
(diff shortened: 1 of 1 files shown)


dancooke:
Any idea when this will be available in a release?

jkbonfield:
> Any idea when this will be available in a release?

Sorry no, but our 3 monthly release cycling has been badly slipping of late, so I would be up for making a minor point release (1.12.1) and backporting this change into it as it's an important bug to fix.

PR #1285, opened 2021-05-06
Work around GCC 11.1 problem by removing `const` from `varint_putNN()` `endp` parameters

jmarshall:
If anyone can face revisiting #1280 again… it turns out that removing the `const` from `const char *endp` in  the `varint_putNN()` functions is a less invasive change than I feared. In particular, no htscodecs update is immediately needed.

The underlying htscodecs `var_put_u32()` etc functions still take a const pointer for their `endp` parameters, but it's fine for callers to pass a non-const pointer to that as the types are compatible in that direction. So we can remove `const` here to avoid the spurious GCC 11.1 warning and leave the corresponding htscodecs change for later (if at all).

As `endp` points to the same array as `cp`, it might as well be `char *` as well, which avoids the spurious warning. As partially noted in _htscodecs/varint.h_, the ideal signatures for the getters and putters would be

```c
int64_t (*varint_get32)(const char **cp, const char *endp, int *err);
int (*varint_put32)(char *cp, char *endp, int32_t val_p);
/* etc */
```

If it was going to be done at all, adding `const` to the getters' `cp` arguments would have to occur after htscodecs had been updated correspondingly and requires rather larger changes in htslib's CRAM code. OTOH happily removing `const` from the putter callers can be done in advance of any corresponding htscodecs change.

While in the neighbourhood, make the "for now" wrappers `static`.

### DIFF
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -1898,11 +1898,6 @@ int cram_xdelta_encode_int(cram_slice *slice, cram_codec *c,
     return -1;
 }
 
-#if __GNUC__ == 11 && __GNUC_MINOR__ == 1
-// See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100417
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
-#endif
 int cram_xdelta_encode_char(cram_slice *slice, cram_codec *c,
                             char *in, int in_size) {
     char *dat = malloc(in_size*5);
@@ -1937,9 +1932,6 @@ int cram_xdelta_encode_char(cram_slice *slice, cram_codec *c,
     free(dat);
     return 0;
 }
-#if __GNUC__ == 11 && __GNUC_MINOR__ == 1
-#pragma GCC diagnostic pop
-#endif
 
 void cram_xdelta_encode_free(cram_codec *c) {
     if (!c) return;
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -739,15 +739,15 @@ static int64_t safe_ltf8_get(char **cp, const char *endp, int *err) {
 }
 
 // Wrapper for now
-int safe_itf8_put(char *cp, const char *cp_end, int32_t val) {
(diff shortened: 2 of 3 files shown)


jkbonfield:
I haven't looked at the details yet, but I like this direction.  I never really intended the pragmas to be a workable long term solution, but a quick hack to get tests passing.

I'd agree the const really isn't helping much here.

daviesrob:
Reading through the lines of [the reply to the gcc bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100417#c1) it looks like passing (pointer, length) would be the most reliable solution, but that really would be a big change.  Meanwhile we can go with this and see how long it sticks.

I also see [another report of the same problem](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100496) which looks like it's related to [libexpat](https://github.com/libexpat/libexpat).  Hopefully that will help encourage the gcc maintainers to produce a sustainable fix for this.

jkbonfield:
I don't see changing from `ptr, end_ptr`  to `ptr, len` as a good move.  We deliberately use `end_ptr` as it's constant and never changes.  As we're doing writes to `*ptr++` and it's constantly changing, if we track a length we'd need to change that to `*ptr++, len--`.  That's extra CPU cycles sometimes in tight loops, added purely to silence one version of one compiler.  Especially given it'll likely get bug-fixed away again.

The correct move is this PR; to mark the data pointed to by `end_ptr` to be non-constant given the function does have the ability to modify that data.

jmarshall:
It's clear that the non-const `char *end_ptr` will always prevent this warning, as the warning is about reading uninitialised data and it won't assume non-const `char *` is only for reading. IMHO the true type invariant for both getters and putters is that `typeof(cp) == typeof(endp)` so IMHO this one should have been `char *` all along. So I don't see deconstifying this as a compromise with the compiler, but rather as a small improvement.

In the end, I don't see this a lot differently than putting extra parentheses around `if ((ret = foo()))` to suppress an instance of a generally very useful “did you mean `==`” warning. (Except that I'm a bit dubious about the value of this warning in user-level code :smile: but it's early days…)

PR #1286, opened 2021-05-12
[index] Initialize the missing entries of the linear index in reverse order

### LINKED ISSUE #486: ioff for linear index appears to be set improperly for empty bins
When an interval of the linear index is empty (no overlapping features), it is being set to the ioff of the previous upstream interval. This means the feature found at that location does not overlap the desired bin and instead can be found far upstream of that bin. I believe the more logical behavior would be to instead set ioff of the empty interval to that of the next downstream interval with a valid ioff. The downstream ioff would correspond to the location of where the first overlapping feature could occur if it existed.

Without the proper setting, chunks in the bin index appear as if they could belong to the empty location because ioff occurs too far upstream so more cnk_end locations artificially pass the linear index filter. If ioff is instead set to the next valid downstream ioff as mentioned, then these chunks can be properly ignored. Being able to assign chunks accurately is important for example when estimating bin content using the index as a proxy for coverage (i.e. estimating attributes of a dataset without having to parse it).

valeriuo:
For each contig, the linear index is an array of the smallest file offsets corresponding to alignments that cover a particular contig window, each 2<sup>min_shift</sup> bp wide. In case there are no alignments covering a particular genomic window, the linear index entries are filled from adjacent valid entries.
Whereas before, the empty linear index entries were filled with values of the nearest valid entry from the **left**, the current PR changes this procedure to use the nearest valid entry from the **right**, making the linear index values of empty windows slightly larger. This change has the potential to improve some reading operations by skipping over unnecessary data. However, in practice, most of HTSlib indexed reads use iterators, which rely on actual index chunks of existing alignments. So the linear index value is used as a starting point for searching valid chunks and not as an actual file offset. Since there are no valid chunks corresponding to the empty windows, there will be no change in performance for iterator reads.

The `dump_index` method has been renamed into `idx_dump` and a few new display capabilities have been added to it:

- showing the index level of a bin
- showing the genomic region covered by a bin

The `hts_idx_save_core` has been renamed into `idx_save_core` to reflect its static use and to match the reading equivalent `idx_read_core`.

The new `hts_bin_level` method has been added. It computes the index level of a bin.

Fixes #486

### DIFF
--- a/htslib/hts.h
+++ b/htslib/hts.h
@@ -845,8 +845,10 @@ typedef struct hts_itr_t {
 
 typedef hts_itr_t hts_itr_multi_t;
 
-    #define hts_bin_first(l) (((1<<(((l)<<1) + (l))) - 1) / 7)
-    #define hts_bin_parent(l) (((l) - 1) >> 3)
+/// Compute the first bin on a given level
+#define hts_bin_first(l) (((1<<(((l)<<1) + (l))) - 1) / 7)
+/// Compute the parent bin of a given bin
+#define hts_bin_parent(b) (((b) - 1) >> 3)
 
 ///////////////////////////////////////////////////////////
 // Low-level API for building indexes.
@@ -1446,10 +1448,26 @@ static inline int hts_reg2bin(hts_pos_t beg, hts_pos_t end, int min_shift, int n
     return 0;
 }
 
+/// Compute the level of a bin in a binning index
+static inline int hts_bin_level(int bin) {
+    int l, b;
+    for (l = 0, b = bin; b; ++l, b = hts_bin_parent(b));
+    return l;
+}
+
+/// Compute the corresponding entry into the linear index of a given bin from
+/// a binning index
+/** @param bin    The bin number
+ *  @param n_lvls The index depth (number of levels - 0 based)
+ *  @return       The integer offset into the linear index
+ *
+ *  Explanation of the return value formula:
+ *  Each bin on level l covers exp(2, (n_lvls - l)*3 + min_shift) base pairs.
+ *  A linear index entry covers exp(2, min_shift) base pairs.
+ */
 static inline int hts_bin_bot(int bin, int n_lvls)
 {
-    int l, b;
-    for (l = 0, b = bin; b; ++l, b = hts_bin_parent(b)); // compute the level of bin
+    int l = hts_bin_level(bin);
     return (bin - hts_bin_first(l)) << (n_lvls - l) * 3;
 }
 
--- a/htslib/hts.h
+++ b/htslib/hts.h
@@ -1455,9 +1455,10 @@ static inline int hts_bin_level(int bin) {
     return l;
 }
 
-/// Compute the corresponding entry into the linear index of a given bin from
-/// a binning index
-/** @param bin    The bin number
+//! Compute the corresponding entry into the linear index of a given bin from
+//! a binning index
+/*!
+ *  @param bin    The bin number
  *  @param n_lvls The index depth (number of levels - 0 based)
  *  @return       The integer offset into the linear index
  *
--- a/hts.c
+++ b/hts.c
@@ -2117,7 +2117,7 @@ static void update_loff(hts_idx_t *idx, int i, int free_lidx)
     int l;
     // the last entry is always valid
     for (l=lidx->n-2; l >= 0; l--) {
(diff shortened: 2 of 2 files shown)


htslib/hts.h
```diff
@@ -1446,10 +1448,26 @@ static inline int hts_reg2bin(hts_pos_t beg, hts_pos_t end, int min_shift, int n
...
 
+/// Compute the level of a bin in a binning index
+static inline int hts_bin_level(int bin) {
+    int l, b;
+    for (l = 0, b = bin; b; ++l, b = hts_bin_parent(b));
+    return l;
+}
+
+/// Compute the corresponding entry into the linear index of a given bin from
+/// a binning index
```
jmarshall, line 1459: This style of brief description has to be on a single `///` line. (See [doxygen docs](https://www.doxygen.nl/manual/docblocks.html#cppblock), “for the brief description”, third option.)

jkbonfield:
I'm struggling to observe any impact on this.

I have 10,000 1KB regions spread across chr22 on a TopHat aligned Exome.  It has ~11% of bases covered, so there should be lots of holes for the linear index to make an impact on.  Using `strace` to dump out all the lseeks I cannot find a single difference, both with `test_view` and `test_view -M` (although I didn't expect the latter to differ anyway).  This was with a .bai index.  I tried with .csi and there's no difference either.    Although there is a big difference of course between using .bai and .csi, sadly with the latter having 23% more seeks and 16% more bytes read.  Losing the linear index in CSI was a bad move. :-(

I can't really understand why there is no difference.  The crux of it is seeding the empty linear index slots with the previous non-empty index item vs the next non-empty index item.  The latter definitely sounds like the right thing to do, but I don't understand how it doesn't become a change in seek location.

I also tested the impact on rebuilding the index.  That *did* change things, marginally for the better.  On the old index vs new index I get:

```
# Old
    Num. seeks   	8359
    Num. read    	33762
    Bytes read   	1098949004
    
 # New
    Num. seeks   	8292
    Num. read    	33651
    Bytes read   	1095311756
```

Maybe I misunderstand how this is being called then.  I assumed the linear index contained empty items on disk and the empty bits got populated on reading, but this doesn't appear to be the case.  I can confirm that the alignments returned by these 10,000 queries are identical.

I'll continue digging and use your dump index tool to understand it better (thanks).

There are still some other useful additions here, such as the improved index dumping, improved documentation and `hts_bin_level` function.  So irrespective of our assessment on `update_loff`, this is worth persuing further.

jkbonfield:
I can see that there are missing bins, which presumably also means missing loff figures, but the bins that are present all have an loff assigned.  I see 1269 loff differences between the two indices, out of 72207.  So subtle, but there. 

I've looked at the code and see `hts_read_core` contains this:

https://github.com/valeriuo/htslib/blob/lindex_rev/hts.c#L2592-L2593
```
            for (j = 1; j < l->n; ++j) // fill missing values; may happen given older samtools and tabix
                if (l->offset[j] == 0) l->offset[j] = l->offset[j-1];
```

This appears to be a duplicate of the bit of code you changed, which is called by the index generation function.  Hence why we see no difference when running on an existing index.  However I tried changing it to the same logic and it made no difference still:

```
-            for (j = 1; j < l->n; ++j) // fill missing values; may happen given older samtools and tabix
-                if (l->offset[j] == 0) l->offset[j] = l->offset[j-1];
+            for (j=l->n-2; j >= 0; j--)
+                if (l->offset[j] == 0) l->offset[j] = l->offset[j+1];
```

valeriuo:
> I also tested the impact on rebuilding the index. That _did_ change things, marginally for the better. On the old index vs new index I get:

Indeed. You can only see a difference by regenerating the index file.

> Maybe I misunderstand how this is being called then. I assumed the linear index contained empty items on disk and the empty bits got populated on reading, but this doesn't appear to be the case. I can confirm that the alignments returned by these 10,000 queries are identical.

No, the linear index on the disk has no holes. So the new code and the old code will return the same index structure in memory.

> I've looked at the code and see `hts_read_core` contains this:
> 
> https://github.com/valeriuo/htslib/blob/lindex_rev/hts.c#L2592-L2593

That probably has no effect at all with the current index files that have the full linear index on disk, but I will change it to reflect the new filling strategy.

jkbonfield:
Rebased and merged as 1830551

PR #1288, opened 2021-05-17
[vcf] Improve method `bcf_remove_allele_set`

### LINKED ISSUE #1259: inconsistent behavior in `bcf_remove_allele_set` can lead to malformed GT fields
Consider the file:
```
##fileformat=VCFv4.2
##reference=file://some/path/human_g1k_v37.fasta
##contig=<ID=22,length=51304566>
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	SAMPLE1	SAMPLE2	SAMPLE3
22	16051457	.	A	G,C	3.42882	PASS	.	GT	0/1	0/2	0/1
```

if we call `bcf_remove_allele_set` to remove the `G` allele, we end up with: 

```
22	16051457	.	A	C	3.42882	PASS	.	GT	0/0	0/1	0/0
```

as an aside I feel like ideally it should be replaced by something like: 
```
22	16051457	.	A	C	3.42882	PASS	.	GT	./0	0/1	./0
```
but let's set that aside for now.  If we try to remove the `C` allele, the GT fields are entirely unchanged, this leaves the  `2` referring to nothing :
```
22	16051457	.	A	G	3.42882	PASS	.	GT	0/1	0/2	0/1
```
I believe this is because the `if ( i<line->n_allele )` [on line 500 of vcfutils.c](https://github.com/samtools/htslib/blob/9672589346459d675d62851d5b7b5f2e5c919076/vcfutils.c#L500) doesn't take into account that `n_allele` is updated by a call to `bcf_update_alleles_str` [on line 283](https://github.com/samtools/htslib/blob/9672589346459d675d62851d5b7b5f2e5c919076/vcfutils.c#L283) (i.e it should be `if (i<nR_ori)` instead?).  I'd make a pull request but I'm not 100% sure that's what's going on here.

Thanks!

valeriuo:
1. Fixes a bug, which caused the iteration over the alleles to stop too early, due to a hidden updated of allele number.
2. Marks removed alleles as proper 'missing'.
3. Does the lazy unpack automatically, in case the user forgot to do it beforehand.

Fixes #1259

### DIFF
--- a/vcfutils.c
+++ b/vcfutils.c
@@ -258,6 +258,8 @@ int bcf_remove_allele_set(const bcf_hdr_t *header, bcf1_t *line, const struct kb
 
     // create map of indexes from old to new ALT numbering and modify ALT
     kstring_t str = {0,0,0};
+    if (!line->d.allele)
+        bcf_unpack(line, BCF_UN_STR);
     kputs(line->d.allele[0], &str);
 
     int nrm = 0, i,j;  // i: ori alleles, j: new alleles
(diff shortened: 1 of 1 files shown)


pd3:
Thank you

PR #1295, opened 2021-06-04
[index] New methods and a linear index fix

valeriuo:
Adds new method `hts_idx_nseq`, which returns the number of covered contigs from an index structure.
~~Exports methods `hts_idx_getfn` and `hts_idx_locatefn`, which were previously internal.~~
Fixes a bug introduced by 18305510d010aab54706fe4abec4d90cd6462ecc. Older indices might encode missing entries with a 0, but the entries at the beginning of the linear index could contain 0 as a valid offset and, as such, are no longer considered missing. Thus, the linear index update from right to left stops at the first non-0 entry.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,9 @@ Noteworthy changes in release a.b
 Features and Updates
 --------------------
 
+* New method `hts_idx_nseq` returns the number of contigs covered by reads
+  from an index structure.
+
 * In case a PG header line has multiple ID tags supplied by other applications,
   the header API now selects the first one encountered as the identifying tag
   and issues a warning when detecting subsequent ID tags.
--- a/hts.c
+++ b/hts.c
@@ -2577,7 +2577,7 @@ static int idx_read_core(hts_idx_t *idx, BGZF *fp, int fmt)
             if (is_be) swap_bins(p);
         }
         if (fmt != HTS_FMT_CSI) { // load linear index
-            int j;
+            int j, k;
             uint32_t x;
             if (bgzf_read(fp, &x, 4) != 4) return -1;
             if (is_be) ed_swap_4p(&x);
@@ -2589,7 +2589,8 @@ static int idx_read_core(hts_idx_t *idx, BGZF *fp, int fmt)
             if (l->offset == NULL) return -2;
             if (bgzf_read(fp, l->offset, l->n << 3) != l->n << 3) return -1;
             if (is_be) for (j = 0; j < l->n; ++j) ed_swap_8p(&l->offset[j]);
-            for (j = l->n-1; j > 0; j--) // fill missing values; may happen given older samtools and tabix
+            for (k = j = 0; j < l->n && l->offset[j] == 0; k = ++j); // stop at the first non-zero entry
+            for (j = l->n-1; j > k; j--) // fill missing values; may happen given older samtools and tabix
                 if (l->offset[j-1] == 0) l->offset[j-1] = l->offset[j];
             update_loff(idx, i, 0);
         }
--- a/hts.c
+++ b/hts.c
@@ -2723,6 +2723,18 @@ const char **hts_idx_seqnames(const hts_idx_t *idx, int *n, hts_id2name_f getid,
     return names;
 }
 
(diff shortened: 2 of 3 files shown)


jmarshall:
`seqno` suggests to me the sequence number of a particular entry in a list or, e.g., the `IDX` field identifying BCF strings. But the `hts_idx_seqno()` function seems to be instead “the number of sequences”, so perhaps it would be better named `hts_idx_nseq()` similarly to `faidx_nseq()` — or perhaps even better `hts_idx_get_nseq()` similarly to the existing `hts_idx_get_stat()` and `hts_idx_get_n_no_coor()`.

What is this function for? It looks like it could be used to set the bound for a `hts_idx_get_stat()` loop, but then it should just be `return idx->n`. The proposed implementation appears to be “return the number of chromosomes in the index that actually have reads on them“ or similar?

What is the reason for making the `hts_idx_getfn()` and `hts_idx_locatefn()` functions public? Could the desired functionality instead be integrated into the existing public `hts_idx_load()` family? e.g. perhaps an `hts_idx_load4()` that also has a `fnidxout` parameter?

valeriuo:
> so perhaps it would be better named hts_idx_nseq() similarly to faidx_nseq() — or perhaps even better hts_idx_get_nseq()

Point taken on the name.

> The proposed implementation appears to be “return the number of chromosomes in the index that actually have reads on them“ or similar?

Yes. It is basically `hts_idx_seqnames`, without the need for a header.

> What is the reason for making the hts_idx_getfn() and hts_idx_locatefn() functions public? Could the desired functionality instead be integrated into the existing public hts_idx_load() family? e.g. perhaps an hts_idx_load4() that also has a fnidxout parameter?

A use case came up (which will be visible in the next bcftools PR), in which there was a need to know the index file name before trying to open it and read it. Any reasons not to expose these methods?

jmarshall:
Perhaps this will all become clearer with the upcoming bcftools PR.

Reasons not to expose these internal methods in their existing form would be: that in revisiting them for addition to the supported API, a better design suggests itself; or that their functionality can be adequately supplied by extending existing public functions; or that their addition makes the profusion of `hts_idx_loadN()` functions even more confusing; or that they are not really necessary after all (e.g., if the new bcftools code always opens the index after finding out its filename, then it could be refactored to not need the preflight).

If a function with this behaviour is indeed needed in the public API, I'd suggest a single function with a clearer name that re-uses the `HTS_IDX_SAVE_REMOTE`/etc flags, and that documents whether the caller needs to free the returned filename etc.

jkbonfield:
> Reasons not to expose these internal methods in their existing form would be: that in revisiting them for addition to the supported API, a better design suggests itself; or that their functionality can be adequately supplied by extending existing public functions; or that their addition makes the profusion of `hts_idx_loadN()` functions even more confusing; or that they are not really necessary after all (e.g., if the new bcftools code always opens the index after finding out its filename, then it could be refactored to not need the preflight).

The problem perhaps needs fixes elsewhere.  The index handling for vcf/bcf is, frankly, a dog's breakfast!

Eg `tbx_index_load3` can load vcf.gz.tbi and vcf.gz.csi, but not bcf.csi. I think this is because the VCF csi and BCF csi are actually differnt formats (IIRC the contents of the undocumented binary blob in the header differs).  Conversely `hts_index_load3` can load all .csi up, but only *some* .tbi.

The naive solution here is the call both and use whatever works, however this generates warning messages.  I believe the intention here is simply to expose the code that finds the default matching index and returns it so the calling code can then call the correct load function.  On the surface that seems emminently sensible, but possibly the right strategy would be to fix the index load functions to do that themselves.  I'm not sufficiently familier though with them to know if their returned data is sufficiently different to be a non-starter.

What it probably needs is a unified in-memory index object that can load from any of tbi, csi or bai.  But again I don't know the internals enough to see whether that's even possible.  I've normally ran away screaming after going near any of that code. :-)

Another alternative solution is to make them silent, so code can still do the "try it and see if it works" strategy without spewing inappropriate warnings out to the user.

See https://github.com/samtools/bcftools/pull/1468 (old) and https://github.com/samtools/bcftools/pull/1505 (new) for context.

PR #1298, opened 2021-06-11
Fix the filter expression "cigar" on unmapped reads.

jkbonfield:
Our error checking for kputw/kputc failing also wrongly claimed that the empty string was also a failure.

I considered returning "*" for empty, but it's not really true as in SAM formally that is unknown.  (Although genuine unknown would be considered to be empty, so maybe it's the same thing.)

To check for not-something, use e.g.

    samtools view -e 'cigar!="" && cigar!~"D"'

Fixes samtools/samtools#1445

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1153,12 +1153,16 @@ static int bam_sym_lookup(void *data, char *str, char **end,
             ks_clear(&res->s);
             uint32_t *cigar = bam_get_cigar(b);
             int i, n = b->core.n_cigar, r = 0;
-            for (i = 0; i < n; i++) {
-                r |= kputw (bam_cigar_oplen(cigar[i]), &res->s);
-                r |= kputc_(bam_cigar_opchr(cigar[i]), &res->s);
+            if (n) {
+                for (i = 0; i < n; i++) {
+                    r |= kputw (bam_cigar_oplen(cigar[i]), &res->s) < 0;
+                    r |= kputc_(bam_cigar_opchr(cigar[i]), &res->s) < 0;
+                }
+                r |= kputs("", &res->s) < 0;
+            } else {
+                r |= kputs("*", &res->s) < 0;
             }
-            kputs("", &res->s);
-            return r ? 0 : -1;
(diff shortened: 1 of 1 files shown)


daviesrob:
This is working around the actual problem, which is that (thanks to kstring's rather inconsistent error return values) `r` will currently be non-zero on both success and failure to add to the string in [`kputc_()`](https://github.com/samtools/htslib/blob/723e261e/htslib/kstring.h#L253).  The original code then treated `r == 0` as the error case, which isn't right as it can only happen for an empty CIGAR.

For the `kputc_()` (and the `kputs()` where a check is currently missing) it should do:
```
r |= kputc_(bam_cigar_opchr(cigar[i]), &res->s) < 0;
```

so that `r` remains zero on success (the test in the `return` statement will of course need to be inverted).  Then the no-CIGAR state will be OK with no additional checks.

jmarshall:
In SAM, there is no such thing as an empty CIGAR string (see the regexp in the table at the start of §1.4).

Internally in htslib and in the BAM format (the spec is vague about this; it probably deserves at least a footnote) this is represented with `n_cigar = 0`.

To the extent that this filter expression returns the textual representation of the CIGAR field, I think it should return "`*`" when `n_cigar == 0`, as that is the equivalent textual representation.

(For expressions matching against CIGAR operators, happily it makes no difference of course.)

jkbonfield:
I did debate using `*` as mentioned in the commit message.  I thought that maybe the empty string is cleaner, but I take your point that the other string form matches SAM so maybe it should be `*`.  We're not printing this up anywhere - it's just a string for regexp matching (and in that regard the printable form perhaps isn't ideal, but it's easy to explain at least).  What do others think?

jmarshall:
People may also write `-e 'cigar != "*"'` and will need to know whether to use `""` or `"*"` (IMHO the latter is more familiar).

jkbonfield:
Persuaded. :-)

Both your comments have been addressed.

PR #1300, opened 2021-06-16
Fix typos in INSTALL [minor]

jmarshall:


jmarshall:
New merging speed record? :smile: (BTW the :x: is because the PR changed state while appveyor was still running, so no big deal.)

jkbonfield:
I considered it not worth waiting for the checks to finish given the nature of the changes!  :-)

PR #1301, opened 2021-06-17
On MinGW-i686, testing for recv() requires the system header

jmarshall:
MinGW's 32-bit version of `<winsock2.h>` declares `recv()` as `__stdcall__`. When `AC_SEARCH_LIBS` does a trial link without declaring `recv()` at all, linking fails due to the mismatch of calling conventions — as seen in msys2/MINGW-packages#8207.

To fix this the non-intrusive way: Before reporting the error with `MSG_ERROR`, try again to link against ws2_32 this time including the system header to get the annotated declaration. This produces the right `$LIBS` on MinGW-i686, and leaves the `MSG_ERROR` failure unchanged on other platforms.

I considered adding an `HTS_SEARCH_LIBS` function that would be like `AC_SEARCH_LIBS` but take `AC_LANG_PROGRAM` output instead of just a function name, but that would be a more intrusive change.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -234,7 +234,16 @@ AC_SEARCH_LIBS([recv], [socket ws2_32], [
 if test "$ac_cv_search_recv" != "none required"
 then
   static_LIBS="$static_LIBS $ac_cv_search_recv"
-fi], [MSG_ERROR([unable to find the recv() function])])
+fi],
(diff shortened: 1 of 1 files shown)


valeriuo:
Merged as ba109a49f37e89337d05ed466df7eaa444df9b43

PR #1302, opened 2021-06-17
Make version.sh work from a separate build directory

jmarshall:
Fix a bug in the separate build directory handling: `version.sh` reported the master release version rather than a git commit description when run from a separate build directory.


daviesrob:
Thanks, the fix looks good.

Issue #1290, opened 2021-05-22
tabix failed to return the contents


Shicheng-Guo:
I have a tab index file generated by `tabix`: `icd10-B02-both_sexes.tsv.bgz` and `icd10-B02-both_sexes.tsv.bgz.tbi`
How tabix don't return me what I should received. 

the 1st issue:
`tabix icd10-B02-both_sexes.tsv.bgz 1:1-323668000` will return everything well
However, 
`tabix icd10-B02-both_sexes.tsv.bgz 1:1-3236680000` will return nothing. 

```
(base) [sguo2@login01 broad]$ zcat icd10-B02-both_sexes.tsv.bgz | awk 'NR==21234569'
13      31892114        G       GA      2.388e-01       2.269e-01       6.583e-02       6.671e-02       3.238e-01       false
(base) [sguo2@login01 broad]$ tabix icd10-B02-both_sexes.tsv.bgz 13:31892113-31892115
(base) [sguo2@login01 broad]$
```
with above command, you can find there is a record in line 21234569, however, tabix return nothing. 
Do you have any suggestion? 
Thanks. 
Shicheng

```
Program: tabix (TAB-delimited file InderXer)
Version: 0.2.5 (r1005)

Usage:   tabix <in.tab.bgz> [region1 [region2 [...]]]

Options: -p STR     preset: gff, bed, sam, vcf, psltbl [gff]
         -s INT     sequence name column [1]
         -b INT     start column [4]
         -e INT     end column; can be identical to '-b' [5]
         -S INT     skip first INT lines [0]
         -c CHAR    symbol for comment/meta lines [#]
         -r FILE    replace the header with the content of FILE [null]
         -B         region1 is a BED file (entire file will be read)
         -0         zero-based coordinate
         -h         print also the header lines
         -H         print only the header lines
         -l         list chromosome names
         -f         force to overwrite the index

```

```
(base) [sguo2@login01 broad]$ tabix icd10-B02-both_sexes.tsv.bgz 1:1-323668000
1       11063   T       G       3.026e-05       4.800e-05       -4.104e+00      1.288e+01       7.499e-01       true
1       13259   G       A       1.210e-05       2.784e-04       -1.131e+00      1.756e+00       5.195e-01       true
1       17641   G       A       6.052e-05       8.322e-04       -1.098e+00      1.072e+00       3.058e-01       true
1       30741   C       A       NA      NA      NA      NA      NA      NA
[...]

PR #1303, opened 2021-06-21
Summer 2021 NEWS update

whitwham:
Please take a look for errors, omissions or entirely fictional updates.

This will be kept updated until we are ready for a release.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -38,9 +38,8 @@ Features and Updates
   
 * Initialise the missing entries of the linear index in reverse order.
   Populates empty linear index entries from the right rather from the left.
-  Two functions have been renamed, dump_index is now idx_dump and
-  hts_idx_save_core is idx_save_core.  The new hts_bin_level function computes
-  the index level of a bin.  Thanks to @carsonh for reporting the issue.
+  The new hts_bin_level function computes the index level of a bin.
+  Thanks to @carsonh for reporting the issue.
   (#1286; fixes #486)
   
 * Related to the above, the new method, hts_idx_nseq, returns the total number
--- a/NEWS
+++ b/NEWS
@@ -63,6 +63,9 @@ These are compiler, configuration and makefile based changes.
 * Support building HTSlib in a separate directory. Thanks to John Marshall.
   (#1277; fixes #231)
 
+* Support building HTSlib on MinGW 32-bit environments. Thanks to John Marshall.
+  (#1301)
+
 Bug fixes
 ---------
 
--- a/NEWS
+++ b/NEWS
@@ -13,18 +13,21 @@ Features and Updates
   remote index file by default.
   (#1266; fixes #380)
 
-* Transparently treat FASTQ as unmapped data as if it was SAM, BAM or CRAM.
+* Support reading and writing FASTQ format in the same way as SAM, BAM or CRAM.
+  Records read from a FASTQ file will be treated as unmapped data.
   (#1156)
 
 * Added GCP requester pays bucket access.  Thanks to @indraniel.
   (#1255)
 
-* Made mpileup's overlap removal choose a random sequence.
+* Made mpileup's overlap removal choose which copy to remove at random instead
+  of always removing the second one.  This avoids strand bias in experiments
+  where the +ve and -ve strand reads always appear in the same order.
   (#1273; fixes samtools/bcftools#1459)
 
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -4,12 +4,100 @@ Noteworthy changes in release a.b
...
+* Added bracket handling to bcf_hdr_parse_line.  Thanks to Alberto Casas Orrtiz.
+
+Build changes
+-------------
+
+These are compiler, configuration and makefile based changes.
+
+* Added a curl/curl.h check to configure and improved INSTALL documentation on
+  build options.  Thanks to John Marshall.
+  (#1265)
```
jmarshall, line 58: I think the reporter should also be credited for this one:
```suggestion
  build options.  Thanks to Melanie Kirsche and John Marshall.
  (#1265; fixes #1261)
```

NEWS
```diff
@@ -4,12 +4,100 @@ Noteworthy changes in release a.b
...
+* Fix to make samtools cat work on CRAMs again.
+  (#1276; fixes samtools/samtools#1420)
+  
+* Fix for a double memory free in SAM header creation.  Thanks to @ihsineme.
+  (#1274)
+  
+* Prevent assert in bcf_sr_set_regions.  Thanks to Dr K D Murray.
+  (#1270)
+  
+* Fixed the update of the minimum offset when unmapped reads are present.  This
```
jmarshall, line 90: This is a serious bug that should be more clearly described, placed at the top of the list, and probably noted in samtools's _NEWS_ as well. Perhaps something like

> Fixed `hts_itr_query()` et al region queries: fixed bug introduced in HTSlib 1.12, which lead to iterators producing very few reads for some queries (especially for larger target regions) when unmapped reads were present. (HTSlib 1.11 had a related problem in which iterators would omit a few unmapped reads that should have been produced; cf #1142.)

This has probably been reported multiple times: #1279 as listed (first and best researched, so the credit for Daniel is certainly deserved), but also samtools/samtools#1451, possibly #1290, and probably samtools/samtools#1457.

Over in samtools, this could be described in terms of region queries instead:

> Fixed `samtools view FILE REGION` and `mpileup -r REGION` etc region queries: fixed bug introduced in 1.12, which lead to region queries producing very few reads for […]

valeriuo: Changed. Thanks!

NEWS
```diff
@@ -4,12 +4,104 @@ Noteworthy changes in release a.b
...
+  Populates empty linear index entries from the right rather from the left.
+  The new hts_bin_level function computes the index level of a bin.
+  Thanks to @carsonh for reporting the issue.
+  (#1286; fixes #486)
+
+* Related to the above, the new method, hts_idx_nseq, now returns the total
+  number of contigs from an index.
+  (#1295 and #1299)
+
+* Added bracket handling to bcf_hdr_parse_line.  Thanks to Alberto Casas Orrtiz.
```
jmarshall, line 49:
```suggestion
* Added bracket handling to bcf_hdr_parse_line.  Thanks to Alberto Casas Ortiz.
```

NEWS
jmarshall: Thanks for applying. It'd probably be worth listing the additional bug report numbers, and I'm afraid it should be `s/lead/led/` — I'm pretty good at mixing that one up :smile:

NEWS
whitwham: Good catch.  I'm too heavy on the 'R' key.

NEWS
valeriuo: Perhaps the other bug reports should be mentioned in the SAMtools news, as they are related to running samtools commands.

NEWS
valeriuo: I will update my latest commit to fix this.

NEWS
jmarshall: I guess where HTSlib bugs noticeably affect particular samtools outputs, it can be worth mentioning them in the samtools context. Alas, who knows whether anyone actually reads the release notes for either samtools or htslib… 🤷

NEWS
jmarshall: (More for the trivial commit: there were a few trailing spaces near the Hulselmans entry too.)

daviesrob:
I've added a few adjustments, to better describe the outcome of some of the changes listed from an end-user perspective.

PR #1306, opened 2021-07-07
Update htscodecs module to release 1.1.1

daviesrob:
Fixes a bug with name tokeniser and variable number of names per slice.

Adds NEWS items.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -63,6 +63,8 @@ Build changes
 
 These are compiler, configuration and makefile based changes.
 
+* HTSlib now uses libhtscodecs release 1.1.1.
+
 * Added a curl/curl.h check to configure and improved INSTALL documentation on
   build options.  Thanks to Melanie Kirsche and John Marshall.
   (#1265; fixes #1261)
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -63,6 +63,8 @@ Build changes
 
 These are compiler, configuration and makefile based changes.
 
+* HTSlib now uses libhtscodecs release 1.1.1
```
jmarshall, line 66: Fullstop? (or was that intentional as there's so many dots in the last word already :smile:)

NEWS
```diff
@@ -112,6 +114,14 @@ Bug fixes
...
   (#1298, fixes samtools/samtools#1445)
 
+* Bug fixes in the bundled copy of htscodecs:
+
+  - Fixed an uninitialized access in the name tokenizer decoder.
+    (samtools/htscodecs#23)
+
+  - Fixed a bug with name tokeniser and variable number of names per slice,
+    causing it to incorrectly report an error on certain valid inputs.
+    (samtools/htscodecs#24)
```
jmarshall, line 125: Usually two blank lines before the previous release.

NEWS
daviesrob: Kind-of due to the number of dots, but I guess I may as well go for it.

NEWS
daviesrob: Adjusted...

jmarshall: Sorry for the picky comments :)  As a penance I have checked out the PR to run a spelling checker on it, and noticed it's got `tokeniser` spelt once each way (s vs z)…

jkbonfield: All part of our drive for greater inclusivity ;-)

daviesrob: Oops, yes.  It should use `tokeniser` I guess, as that's the spelling in the library.

PR #1307, opened 2021-07-07
Add "endpos" filter expression

jmarshall:
For your amusement at this time, here is a PR that implements `endpos` as a filter expression, in response to <https://twitter.com/gringene_bio/status/1412686733225857024>.

See also the companion documentation PR samtools/samtools#1464.

"endpos" is the (1-based inclusive) position of the rightmost mapped base of the read, as measured using the CIGAR string. (`bam_endpos()` returns 0-based exclusive, which is the same thing.) For unmapped reads, it is the same as "pos".

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1166,6 +1166,14 @@ static int bam_sym_lookup(void *data, char *str, char **end,
         }
         break;
 
+    case 'e':
+        if (memcmp(str, "endpos", 6) == 0) {
+            *end = str+6;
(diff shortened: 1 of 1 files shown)


jkbonfield:
I think the ongoing discussion around endpos vs posend is a good one too.  I'm ambivalent really.

Anyway it's not something we should spend time on for this release, but I'm fully up for adding this feature.  A bit of a no-brainer really.  Thanks.

PR #1310, opened 2021-07-23
Pkg config fix

jkbonfield:
This has two comits - please see the commit messages for the gory details.

The first commit is a trivial fix to make `pkg-config --static --libs` work, but only with autoconf builds.  It adds `-lcurl` and removes a duplicate `-lhts` which trips up libtool.

The second commit takes a different direction and completely removes the `private_LIBS` concept added in 57c71055.  (I strongly suspect that commit never actually worked with static libs anyway.)  This fixes non-autoconf builds too, as well as considerably  simplifying the process.  It does however mean we get some duplication in the library list, but it appears harmless (as far as the test systems are concerned).

We need to decide whether we wish to continue supporting non-autoconf setups, and if so whether we care about pkg-config in such environments.  If not then maybe the first commit alone is sufficient and ditch the second.  Otherwise I'd suggest squashing both together unless we can find some overwhelming reason why we need to keep the concept of private_LIBS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,7 +256,7 @@ config.h:
 # on htslib.pc.in listed, as if that file is newer the usual way to regenerate
 # this target is via configure or config.status rather than this rule.
 htslib.pc.tmp:
-	sed -e '/^static_libs=/s/@static_LIBS@/$(htslib_default_libs)/;s#@[^-][^@]*@##g' $(srcprefix)htslib.pc.in > $@
+	sed -e '/^static_libs=/s/@static_LIBS@/$(htslib_default_libs)/;s/@private_LIBS@/$(htslib_default_libs)/;s#@[^-][^@]*@##g' $(srcprefix)htslib.pc.in > $@
 
 # Create a makefile fragment listing the libraries and LDFLAGS needed for
 # static linking.  This can be included by projects that want to build
--- a/configure.ac
+++ b/configure.ac
@@ -388,6 +388,7 @@ dnl -lcurl is only needed for static linking if hfile_libcurl is not a plugin
   if test "$libcurl" = enabled ; then
     if test "$enable_plugins" != yes ; then
       static_LIBS="$static_LIBS -lcurl"
+      private_LIBS="$private_LIBS -lcurl"
     fi
   fi
 fi
--- a/htslib.pc.in
+++ b/htslib.pc.in
@@ -11,5 +11,5 @@ Description: C library for high-throughput sequencing data formats
 Version: @-PACKAGE_VERSION@
 Cflags: -I${includedir}
 Libs: -L${libdir} -lhts
-Libs.private: -L${libdir} @private_LIBS@ -lhts -lm -lpthread
+Libs.private: -L${libdir} @private_LIBS@ -lm -lpthread
 Requires.private: zlib @pc_requires@


jmarshall:
The extra `-lhts` is because when you ask for `--static --libs` it prints out `concat(Libs, Libs.private)` so the first one is coming from `Libs`. I wonder whether this is something that has changed over the years with pkg-config… the first bug fix in [0.28](https://gitlab.freedesktop.org/pkg-config/pkg-config/-/blob/master/NEWS#L78) seems relevant and _htslib.pc.in_ might have been written with 0.27, which may have hidden these duplicates (not particularly correctly!).

Probably it was just an oversight not to substitute `@private_LIBS@` in the _Makefile_ `htslib.pc.tmp` rule.

jkbonfield:
> The extra `-lhts` is because when you ask for `--static --libs` it prints out `concat(Libs, Libs.private)` so the first one is coming from `Libs`. I wonder whether this is something that has changed over the years with pkg-config… the first bug fix in [0.28](https://gitlab.freedesktop.org/pkg-config/pkg-config/-/blob/master/NEWS#L78) seems relevant and _htslib.pc.in_ might have been written with 0.27, which may have hidden these duplicates (not particularly correctly!).

That was my assumption on where the duplicate came from, but I hadn't thought about us adding an explicit 2nd copy as a workaround for a pkg-config bug.  If we conclude we need a 2nd copy to cope with an old pkg-config, then I guess we could put it right at the start so it doesn't matter which one libtool nukes.

> Probably it was just an oversight not to substitute `@private_LIBS@` in the _Makefile_ `htslib.pc.tmp` rule.

I'm sure we could also fix the Makefile too, but I still rather dislike having to maintain two rather different sets of build systems.  I still don't really know what audience we're catering to here that refuse to or are unable to run configure.  If it's pandering to people that love to cargo-cult cut-and-paste the entire source tree into their package, then pfft!  I'd rather make life harder! :P

daviesrob:
I agree, `-lhts` shouldn't be in `Libs.private`.  I think for curl we should add `libcurl` to `Requires.private` (via `pc_requires` in `configure.ac`) instead of adding it to `Libs.private`.  The difference is that `Requires.private` allows `pkg-config` to add all of libcurl's dependencies from its own `libcurl.pc` file, which is needed for a fully-static build.  As libcurl [comes with a pkg-config configuration](https://github.com/curl/curl/blob/master/libcurl.pc.in) we can assume that `libcurl.pc` will always be present for a libcurl installation.

jkbonfield:
>  I think for curl we should add `libcurl` to `Requires.private` (via `pc_requires` in `configure.ac`) instead of adding it to `Libs.private`. The difference is that `Requires.private` allows `pkg-config` to add all of libcurl's dependencies from its own `libcurl.pc` file, which is needed for a fully-static build. As libcurl [comes with a pkg-config configuration](https://github.com/curl/curl/blob/master/libcurl.pc.in) we can assume that `libcurl.pc` will always be present for a libcurl installation.

Nice idea, but it fell flat on the very first system I tried!  (Ubuntu 18.04.5 LTS).

If we're building statically against htslib, then we get no ability to say "static htslib, but dynamic curl".  Thus we also get static curl.  On my system that expands to:

```
$ pkg-config libcurl --static --libs
-L/usr/lib/x86_64-linux-gnu/mit-krb5 -lcurl -lnghttp2 -lidn2 -lrtmp -lpsl -lssl -lcrypto -lssl -lcrypto -Wl,-Bsymbolic-functions -Wl,-z,relro -lgssapi_krb5 -lkrb5 -lk5crypto -lcom_err -llber -lldap -llber -lz
```

All well and good, but half of these packages haven't been installed.

```
$ echo 'int main(void) {return 0;}' > _.c
$ cc _.c `pkg-config libcurl --static --libs`
/usr/bin/ld: cannot find -lnghttp2
/usr/bin/ld: cannot find -lrtmp
/usr/bin/ld: cannot find -lpsl
/usr/bin/ld: cannot find -lgssapi_krb5
/usr/bin/ld: cannot find -lkrb5
/usr/bin/ld: cannot find -lk5crypto
/usr/bin/ld: cannot find -lcom_err
collect2: error: ld returned 1 exit status
```

Frankly, I go back to the statement I mentioned on Bambi - pkg-config is just a disaster area!
Therefore I'd prefer to keep -lcurl in our htslib.pc as it's simply more robust than attempting static curl and all the nightmares that brings.

daviesrob:
Your test program was a bit too simple.  As it didn't try to use any `libcurl` symbols, the linker doesn't try to import anything.  You also didn't use `-static` so it will have attempted a dynamic link.

A better test program is:
```sh
$ printf '#include <curl/curl.h>\nint main(void) { curl_global_init(0); curl_global_cleanup(); return 0; }\n' > test.c
```
It's also possible grab everything (in theory) needed to build a static version using:
```sh
$ sudo apt build-dep libcurl4-openssl-dev
$ sudo apt install libcurl4-openssl-dev
```

Now building a dynamically-linked executable works as expected:
```sh
ubuntu@rmd-tmp:~$ cc -o test test.c -lcurl
ubuntu@rmd-tmp:~$ file test
test: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=f712088d96559c87760a2a1fdc208f7ce14096e2, not stripped
```

Also with `pkg-config`:
```sh
$ cc -o test test.c `pkg-config --cflags --libs libcurl`
$ file test
test: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=f712088d96559c87760a2a1fdc208f7ce14096e2, not stripped
```

Adding `--static` to `pkg-config` also works, but builds a dynamic executable again:
```sh
$ cc -o test test.c `pkg-config --static --cflags --libs libcurl`
$ file test
test: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=5bf9fd8522da61fc9428f0491c4f8530eda652c0, not stripped
```

Building with `-static` but without listing the dependencies fails spectacularly:
```sh
$ cc -static -o test test.c -lcurl
/usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/libcurl.a(libcurl_la-netrc.o): In function `Curl_parsenetrc':
(.text+0x3a8): warning: Using 'getpwuid_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/libcurl.a(libcurl_la-curl_addrinfo.o): In function `Curl_getaddrinfo_ex':
(.text+0x83): warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/libcurl.a(libcurl_la-url.o): In function `Curl_disconnect':
(.text+0xb4c): undefined reference to `idn2_free'
(.text+0xb68): undefined reference to `idn2_free'
(.text+0xb84): undefined reference to `idn2_free'
(.text+0xba0): undefined reference to `idn2_free'
```
... etc, etc.  The entire error message is 504 lines long!

So, will `pkg-config --static` rescue us?  Well, regrettably no:

```sh
$ cc -static -o test test.c `pkg-config --static --cflags --libs libcurl`
/usr/bin/ld: cannot find -lgssapi_krb5
/usr/bin/ld: cannot find -lkrb5
/usr/bin/ld: cannot find -lk5crypto
 ... lots of warnings about deprecated functions and glibc runtime compatibility skipped ...
collect2: error: ld returned 1 exit status
```

So where do we get the static versions of those kerberos libraries from?  After some searching, I found [this bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=439039) which suggests that they haven't been available for quite some time and it's therefore not possible to build a static binary with debian / ubuntu `libcurl.a`. It would probably work if you built your own `libcurl` without Kerberos support.

So, anyone building programs using `libhts` with `pkg-config` would be best advised to either use dynamic linking, or turn on external plug-ins so that all of the `libcurl` dependencies are pushed into a separate module (which is of course itself dynamically loaded).

daviesrob:
The trouble is `pkg-config --static` doesn't work that way.  It seems to be all or nothing, there's no way to say you want to link one library statically and another one dynamically.  If fact, all the `--static` option does is add link options for the dependencies.  It also used `-lname` for everything, so if you have both `.so` and `.a` files around the linker will pick up the shared ones unless you use the `-static` compiler option to force it to use the archive files instead.  If you want to use archive files for only some of the libraries `pkg-config` would need to specify their `.a` files directly, which is not something that it appears to do.

jkbonfield:
I think you're confusing my intentions here.

I can see three situations.

1) Dynamic linking.  That's trivial and works already.

2) Linking against a static htslib and dynamic everything else.  This is how Samtools builds right now if it has htslib along side of within it.  It's suitable for building e.g. a samtools binary that you wish to copy to other *related* systems.  It's not truely universal run-anywhere, but is generally going to work in most environments.

3) A totally static binary where there are zero dynamic libraries.  This is for people who wish to build a universal run-anywhere binary.  However it's complex and tricky to do due to OS interfaces potentially differing.  If this is your aim then you're best off using one of the dedicated frameworks for making such binaries to avoid the various pitfalls.  Eg https://justine.lol/cosmopolitan


So what have we learn above?  Well, 1) works and did already before this PR.  3) doesn't work on default OS installs without having to do lots of extra package management, hence I discounted it.  2) is where I'm pitching this PR at.

Your conclusion appears to be 3 doesn't work, so we should do 1 only (_"So, anyone building programs using libhts with pkg-config would be best advised to either use dynamic linking, or turn on external plug-ins so that all of the libcurl dependencies are pushed into a separate module (which is of course itself dynamically loaded"_).   If that's the case, we should just delete all the static stuff from htslib.pc as leaving a non-working setup is just a trap.

However I still think option 2 is perfectly justified, especially as we do this ourselves already!  Why is it acceptable for Samtools to build this way but not other applications?

daviesrob:
I understand what you want to do.  Unfortunately `pkg-config` does not really support option 2.  It only does 1 and 3.

jkbonfield:
Well, it does support it if you merge this PR!  I've tested it with bambi by installing htslib, removing the .so files, and changing bambi's htslib config to use `LIBS="$(pkg-config htslib --static --libs)"` and it works.

It may not be what pkg-config intended as usage, but pragmatically it gives us something we can't do right now, plus doing nothing is unacceptable as currently it just plain doesn't work at all!  Can we please try and at least move forward with this as a bug fix, even if it's not perfection?

daviesrob:
Try it with both `.so` and `.a` files installed, which is usual for an HTSlib installation.  I expect you'll end up linking to `libhts.so` which is not what you want to happen.

Fundamentally `pkg-config` does not support this use-case, which one of the reasons why SAMtools and BCFtools don't use it in the way you suggest.

jkbonfield:
Fair enough, closing this then as it appears to be impossible to fix.

PR #1317, opened 2021-08-09
Make bam_tag2cigar's CG tag handling more lenient.

jkbonfield:
The SAM specification explicitly states the type for this tag is
CG:B:I.  However Htsjdk (sometimes or always?) writes these tags as
CG:B:i.  Htslib then silently ignores the tag, assuming it's some
unofficial (and incorrect) abuse of an uppercase tag for some local
purpose.

Given that there is data published in the wild using the incorrect
data type, it would be less problematic for us to simply handle the
incorrect value sign than the minimal risk of misinterpretting
someone's private tag data as CIGAR. (Plus they'd bring such woe onto
themselves by using the official name-space.)

Fixes samtools/samtools#1477

See also https://github.com/samtools/htsjdk/issues/1560

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -654,7 +654,8 @@ static int bam_tag2cigar(bam1_t *b, int recal_bin, int give_warning) // return 0
         errno = saved_errno; // restore errno on expected no-CG-tag case
         return 0;
     }
-    if (CG[0] != 'B' || CG[1] != 'I') return 0; // not of type B,I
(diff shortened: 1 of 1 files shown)


daviesrob:
This seems safe enough.  I doubt `CG:B:i` has been used for any other purpose, and as you say it would be unwise to do so after `CG` was added to [SAMtags](https://samtools.github.io/hts-specs/SAMtags.pdf).

PR #1320, opened 2021-08-10
Minor tidy ups for methylation test programs

jmarshall:
Add omitted `#include`, and ignore new built test programs.


daviesrob:
Thanks for the update.  I'll look into adding something to the test harness to spot untracked files - it should be easy enough to do there.

PR #1324, opened 2021-08-24
Catch untracked files (and make version.sh work on shallow checkouts)

daviesrob:
This PR contains two commits.  The first fixes `version.sh` so that it works on `--depth=1` clones of the repository.  This both gets rid of the annoying "No names found, cannot describe anything." messages that were printed because `git describe` couldn't find a tag; and it also makes the `version.sh` report a correct-ish version string on such clones.

The second commit adds CI test support to catch any untracked files that make be present after build and test.  Finding such files usually means they need to be added to `.gitignore`.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -48,7 +48,7 @@ test_template: &TEST
   test_script: |
     make test-shlib-exports
     make test
-
+    if test "x$DO_UNTRACKED_FILE_CHECK" = "xyes"; then make check-untracked ; fi
 
 #--------------------------------------------------
 # Task: linux builds.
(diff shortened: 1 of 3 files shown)


daviesrob:
An example of an untracked file being caught can be seen in https://github.com/daviesrob/htslib/runs/3403088881

jmarshall:
Perhaps add something to the untracked message along the lines of “Please either clean up, or add to .gitignore, or for test output files consider naming them to match `*.tmp[.*]`“.

daviesrob:
The updated text is a bit long, but it's a useful hint so I've added it, thanks.

PR #1325, opened 2021-08-26
Add FASTQ_OPT_NCBI option for parsing of NCBI's SRA data.

jkbonfield:
This variant of FASTQ has the read name as the second field on the name line, just to be awkward.


sam.c
```diff
@@ -3710,11 +3715,25 @@ static int fastq_parse1(htsFile *fp, bam1_t *b) {
...
+    char *name = x->name.s+1;
+    if (x->ncbi_names) {
+        char *cp = strchr(x->name.s, ' ');
+        if (cp) {
+            *cp = '@';
+            i = cp - x->name.s;
+            name = cp+1;
+        }
+    }
+
```
daviesrob, line 3731: A minor point, but it might be good to move this after the `x->nprefix` test below.  As it is, it will "fix" files with lines that don't have the correct prefix, but do start with a leading space, which may or may not be a good thing...

Also, how general do we want this.  Should we be looking for multiple spaces or tabs?

daviesrob (commented):
I'm not sure if we should be specifically calling out the NCBI here.  The SRA toolkit [`fastq-dump`](https://trace.ncbi.nlm.nih.gov/Traces/sra/sra.cgi?view=toolkit_doc&f=fastq-dump) program has a `--origfmt` option that outputs the original read names if they've been stored, so it is in theory possible to work around the problem using NCBI tools.  In practice it seems that their archive doesn't store the names anyway, so neither `fastq-dump` nor this PR can get them back.

The ENA on the other hand does store the read names, at least for data directly submitted to them, so this option is more useful for their data.  Compare for example ERR4204010 from the [NCBI](https://trace.ncbi.nlm.nih.gov/Traces/sra/?run=ERR4204010) and [ENA](https://www.ebi.ac.uk/ena/browser/view/ERR4204010?show=reads).

Although I'm not entirely sure what we could call the option instead.  Maybe `FASTQ_OPT_SECOND_NAME` or `FASTQ_OPT_NAME2`?

sam.c
jkbonfield: I could rename it FASTQ_OPT_SRA.  I don't really like second name as it's long winded and like it or not this *is* an SRA specific format.  ENA only use it because SRA use it.

jkbonfield: Actually I'll take that back, I see what you mean.  Maybe ERA fixed SRA format by still included the data.  I'm not really sure.  SRA came first though so I mentally peg this as their error.  I'll ponder.

jkbonfield:
The code has been revised.

- Ordering of code adjusted as suggested.  It now also recognises tab and any number of space/tab.  (Although the specific erroneous SRA format-within-format uses 1 space only).

- Renamed the user-visible option to "fastq_name2" and enum to FASTQ_OPT_NAME2.

- Renamed the internal variable to "sra_names".  They have to carry the blame somewhere :-)

daviesrob:
Looks OK.  I've squashed & rebased, plus added a couple of simple tests to exercise the option in a new commit.

PR #1326, opened 2021-08-31
Add hts_flush() API function

jmarshall:
The various underlying file types provide flush routines, but `htsFile` to date does not. 

Also add `sam_flush()` and `bcf_flush()` wrappers. There was a temptation to make this a `static inline` function, but instead I followed the existing pattern for `sam_*` and `bcf_*` wrappers of `hts_*` functions via `#define`.

This API function is needed by an upcoming samtools PR that wants to flush a `samFile`.


valeriuo (approved):

valeriuo:
A NEWS entry about these interfaces would be appreciated.

PR #1328, opened 2021-09-10
Warn if spaces instead of tabs are detected to prevent confusion

pd3:
… as in https://github.com/samtools/bcftools/issues/1575

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -143,21 +143,33 @@ int bcf_hdr_add_sample(bcf_hdr_t *h, const char *s)
     return bcf_hdr_add_sample_len(h, s, 0);
 }
 
-int HTS_RESULT_USED bcf_hdr_parse_sample_line(bcf_hdr_t *h, const char *str)
+int HTS_RESULT_USED bcf_hdr_parse_sample_line(bcf_hdr_t *hdr, const char *str)
 {
-    int ret = 0;
-    int i = 0;
-    const char *p, *q;
-    // add samples
-    for (p = q = str;; ++q) {
-        if ((unsigned char) *q > '\n') continue;
-        if (++i > 9) {
-            if ( bcf_hdr_add_sample_len(h, p, q - p) < 0 ) ret = -1;
-        }
-        if (*q == 0 || *q == '\n' || ret < 0) break;
-        p = q + 1;
+    const char *mandatory = "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO";
+    if ( strncmp(str,mandatory,strlen(mandatory)) )
+    {
+        hts_log_error("Could not parse the \"#CHROM..\" line, either the fields are incorrect or spaces are present instead of tabs:\n\t%s",str);
+        return -1;
     }
 
+    const char *beg = str + strlen(mandatory), *end;
+    if ( !*beg || *beg=='\n' ) return 0;
+    if ( strncmp(beg,"\tFORMAT\t",8) )
+    {
+        hts_log_error("Could not parse the \"#CHROM..\" line, either FORMAT is missing or spaces are present instead of tabs:\n\t%s",str);
+        return -1;
+    }
+    beg += 8;
+
+    int ret = 0;
+    while ( *beg )
+    {
+        end = beg;
+        while ( *end && *end!='\t' && *end!='\n' ) end++;
+        if ( bcf_hdr_add_sample_len(hdr, beg, end-beg) < 0 ) ret = -1;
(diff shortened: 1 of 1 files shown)


vcf.c
```diff
@@ -145,12 +145,34 @@ int bcf_hdr_add_sample(bcf_hdr_t *h, const char *s)
...
+            if ( i==0 && strncmp(p,"#CHROM",q-p) ) hts_log_error("Could not parse the sample line: %s",str);
+            if ( i==1 && strncmp(p,"POS",q-p) ) hts_log_error("Could not parse the sample line: %s",str);
+            if ( i==2 && strncmp(p,"ID",q-p) ) hts_log_error("Could not parse the sample line: %s",str);
+            if ( i==3 && strncmp(p,"REF",q-p) ) hts_log_error("Could not parse the sample line: %s",str);
+            if ( i==4 && strncmp(p,"ALT",q-p) ) hts_log_error("Could not parse the sample line: %s",str);
+            if ( i==5 && strncmp(p,"QUAL",q-p) ) hts_log_error("Could not parse the sample line: %s",str);
+            if ( i==6 && strncmp(p,"FILTER",q-p) ) hts_log_error("Could not parse the sample line: %s",str);
+            if ( i==7 && strncmp(p,"INFO",q-p) ) hts_log_error("Could not parse the sample line: %s",str);
+            if ( i==8 && strncmp(p,"FORMAT",q-p) ) hts_log_error("Could not parse the sample line: %s",str);
+            if ( !hdr_space_warned ) hts_log_warning("The sample names contain spaces, is this intended? Trying to proceed anyway..");
```
jkbonfield, line 165: I assume you added ".." so it the function can add the final ".", but it triggers a failure the test harness.  Maybe just omit the ".." to silence this.

pd3: The dots ".." are removed now

pd3:
Thanks for your HO.

HTSlib and BCFtools often have to work with malformed VCFs. The question is, should HTSlib attempt to do some error recovery and proceed anyway, or be strict. In this case my opinion is that it should try to proceed. Note that your test case ignores the fact that the current develop version accepts everything as long as `#CHROM\tPOS` is correct and does not give the user any clue what may be the problem.

I agree that the PR version should be improved so that the warning message makes a distinction between sample names and the standard VCF columns, or at least mentions both.

jmarshall:
My test case and comment were a review of the proposed PR, not of the status quo on **develop**. As you say, there is room for improvement on the status quo.

jkbonfield:
Given it's malformed data and the PR is already clearly a significant improvement on develop,  I think the minimal change of making the diagnostic warning message more encompassing is a simple fix.  (Ie option 2) Eg mentioning spaces on `#CHROM` line without pinning it down specifically to being within the samples component.  It's stil sufficient for the user to identify the problem.

FWIW I'm not against the idea of fixing the error while emitting a warning.  I don't view this as incompatible with htsjdk as what's then output is compatible with all tools once again (assuming nothing else is broken).  I'm rather ambivalent as to whether it's a good thing to fix-and-warn or hard-error though.  I guess it depends on the prevalence of historic data out there having this defect in which case fix is good, or whether it's some recent tool with buggy output in which case error is good and file a bug report.

jmarshall:
> I don't view this as incompatible with htsjdk as what's then output is compatible with all tools once again (assuming nothing else is broken).

I perhaps shouldn't have mentioned the output, as that's a red herring.

The point is that if people archive files like the example input file on the basis that they work with htslib/bcftools, they or others will later find out that they don't work with the Java tools. This would be an interoperability problem caused by htslib accepting such invalid files.

pd3:
I just modified the code to strictly conform to VCF specification + print an informative error if not, as was discussed yesterday at the meeting.

jkbonfield:
I note it doesn't accept "FORMAT\n", so with zero samples.  I'm assuming this is correct as the specification says "_If genotype data is present in the file, these are followed by a FORMAT column header, then an arbitrary number
of sample IDs_".  While "arbitrary number" doesn't exclude zero, and it says "If" and not "If and only if", I think it's pretty clear that's the intent.

pd3:
I had given it a thought while reviewing the code and on several occasions before. Practically speaking, FORMAT without samples is useless. Formally, the VCF specification says "If genotype data is present in the file, these are followed...". This implies that when there are no genotype data (zero samples), nothing follows the INFO field. Suggesting otherwise is nitpicking.

jkbonfield:
I just noted it incase it was something to consider as the spec doesn't explicitly forbid this case, but I'm glad it's already been considered and we agree too on it not being a problem.

PR #1331, opened 2021-09-15
[bgzip]: Add a keep option to not delete input files

pettyalex:
This is a small convenience feature that I've wished was present: Add a keep option to preserve input files to bgzip. Flag is -k or --keep, just like other compression tools. Default behavior is entirely unchanged.

I didn't see any tests for the bgzip tool, let me know if I missed them and I'll add tests for this.


whitwham:
Looks good, thank you.  I will take a proper look later.

whitwham:
Merged.  Thank you.

PR #1332, opened 2021-09-16
S3 V4 signature to handle no credentials.

whitwham:
Signature v4 requests were trying to create authorisation strings when no credentials were present.  This change adds the v2 lines that handle this. Fixes samtools/samtools#1491.


daviesrob:
I've checked both with and without credentials.  Looks good, thanks.

PR #1336, opened 2021-09-28
Update to bgzip man page.

whitwham:
Added missing --keep and --test options to man page. Updated NEWS file.


jmarshall:
Might as well keep them in alphabetical order (in both places in the man page).

Currently `-g` is out of order in the description list, presumably because of the similarity of `--reindex` and `--rebgzip`, but IMHO that's a bit spurious and `-g` should probably be moved to its proper place in the list.

whitwham:
Good point.  I'll change it.

jmarshall:
Cheers. (Similarly in `bgzip_main_usage()` too I guess.)

Issue #530, opened 2017-05-05
Zstd Support


reedacartwright:
An undergrad is the lab is looking into how modern compression algorithms work with genomic data. Long story short, it looks like there are justifications to use LZ4, ZSTD, and Brotli in different cases.

If we produced a pull request to add support for these libraries to htslib, maybe initially to cram, would it be likely to be accepted?

kyleabeauchamp:
I have also observed good performance on ZSTD, although it appears that the CRAM compressor is also a very modern approach. 

See my issue here on the "unifying compressor API" squash:

https://github.com/quixdb/squash-benchmark/issues/35

kyleabeauchamp:
In particular, I was able to convince the squash benchmark folks to add the genome FASTA data to their compressor benchmark.  I then ran the benchmark on chr1 and observed very good compression and speed for zstd at the highest (21) compression level:

https://gist.github.com/kyleabeauchamp/e5f5d79aa153bc85d854a705a25c9166

reedacartwright:
@kyleabeauchamp My student has been using a modified version of `lzbench` to look at compression of FASTA, VCF, GTF, etc. I'll have her look at `squash` as well.

A programmer in the lab has produced a version of htslib that uses zstd instead of zlib, relying on the wrapper provided by zstd to port the code. https://github.com/zmertens/htslib/tree/zstd  Ideally, we would like to have something in which zstd and zlib work together.

kyleabeauchamp:
Cool.  I'm not actually an HTSLib developer :) but I wanted to voice my support for any project that improves the compression ratio and/or read throughput of HTSLib-supported formats (particularly BAM and CRAM).

jkbonfield:
I've experimented with Zstd and CRAM before.  It's a definite win over zlib, at some compression levels both in speed and size.  I experimented on this in Staden io_lib for the original Scramble implementation of CRAM (https://github.com/jkbonfield/io_lib).  There is an earlier Zstd branch there but also a cram_modules branch which has lots of other experimental plugins.

The problem isn't adding Zstd, which is actually very easy, but being able to put together a group of changes that jointly justify a change to the specification.  We don't want to do this piecemeal with lots of variations of specification, but a significant step forward as we did between CRAM v2 and v3.  I have various other experiments on the go and in particular have a significant improvement for compression of query names (beating everything pretty much bar the excessively slow context mixing algorithms). 

[...]

PR #1340, opened 2021-10-05
Add basic recognition for XZ- & Zstd-compressed data, and D4 format

jmarshall:
There has been some recent discussion of using Zstandard compression (cf #530).

As a very small first step towards this, this adds `xz_compression` and `zstd_compression` to `enum htsCompression` and recognises their magic numbers, as we do for `bzip2_compression`. As htslib already uses liblzma, this also adds `decompress_peek_xz()` (if liblzma is configured in) to peek at the contents. Unlike libbz2's decompression routines, an `lzma_stream_decoder` can decompress a few hundred bytes of a stream without needing to read in an entire multi-hundred-KB block — hence is suitable for use with peeking.

There is also some talk of an intersection between D4 and htsget (though IMHO I'm not yet sure this intersection is non-empty), and of upcoming work in that area. So this also adds basic recognition of D4 format. Note that the D4 spec is fairly incomplete and (as far as I am aware) the format currently has only one implementation. So you may wish to leave the D4 commit off for now. This PR does not set `version.major`/`.minor` for D4; see 38/d4-format#34.


jmarshall:
Also fixed a minor _cram_io.c_ memory leak I noticed while studying the `lzma_*` code.

jkbonfield:
Thanks.

I'm curious where you're going with this, but useful additions to have.

PR #1349, opened 2021-10-29
Fill in cr->rg when writing CRAM v1 [minor]

jmarshall:
I'm sure no-one much cares about writing CRAM v1, but…

It seems like this branch of the `if`-`else` should also be filling in `cr->rg`. Noticed because with `brg` used only in the `assert()`, there is a “variable unused” warning when compiling with `-DNDEBUG`.

I suppose now that `brg` is used in the assignment, it could stay as an `assert()` and produce an assert failure (or a hard crash with `-DNDEBUG`) rather than `goto block_err` and returning an error code. Probably doesn't matter much either way — see paragraph 1 above!


jkbonfield:
Thanks, although I'm tempted to simply nuke all version 1 code, both reading and writing, as it's just another potential attack vector for unverified input and frankly it never existed in the wild as far as I can see.

PR #1351, opened 2021-11-02
Iteration over SAM header lines and header expression filtering

jmarshall:
A proposal to add API functions allowing user code to iterate over all header lines in order. Then with a way of referencing a particular generic header line established, we can add a `sam_hdr_passes_filter()` function to enable using filter expressions on SAM header lines.

Draft because the proposed basic iteration API (and everything else) is still in flux: the `_line_` in the names can probably be removed, and I am considering changing to C++-esque `begin`/`end` terminology (which is an option for a SAM aux field iterator, and it would be good to use the same terminology for both) instead of `first`/`NULL`.


jmarshall:
> Draft because the proposed basic iteration API (and everything else) is still in flux: the _line_ in the names can probably be removed

Updated to a more final version of the proposed API names. Still needs some test cases probably…

jmarshall:
(CI failures are because Cirrus's `catalina-base` image contains an outdated and now-broken brew. See cirruslabs/osx-images#37; apparently Catalina is too old 🤣  and is [deprecated on Cirrus](https://cirrus-ci.org/guide/macOS/).)

PR #1354, opened 2021-11-09
Proposal for an aux tagged field iterator API

### LINKED ISSUE #1319: Wish list: aux tag iterator API
(Put here simply so we don't forget it.)

The (now merged) Samtools PR https://github.com/samtools/samtools/pull/516 copies over the `aux_type2size` and `skip_aux` functions as it has a mechanism of iterating through the tag list.

Something akin to these functions ought to be public, although perhaps not these explicit interfaces.  I think a more iterator focused approach would work.  Eg in a similar fashion to kstring initialisations:

```
bam_tag_iter_t it = BAM_TAG_ITER_INITIALIZE;  // or bam_tag_iter_initialize(&it) if we want it opaque.
uint8_t *tag;
while ((tag = bam_aux_next(b, &it))) {
    // do stuff
}
```

The tag initialise function just nulls the contents.  The `bam_aux_next` function skips to the next tag, or identifies the first one if the pointer hasn't been set.  The BAM data end pointer from `skip_aux` can be internalised into the data type and hidden from the user.   Alternatively we just expose `skip_aux` as it is now and have an `aux_end` function to avoid the mess of grubbing through `b->data + b->l_data` internals.

`aux_type2size` may be useful as-is, as we already have search functions so being able to return how big it is would be helpful for more than this case.  However it could be renamed perhaps - maybe `bam_aux_size`.

jmarshall:
As suggested in #1319, this adds new API functions for iterating through a BAM record's aux fields, inline accessor methods for field tag and type (or code can continue to use `s-2` and `*s`), and a  variant of `bam_aux_del()` that returns the (updated) iterator to the following field (for use in iterator-based loops that delete fields). IMHO there's no need for an additional `bam_tag_iter_t` type, as the `uint8_t*` pointer to aux fields is already all over this part of the API and there's no need for more iterator state anyway.

This draft rewrites `bam_aux_get()` in terms of the new iterator functions. It's slightly debatable whether to do that (and remove the old version) or to leave it alone.

Tossing up between using the C++-style `bam_aux_erase()` nomenclature or calling the new deletion function `bam_aux_del1()` or so. I also have a version with an additional `bam_aux_erase_range()` function that deletes several aux fields at once (so could be used to directly reimplement the way James did the deleting in samtools/samtools#516, which probably reduces the number of `memmove()` calls so is 0.1% more efficient), but not totally sure the slight complication is worth it.

This somewhat parallels PR #1351, but it's a separate part of the API (aux fields vs. headers) so might as well be a separate PR. Needs some test cases once the basic approach has basic approval.

Closes #1319.

### DIFF
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -1438,7 +1438,6 @@ int sam_passes_filter(const sam_hdr_t *h, const bam1_t *b,
 
 /// Converts a BAM aux tag to SAM format
 /*
- * @param b    Pointer to the bam record
  * @param key  Two letter tag key
  * @param type Single letter type code: ACcSsIifHZB.
  * @param tag  Tag data pointer, in BAM format
@@ -1628,6 +1627,29 @@ static inline const uint8_t *sam_format_aux1(const uint8_t *key,
     return NULL;
 }
 
+/// Return a pointer to a BAM record's first aux field
+/** @param b   Pointer to the BAM record
+    @return    Aux field pointer, or NULL if the record has none
+
+When NULL is returned, errno will also be set to ENOENT. ("Aux field pointers"
+point to the TYPE byte within the auxiliary data for that field; but in general
+it is unnecessary for user code to be aware of this.)
+ */
+HTSLIB_EXPORT
+uint8_t *bam_aux_first(const bam1_t *b);
+
+/// Return a pointer to a BAM record's next aux field
+/** @param b   Pointer to the BAM record
+    @param s   Aux field pointer, as returned by bam_aux_first()/_next()/_get()
+    @return    Pointer to the next aux field, or NULL if no next field or error
+
+Whenever NULL is returned, errno will also be set: ENOENT if @p s was the
+record's last aux field; otherwise EINVAL, indicating that the BAM record's
+aux data is corrupt.
+ */
+HTSLIB_EXPORT
+uint8_t *bam_aux_next(const bam1_t *b, const uint8_t *s);
+
 /// Return a pointer to an aux record
 /** @param b   Pointer to the bam record
     @param tag Desired aux tag
@@ -1639,6 +1661,19 @@ static inline const uint8_t *sam_format_aux1(const uint8_t *key,
  */
 HTSLIB_EXPORT
 uint8_t *bam_aux_get(const bam1_t *b, const char tag[2]);
+
+/// Return the aux field's 2-character tag
+/** @param s   Aux field pointer, as returned by bam_aux_first()/_next()/_get()
+    @return    Pointer to the tag characters, NOT NUL-terminated
+ */
+static inline
+const char *bam_aux_tag(const uint8_t *s) { return (const char *) (s-2); }
+
+/// Return the aux field's type character
+/** @param s   Aux field pointer, as returned by bam_aux_first()/_next()/_get()
+    @return    The type character: one of cCsSiI/fd/A/Z/H/B
+ */
+static inline char bam_aux_type(const uint8_t *s) { return *s; }
 
 /// Return a SAM formatting string containing a BAM tag
 /** @param b   Pointer to the bam record
@@ -1750,6 +1785,22 @@ int bam_aux_append(bam1_t *b, const char tag[2], char type, int len, const uint8
 */
 HTSLIB_EXPORT
 int bam_aux_del(bam1_t *b, uint8_t *s);
+
+/// Delete an aux field from a BAM record
+/** @param b   The BAM record to update
+    @param s   Pointer to the aux field to delete, as returned by
+               bam_aux_first()/_next()/_get(); must not be NULL
+    @return    Pointer to the following aux field, or NULL if none or on error
+
+Identical to @c bam_aux_del() apart from the return value, which is an
+aux iterator suitable for use with @c bam_aux_next()/etc.
+
+Whenever NULL is returned, errno will also be set: ENOENT if the aux field
+deleted was the record's last one; otherwise EINVAL, indicating that the
+BAM record's aux data is corrupt.
+ */
+HTSLIB_EXPORT
+uint8_t *bam_aux_remove(bam1_t *b, uint8_t *s);
 
 /// Update or add a string-type tag
 /* @param b    The bam record to update
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -1777,11 +1777,13 @@ HTSLIB_EXPORT
 int bam_aux_append(bam1_t *b, const char tag[2], char type, int len, const uint8_t *data);
 
 /// Delete tag data from a bam record
-/* @param b The bam record to update
-   @param s Pointer to the tag to delete, as returned by bam_aux_get().
-   @return 0 on success; -1 on failure
-   If the bam record's aux data is corrupt, errno is set to EINVAL and this
-   function returns -1;
+/** @param b   The BAM record to update
+    @param s   Pointer to the aux field to delete, as returned by bam_aux_get()
+               Must not be NULL
+    @return    0 on success; -1 on failure
+
+If the BAM record's aux data is corrupt, errno is set to EINVAL and this
+function returns -1.
 */
 HTSLIB_EXPORT
 int bam_aux_del(bam1_t *b, uint8_t *s);
--- a/sam.c
+++ b/sam.c
@@ -4614,31 +4614,42 @@ static inline uint8_t *skip_aux(uint8_t *s, uint8_t *end)
     }
 }
 
+uint8_t *bam_aux_first(const bam1_t *b)
+{
+    uint8_t *s = bam_get_aux(b);
+    uint8_t *end = b->data + b->l_data;
+    if (s >= end) { errno = ENOENT; return NULL; }
+    return s+2;
+}
+
+uint8_t *bam_aux_next(const bam1_t *b, const uint8_t *s)
+{
+    uint8_t *end = b->data + b->l_data;
+    uint8_t *next = s? skip_aux((uint8_t *) s, end) : end;
+    if (next == NULL) goto bad_aux;
+    if (next >= end) { errno = ENOENT; return NULL; }
+    return next+2;
+
+ bad_aux:
+    hts_log_error("Corrupted aux data for read %s", bam_get_qname(b));
+    errno = EINVAL;
+    return NULL;
+}
+
 uint8_t *bam_aux_get(const bam1_t *b, const char tag[2])
 {
-    uint8_t *s, *end, *t = (uint8_t *) tag;
-    uint16_t y = (uint16_t) t[0]<<8 | t[1];
-    s = bam_get_aux(b);
-    end = b->data + b->l_data;
-    while (s != NULL && end - s >= 3) {
-        uint16_t x = (uint16_t) s[0]<<8 | s[1];
-        s += 2;
-        if (x == y) {
+    uint8_t *s;
+    for (s = bam_aux_first(b); s; s = bam_aux_next(b, s))
+        if (s[-2] == tag[0] && s[-1] == tag[1]) {
             // Check the tag value is valid and complete
-            uint8_t *e = skip_aux(s, end);
-            if ((*s == 'Z' || *s == 'H') && *(e - 1) != '\0') {
-                goto bad_aux;  // Unterminated string
-            }
-            if (e != NULL) {
-                return s;
-            } else {
-                goto bad_aux;
-            }
+            uint8_t *e = skip_aux(s, b->data + b->l_data);
+            if (e == NULL) goto bad_aux;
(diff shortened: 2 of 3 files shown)


jmarshall:
(See jmarshall/samtools@bf429a2f78c14737c86f49fa3f1cb0db9123c07c for a draft of the samtools `--keep-tag`/`--remove-tag` functionality implemented via this API.)

jmarshall:
I'll take complete silence as signalling approval of the basic approach :smile:

The new iterator-oriented field deletion functions are now `bam_aux_remove()` and `bam_aux_remove2()`, as `…remove` is fairly widespread in HTSlib API function names. Test cases exercising the new API functions added.

@daviesrob et al: This is now ready for review.

sam.c
```diff
@@ -4581,31 +4581,42 @@ static inline uint8_t *skip_aux(uint8_t *s, uint8_t *end)
...
-            uint8_t *e = skip_aux(s, end);
-            if ((*s == 'Z' || *s == 'H') && *(e - 1) != '\0') {
-                goto bad_aux;  // Unterminated string
-            }
-            if (e != NULL) {
-                return s;
-            } else {
-                goto bad_aux;
-            }
+            uint8_t *e = skip_aux(s, b->data + b->l_data);
```
jkbonfield, line 4612: This feels like an inefficiency as `bam_aux_next` and `skip_aux` are both parsing the next aux tag.  Usually fast, but potentially slower if it's a long string (SA:Z, etc).  I wonder if it should just do something like `s = e` in the loop iterator.

htslib/sam.h
```diff
@@ -1750,6 +1785,36 @@ int bam_aux_append(bam1_t *b, const char tag[2], char type, int len, const uint8
...
+              bam_aux_first()/_next()/_get()
+   @param e   Pointer to the next aux field after the last one to delete,
+              as returned by bam_aux_first()/_next()/_get()
+   @return    Pointer to the following aux field, or NULL if none or on error
+
+When @a e is NULL, removes all fields from @a s onwards. When @s is also NULL,
+no fields are removed.
+ */
+HTSLIB_EXPORT
+uint8_t *bam_aux_remove2(bam1_t *b, uint8_t *s, uint8_t *e);
```
jkbonfield, line 1816: What's the use case for `bam_aux_remove2`?

I can see a case for specifying a list of aux tags we wish to remove and it doing the appropriate logic to remove them all within a single pass and avoid potentially O(N^2) complexity on extremely large sets of aux tags.  However unless I'm misunderstanding this, it looks like it's only removing aux tags that are adjacent to each other which is rather tenuous.  I guess the only use I can see is trivial removal of *all* tags.

Although if it's a general list, then it doesn't entirely remove the `samtools view -x` code as that has both keep-list and remove-list functionality.  If we were sufficiently bothered to replicate it here, then the list would need a negation flag to indicate the keep-or-reject status.

jkbonfield:
I like this feature.  Bar a few niggles, it looks solid.

Thanks

htslib/sam.h
jmarshall: It's akin to C++ iterators, which have a two-argument `erase()` to remove a consecutive range of items from a sequence. However as you don't directly control the ordering of aux fields, it's not as obviously applicable here. The motivating use case was noted in the first comment above:

> I also have a version with an additional ~~`bam_aux_erase_range()`~~`bam_aux_remove2()` function that deletes several aux fields at once (so could be used to directly reimplement the way James did the deleting in https://github.com/samtools/samtools/pull/516, which probably reduces the number of memmove() calls so is 0.1% more efficient), but not totally sure the slight complication is worth it.

There's a draft of that at jmarshall/samtools@aefbf50cf18fef678beb2b1bf364cda68c2e174c though I think it needs minor alterations. (That commit implements `remove_tag` the obvious way, but attempts the optimised way for `keep_tag`.) The idea is you can implement `keep_tag` by iterating over a run of fields you don't want until you see one that you **do** want to keep, and then nuke the whole run of don't-wants at once with `bam_aux_remove2()`; then repeat that until you're at the end of the aux fields.

Essentially `bam_aux_remove2()` falls out of the implementation so might as well be public. Useful for nuke-all-tagged-fields and as a building block for fancy stuff like `keep_tag`.

Or I'm happy to inline the `aux_remove()` helper back into `bam_aux_remove()` again, so it goes away. :shrug:

sam.c
jmarshall: This is the sort of thing I was musing about with “This draft rewrites `bam_aux_get()` in terms of the new iterator functions. It's slightly debatable whether to do that”…

However note that line 4612 is only executed when we've found the right tag, so we're not going to go around the loop again (i.e., `bam_aux_next()` won't be called again).

jkbonfield: Ah yes I'd not considered that the loop is bailing out at this point, so my comment is irrelevant.  Please ignore.

jmarshall: :+1: :smile:

`s = e` instead of `bam_aux_next()` was a nice idea for optimisation though… if the code were different, it would be something to keep in mind…

htslib/sam.h
jkbonfield: The problem is that if we're removing X, Y and Z from tags ABCXYDEFZGHI then the first `XY` removal moves DEFZGHI down 2 places, and then the Z moves GHI down 1 place again.   This is because `bam_aux_remove` always moves up to the end (`data + l_data`) and not up to the next point of interest.  Hence GHI has been moved twice.  

The efficient algorithm however is to move DEF down 2 and then GHI down 3.   Hence your reimplementation is not the same complexity as the original.

In general the algorithm is O(N^2) complexity, although in practice it's probably irrelevant as N is usually small.  There may be some cases though, such as the large PacBio tags, where N becomes quite significant.

jmarshall: Such is the cost of abstraction. The raw pointer version has the liberty of leaving a gap while it is working; the API version must maintain the invariant that the `bam1_t` has a contiguous vector of valid tag fields between API calls.

Perhaps what would be more useful would be a `bam_aux_remove_if(…)` function that took a callback saying whether to remove a particular aux field, and iterated over the whole list itself within the one API call.

jkbonfield: Regarding abstraction, that's why I said I felt the `bam_aux_remove2` function to be of questionable benefit.  Perhaps some special corner cases may work, such as removing all bar the first tag, but I don't really see the benefit of extending the API with something so limited.  If we wanted to have an API which always left bam1_t in a consistent manner, then it should be as I said before - a remove function that takes a list of tag names rather than a single one (and perhaps a keep / remove flag to negate it).  Plus we already have this implemented in samtools view.

We could perhaps do it with `bam_aux_remove_if` callbacks, but it seems like a lot of additional faff to use over a basic array argument.

jmarshall: I've removed `bam_aux_remove2()`, which was amongst other things a pain to use, and reunified `bam_aux_remove()` accordingly. (See force push to 99954c125b47166945a751b3a0d1fa20284b4cac below.)

I'll add something else more useful for implementing stuff like `keep_tag`/`remove_tag` in its place.

jkbonfield:
Bumping this again to see where we're at.

> I've removed bam_aux_remove2(), which was amongst other things a pain to use, and reunified bam_aux_remove() accordingly. (See force push to https://github.com/samtools/htslib/commit/99954c125b47166945a751b3a0d1fa20284b4cac below.)
>
> I'll add something else more useful for implementing stuff like keep_tag/remove_tag in its place.

I haven't refreshed my mind of the discussions and code yet, but was this something you were planning on doing still, and if so is it worth waiting for it or would you view it as part 2 of the API, to arrive in a subsequent PR?  (If so I assume it's a purely additional thing and not something that would need changing of the existing API.)

jmarshall:
Don't refresh your memory of the code as it stands in this PR. Back in April, I got most of the way through refactoring it as discussed. I may have time to finish that up in the next few days.

jmarshall:
Oops, actually it is #1361 that has been substantially refactored from the current state of its PR.

I have an in-progress commit adding `bam_aux_remove_if()` to the existing PR. I was planning on adding it to this PR (as there's some risk it might involve tweaks to the existing proposed addition), or it probably could be done as a Part II.

I may have time to finish that up in the next few days, so maybe put this one on your list to have another look at on Monday :smile:

jkbonfield:
Thanks.  I did check if it was draft before commenting, as I had a recollection it was, but as you noted it's the samtools side that is the draft bit.

I'm refreshing my memory of what this does anyway as it'll help get my head in the zone for the upcoming changes.

sam.c
```diff
@@ -4614,23 +4625,26 @@ uint8_t *bam_aux_get(const bam1_t *b, const char tag[2])
...
+}
+
+uint8_t *bam_aux_remove(bam1_t *b, uint8_t *s)
+{
+    uint8_t *end = b->data + b->l_data;
+    uint8_t *next = skip_aux(s, end);
+    if (next == NULL) goto bad_aux;
+    memmove(s-2, next, end - next);
+    b->l_data -= next - (s-2);
+    if (next >= end) { errno = ENOENT; return NULL; }
```
jkbonfield, line 4641: Nit-picking: If `next > end` then the memmove has attempting moving a negative size.  If `next == end` then memmove is a no-op.  Both of these imply memmove should be after this check.

In reality though `next <= end` in all scenarios due to the error checking in `skip_aux`, so it just *looks* wrong rather than actually being wrong.

sam.c
```diff
@@ -4614,23 +4625,26 @@ uint8_t *bam_aux_get(const bam1_t *b, const char tag[2])
...
-    int l_aux = bam_get_l_aux(b);
-    aux = bam_get_aux(b);
-    p = s - 2;
-    s = skip_aux(s, aux + l_aux);
-    if (s == NULL) goto bad_aux;
-    memmove(p, s, l_aux - (s - aux));
-    b->l_data -= s - p;
-    return 0;
+    s = bam_aux_remove(b, s);
+    return (s || errno == ENOENT)? 0 : -1;
```
jkbonfield, line 4631: What if errno is already `ENOENT` on entry?  This is a change of behaviour as it will now fail in situations it previously worked.

I think we should take a temporary copy of errno, set errno to zero and check after `bam_aux_remove`,  with a reset after if appropriate.

sam.c
jmarshall: Good spotting — this indicates a bug, probably introduced during that refactoring!

That `if` is supposed to return NULL or `s` depending on whether `s` was previously the last entry. Probably `end` needs to be recalculated now that `l_data` has changed, or something equivalent.

jmarshall:
(That force-push was just to bring this up to current **develop** unchanged. Further pushes will make actual changes.)

sam.c
jmarshall: The post-condition of `bam_aux_remove()` is that either the return value is non-NULL or `errno` will have been set inside `bam_aux_remove()` to indicate why it returned NULL. So there is no change in behaviour as any pre-existing `errno` value is never consulted.

(Both `bam_aux_del()` and `bam_aux_remove()` have a pre-condition that their `s` argument must be non-NULL — and `bam_aux_del()` has always has this requirement. I should probably add a note to the doco for both functions making this explicit.)

jkbonfield: Sorry that's my silly mental-block or misreading of the OR statement.  You're correct.  Thanks

jmarshall: “Non-NULL” note added to both functions.

sam.c
jmarshall: I think this was not a bug in the end, because as you say `next > end` can't happen. But I have reorganised the code and think it's probably clearer now anyway.

jkbonfield:
Thanks.  I'll review the first couple commits with a view to merging them in isolation.

jmarshall:
Gah, I'll remove that last commit and bring `bam_aux_remove_if()` back as a separate PR later — this one has been hanging around quite long enough!

So you can review this part 1 as a normal PR…

jkbonfield:
Hah beat me to it.  I was going to squash and merge outside of this PR so as not to confuse your own branch (after testing it in anger with another samtools branch).  Are you happy for me to squash and push over the top of your branch so this tidies itself up automagically on merge?  (Or do it yourself - I don't care either way)

jmarshall:
I think I know which other samtools branch that is :smile:

Feel free to push to this PR's branch. Or use the “Squash and merge” UI button in due course, which will turn the PR happily purple/merged too.

jkbonfield:
Yeah the draft sanitize function.  We also discussed whether it belongs in htslib.  My preference would be there and maybe enabled by an `hts_set_opt` call so any program can read *AM records with some basic aligner (*cough bwa*) fix-ups applied and be sure it doesn't have to worry about silly things like bases aligning off the ends of chromosomes or unmapped data with silly mapping qualities and non-sensical cigar strings.

But for now it's a useful test of an external tool against this API too. :)

jkbonfield:
Ah frogger-it... meant squash and merge rather than just rebase.  Oh well, it is as it is :)

PR #1358, opened 2021-11-10
Make bcf_remove_allele_set() set work on unpacked records

### LINKED ISSUE #1353: vcfutils::bcf_remove_allele_set should unpack to BCF_UN_INFO
`bcf_remove_allele_set` currently checks that the record is unpacked up to `BCF_UN_STR`, but it later modifies the INFO fields with Number=A, R, or G. So it should unpack up to `BCF_UN_INFO`.

https://github.com/samtools/htslib/blob/9045785cef02e7cd97b0b4caecbdcaf6217a986a/vcfutils.c#L262

pd3:
Resolves #1353

### DIFF
--- a/vcfutils.c
+++ b/vcfutils.c
@@ -256,10 +256,10 @@ int bcf_remove_allele_set(const bcf_hdr_t *header, bcf1_t *line, const struct kb
     int *map = (int*) calloc(line->n_allele, sizeof(int));
     uint8_t *dat = NULL;
 
+    bcf_unpack(line, BCF_UN_ALL);
+
     // create map of indexes from old to new ALT numbering and modify ALT
     kstring_t str = {0,0,0};
-    if (!line->d.allele)
-        bcf_unpack(line, BCF_UN_STR);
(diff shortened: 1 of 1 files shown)


vcfutils.c
```diff
@@ -256,10 +256,10 @@ int bcf_remove_allele_set(const bcf_hdr_t *header, bcf1_t *line, const struct kb
     int *map = (int*) calloc(line->n_allele, sizeof(int));
     uint8_t *dat = NULL;
 
+    bcf_unpack(line, BCF_UN_FMT);
```
daviesrob, line 259: Should this be `BCF_UN_ALL`?  Looking at [`bcf_unpack()`](https://github.com/samtools/htslib/blob/f867feab/vcf.c#L3060), `BCF_UN_FMT` will only unpack the `indiv` data, but this needs to also access `info` and `allele`.

pd3: Of course it should, duh

pd3: This is now replaced with the correct `BCF_UN_ALL`

PR #1360, opened 2021-11-17
Add pathA##idx##pathB support for CRAM indices.

jkbonfield:
Fixes samtools/samtools#1541

The original report was for s3: URIs, but it turns out this is simply lacking in local files too.   It seems surprising we never implemented this, but even in 1.10 (where `##idx##` was first added) CRAM support was missing. 7e583ccea8e8a1a31c16bf5224d8f7785182dd82 stated "it works transparently for all file types and all programs".  I guess we never actually checked it.

### DIFF
--- a/cram/cram_index.c
+++ b/cram/cram_index.c
@@ -173,6 +173,11 @@ int cram_index_load(cram_fd *fd, const char *fn, const char *fn_idx) {
 
     idx_stack[idx_stack_ptr] = idx;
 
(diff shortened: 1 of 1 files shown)


jkbonfield:
Note given it was just `hts_idx_load` that had this functionality, and now it's also `cram_index_load`, my guess is "all formats" still isn't true as we have other index types such as gzi (and maybe tbi) which aren't part of the standard index formats.  I think they're substantially less common though.

Issue #1362, opened 2021-11-30
bcf_sr_seek() still fails


freeseek:
This issue seems related to issue #691 but I was not able to understand the issue in the source code:

Say you have a simple VCF:
```
(echo "##fileformat=VCFv4.2"
echo "##contig=<ID=chr1>"
echo "##contig=<ID=chr2>"
echo -e "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO"
echo -e "chr1\t10000\tx\tA\tC\t.\t.\t.") | bcftools view --no-version -Oz -o A.vcf.gz
bcftools index -ft A.vcf.gz
```

And a `test.c` BCFtools plugin with the following code:
```c
#include <htslib/synced_bcf_reader.h>
#include "bcftools.h"

const char *about(void) { return "\n"; }

int run(int argc, char *argv[]) {
  bcf_srs_t *sr = bcf_sr_init();
  bcf_sr_set_opt(sr, BCF_SR_REQUIRE_IDX);
  if (!bcf_sr_add_reader(sr, "A.vcf.gz")) error("Failed to open %s: %s\n", "A.vcf.gz", bcf_sr_strerror(sr->errnum));
  bcf_hdr_t *hdr = bcf_sr_get_header(sr, 0);
  int rid = 0;

  int i;
  bcf_sr_seek(sr, bcf_hdr_id2name(hdr, rid), 0);
  for (i = 0; bcf_sr_next_line(sr); i++) {
    fprintf(stderr, "round1 chr=%s, pos=%ld\n", bcf_hdr_id2name(hdr, rid), bcf_sr_get_line(sr, 0)->pos + 1);
  }

  bcf_sr_seek(sr, bcf_hdr_id2name(hdr, rid), 0);
  for (i = 0; bcf_sr_next_line(sr); i++) {
    fprintf(stderr, "round2 chr=%s, pos=%ld\n", bcf_hdr_id2name(hdr, rid), bcf_sr_get_line(sr, 0)->pos + 1);
  }

  return 0;
}
```
The plugin seeks at the beginning of contig `chr1` two times and each time it tries to read all the records.

When running the plugin with HTSlib 1.13:
```
$ ./bcftools +plugins/test.so 
round1 chr=chr1, pos=10000
round2 chr=chr1, pos=10000
```

When running the plugin with HTSlib 1.14:
```
$ ./bcftools +plugins/test.so 
round1 chr=chr1, pos=10000
```
[...]

PR #1363, opened 2021-12-09
Make repeated seeks possible

pd3:
Repeated seeks with implicitly created region list wouldn't initialize all internal structures to the original clean state, as demonstrated by the issue #1362, resolved by this commit

### DIFF
--- a/synced_bcf_reader.c
+++ b/synced_bcf_reader.c
@@ -75,6 +75,7 @@ static int _regions_add(bcf_sr_regions_t *reg, const char *chr, hts_pos_t start,
 static bcf_sr_regions_t *_regions_init_string(const char *str);
 static int _regions_match_alleles(bcf_sr_regions_t *reg, int als_idx, bcf1_t *rec);
 static void _regions_sort_and_merge(bcf_sr_regions_t *reg);
+static int _bcf_sr_regions_overlap(bcf_sr_regions_t *reg, const char *seq, hts_pos_t start, hts_pos_t end, int missed_reg_handler);
 
 char *bcf_sr_strerror(int errnum)
 {
@@ -838,6 +839,7 @@ static void bcf_sr_seek_start(bcf_srs_t *readers)
     for (i=0; i<reg->nseqs; i++)
         reg->regs[i].creg = -1;
     reg->iseq = 0;
+    reg->prev_seq = -1;
 }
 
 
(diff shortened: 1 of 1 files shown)


daviesrob:
This fixes the reported seeking bug, but I'm worried that it might have odd interactions with the `missed_reg_handler` callback should anyone use it.  Admittedly it doesn't seem to have been used since bcftools commit [8457b56 ](https://github.com/samtools/bcftools/commit/8457b56), but it would be good to ensure that it behaves correctly.  As far as I can tell, it would be best to not call `missed_reg_handler` at all when using `bcf_sr_seek()`?

pd3:
I just force-pushed a fix to that, now also `missed_reg_handler` should do its job when `bcf_sr_seek()` is used

daviesrob:
I thought of that as a fix, but I'm not sure if it's enough as while it stops `missed_reg_handler` getting called on the current chromosome, it also gets called [later on](https://github.com/samtools/htslib/blob/d3561857/synced_bcf_reader.c#L1441) on the target of the seek.  So I think a seek to a high-ish position could call it, depending on the ranges present?

pd3:
You are right. After a seek it will now behave as if this was the first region at the chromosome and will trigger the callback. I don't know what the right behavior should be. Since it is not used anymore, maybe we just let be for now and wait until a real world use case comes?

daviesrob:
I'd be inclined to just disable it for the `bcf_sr_seek()` case.  The cleanest way might be to rename `bcf_sr_regions_overlap()`, make it static and add an extra parameter to disable the callbacks.  Then make a new `bcf_sr_regions_overlap()` which calls the new function with callbacks enabled so the current API is unchanged.

pd3:
Or just call `assert` from `bcf_sr_seek()` if the callback is set

daviesrob:
No, calling `assert()` would not be appropriate.

pd3:
> No, calling `assert()` would not be appropriate.

I don't understand. The current code is broken, therefore keeping the current behavior should not be appropriate. Trying to use the code with `missed_reg_handler` will likely cause unexpected behavior and this is in the programming space, not user space. In my view `assert` and/or API documentation is the right place.

daviesrob:
Compiling the library with `NDEBUG` defined will make any `assert()` calls go away.  It should therefore only be used to detect bugs inside HTSlib itself, and is not a reliable mechanism to report API misuse.

pd3:
Okay, I modified it using a wrapper around a static function.

daviesrob:
Looks OK now, thanks.

PR #1365, opened 2021-12-11
`-Wdouble-promotion` clean API

SoapGentoo:
We have a GPU codebase that we use with htslib where we have to be very careful with `float`<->`double` conversions, for which we enable `-Wdouble-promotion`. Unfortunately, the publicly facing API has two lines that trigger this, mainly due to the fact that `printf` doesn't have float specifiers. This PR has two commits:

* The first one makes the public API `-Wdouble-promotion` clean by adding explicit casts.
* The second one fixes a simple `-Wdouble-promotion` issue. This one can be dropped, but it's not very invasive.

I tried fixing the whole codebase, but this required more casts. If desired, I can do those too.

### DIFF
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -1492,7 +1492,8 @@ static inline const uint8_t *sam_format_aux1(const uint8_t *key,
         ++s;
     } else if (type == 'f') {
         if (end - s >= 4) {
-            ksprintf(ks, "f:%g", le_to_float(s));
+            // cast to avoid triggering -Wdouble-promotion
+            ksprintf(ks, "f:%g", (double)le_to_float(s));
             s += 4;
         } else goto bad_aux;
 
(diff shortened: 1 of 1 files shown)


jkbonfield:
Thanks for raising this.  We'll discuss it on our weekly meeting (Tuesday).

Fixing the whole code base is probably quite an undertaking and I'm not sure we really wish to have the maintenance overhead of avoiding automatic float to double promotion everywhere (I count 81 one of them, and likely many more in samtools/bcftools if we touch those), but if this is all that's needed for your work then it seems reasonable to me.

errmod.c
```diff
@@ -201,7 +201,7 @@ int errmod_cal(const errmod_t *em, int n, int m, uint16_t *bases, float *q)
             } else q[j*m+k] = q[k*m+j] = -4.343 * em->lhet[cjk<<8|aux.c[k]]; // all the bases are either j or k
         }
         /* clamp to greater than 0 */
-        for (k = 0; k < m; ++k) if (q[j*m+k] < 0.0) q[j*m+k] = 0.0;
+        for (k = 0; k < m; ++k) if (q[j*m+k] < 0.0f) q[j*m+k] = 0.0f;
```
jkbonfield, line 204: Out of curiosity, why is it you need to fix this assignment, but not the `-4.343 * em->...` line above?

SoapGentoo: @jkbonfield good question, I just went with the warnings GCC gave me. Anyhow, I've dropped the commit changing this, seeing it's not in the public API and therefore won't further the actual point of this PR.

PR #1370, opened 2022-01-06
Prevent empty FORMAT fields and output ":.:" instead of "::"

pd3:


daviesrob:
Rebased to fix the unrelated Appveyor problem.

PR #1374, opened 2022-01-10
Remove static from array type

BickfordA:
While this is valid in C99 I believe it was made optional in C11.  While this works in gcc and clang it does not work in MSVC. I know the officially supported build platform is MSYS on windows, but MSVC now supports compiling C11 and C17 and this would be one of the few changes needed to support it.

### DIFF
--- a/hts_internal.h
+++ b/hts_internal.h
@@ -127,7 +127,7 @@ int bgzf_idx_push(BGZF *fp, hts_idx_t *hidx, int tid, hts_pos_t beg, hts_pos_t e
  */
 void bgzf_idx_amend_last(BGZF *fp, hts_idx_t *hidx, uint64_t offset);
 
-static inline int find_file_extension(const char *fn, char ext_out[static HTS_MAX_EXT_LEN])
+static inline int find_file_extension(const char *fn, char ext_out[HTS_MAX_EXT_LEN])
 {
     const char *delim = fn ? strstr(fn, HTS_IDX_DELIM) : NULL, *ext;
     if (!fn) return -1;


jkbonfield:
Thanks.

Interesting.  I've never seen that syntax before.  I suspect it snuck in by accident and stayed there because it works (for us) as we don't use that elsewhere.

daviesrob:
The `static` is there because it forces the parameter to be an array of at least `HTS_MAX_EXT_LEN` bytes (http://port70.net/~nsz/c/c99/n1256.html#6.7.5.3p7 and http://port70.net/~nsz/c/c11/n1570.html#6.7.6.3p7).  It makes the interface safer to use and should not be removed.

jkbonfield:
I know what it does, but we've never used it before.  Is this an official change in policy then?

We've never strictly enforced non-NULL compiler checking before on things taking arrays.  There are countless interfaces that do this.

jmarshall:
I too was unaware of this C99 syntax prior to its use in PR #1096. It remains described in the sections @daviesrob mentioned in the C11, C17, and current C2X draft (n2731.pdf) standards. So more details on your belief that it's optional in later standards would be useful — is it just that MSVC does not support this part of C11 and C17?

jkbonfield:
We discussed this in our weekly meeting and decided to close it.

The decision was we only support POSIX/C99 compatible compilers for building the software.  There was some discussion whether we should be a bit more flexible about non-standard compilers in third party tools that use the library (so linking against it and including public headers), but it's a moot point here as this is an internal header.

BickfordA:
Thanks for taking the time to dig into this! I realize it is an internal header but it was the only thing i had to change (apart from ssize_t ;) ) to get it to compile on MSVC, so i thought it couldn't hurt to ask.  My thoughts were that it might fall under the VLA section of the C spec, though you have convinced me otherwise.  It definitely seems more correct to have the `static` key word. Here is hoping MSVC gets better support for it, at least there has been a little movement with MSVC and the c11/c17 standards.

BickfordA:
Looks like there is an open issue, though not a lot of traction. It has been open since 2020.  I moved some of the specifics that @daviesrob  and @jmarshall  provided:

https://developercommunity.visualstudio.com/t/C11:-static-inside-array-parameter-squar/1475168

Sorry for the rabbit hole, this is definitely an MSVC issue. 

Thanks again!

jkbonfield:
How are you driving MSVC?  I tried it briefly and it blows up all over the shop:

- lack of pthreads
- no M_LN2, M_LN10, M_SQRT in math.h
- no STDIN_FILENO and STDOUT_FILENO
- no SSIZE_MAX (as well as ssize_t obviously)

I haven't got much further than those so far, but it looks like a big uphill struggle.  This was the latest 2022 cl.exe compiler.  Maybe there are specific command line arguments to enable POSIX support that I'm not using?

BickfordA:
Right, you have to bring your own POSIX environment. 

* pthreads: I used the winpthreads library from msys2.

* For the math functions you need to define `_USE_MATH_DEFINES` https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170

* STDIN_FILENO and STDOUT_FILENO are in a moc unistd.h header as:
```
#define STDIN_FILENO  0
#define STDOUT_FILENO 1
#define STDERR_FILENO 2

#define R_OK 4
````
This project provides a lot of headers which i found useful: https://github.com/win32ports/

* `SSIZE_MAX` is `MAXSSIZE_T` which comes from `basetsd.h` which i added to the config.h file (yes it is a bit of a hack).

Definitely a hill climb, but the hill is smaller than it used to be. It is definitely a bespoke process and I understand not wanting to support it, but you can almost do it without making any changes to the library.

jkbonfield:
So this is indeed more than the minimal changes in this PR.

That said, if it weren't for this use of "static" the entire rest of the changes could pretty much be forced into CFLAGS (I assume MSVC has a pre-include style option to force an include file to be used before every compile, but I haven't checked that).  We did discuss replacing that `static` with something like `HTS_STATIC_ARRAY` that can be defined in an OS specific way, but it was rejected, so I think this is probably dead in the water. Sorry

Issue #1025, opened 2020-02-14
ssize_t missing definition


sg-james:
### Background
- Windows 10
- mingw compiled htslib (latest develop)
- msvc 2017 (main program)

### Problem:

My program cannot link / use headers: `bgzf.h` and `kstring.h` since they `ssize_t`, which is not defined for this platform.

### Suggestion:

Could we add a compatibility define somewhere like `hts_defs.h`? e.g.

```cpp
// hts_defs.h

#if defined(_WIN32) && defined(_MSC_VER) && !defined(ssize_t)
#include <BaseTsd.h>
typedef SSIZE_T ssize_t;
#endif
```

### Workaround

I'm currently using this as a work around:
```cpp
// myfile.cpp
#include "aaa_htslib_compat.h" // include before any htslib headers
#include <htslib/bgzf.h"
```
but I have to remember to do this for each and every file that needs compatibility.

Given how close default htslib headers are to "just working", would a PR for something like this be accepted?

jkbonfield:
I don't know when our use of `ssize_t` started, but it predates 2012 when this code was initially checked in to git.  The type is not a formal C99 data type, but comes from POSIX.  Our goal is to support C99/POSIX compliant platforms as far as possible, which (mostly) covers Mingw on windows.  We don't have plans for MSVC compatibiility as it's more work for us and broadly speaking not something we wish to maintain.  My fear and past experience of dealing with MSVC is it's close to "just working" right up until the next thing and the next thing.

Given apparently Microsoft chose to create a `SSIZE_T` instead of a `ssize_t` type, could you not just use the msvc equivalent of `-Dssize_t=SSIZE_T` in CFLAGS?

sg-james:
> or put those into your source prior to inclusion of bgzf.h

Sorry, I didn't fully specify the contents of my `#include "aaa_htslib_compat.h". It is just the additional code block under "Suggestion", which works quite well, but is prone to forgetfulness. 

I was just wondering how much MSVC support / compat would be acceptable, and since I can live without it, I can understand not going down the road to "just working".

Thanks for taking the time to explain

jmarshall:
Harmonising this sort of platform deficiency is the role of _htslib/hts_os.h_, so something like this wouldn't be completely out of place in that header. However the other _htslib/*.h_ headers don't include that one themselves and shouldn't start doing so, so your code would need to `#include <htslib/hts_os.h>` itself before including other HTSlib headers.

[...]

PR #1375, opened 2022-01-10
Include definition for MSVC ssize_t

BickfordA:
Adding the definition for ssize_t would allow for compiling with MSVC c11/c17.

### DIFF
--- a/htslib/hts_defs.h
+++ b/htslib/hts_defs.h
@@ -103,6 +103,11 @@ DEALINGS IN THE SOFTWARE.  */
 #define HTS_FORMAT(type, idx, first)
 #endif
 
+#if defined(_MSC_VER)
+#include <BaseTsd.h>
+typedef SSIZE_T ssize_t;
+#endif
+
 #if defined(_WIN32) || defined(__CYGWIN__)
 #if defined(HTS_BUILDING_LIBRARY)
 #define HTSLIB_EXPORT __declspec(dllexport)


jmarshall:
See the previous discussion in #1025.

jkbonfield:
Our main work is on Unix/MacOS, and Windows support so far has been using Unix-derived compilers, such as the MinGW tool chain.  We just don't have the capacity in group to fully support all the nuances of another platform.

However, if it really is just 1 or 2 very minor things, then perhaps we can add definitions which aid compilation.  However it's not something we're likely to be doing in a continuous integration service (unless you wish to contribute magic to make that work too) so it's always potentially going to break again.

Can you confirm how many other problems there are when compiling with MSVC?  Ie when doing a "make -k" to let it continue beyond errors, how many types of error are there?

BickfordA:
Thanks for looking at this! This and the static array initialization reference in: https://github.com/samtools/htslib/pull/1374 were the only changes i needed.  Everything else could be handled with `#defines` and some options in the config.h.

I also saw this ticket which looks like maybe some forward movement in this direction: https://github.com/samtools/htslib/pull/1377

Happy to help! I also could put the block behind a flag so that it could be turned on/off if people don't want the behavior. 

I currently am using CMake to drive the compilation (so some files are not being built tests, plugins etc) but i could look at using `nmake` and getting the out put of the compilation if that would be helpful.

jkbonfield:
Closing as replaced by #1377 (now merged).

PR #1378, opened 2022-01-12
Capturing Error Messages 

BickfordA:
Super excited to see the new logging functions, and consolidation of the error reporting! When embedding htslib it would be really nice to be able to capture the errors if a function has an error code. 

These changes would keep a thread local buffer of the last error message which could be checked in the event that a operation fails. Sort of like errno but more descriptive?

Not sure if you want these changes maybe they could be inspiration for another way of getting the error? Let me know if you have any feedback.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -4809,7 +4809,7 @@ void hts_log(enum htsLogLevel severity, const char *context, const char *format,
     errno = save_errno;
 }
 
-int get_last_err(const char *message, int len)
+int get_last_err(char *message, int len)
 {
     char *thread_log_buffer = NULL;
 
--- a/hts.c
+++ b/hts.c
@@ -81,6 +81,12 @@ KHASH_INIT2(s2i,, kh_cstr_t, int64_t, 1, kh_str_hash_func, kh_str_hash_equal)
 HTSLIB_EXPORT
 int hts_verbose = HTS_LOG_WARNING;
 
+HTSLIB_EXPORT
+const int hts_log_buffer_size = 256;
+
+static pthread_key_t threadlocal_log_key;
+static pthread_once_t threadlocal_log_init_done = PTHREAD_ONCE_INIT;
+
 const char *hts_version()
 {
     return HTS_VERSION_TEXT;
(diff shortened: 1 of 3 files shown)


daviesrob:
This looks like messages from threads could be trapped inside buffers that can't be easily accessed.  I think it might be better if the messages from all threads could be combined into a stream.

This could be done now without making any changes to HTSlib by redirecting `STDERR` to a self-pipe.  It would be necessary to ensure that the reading end of the pipe got drained promptly.

If we do make changed to HTSlib, we could try making a global thread-safe ring buffer to capture all of the messages.  Or we could allow the messages to be redirected to an `hFILE`, and make some sort of plugin to allow the messages to be captured into a memory buffer.

BickfordA:
The problem I'm having is differentiating the error messages when reading multiple files from different threads.  I dont think there is a way to redirect stderr on a thread local basis, i think it only works on a per process level. Maybe there is a clever way to do this though?  

With the proposed after a failed api call results in a nonzero return value, you could check the last error message written to the thread local string.  Since any other concurrent calls would be coming from separate threads you don't have to worry about them changing the error state. There might be a better way to do this though. 

Here is the problem i was trying to solve:

If reading three files A,B,C in concurrently in different threads. If file B and C have errors, I would like to be able to separate the errors from B , and the errors from C.  And be able to get the errors for B after it fails without having to worry about them being overwritten by the errors from C.

If there was a global thread-safe ring buffer or redirection to an hFile could there be some way to tell which thread generated the error so that the error message could be match back to the thread/function.

daviesrob:
I guess this might work if you started a thread yourself, and then used it exclusively for a single file.  That doesn't really match how HTSlib operates, where thread pool workers will pick up decoding jobs as they arrive, and don't have any link to a specific file.

I think what you really need is a way of linking the messages to a specific file handle, or possibly API call.  I'm afraid it might not be too easy, given the constraints on how the library works at the moment.  We'll have to think about it, but I fear that there won't be an easy solution to this problem.

BickfordA:
That is how i have been using the library, with a dedicated thread for a single file operation (where a file operation could be indexing a file, reading a region of the file, etc). That's a good point, it doesn't work when there are worker threads that have an error.  Though failures in the work threads do trickle up the the overarching job.  

It isn't ideal though because some of the error context is lost. It doesn't seem like there is a easy solution, at least not yet  :)

daviesrob:
Closing due to the problems mentioned above with messages being trapped in per-thread buffers, and the mismatch with how the thread pool works.  We'll have a think to see if we can come up with a better way of linking messages to individual file handles.

PR #1381, opened 2022-01-14
Update docs: Iterate through bam1_t records

BickfordA:
I think this needs to iterate through the records, as it is just passing the index right now.

### DIFF
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -938,7 +938,7 @@ void bam_destroy1(bam1_t *b);
 
  cleanup:
    for (size_t i = 0; i < nrecs; i++)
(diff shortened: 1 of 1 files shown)


daviesrob:
Yes, I think you're right.  Although looking at it, the entire `for` loop should also have `if (recs) { ... }` around it due to the test on [line 905](https://github.com/samtools/htslib/blob/ba2e5d9a01c0b5c177acfff20073517293a2d5da/htslib/sam.h#L905).

daviesrob:
I've taken the liberty of adding a commit with my suggested `if (recs) { ... }` change...

BickfordA:
Nice, thanks!

PR #1388, opened 2022-01-26
Improve handling of CRAM "alignment span" for unmapped data.

### LINKED ISSUE #1387: Incorrect count of unmapped reads in a CRAM file using samtools view
Using samtools 1.14, Darwin 20.6.0.

In the [disq](https://github.com/disq-bio/disq) repo, we cross-check some test results using both htsjdk and samtools to ensure both toolkits give the same results. Attached is an htsjdk-generated .cram/.crai (`incorrect_unmapped_count.cram`), and a reference and index. The  cram file contains one container/slice with 2000 mapped reads, and second (unmapped) container/slice containing only 2 unmapped reads. The command:

`samtools view -c -T test.fa incorrect_unmapped_count.cram '*'`

returns 0 unmapped records, but htsjdk finds 2, which I think is correct. If I use samtools to convert the entire cram to a sam, the unmapped reads seem to be intact. As far as I can tell, the .crai looks correct. 

I'm not sure if the problem is in htsjdk, samtools, or user error, but I can't spot the problem. Any help appreciated.

[incorrect_unmapped_count.zip](https://github.com/samtools/samtools/files/7935417/incorrect_unmapped_count.zip)

jkbonfield:
When writing, the specification states unmapped data must have alignment start and alignment span both zero.  Previously span was 1 (from end-start+1).

When reading, the code that filters data by the requested range now has an exception for ref_seq -1 (unmapped) so span is not checked.  This fixes the check for end-start-1 failing.  Note this previous triggered issues on files (correctly) output from htsjdk.

Fixes #1387

### DIFF
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -1397,6 +1397,9 @@ static int add_read_names(cram_fd *fd, cram_container *c, cram_slice *s,
     return -1;
 }
 
+// CRAM version >= 3.1
+#define CRAM_ge31(v) ((v) >= 0x301)
+
 /*
  * Encodes all slices in a container into blocks.
  * Returns 0 on success
@@ -1524,6 +1527,12 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
 
         if (c->multi_seq) {
             s->hdr->ref_seq_id    = -2;
+            s->hdr->ref_seq_start = 0;
+            s->hdr->ref_seq_span  = 0;
+        } else if (c->ref_id == -1 && CRAM_ge31(fd->version)) {
+            // Spec states span=0, but it broke our range queries.
+            // See commit message for this and prior.
+            s->hdr->ref_seq_id    = -1;
             s->hdr->ref_seq_start = 0;
             s->hdr->ref_seq_span  = 0;
         } else {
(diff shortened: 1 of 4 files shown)


jkbonfield:
Hmm, maybe we shouldn't fix the output to match the spec.  That sounds radical and in theory the fix should happen, but doing so also means more files which trigger the bug in #1387 with older htslibs / samtools.  The consequence of not writing span=0 is, as far as I can see, inconsequential (given no one spotted it until now).

No right answer here.  Thoughts?

jkbonfield:
After discussion today at our weekly Samtools meeting and also the GA4GH File Formats meeting we took a pragmatic middle ground.

- Htslib will now correctly do range queries on unmapped data regardless of span contents (also further improved on previous fix).
- Writing CRAM 3.0 will continue to output span=1, to ensure the best compatibility with legacy software.  Failure to do so would mean a much wider proliferation of CRAM files that the legacy samtools versions couldn't query properly.
- Writing CRAM 3.1 onwards will output span=0, as these versions are not in regular use.
- The CRAM specification will be amended to state the start and span should not be used when decoding unmapped data, that the recommended values are (still) start=0 and span=0 for such data, but that other values are permitted.

This spec relaxation is a cheat so that this PR and all previous htslib's are now technically spec compliant, so we don't feel compelled to be purists and fix the software and in the process harm users of earlier versions.

We did ponder simply stating that span=1 should be the recommended (not required) value for unmapped data, but it feels a bit unfair on implementations that did the correct thing.  I'd have been more inclined if it was an issue with regular index querying, but explicitly fetching unmapped only is a less common operation and the workaround of updating the library is not an insurmountable hill to climb.

cram/cram_encode.c
```diff
@@ -1923,8 +1930,13 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
     }
 
     c->ref_seq_id    = c->slices[0]->hdr->ref_seq_id;
-    c->ref_seq_start = c->slices[0]->hdr->ref_seq_start;
-    c->ref_seq_span  = c->slices[0]->hdr->ref_seq_span;
+    if (c->ref_seq_id == -1 && CRAM_ge31(fd->version)) {
```
daviesrob, line 1933: It might be a good idea to have a short comment here, explaining why there's a check on the file version (and also in `cram_update_current_slice`).

jkbonfield: Could do, but it'd probably be best to comment it to simply see the relevant commit message as it'd be unwieldy to document the full reasoning.  I'll try and think of something appropriate.

Issue #1065, opened 2020-05-13
bcf_hdr_parse_line() and bcf_hrec_add_key() in vcf.c do not check for INFO and FORMAT keys matching VCF specifications


freeseek:
If I attempt to add a header string to a VCF header using `bcf_hdr_append()`, the string gets passed to the function `bcf_hdr_parse_line()` which parses the key and passes it to `bcf_hrec_add_key()`. However, in none of these steps the key is checked for consistency with the VCF specification which says that `INFO  keys  must  match  the  regular  expression^([A-Za-z][0-9A-Za-z.]*|1000G)$, please  note  that  “1000G”  is  allowed  as  a  special  legacy  value.`

Ideally, `bcf_hdr_parse_line()` is the place where this check should take place. There is some code that seems like it should do that in `bcf_hdr_parse_line()`:
```
        // ^[A-Za-z_][0-9A-Za-z_.]*$
        if (p==q && *q && (isalpha_c(*q) || *q=='_'))
        {
            q++;
            while ( *q && (isalnum_c(*q) || *q=='_' || *q=='.') ) q++;
        }
```
However, this code gets applied to what is left of the key (i.e., the `ID` part). So if I was to invoke:
```
        bcf_hdr_append(hdr, "##INFO=<ID=XXX:1");
```
The above code would be applied to the `ID` part of the string which would be valid and indeed the code would add this line to the header and the INFO field in the VCF dictionary structure. This does not seem like the intended behavior.

freeseek:
Here is code to recreate the issue. Create a minimal `main.c` file:
```
$ echo '#include <htslib/vcf.h>
#include <htslib/hts.h>

int main(int argc, char *argv[]) {
  htsFile *fh = hts_open("-", "wv");
  if ( !fh ) {
    fprintf(stderr, "Unable to open output VCF file\n");
    exit(-1);
  }
  bcf_hdr_t *hdr = bcf_hdr_init("w");
  bcf_hdr_append(hdr, "##INFO=<ID=XXX:1,Number=1,Type=Integer");
  if (bcf_hdr_write(fh, hdr) < 0) {
    fprintf(stderr, "Unable to write to output VCF file\n");
    exit(-1);
  }
  bcf_hdr_destroy(hdr);
  hts_close(fh);
  return 0;
}' > main.c
```

Compile using the current version of HTSlib:
```
$ HTSDIR="..."
[...]

PR #1389, opened 2022-01-26
Check for incorrect contig names and INFO/FORMAT tags

pd3:
This commit will only result in warnings printed on stderr. This
should be improved to propagate the error all the way up to the
caller and let it decide whether to throw an error or proceed despite
the error.

See
https://github.com/samtools/htslib/issues/1065
https://github.com/samtools/samtools/issues/1573

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -1,7 +1,7 @@
 /*  vcf.c -- VCF/BCF API functions.
 
     Copyright (C) 2012, 2013 Broad Institute.
-    Copyright (C) 2012-2021 Genome Research Ltd.
+    Copyright (C) 2012-2022 Genome Research Ltd.
     Portions copyright (C) 2014 Intel Corporation.
 
     Author: Heng Li <lh3@sanger.ac.uk>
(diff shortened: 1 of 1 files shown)


daviesrob:
Yes, I think I'd prefer pre-computed tables.  I find perl is handy for generating them, e.g.:
```
perl -le '@v = (split(//, q[!#$%&*+./:;=?@^_|~-]), "a"..."z", "A"..."Z", "0"..."9"); @a = (0) x 256; foreach $c (@v) { $a[ord($c)] = 1; } print join(", ", @a)' | fold -w 48
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
0, 1, 0, 1, 1, 1, 1, 0, 0, 0, 1, 1, 0, 1, 1, 1, 
1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 0, 1, 
1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 
1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 1, 1, 
0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 
1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 0, 1, 0, 
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
```

It might also be a good idea to put it in `hts.c` so it can be used for SAM if we ever get around to putting the same validation there.  I would keep it private to the library though (so no `HTSLIB_EXPORT`).

Note that the look-up tables need to be size 256, not 255.

vcf.c
```diff
@@ -376,6 +376,96 @@ int bcf_hrec_find_key(bcf_hrec_t *hrec, const char *key)
...
+static int bcf_hrec_set_type(bcf_hrec_t *hrec)
+{
+    if ( !strcmp(hrec->key, "contig") ) hrec->type = BCF_HL_CTG;
+    else if ( !strcmp(hrec->key, "INFO") ) hrec->type = BCF_HL_INFO;
+    else if ( !strcmp(hrec->key, "FILTER") ) hrec->type = BCF_HL_FLT;
+    else if ( !strcmp(hrec->key, "FORMAT") ) hrec->type = BCF_HL_FMT;
+    else if ( hrec->nkeys>0 ) hrec->type = BCF_HL_STR;
+    else hrec->type = BCF_HL_GEN;
+    return 0;
+}
```
jkbonfield, line 388: It probably doesn't matter at all as there aren't many fields to check, but we should probably test the performance impact of this test.

For large if/else-if strcmp lists I tend to use a switch on the first char.  Eg https://github.com/samtools/htslib/blob/develop/sam.c#L1149 (although I see I ought to do the same for the flag name bit).

Ie benchmark it first - likely no change needed.

vcf.c
```diff
@@ -376,6 +376,96 @@ int bcf_hrec_find_key(bcf_hrec_t *hrec, const char *key)
...
+
+    Currently the return status is not checked by the caller
+    and only a warning is printed on stderr. This should be improved
+    to propagate the error all the way up to the caller and let it
+    decide what to do: throw an error or proceed anyway.
+ */
+static int bcf_hrec_check(bcf_hrec_t *hrec)
+{
+    int i;
+    static uint8_t valid_ctg[255], valid_tag[255], valid_set = 0;
```
jkbonfield, line 403: As John / Rob points out - issues with this code.

vcf.c
```diff
@@ -376,6 +376,96 @@ int bcf_hrec_find_key(bcf_hrec_t *hrec, const char *key)
...
+        if ( val[0]=='*' || val[0]=='=' || !valid_ctg[(int)val[0]] ) goto err_invalid_ctg;
+        while ( *(++val) )
+            if ( !valid_ctg[(int)*val] ) goto err_invalid_ctg;
+        return 0;
+    }
+    if ( hrec->type==BCF_HL_INFO )
+    {
+        i = bcf_hrec_find_key(hrec,"ID");
+        if ( i<0 ) goto err_missing_id;
+        char *val = hrec->vals[i];
```
jkbonfield, line 437: This needs to be unsigned char, to avoid underflow of the valid_ctg array.  Similarly all the other lookups to this array.

Note casting it to int may remove the compiler warning, but it won't solve the underlying issue.

vcf.c
```diff
@@ -376,6 +376,96 @@ int bcf_hrec_find_key(bcf_hrec_t *hrec, const char *key)
     return -1;
 }
 
+static int bcf_hrec_set_type(bcf_hrec_t *hrec)
```
jkbonfield, line 379: This may as well be void, as the return isn't used nor can it ever change.

vcf.c
```diff
@@ -483,7 +575,16 @@ bcf_hrec_t *bcf_hdr_parse_line(const bcf_hdr_t *h, const char *line, int *len)
...
-        if ( *q=='>' ) { nopen--; q++; }
+        if ( *q=='>' )
+        {
+            if (nopen) nopen--;     // this can happen with nested angle brackets <>
+            q++;
+        }
+    }
+    if ( nopen && warn_unfinished_line )
+    {
+        warn_unfinished_line = 0;
```
jkbonfield, line 586: Part of me feels the guard for whether we've warned once already isn't necessary.

If I have dozens of invalid lines in my file, it's not really very helpful to get a fault reported, fix it, run the software and get the next line reported, fix, repeat...

Besides, frankly if people have hundreds and hundreds of badly formatted headers then beating them around the head repeatedly for each and every one may be considered to be of benefit to the community. :-)

jkbonfield (commented):
Note, this is just reading on github.  I haven't checked it out to build & test.

vcf.c
pd3: This is reading a VCF header. Note that this code is there elsewhere already, it just has been refactored here. This not a performance concern.

vcf.c
```diff
@@ -376,6 +376,126 @@ int bcf_hrec_find_key(bcf_hrec_t *hrec, const char *key)
...
+/**
+    The arrays were generated with
+
+    valid_ctg:
+        perl -le '@v = (split(//,q[!#$%&*+./:;=?@^_|~-]),"a"..."z","A"..."Z","0"..."9"); @a = (0) x 256; foreach $c (@v) { $a[ord($c)] = 1; } print join(", ",@a)' | fold -w 48
+
+    valid_tag:
+        perl -le '@v = (split(//,q[_.]),"a"..."z","A"..."Z","0"..."9"); @a = (0) x 256; foreach $c (@v) { $a[ord($c)] = 1; } print join(", ",@a)' | fold -w 48
+*/
+static uint8_t valid_ctg[256] =
```
jmarshall, line 399:
```suggestion
static const uint8_t valid_ctg[256] =
```

jmarshall: The other table needs the same tweak, and there's no need for a credit for such a minor change. I'd appreciate the `Co-authored-by` tag being removed when these commits are further edited (e.g. on merge) — if this _had_ been co-authored by me, there would be one table with four bits used, for CONTIG_FIRST_CHAR/CONTIG_CHAR/TAG_FIRST_CHAR/TAG_CHAR :smile:

daviesrob:
Squashed and rebased so that the tests will all work.  I've also taken the liberty of making the other table const as well.

daviesrob:
Now merged, leaving for later the part about exactly how non-compliant identifiers should be dealt with.  One possibility might be to turn on strict mode only for VCF 4.3, given that the earlier versions didn't specify any restrictions on the names.

pd3:
> [...] turn on strict mode only for VCF 4.3, given that the earlier versions didn't specify any restrictions on the names.

I wouldn't bother, the lack of specification in previous versions was not supposed to mean that anything is allowed.

PR #1396, opened 2022-02-14
Fix hts_parse_decimal and make synced BCF targets parsing more robust

pd3:
The function bcf_sr_regions_init() autodetects the format of input regions,
trying to decide between CHR,POS vs CHR,BEG,END formats. However, the function
hts_parse_decimal() was recognising standalone `G` as `0` which resulted in
`chr1 1234 G` being interpretted as `chr1 1234 0`.

This commit makes the digit parsing more strict, requiring at least one digit
to appear before the exponent or unit part.

See also samtools/bcftools#1598

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -3482,7 +3482,7 @@ static inline long long push_digit(long long i, char c)
 long long hts_parse_decimal(const char *str, char **strend, int flags)
 {
     long long n = 0;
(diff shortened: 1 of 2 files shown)


pd3:
Whops, did not pass htslib tests...

PR #1400, opened 2022-02-15
Check number syntax is valid in hts_parse_decimal()

### LINKED ISSUE #1395: hts_parse_decimal() silently fails on non-numeric strings starting with G
The function does not reject strings like `G A`, thinking it was able to parse a coordinate due to https://github.com/samtools/htslib/blob/1c8829a9a189004c6229bab923753aa5569cff0c/hts.c#L3513

This leads to problems reported in https://github.com/samtools/bcftools/issues/1598 where region format is autodetected by first attempting to read `chr beg end` and, upon failure, reads `chr pos`. One relevant part of code that can be mislead by this is https://github.com/samtools/htslib/blob/1c8829a9a189004c6229bab923753aa5569cff0c/synced_bcf_reader.c#L1126-L1130

jmarshall:
Motivated by #1395, this alters `hts_parse_decimal()` to validate that the number parsed is indeed a validly presented numeric value and to have a way of reporting failure. What we want to accept and the returned `strend` value are carefully considered, and test cases are added accordingly.

Follows `strtol()` et al's conventions, setting `strend` to the original `str` (or emitting a warning if `strend` is NULL) if the input string does not form a valid numeric constant. In particular, require it to have a non-zero number of digits.

Merges E and k/M/G handling, so that at most one can be used. In particular, "1e+3k" is non-standard and ambiguous, and should be rejected.

Adds documentation to _htslib/hts.h_ describing the numeric syntax parsed and the two modes of operation (as determined by whether `strend` is NULL or not).

Add test cases, in particular verifying that a sole "G" is rejected. Fixes #1395. For simplicity, the tests are added to _test/sam.c_ which already contains similar API tests. If preferred, I could create a new _test/hts.c_ containing these tests instead.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -3482,39 +3482,32 @@ static inline long long push_digit(long long i, char c)
 long long hts_parse_decimal(const char *str, char **strend, int flags)
 {
     long long n = 0;
-    int decimals = 0, e = 0, lost = 0, has_digit = 0;
+    int digits = 0, decimals = 0, e = 0, lost = 0;
     char sign = '+', esign = '+';
-    const char *s;
+    const char *s, *str_orig = str;
 
     while (isspace_c(*str)) str++;
     s = str;
 
     if (*s == '+' || *s == '-') sign = *s++;
     while (*s)
-        if (isdigit_c(*s)) n = push_digit(n, *s++), has_digit = 1;
+        if (isdigit_c(*s)) digits++, n = push_digit(n, *s++);
         else if (*s == ',' && (flags & HTS_PARSE_THOUSANDS_SEP)) s++;
         else break;
 
     if (*s == '.') {
         s++;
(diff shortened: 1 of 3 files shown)


jmarshall:
Rebased past PR #1396, which has since been merged.

IMHO this PR further improves `hts_parse_decimal()` over current **develop**, by carefully considering how the returned `strend` value can be used by the caller to detect failure and by considering the diagnostics that should be emitted in the `strend == NULL` mode of operation. In particular, current **develop**'s early return for the `!has_digit` case:

* does not emit a diagnostic (in `strend == NULL` mode) when no valid number has been parsed, which would be the user's only way of noticing the problem in this mode.
* does not return a `strend` value that can easily be used to detect parse failure when the string begins with whitespace (as `str` has already been incremented past the whitespace, so the returned `*strend` is not equal to the caller's `str`).

PR #1407, opened 2022-02-23
Use a constant printf format string in `hts_feature_string()` [minor]

jmarshall:
Refactor this code to avoid a `-Wformat-nonliteral` warning. Since PR #1406 was applied, this is the only instance of this warning produced by the HTSlib code base.

With this false positive corrected, it would be possible to turn `-Wformat=2 -Werror` on in a CI build.
(For GCC builds only; Clang's equivalent option has false positives as it does not suppress the warning for `va_list`-using functions.)

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -168,18 +168,12 @@ const char *hts_test_feature(unsigned int id) {
 // to find the configuration parameters.
 const char *hts_feature_string(void) {
     static char config[1200];
-    const char *fmt=
+    const char *flags=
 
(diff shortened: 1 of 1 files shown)


daviesrob:
Looks like a good idea, as it will allow us to eliminate an entire class of bug.  The only side-effect is it changes the order of items in the feature string - hopefully no-one relies on that.

I'll merge this and look at adding a `-Wformat=2` to the test harness in a separate PR.

PR #1410, opened 2022-02-25
Changes a data type in sam.h to avoid compiler warnings

ubuntolog:
While compiling on Windows 10, we are getting multiple warnings from a single method.
image

This PR solves this issue by changing a data type for sub_type_size variable. It looks like the problem has something to do with a specific version of a compiler. We could not reproduce it on a Linux machine.

### DIFF
--- a/htslib/sam.h
+++ b/htslib/sam.h
@@ -1524,7 +1524,7 @@ static inline const uint8_t *sam_format_aux1(const uint8_t *key,
         ++s;
     } else if (type == 'B') {
         uint8_t sub_type = *(s++);
-        int sub_type_size;
+        unsigned sub_type_size;
 
         // or externalise sam.c's aux_type2size function?
         switch (sub_type) {
@@ -1547,7 +1547,7 @@ static inline const uint8_t *sam_format_aux1(const uint8_t *key,
             goto bad_aux;
         n = le_to_u32(s);
         s += 4; // now points to the start of the array
-        if ((end - s) / sub_type_size < n)
+        if ((size_t)(end - s) / sub_type_size < n)
             goto bad_aux;
         r |= kputsn_("B:", 2, ks) < 0;
         r |= kputc(sub_type, ks) < 0; // write the type


daviesrob:
The solution looks OK, but I'm surprised it makes a difference, given that `end - s` will be a `ptrdiff_t`, and so already signed.  Presumably your compiler failed to stop that `sub_type_size` can only be 1, 2 or 4 when it gets to the offending line, but is happy that `end - s` is positive.  Out of interest, exactly which compiler was it, and what optimisation options did you use?

jmarshall:
I suspect the platform problem here is that `ptrdiff_t` on this platform is surprisingly narrow — only 32 bits at most. If you hackily change `n`'s declaration to `uint64_t`:

```diff
diff --git a/htslib/sam.h b/htslib/sam.h
@@ -1545 +1545 @@
-        uint32_t i, n;
+        uint64_t i, n;
```

you can reproduce the `gcc -Wsign-compare` warning on Linux, and experiment easily with fixes.

I think I would suggest changing `sub_type_size` to `unsigned` and tell the compiler that `(end - s)` is non-negative, either by casting to `size_t` or by introducing a `size_t` temporary variable to assign it to.

daviesrob:
Ah, yes, so it could be 32-bit Windows?

ubuntolog:
> Ah, yes, so it could be 32-bit Windows?

Yes, for historical reasons we still use a 32-bit version of our app. We are planing to switch to 64 soon. Thank you for such a quick reply

jmarshall:
BTW buried in the screenshot is the information that this is included from a file named _VariantList.cpp_ and hence that the warning is coming from a C++ compiler. (Though C can be coerced into producing a similar warning. So fixes ought to be tested in both C and C++ compilers.)

ubuntolog:
> The solution looks OK, but I'm surprised it makes a difference, given that `end - s` will be a `ptrdiff_t`, and so already signed. Presumably your compiler failed to stop that `sub_type_size` can only be 1, 2 or 4 when it gets to the offending line, but is happy that `end - s` is positive. Out of interest, exactly which compiler was it, and what optimisation options did you use?

We are using MinGW on Windows. This is the information about the compiler:
image

I have made some adjustments in the PR: changed the data type back and used casting

daviesrob:
That was a C++ style cast you added.  The C way to do it would be `(size_t) (end - s)`

daviesrob:
Also, I think you probably still want to change `sub_type_size` to `uint32_t` to get the effect you want.

ubuntolog:
> Also, I think you probably still want to change `sub_type_size` to `uint32_t` to get the effect you want.

Thank you for the help. I do not have a lot of experience with C.

jmarshall:
Also remove the added `#include`, which was a red herring. IMHO changing `sub_type_size` to just plain `unsigned` would be slightly better — there's no need for a specific size of integer here.

ubuntolog:
On Windows we use
image

After the changes I have made there are no more warnings

jmarshall:
> After the changes I have made there are no more warnings

If you added `-Wconversion` to your MinGW GCC flags, I think you would see another warning. Might as well squash it too while you're at it :smile:

ubuntolog:
> > After the changes I have made there are no more warnings
> 
> ~If you added `-Wconversion` to your MinGW GCC flags, I think you would see another warning.~ Might as well squash it too while you're at it.
> 
> Hmmm… actually this `-Wconversion` warning appears with `gcc` but not `g++`. Well, someone should do it for the C programmers smile

I will try to fix the warnings

ubuntolog:
> > > After the changes I have made there are no more warnings
> > 
> > 
> > ~If you added `-Wconversion` to your MinGW GCC flags, I think you would see another warning.~ Might as well squash it too while you're at it.
> > Hmmm… actually this `-Wconversion` warning appears with `gcc` but not `g++`. Well, someone should do it for the C programmers smile
> 
> I will try to fix the warnings

I changed the type of **sub_type_size** from **int** to **uint32_t**. The warning is gone. However, you did not want to change the type

jmarshall:
> I changed the type of sub_type_size from int to uint32_t. The warning is gone. However, you did not want to change the type

Me, I did want to change the type of `sub_type_size`, but from `int` to `unsigned` — which is slightly more general. 🤷

jmarshall:
FWIW :+1: looks perfect to me — apologies for all the quibbling!

ubuntolog:
> FWIW +1 looks perfect to me — apologies for all the quibbling!

No problem! Thank you for your feedback. Now I know a bit more about C :)

daviesrob:
Thanks, looks OK now, and I've confirmed that the issue existed and was fixed by this on a 32-bit platform.

Issue #1397, opened 2022-02-14
hts_log_error() extension for better error reporting


lczech:
Hi there!

This is more of a general suggestion to improve error reporting, but I'll illustrate it with a concrete example.

I have started using the sam/bam/cram functionality recently, and got an (expected) error "The input is not sorted" on an unsorted bam file during testing. This error is reported in `bam_plp_push()` [here](https://github.com/samtools/htslib/blob/ab6065cd02a7cb4f104f2d20f25792ebc4c76406/sam.c#L5705-L5714), but not propagated from there to the original caller of the function - the error status gets lost.

My code calls `bam_plp_auto()`, and checks the result for null to detect whether the current position is usable (I gathered that this is the way this is to be used). That function returns a `bam_plp_s*` (typedef'ed as `bam_plp_t`), which however is only declared in the `sam.h` header, and then defined in `sam.c` [here](https://github.com/samtools/htslib/blob/ab6065cd02a7cb4f104f2d20f25792ebc4c76406/sam.c#L5295-L5313). Hence, the `bam_plp_t` field `error`, which is correctly set by `bam_plp_push()`, is not accessible from my code to check whether an error occurred.

Furthermore, `hts_log()` [here](https://github.com/samtools/htslib/blob/4469ab18377e7fc074c0e38e8aba0f4e886daaaf/hts.c#L4752) prints the error directly to stderr, and hence further complicates for me to catch this and report any errors (in my case, via throwing an exception in C++).

**It seems hence impossible for me to detect in my code whether a file has simply ended or has an error during iteration.**

This all is of course only an example, and seems applicable to many other functions as well. Am I overlooking a simple solution here? What would be a good way of solving this?

One idea that might improve the situation is to offer to set a callback function that gets called in `hts_log()` with the `htsLogLevel` and the final error message. Users could then set this callback, and react in their own code to the error message, for example by throwing an exception (or whatever C users do in that situation :-D).

Thank you in advance and all the best
Lucas

whitwham:
Let's start with your concrete example.

The two occasions where `bam_plp_auto` is used in samtools the error detection is done by looking at the last argument.  If it is < 0 then an error has occurred.  This is not documented anywhere.  As my colleague said in answer to your other question (#1398); our documentation is poor in many places and very poor for this set of functions.

For the iterator we could add a couple of accessor functions for the `eof` of and `error` values of `bam_plp_t`.  They might be helpful.

[...]

PR #1414, opened 2022-03-23
Proposal to add hts_log callback function and hts_log_set_logger()

jmarshall:
A somewhat tidied-up version of @lczech's concrete proposal in #1397 for overriding the default printing of hts_log messages to stderr. Draft, as the new function (and associated typedef) would require documenting in _hts_log.h_.

I don't know whether the `hts_log_*` functions are called from threads that HTSlib has started. If so, this callback function would be called from those threads — which might be “interesting”. This PR doesn't consider what to do in this case.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -4754,19 +4754,32 @@ static char get_severity_tag(enum htsLogLevel severity)
     return '*';
 }
 
+static hts_log_func *hts_logger = NULL;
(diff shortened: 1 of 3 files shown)


jmarshall:
@lczech or the maintainers should feel free to do so, and may find this PR (or just the commit) a useful starting point.

PR #1419, opened 2022-04-05
Minor test/test.pl improvements

jmarshall:
Prevent a very minor Perl warning that I noticed in passing:

```
…
cd test && REF_PATH=: ./test.pl ${TEST_OPTS:-}
"my" variable @p masks earlier declaration in same scope at ./test.pl line 617.
…
```

While checking differences between reruns I also noticed that the `test-bcf-sr` test cases select random numbers based partly on the current wall time (and without calling `srand()` first):

```perl
sub test_bcf_sr_sort
{
    my ($opts, %args) = @_;
    for (my $i=0; $i<10; $i++)
    {
        my $seed = int(rand(time));
        my $test = 'test-bcf-sr';
        my $cmd  = "$$opts{path}/test-bcf-sr.pl -t $$opts{tmp} -s $seed";
        # …
```

I didn't look at what that `-s` argument is used for, but this does theoretically mean that you could see a test failure and then be unable to reproduce it by rerunning the test suite.

### DIFF
--- a/test/test.pl
+++ b/test/test.pl
@@ -32,6 +32,7 @@ use File::Temp qw/ tempfile tempdir /;
 use IO::Handle;
 
 my $opts = parse_params();
+srand($$opts{seed});
 
 test_bgzip($opts, 0);
 test_bgzip($opts, 4);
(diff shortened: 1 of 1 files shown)


jmarshall:
_test-bcf-sr.pl_ uses its `-s` option to seed its own `srand`/`rand` random number generator.

Added another commit to use a repeatable sequence of seed values in the ten randomly-generated `test-bcf-sr` test cases. (Currently this `test_bcf_sr_sort()` is the only user of `rand()` in _test/test.pl_.)

daviesrob:
Looks good, thanks.

PR #1423, opened 2022-04-12
Permit MM tags containing "." and "?" suffixes.

### LINKED ISSUE #1418: Support optional `.` and `?` in Mm tag
@jts proposed changes to the Mm/Ml tag specification which allow to explicitely record what can be assumed about bases for which no entry in the tag is made. These changes are now in the SAMtags specification, but not supported by the tag parsing code in htslib.

A check needs to be performed around here:
https://github.com/samtools/htslib/blob/9bcb2d2e3c0a84b1ba6c0e7cc89a4e82c055f9fa/sam.c#L6175

to see if a `.` or `?` is present and (minimally) skip to the next character. Currently a file containing this optional information will cause the parser to trip up eventually emitting the error.

```"Insufficient number of entries in ML tag"```

### PR DESCRIPTION
**Pull request by jkbonfield**
These define explicit vs implicit coordinates.  They are now part of the MM specification, but we don't do anything with this data yet.  This PR simply permits them to be parsed without choking, and ignores the additional markup.  A subsequent PR will improve on this.

Fixes #1418

### CODE REVIEW

File: sam.c [hunk 1]
```diff
@@ -6181,13 +6181,27 @@ int bam_parse_basemod(const bam1_t *b, hts_base_mod_state *state) {
                 cp = cp_end;
                 ms = cp-1;
             } else {
-                while (*cp && *cp != ',' && *cp != ';')
+                while (*cp && isalpha(*cp))
```
jmarshall, line 6184:
```suggestion
                while (*cp && isalpha_c(*cp))
```

And the same applies to the `isdigit()` a few lines above.

PR #1435, opened 2022-05-13
Fix clang builds under mingw.

### LINKED ISSUE #1433: Compiling with clang on Windows does not work
Hi!

I was trying to build htslib on Windows (MSYS2, clang 14.0.0) using clang but it does not work at the moment.

Compilation produces several errors regarding dllexport:

`hfile.c:105:8: warning: redeclaration of 'hfile_init' should not add 'dllexport' attribute [-Wdll-attribute-on-redeclaration]
hFILE *hfile_init(size_t struct_size, const char *mode, size_t capacity)
       ^

./hfile_internal.h:98:8: note: previous declaration is here
hFILE *hfile_init(size_t struct_size, const char *mode, size_t capacity);
       ^

hfile.c:153:6: error: redeclaration of 'hfile_destroy' cannot add 'dllexport' attribute
void hfile_destroy(hFILE *fp)
     ^

./hfile_internal.h:110:6: note: previous declaration is here
void hfile_destroy(hFILE *fp);
     ^

hfile.c:203:5: error: redeclaration of 'hfile_set_blksize' cannot add 'dllexport' attribute
int hfile_set_blksize(hFILE *fp, size_t bufsiz) {
    ^

./hfile_internal.h:52:5: note: previous declaration is here
int hfile_set_blksize(hFILE *fp, size_t bufsiz);`

And so on. Any ideas how to fix this?

jkbonfield:
Under mingw clang requires dllexport to be applied to both function
declarations and function definitions.  (Gcc is happy for the
definition only to have the dllexport attribute.)

Note this exports several "internal" functions, including some
apparently unused anywhere (hts_json_*).   Perhaps they were once used
in htslib-plugins, but no more.

I haven't taken the decision to remove these though, but it's worth
considering for whenever we next do an ABI breaking change.  Either
that or stop pretending that their internal only when clearly they are
not, and move their API to the external htslib/*.h files instead.
These appear to be somewhat in Limbo right now.

Fixes #1433

### DIFF
--- a/hfile_internal.h
+++ b/hfile_internal.h
@@ -90,11 +90,13 @@ struct hFILE_backend {
 
 /* May be called by hopen_*() functions to decode a fopen()-style mode into
    open(2)-style flags.  */
+HTSLIB_EXPORT
 int hfile_oflags(const char *mode);
 
 /* Must be called by hopen_*() functions to allocate the hFILE struct and set
    up its base.  Capacity is a suggested buffer size (e.g., via fstat(2))
    or 0 for a default-sized buffer.  */
+HTSLIB_EXPORT
 hFILE *hfile_init(size_t struct_size, const char *mode, size_t capacity);
 
 /* Alternative to hfile_init() for in-memory backends for which the base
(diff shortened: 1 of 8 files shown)


jkbonfield:
Note it fails the tests using gcc-12.  I think these will be fixed by https://github.com/samtools/htscodecs/pull/51, but haven't verified it.  (Or alternatively use -O3 which is preferred anyway for htscodecs.)

PR #1441, opened 2022-05-27
Add an "sclen" expression keyword.

### LINKED ISSUE #1436: [FEATURE REQUEST] Add input-fmt-option to show number of match bases?
The document is a little confusion for `qlen` and `rlen`. 

- qlen: no. query bases
- rlen: no. reference bases

It took me some time to realize that qlen includes match bases and soft-clipped bases, while  rlen includes the intron skip. So, there is  no parameters recording the number of matched bases, **exclude** the soft-clipped and intron-skip region.  It is possible to have another parameter for "number of matched bases", say `mlen`?

Thanks!

jkbonfield:
This is the length of soft-clips, both left and right end.
It may be combined with qlen (qlen-sclen) to obtain the number of
bases in the query sequence that have been aligned to the genome.  Ie
it provides a way to compare local-alignment vs global-alignment
length.

Fixes #1436


daviesrob:
Looks OK at a first glance.  Could you add something to `test/sam_filter` to exercise it a bit, please?

PR #1456, opened 2022-06-16
Fix reverse breakend detection and test

### LINKED ISSUE #1455: bcf_set_variant_type does not correct set type BND for reverse strand breakends
https://github.com/samtools/htslib/blob/58d9f604a0b9de5be021c9e9a79d1dc5d5892503/vcf.c#L4182-L4199

A forward strand breakend like:
REF N
ALT N]16:33625444]
will be correctly detected as VCF_BND.

A reverse strand breakend like: 
REF N
ALT [16:33531168[N
will be incorrectly detected as VCF_INDEL.

This is because it should be working right to left for the reverse strand INDEL.

mp15:
fixes #1455


daviesrob:
I think including `vcf.c` in your test didn't go down too well.  You might need to include a few more headers to get it to work properly.

mp15:
Fixed it, it caught me by surprise that the order you include <config.h> and <string.h> matters. I assume config.h is defining something to bring POSIX functions in whilst in pedantic mode.

daviesrob:
It looks like this should fix the problem, although I'm not sure if I agree with the terminology in the comments.  Shouldn't the relationships be "joined before" and "joined after" (as suggested by the [specification](https://samtools.github.io/hts-specs/VCFv4.3.pdf#subsection.5.4))?

mp15:
Done. Also reworked the test so that it covers most of the bcf_set_variant_type function.

daviesrob:
Thanks.  I've taken the liberty of pushing a couple of minor adjustments.  I'll squash them on merge.

test/test.pl
```diff
@@ -1055,3 +1056,17 @@ sub test_realn {
...
+    my $test = 'test-bcf_set_variant_type';
+    my $cmd  = "$$opts{path}/test-bcf_set_variant_type";
+    print "$test:\n";
+    print "\t$cmd\n";
+    my ($ret,$out) = _cmd($cmd);
+    if ( $ret ) {
+        print $out;
+        failed($opts,$test);
+    } else { passed($opts,$test); }
+}
```
jmarshall, line 1072: Missing newline at EOF.

daviesrob: Now fixed...

PR #1457, opened 2022-06-17
Set tab delimiter in manpage for tabix GFF3 sort

cmdcolin:
This can help if there are spaces in the GFF3 file e.g. in column 2 or 3. Was found in the wild by my co-worker on a gff file here https://ftp.ncbi.nlm.nih.gov/genomes/all/GCF/000/002/285/GCF_000002285.3_CanFam3.1/GCF_000002285.3_CanFam3.1_genomic.gff.gz

Another alternative but similar command is

```
awk '$1 ~ /^#/ {print $0;next} {print $0 | "sort -t\"\t\" -k1,1 -k4,4n"}' file.gff > file.sorted.gff

```

But, the PR here keeps it as is

### DIFF
--- a/tabix.1
+++ b/tabix.1
@@ -169,7 +169,7 @@ The default is 3, which turns on error and warning messages;
 Values higher than 3 produce additional informational and debugging messages.
 .PP
 .SH EXAMPLE
-(grep ^"#" in.gff; grep -v ^"#" in.gff | sort -k1,1 -k4,4n) | bgzip > sorted.gff.gz;
+(grep ^"#" in.gff; grep -v ^"#" in.gff | sort -t$'\t' -k1,1 -k4,4n) | bgzip > sorted.gff.gz;
 
 tabix -p gff sorted.gff.gz;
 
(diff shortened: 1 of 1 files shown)


daviesrob:
This is a good idea, but `$'\t'` to get the tab is a bash-ism.  ```"`printf '\t'`"``` is more portable, and even works in csh.

I quite like the `awk` version as it only makes one pass through the file, although sadly it seems to be slightly slower than making two passes with `grep`.  Presumably that's because `awk` is doing more work splitting the input up.

cmdcolin:
as long as we are looking at compatibility, the current one actually has issue in zsh (needed the ^ to be inside of quotes I think) so updated again :)

also incorporated the printf \t! good catch

cmdcolin:
also interesting that the awk is a bit slower. also, if could be helpful to make typing out the command easier. for that, can potentially offer people a bash command like:

```
function gffsort() {
  grep "^#" $1;
  grep -v "^#" $1 | sort -t"`printf '\t'`" -k1,1 -k4,4n;
}
```

then 

```
gffsort input.gff | bgzip> out.gff.gz
tabix -p gff out.gff.gz
```

daviesrob:
The update is mostly OK, apart from the `\t` got interpreted as a troff macro, and disappeared from the formatted output.  I'll push a tiny fix.

The slowness with `awk` is most likely due to it splitting the entire line into fields.  `grep` can get away with just looking at the first character, which is much faster.

PR #1458, opened 2022-06-20
Remove `RANLIB`

SoapGentoo:
* `ar` is guaranteed to generate the symbol table on POSIX:
  - https://stackoverflow.com/a/38244200
  - https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ar.html
  - We invoke `ar` with `-s` to force generation of the symbol table on OSes where this isn't the default behaviour (old BSD).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,6 @@
 
 CC     = gcc
 AR     = ar
-RANLIB = ranlib
 
 # Default libraries to link if configure is not used
 htslib_default_libs = -lz -lm -lbz2 -llzma -lcurl
@@ -319,8 +318,7 @@ plugins: $(BUILT_PLUGINS)
 
 libhts.a: $(LIBHTS_OBJS)
 	@-rm -f $@
-	$(AR) -rc $@ $(LIBHTS_OBJS)
-	-$(RANLIB) $@
+	$(AR) -rcs $@ $(LIBHTS_OBJS)
 
 print-config:
 	@echo LDFLAGS = $(LDFLAGS)
--- a/config.mk.in
+++ b/config.mk.in
@@ -38,7 +38,6 @@ datarootdir  = @datarootdir@
 mandir       = @mandir@
 
 CC     = @CC@
-RANLIB = @RANLIB@
 
 CPPFLAGS = @CPPFLAGS@
 CFLAGS   = @CFLAGS@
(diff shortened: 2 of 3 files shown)


jkbonfield:
Is there a specific reason why you need this change?  Granted everyone should be using a POSIX system these days, but certainly in the past there were a plethora of systems which didn't support `ar -s` (eg SGI Irix).  It's unlikely we have many, or indeed any, users still using systems that need ranlib, but equally so I think it's probably unlikely we have any users where ranlib doesn't work either.

SoapGentoo:
I didn't need the change, but it removes code that isn't needed anymore, makes `configure` a bit faster, and I strongly doubt anyone is using htslib on IRIX (given that it likely never had full C99 support). You can close the PR if you prefer.

SoapGentoo:
Also, [IRIX seems to support `ar -s`](http://polarhome.com/service/man/?qf=ar&tf=2&of=IRIX&sf=1)

jmarshall:
When these build rules were introduced in f455afd08c8e5613e678bcd363a036f3c32a2dbd, the specific commands used were modelled on best practice as reflected by GCC's makefiles. FWIW GCC still uses `AC_PROG_RANLIB` and `-$(RANLIB)`.

SoapGentoo:
Interestingly, [even FreeBSD stopped using ranlib](https://reviews.freebsd.org/D26044?id=75747).

jkbonfield:
I concede it's highly unlikely people are using a system that requires ranlib, but we cannot be certain.  The question therefore is whether it's more likely there is a system that doesn't have ranlib than one that requires it, vs the speed benefit.

I measured the speed change to configure and build using 8 threads and it was about 0.7s difference, or about 4%.  Without 8 threads that obviously drops enormously as the wall-clock time becomes more dominated by the compiler.  So the speed of compile reasoning is somewhat spurious.

Thank you for the PR, but I'm tempted therefore to decline it unless we can find evidence that there are systems that do not have ranlib installed and hence this is an improvement to portability.

jmarshall:
> systems that do not have ranlib installed and hence this is an improvement to portability.

Note that `AC_PROG_RANLIB` handles this scenario by setting `@RANLIB@` to `:` if no ranlib utility is found. (Though I guess it makes some difference for without-configure builds, though those are already unsupported on sufficiently exotic platforms.)

Issue #1475, opened 2022-07-20
0 coordinate overflows in VCF format


Rapsssito:
As stated here: https://github.com/samtools/bcftools/issues/1753. There is an overflow issue when treating a VCF file with zero positions. Here is an example:

### Unsorted VCF
```
MT	0	1915430843:1	N	]MT:16640]N	99	PASS	.
MT	0	3110098432:1	N	]MT:16638]N	99	PASS	.
MT	16638	3110098432:2	G	G[MT:0[	99	PASS	.
MT	16640	1915430843:2	G	G[MT:0[	99	PASS	.
```

#### Commands
```
bcftools sort test.vcf -o test.sorted.vcf
```

### Sorted VCF
```
MT	4294967296	3110098432:1	N	]MT:16638]N	99	PASS	.
MT	4294967296	1915430843:1	N	]MT:16640]N	99	PASS	.
MT	16638	3110098432:2	G	G[MT:0[	99	PASS	.
MT	16640	1915430843:2	G	G[MT:0[	99	PASS	.
```

Note how the `0` from `3110098432:1` has been wrongly changed to `4294967296`.

As described by @pd3 in https://github.com/samtools/bcftools/issues/1753#issuecomment-1190071361:
> This problem is not specific to `bcftools sort` but to a round trip through BCF via htslib. The format stores 0-based position in `hts_pos_t` type and 0 coordinate overflows. This needs to be fixed in htslib.

@daviesrob, could you check this out?

daviesrob:
Fixed by #1476

PR #1476, opened 2022-07-20
Support 0 coordinate in BCF

pd3:
The 0 coordinate is valid in VCF specification, but the round-trip
VCF -> BCF -> VCF turns MT:0 into MT:4294967296. Add a check to
detect this overflow.

See https://github.com/samtools/htslib/issues/1475 and https://github.com/samtools/bcftools/issues/1753

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -1413,6 +1413,7 @@ static inline int bcf_read1_core(BGZF *fp, bcf1_t *v)
     if (ks_resize(&v->indiv, indiv_len ? indiv_len : 1) != 0) return -2;
     v->rid  = le_to_i32(x + 8);
     v->pos  = le_to_u32(x + 12);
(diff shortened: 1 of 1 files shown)


daviesrob:
This need to be rebased on the current `develop` branch to get the tests to work properly.

The BCF specification [says](https://samtools.github.io/hts-specs/VCFv4.3.pdf#subsubsection.6.3.1) that `POS` should be an `int32_t`, so should we actually switch to `le_to_i32()` for this value?  It would put a rather bigger restriction on the longest length supported, but going beyond `INT_MAX` is likely to cause problems in other areas (notably `END` tags) so it might not be a problem in practice.

pd3:
I also think it's a result of lazy definition, there is no reason for negative coordinates, even on circular chromosomes. Forced pushed a code that is in sync with the latest develop.

daviesrob:
I've now checked that this fixed the problem reported on `bcftools sort`, so:
Fixes #1475
Fixes samtools/bcftools#1753

PR #1478, opened 2022-07-20
Improve filter expression language consistency

jkbonfield:
This tries to follow the principles of 3-state logic used in other systems, such as SQL.  See https://en.wikipedia.org/wiki/Null_(SQL)#Comparisons_with_NULL_and_the_three-valued_logic_(3VL)

There is some controversy there and a suggestion that 4-state logic is better, which we sort of already have with the "null-but-true" concept, which may impact below.

The primary changes here are:

- New `hts_expr_val_exists` and `hts_expr_val_undef` inline functions to check if a value exists and to clear it.  These simply logic elsewhere.  This fixes SAM filter expressions like `[AB] / [BC] > 1` where tags AB or BC don't exist.  Similar updates have been applied to many logic and mathematical operators.  To avoid spammage these are silent, failing the expression without complaint due to undefined values.  Additionally `hts_exp_val_existsT` for conditionals (&&, ||) is true if the data exists or it is true (so null-but-true works).
- `hts_expr_val_undef` now uses NAN as a definition of undefined, rather than purely a false empty string.  I do this because we automatically gain some consistency with how NAN works on mathematical operators which we previously didn't have.  Eg `sqrt(-1) == sqrt(-1)` is false according to the strict comparison rules for NAN.
- In addition to this, <, >, <=, >=, == and != also fail for undefined values (due to the use of NAN).
- A new function `exists(var)` is used to test for the existance (whether it is defined) of a variable.  This returns a boolean, which removes the need for the null-but-true and zero-but-true nature used in current tag querying.  We still have this null-but-true concept as we've previously documented the difference between filter expression `[X0]` (has tag X0) and `[X0] && [X0]!=0` (has tag X0 with a non-zero value).  This still works, but is now much easier to write using `[X0] != 0` rather than it automatically including `!exists[X0] || [X0] != 0`.  This is therefore an incompatibility, but probably a sensible one.

  What we previously struggled to do was to check a tag does not exist.  It was possible via `!([X0] || [X1]==0)`, but that was highly cryptic and I doubt many figured it out!  Now `!exists([X0])` is a far simpler solution.  NOTE however the old cryptic method no longer works, as null || false is still null (not false) and !null is also null due to the afforementioned logic.  (I'm prepared to change this though; maybe we need null-and-false vs null-and-true so we can explicitly distinguish from null-is-unknown.)
  
  **INCOMPATIBILITY**:  Note previously "[X1] != 0" meant X1 is non-zero or X1 doesn't exist.  This now gets interpreted only as X1 is non-zero.  I believe this change is a considerable improvement and worth breaking compatibility over.
  
- A new function `default(var1,var2)` which returns the first value that is defined of the pair.  So `exists(var1)?var1:var2` (if we had a ?: operator that is).
- New mathematic operators pow, exp, sqrt, log. These perhaps should have been their own commit, but I added them during this work purely because they're a convenient mechanism for getting +/- inf and nan into things for validation testing.

If I use the old test_expr.c test harness on the new hts_expr.c code now also passes all tests.

Fixes https://github.com/samtools/samtools/issues/1677

### DIFF
--- a/hts_expr.c
+++ b/hts_expr.c
@@ -23,7 +23,6 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.  */
 
 // TODO:
-// - add maths functions.  pow, sqrt, log, ?
 // - ?: operator for conditionals?
 
 #define HTS_BUILDING_LIBRARY // Enables HTSLIB_EXPORT, see htslib/hts_defs.h
@@ -163,10 +162,52 @@ static int func_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
         }
         break;
 
+    case 'd':
+        if (strncmp(str, "default(", 8) == 0) {
+            if (expression(filt, data, fn, str+8, end, res)) return -1;
+            if (**end != ',')
+                return -1;
+            (*end)++;
+            hts_expr_val_t val = HTS_EXPR_VAL_INIT;
+            if (expression(filt, data, fn, ws(*end), end, &val)) return -1;
+            func_ok = 1;
+            if (!hts_expr_val_existsT(res)) {
+                kstring_t swap = res->s;
+                *res = val;
+                val.s = swap;
+                hts_expr_val_free(&val);
+            }
+        }
+        break;
+
+    case 'e':
+        if (strncmp(str, "exists(", 7) == 0) {
+            if (expression(filt, data, fn, str+7, end, res)) return -1;
+            func_ok = 1;
+            res->is_true = res->d = hts_expr_val_existsT(res);
+            res->is_str = 0;
+        } else if (strncmp(str, "exp(", 4) == 0) {
+            if (expression(filt, data, fn, str+4, end, res)) return -1;
+            func_ok = 1;
+            res->d = exp(res->d);
+            res->is_str = 0;
+            if (isnan(res->d))
+                hts_expr_val_undef(res);
+        }
+
+        break;
+
     case 'l':
         if (strncmp(str, "length(", 7) == 0) {
             if (expression(filt, data, fn, str+7, end, res)) return -1;
             func_ok = expr_func_length(res);
+        } else if (strncmp(str, "log(", 4) == 0) {
+            if (expression(filt, data, fn, str+4, end, res)) return -1;
+            func_ok = 1;
+            res->d = log(res->d);
+            res->is_str = 0;
+            if (isnan(res->d))
+                hts_expr_val_undef(res);
         }
         break;
 
@@ -177,6 +218,44 @@ static int func_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
         } else if (strncmp(str, "max(", 4) == 0) {
             if (expression(filt, data, fn, str+4, end, res)) return -1;
             func_ok = expr_func_max(res);
+        }
+        break;
+
+    case 'p':
+        if (strncmp(str, "pow(", 4) == 0) {
+            if (expression(filt, data, fn, str+4, end, res)) return -1;
+            func_ok = 1;
+
+            if (**end != ',')
+                return -1;
+            (*end)++;
+            hts_expr_val_t val = HTS_EXPR_VAL_INIT;
+            if (expression(filt, data, fn, ws(*end), end, &val)) return -1;
+            if (!hts_expr_val_exists(res) || !hts_expr_val_exists(&val)) {
+                hts_expr_val_undef(res);
+            } else if (res->is_str || val.is_str) {
+                hts_expr_val_free(&val); // arith on strings
+                return -1;
+            } else {
+                func_ok = 1;
+                res->d = pow(res->d, val.d);
+                hts_expr_val_free(&val);
+                res->is_str = 0;
+            }
+
+            if (isnan(res->d))
+                hts_expr_val_undef(res);
+        }
+        break;
+
+    case 's':
+        if (strncmp(str, "sqrt(", 5) == 0) {
+            if (expression(filt, data, fn, str+5, end, res)) return -1;
+            func_ok = 1;
+            res->d = sqrt(res->d);
+            res->is_str = 0;
+            if (isnan(res->d))
+                hts_expr_val_undef(res);
         }
         break;
     }
@@ -286,32 +365,46 @@ static int unary_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
                       char *str, char **end, hts_expr_val_t *res) {
     int err;
     str = ws(str);
-    if (*str == '+') {
+    if (*str == '+' || *str == '-') {
         err = simple_expr(filt, data, fn, str+1, end, res);
-        err |= res->is_str;
-        res->is_true = res->d != 0;
-    } else if (*str == '-') {
-        err = simple_expr(filt, data, fn, str+1, end, res);
-        err |= res->is_str;
-        res->d = -res->d;
-        res->is_true = res->d != 0;
+        if (!hts_expr_val_exists(res)) {
+            hts_expr_val_undef(res);
+        } else {
+            err |= res->is_str;
+            if (*str == '-')
+                res->d = -res->d;
+            res->is_true = res->d != 0;
+        }
     } else if (*str == '!') {
         err = unary_expr(filt, data, fn, str+1, end, res);
         if (res->is_true) {
+            // Any explicitly true value becomes false
             res->d = res->is_true = 0;
-            res->is_str = 0;
+        } else if (!hts_expr_val_exists(res)) {
+            // We can also still negate undef values by toggling the
+            // is_true override value.
+            res->d = res->is_true = !res->is_true;
         } else if (res->is_str) {
-            res->is_str = 0;
+            // !null = true, !"foo" = false, NOTE: !"" = false also
             res->d = res->is_true = (res->s.s == NULL);
         } else {
             res->d = !(int64_t)res->d;
             res->is_true = res->d != 0;
         }
+        res->is_str = 0;
     } else if (*str == '~') {
         err = unary_expr(filt, data, fn, str+1, end, res);
-        err |= res->is_str;
-        res->d = ~(int64_t)res->d;
-        res->is_true = res->d != 0;
+        if (!hts_expr_val_exists(res)) {
+            hts_expr_val_undef(res);
+        } else {
+            err |= res->is_str;
+            if (!hts_expr_val_exists(res)) {
+                hts_expr_val_undef(res);
+            } else {
+                res->d = ~(int64_t)res->d;
+                res->is_true = res->d != 0;
+            }
+        }
     } else {
         err = simple_expr(filt, data, fn, str, end, res);
     }
@@ -338,7 +431,9 @@ static int mul_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
         str = ws(str);
         if (*str == '*' || *str == '/' || *str == '%') {
             if (unary_expr(filt, data, fn, str+1, end, &val)) return -1;
-            if (val.is_str || res->is_str) {
+            if (!hts_expr_val_exists(&val) || !hts_expr_val_exists(res)) {
+                hts_expr_val_undef(res);
+            } else if (val.is_str || res->is_str) {
                 hts_expr_val_free(&val);
                 return -1; // arith on strings
             }
@@ -348,12 +443,15 @@ static int mul_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
             res->d *= val.d;
         else if (*str == '/')
             res->d /= val.d;
-        else if (*str == '%')
-            res->d = (int64_t)res->d % (int64_t)val.d;
-        else
+        else if (*str == '%') {
+            if (val.d)
+                res->d = (int64_t)res->d % (int64_t)val.d;
+            else
+                hts_expr_val_undef(res);
+        } else
             break;
 
-        res->is_true = res->d != 0;
+        res->is_true = hts_expr_val_exists(res) && (res->d != 0);
         str = *end;
     }
 
@@ -378,9 +476,12 @@ static int add_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
     hts_expr_val_t val = HTS_EXPR_VAL_INIT;
     while (*str) {
         str = ws(str);
+        int undef = 0;
         if (*str == '+' || *str == '-') {
             if (mul_expr(filt, data, fn, str+1, end, &val)) return -1;
-            if (val.is_str || res->is_str) {
+            if (!hts_expr_val_exists(&val) || !hts_expr_val_exists(res)) {
+                undef = 1;
+            } else if (val.is_str || res->is_str) {
                 hts_expr_val_free(&val);
                 return -1; // arith on strings
             }
@@ -393,7 +494,11 @@ static int add_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
         else
             break;
 
-        res->is_true = res->d != 0;
+        if (undef)
+            hts_expr_val_undef(res);
+        else
+            res->is_true = res->d != 0;
+
         str = *end;
     }
 
@@ -412,11 +517,14 @@ static int bitand_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
     if (add_expr(filt, data, fn, str, end, res)) return -1;
 
     hts_expr_val_t val = HTS_EXPR_VAL_INIT;
+    int undef = 0;
     for (;;) {
         str = ws(*end);
         if (*str == '&' && str[1] != '&') {
             if (add_expr(filt, data, fn, str+1, end, &val)) return -1;
-            if (res->is_str || val.is_str) {
+            if (!hts_expr_val_exists(&val) || !hts_expr_val_exists(res)) {
+                undef = 1;
+            } else if (res->is_str || val.is_str) {
                 hts_expr_val_free(&val);
                 return -1;
             }
@@ -426,6 +534,8 @@ static int bitand_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
         }
     }
     hts_expr_val_free(&val);
+    if (undef)
+        hts_expr_val_undef(res);
 
     return 0;
 }
@@ -440,11 +550,14 @@ static int bitxor_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
     if (bitand_expr(filt, data, fn, str, end, res)) return -1;
 
     hts_expr_val_t val = HTS_EXPR_VAL_INIT;
+    int undef = 0;
     for (;;) {
         str = ws(*end);
         if (*str == '^') {
             if (bitand_expr(filt, data, fn, str+1, end, &val)) return -1;
-            if (res->is_str || val.is_str) {
+            if (!hts_expr_val_exists(&val) || !hts_expr_val_exists(res)) {
+                undef = 1;
+            } else if (res->is_str || val.is_str) {
                 hts_expr_val_free(&val);
                 return -1;
             }
@@ -454,6 +567,8 @@ static int bitxor_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
         }
     }
     hts_expr_val_free(&val);
+    if (undef)
+        hts_expr_val_undef(res);
 
     return 0;
 }
@@ -468,11 +583,14 @@ static int bitor_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
     if (bitxor_expr(filt, data, fn, str, end, res)) return -1;
 
     hts_expr_val_t val = HTS_EXPR_VAL_INIT;
+    int undef = 0;
     for (;;) {
         str = ws(*end);
         if (*str == '|' && str[1] != '|') {
             if (bitxor_expr(filt, data, fn, str+1, end, &val)) return -1;
-            if (res->is_str || val.is_str) {
+            if (!hts_expr_val_exists(&val) || !hts_expr_val_exists(res)) {
+                undef = 1;
+            } else if (res->is_str || val.is_str) {
                 hts_expr_val_free(&val);
                 return -1;
             }
@@ -482,6 +600,8 @@ static int bitor_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
         }
     }
     hts_expr_val_free(&val);
+    if (undef)
+        hts_expr_val_undef(res);
 
     return 0;
 }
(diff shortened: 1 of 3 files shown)


jkbonfield:
After some corrections and rebasing I think I'm ready for this to be reviewed.

It's almost entirely compatible with the old mechanism apart from expressions genuinely involving NaN (eg 0/0) which were considered true previously and are now false, and the `[X1] != 0` example quoted above.

This means we probably want a slight update to the samtools man page too.  Although what it states there is all legal, some of it is out of date and it misses the newer functionality.

test/test_expr.c
```diff
@@ -64,14 +64,17 @@ int lookup(void *data, char *str, char **end, hts_expr_val_t *res) {
         kputs("", ks_clear(&res->s));
     } else if (strncmp(str, "null-but-true", 13) == 0) {
         *end = str+13;
+        hts_expr_val_undef(res);
```
daviesrob, line 67: Maybe this should be done using a `nan-but-true`, leaving the `null-but-true` test as-is.

Note that `nan-but-true` could occur, if a floating-point tag exists with a `nan` in it.

daviesrob (commented):
I notice that if [`bam_sym_lookup()`](https://github.com/samtools/htslib/blob/b7addd3151/sam.c#L1438) were changed to return NaN for missing tags instead of NULL strings, it would go a long way to solve the problem of having to free the kstring each time `hts_filter_eval2()` is called.

hts_expr.c
```diff
@@ -634,16 +791,37 @@ static int and_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
         str = ws(*end);
         if (str[0] == '&' && str[1] == '&') {
             if (eq_expr(filt, data, fn, str+2, end, &val)) return -1;
-            res->is_true = res->d =
-                (res->is_true || (res->is_str && res->s.s) || res->d) &&
-                (val.is_true  || (val.is_str && val.s.s) || val.d);
-            res->is_str = 0;
+            if (!hts_expr_val_existsT(res) || !hts_expr_val_existsT(&val)) {
+                hts_expr_val_undef(res);
+                res->d = 0;
```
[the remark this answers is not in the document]
jkbonfield: It's deliberate, as stated in the commit message.

You've moved on from quoting SQL to quoting somewhere else as the source of 3VL, but the SQL one was potentially more interesting.  Specifically it stated that NOT(unknown) is unknown.  That is obviously logical from a boolean value view point as bool can be {0,1} and when negating it's {1,0} which is still basically all possibilities.

However it then went on to say that NOT used in a comparison would yield the opposite case, so if (unknown) will fail and if (NOT(unknown)) would pass.  That is also clearly useful as it means negation always takes the opposite code path even when we don't know what the value is.

I applied the same logic to the AND and OR case.  I could potentially set is_true 0/1 and leave the value as unknown, but I felt it's not desparately helpful.  I came to this conclusion when looking at how things like NAN work in C.  "NAN || 0" is 1, not NAN, despite NAN being considered true.  It's canonicallised it basically.  I took the same approach.  We're not bit-wise ANDing and ORing, but as a logic statement, and if it's going to return true/false as far as a program execution goes then it's also less surprising if it returns 1/0 in value too.

hts_expr.c
```diff
@@ -286,32 +365,43 @@ static int unary_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
...
         err = simple_expr(filt, data, fn, str+1, end, res);
-        err |= res->is_str;
-        res->is_true = res->d != 0;
-    } else if (*str == '-') {
-        err = simple_expr(filt, data, fn, str+1, end, res);
-        err |= res->is_str;
-        res->d = -res->d;
-        res->is_true = res->d != 0;
+        if (!hts_expr_val_exists(res)) {
+            hts_expr_val_undef(res);
```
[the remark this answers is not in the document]
jkbonfield: Take a look at the code.  Undef resets more than the states that are checked for in `val_exists`.  Possibly you're correct that it's always the case that if `hts_expr_val_exists` is false then the other unchecked for parts of `res` are already also undefined, but it wasn't something I wanted to assume.

hts_expr.c
```diff
@@ -353,7 +447,7 @@ static int mul_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
         else
             break;
 
-        res->is_true = res->d != 0;
+        res->is_true = hts_expr_val_existsT(res) && (res->d != 0);
```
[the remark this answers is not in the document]
jkbonfield: Agreed `x%0` is an easy case to catch.

I cannot recall why I explicitly chose `hts_expr_val_existsT` here but I can't find any way for it to make a difference so changing it is fine.

hts_expr.c
```diff
@@ -500,33 +616,60 @@ static int cmp_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
...
+            res->is_true=res->d
+                = res->is_str && res->s.s && val.is_str && val.s.s
+                ? strcmp(res->s.s, val.s.s) < 0
+                : !res->is_str && !val.is_str && res->d < val.d;
+            res->is_str = 0;
+        }
     }
+
+    if (cmp_done && (!hts_expr_val_exists(&val) || !hts_expr_val_exists(res)))
+        hts_expr_val_undef(res);
```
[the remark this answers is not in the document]
jkbonfield: Same answer as above - it's just defensive belt & braces style to sanitize the state, given undef clears more than the exists check checks for.

hts_expr.c
```diff
@@ -546,34 +689,45 @@ static int eq_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
...
+                res->is_true = res->d = res->is_str
+                    ? (res->s.s && val.s.s ?strcmp(res->s.s, val.s.s) != 0 :1)
+                    : res->is_str != val.is_str || res->d != val.d;
+            }
         }
         res->is_str = 0;
 
     } else if ((str[0] == '=' && str[1] == '~') ||
                (str[0] == '!' && str[1] == '~')) {
+        eq_done = 1;
```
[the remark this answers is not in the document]
jkbonfield: The check already exists with `!res->is_str`.  I could add an additional exists check, but it wouldn't change anything I think.

Non-const string is an interesting one.  You're correct it would fail, eg with `samtools view -e '[AB] =~ [CD]'` with tags AB being a string and CD being a regexp to match it.  Cryptic!  We'd need to cache the string too and check it so we can invalidate the `regcomp` result, but for now I'm inclined to simply document it as only accepting constant strings.

hts_expr.c
```diff
@@ -705,7 +883,7 @@ static int hts_filter_eval_(hts_filter_t *filt,
     if (res->is_str) {
         res->is_true |= res->s.s != NULL;
         res->d = res->is_true;
-    } else {
+    } else if (hts_expr_val_exists(res)) {
```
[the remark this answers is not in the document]
jkbonfield: Not unless you want to remove the concept of `null-but-true`.

test/test_expr.c
jkbonfield: I switched to using nan at your own request...

I'm not inclined to switch back again now.  I don't think we really need to worry too much about real nans as it's such a corner case we can simply document it as nan being considered to be false by default (which it is in hts_expr.h).

I thought I'd already updated the expression syntax documentation too, but apparently not.  (It's rather annoying that most of htslib's documentation is in samtools, but that's a different issue.)

jkbonfield:
Adding documentation I'm struck by null being "false/null" while !null being "true/1".  I'm wondering what the purpose was of that, but we could just do `is_true ^= 1` so it becomes "true/null".  This seems to be more logical, although for numeric values `!2` becomes `0` and `!!2` therefore becomes `1`, so maybe that's why I chose to make `!null` to be 0 and `!!null` to be 1.

Thoughts?

hts_expr.c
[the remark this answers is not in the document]
daviesrob: The `hts_expr_val_existsT()` is still here...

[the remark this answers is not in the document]
jkbonfield: Oops sorry.  I was toggling it on/off to test the effect and apparently left it in the original style.  Fixed and pushed again.

hts_expr.c
```diff
@@ -338,7 +428,11 @@ static int mul_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
         str = ws(str);
         if (*str == '*' || *str == '/' || *str == '%') {
             if (unary_expr(filt, data, fn, str+1, end, &val)) return -1;
-            if (val.is_str || res->is_str) {
+            if (!hts_expr_val_exists(&val) || !hts_expr_val_exists(res)) {
+                hts_expr_val_undef(res);
+                hts_expr_val_free(&val);
+                return 0;
```
daviesrob, line 434: I just noticed that this short-circuits, which can lead to a parsing failure if you hit a null value:
```
$ ./test/test_expr 'null * 1 * 1'
Unable to parse expression at null * 1 * 1
```

Compare:
```
$ ./test/test_expr 'null + 1 + 1'
false	nan
```

jkbonfield: Just commenting out the free and return solves it, but this is slightly different to how the add code works which sets `undef=1`.   Given nan +/- anything is nan then it won't change the `d` value and `is_true` becomes 0 in both cases.

I am wondering though if `null-but-true + 1` and `null-but-true * 1` should be true still.  I expect there are lots of bizarre corner cases like that.  Practically speaking, false is better as `[NM]+1 < 11` and `[NM] < 10` give the same result now, but wouldn't I think if `null-but-true + 1` is true.

daviesrob: Looks like that fixed it.

It's possible to get NAN with is_true = 1 by making an aux tag with NAN in it: `NT:f:nan`.  It stays true long enough for `exists()` and the logical-and/or operators to work.  Everything else thinks it's false, which is probably OK, apart maybe from `default()`.  If the tag exists, shouldn't `default()` return its value, even if it's `nan`?

This also leads to another slight `default()` oddity, as it also checks the second value with `hts_expr_val_exists()`:
```
$ ./test/test_expr 'default(null, 123)'
true	123    << Expected
$ ./test/test_expr 'default(null-but-true, 123)'
true	123    << Arguable, maybe wrong?
$ ./test/test_expr 'default(null, null-but-true)'
false	nan    << Not the second argument
$ ./test/test_expr 'default(null-but-true, null)'
true	nan    << Not the second argument
```
The last two look inconsistent with the first two.  I think `default()` should only test `hts_expr_val_existsT(res)` and always return `val` if false.  (Or `hts_expr_val_exists(res)` if `null-but-true` is more null than true in this context).

jkbonfield: I'm not sure it's worth worrying overly about genuine NANs.

We took the decision to basically relegate genuine NANs when we decided that NAN was to be our "null" state instead of a null string (and something you suggested IIRC).  Specifically using NAN instead of null meant nan is false in conditionals (unlike C where it's true).  I was OK with that as realistically it's not something that crops up in the wild, at least not deliberately so!  So I'm not concerned with getting cases like `NT:f:nan` working fully. 

It's possible we should improve the consistency of null-but-true though and how it propagates through things.  Mostly it's fine, but there's perhaps some debate over `!null-but-true` being false/0 rather than false/null.  I can see `default` changes are justified.  I'll think some more about that and test other cases too.

Regarding `default(null-but-true, 123)`, currently all nulls, whether true or false, are still considered as null.  The only difference is how they work in conditionals and the negation operator.  I think that's fine.   Although maybe what you're saying is "null-but-true" is infact a genuine NAN, and "null" is our non-NAN "null" equivalent.  That's very mirky though and it doesn't make sense for the NOT operator to toggle between a null state and a nan state.  I'd rather not sail in those mirky waters and think the current action of `default` is the correct one.

daviesrob: It matters when you have a tag like `fN:f:nan`, as for those `default` disagrees with `exists`:
```
$ ./test/test_view -i filter='exists([fN])' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800
SRR065390.14978392	16	CHROMOSOME_I	2	1	27M1D73M	CCTAGCCCTAACCCTAACCCTAACCCTAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAA	#############################@B?8B?BA@@DDBCDDCBC@CDCDCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC	XN:i:0	XO:i:1	fN:f:nan	f0:f:0

$ ./test/test_view -i filter='default([fN], 0)' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800
```

Compare with a tag that isn't `nan`:
```
$ ./test/test_view -i filter='exists([f0])' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800
SRR065390.14978392	16	CHROMOSOME_I	2	1	27M1D73M	CCTAGCCCTAACCCTAACCCTAACCCTAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAA	#############################@B?8B?BA@@DDBCDDCBC@CDCDCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC	XN:i:0	XO:i:1	fN:f:nan	f0:f:0

$ ./test/test_view -i filter='default([f0], 0)' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800
SRR065390.14978392	16	CHROMOSOME_I	2	1	27M1D73M	CCTAGCCCTAACCCTAACCCTAACCCTAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAA	#############################@B?8B?BA@@DDBCDDCBC@CDCDCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC	XN:i:0	XO:i:1	fN:f:nan	f0:f:0
```

And one that isn't there:
```
$ ./test/test_view -i filter='exists([XX])' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800

$ ./test/test_view -i filter='default([XX], 0)' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800
```

jkbonfield: I'm aware it matters for files with NAN tags, but as I stated before, we already took the decision to treat them as nulls and basically accept that true NANs are something out of the scope of our filtering language.

If you look back on our internal Slack history you'll see you suggested I switch to using NANs.  I raised a concern as reusing NAN to mean something else has already bitten us before in bcftools, but you countered that by saying bcftools was using them courageously for something else (I won't dispute that!), so I acquiesced.

Either NAN is null, or NAN is NAN.  It cannot be both without creating the same nightmare scenario we have in bcftools.  I've already agreed however that `default` could have better behaviour and I'll work on that.

I did however somewhat forsee this and instead of using isnan or `var.d=nan` I added the `hts_expr_val_exists` and `hts_expr_val_undef` functions to test and set the null state, so we could potentially revert the decision to use NAN and switch back to a null string (although you disliked that as it breaks reuse of the kstring) or just add an entirely new field for "is_null".  It'd need a lot more testing though and some things we're simply exploiting with NAN such as equiality/</> comparisons failing (which I believe was your original intention when suggesting I switch to them).

Frankly I think we're nitpicking on things that aren't really relevant.  I'm OK with simply stating that the filtering language doesn't support NAN and be done with it.

daviesrob: The problem with BCF was due to it using quiet and signalling NANs, which is a completely different matter.  The issue here is about supporting the `is_true` state, and consistently making `[XX]` report the presence or absence of a tag irrespective of what it contains.

jkbonfield: Maybe what we need is a clear specification of what should be supported.  Eg if we actually want to support NAN in tags and be able to compare them, check for them, etc, then:

- We need an `isnan` function, as comparing to nan doesn't work by design.
- We need another way to represent null which isn't the same nan value.
- We need to decide whether NAN is true or false in booleans.  Currently it's false, but C defines it to be true.  I'm unsure if the IEEE FP stuff defines it either way.
- Do we care about +nan vs -nan?  They're different values technically (see 0/0 which is -nan).

I also think this is now likely to be something to solve in the next release, as we can't keep delaying this one waiting for things in the pipeline to be completed.

jkbonfield: > The problem with BCF was due to it using quiet and signalling NANs, which is a completely different matter. The issue here is about supporting the `is_true` state, and consistently making `[XX]` report the presence or absence of a tag irrespective of what it contains.

I understand what you're saying I think.  You mean `[XX]` when the tag doesn't exist is a NAN with `is_true=0` and `[XX]` when the tag does exist with a NAN value is NAN with `is_true=1`.  However that way lies problems surely.

We now have null vs nan both using `var.d=nan`, and only the `is_true` flag being able to distinguish between null vs nan.  That then makes things like `![XX]` problematic, as the not operator is now toggling between null and nan and not between null-is-false and null-is-true as it does now.

So I go back to my original assertion - if we wanted to use nan to represent nulls, then we had already decided to not support nan.  That seems quite obvious.  We can still go back and change this to use another flag as I wrote it with the expectation something like this would then crop up, but it would have been far simpler to not start with here in the first place.

jkbonfield: A demonstration of the problems.

`exists()` currently does: `res->d = hts_expr_val_exists(res);`.  Thus it doesn't modify `is_true`.  This is probably an error.

It means:

```
@ seq4d[samtools.../htslib]; ./test/test_expr 'exists(null)'
false	0
@ seq4d[samtools.../htslib]; ./test/test_expr 'exists("foo")'
true	1
@ seq4d[samtools.../htslib]; ./test/test_expr 'exists("12")'
true	1
@ seq4d[samtools.../htslib]; ./test/test_expr 'exists(0)'
true	1
@ seq4d[samtools.../htslib]; ./test/test_expr 'exists(null)'
false	0
@ seq4d[samtools.../htslib]; ./test/test_expr 'exists(null-but-true)'
true	0  <--- wrong IMO
@ seq4d[samtools.../htslib]; ./test/test_expr '!exists(null)'
true	1
@ seq4d[samtools.../htslib]; ./test/test_expr '!exists(null-but-true)'
true	1
```

Negation works though, as we're negating the exists value and not the truth state (as the value exists and the override is only used for 0-but-true or null-but-true).

We can't do searches correctly though:

```
@ seq4d[samtools.../htslib]; cat _.sam
@SQ	SN:I	LN:1009800
+nan	16	I	2	1	5M	*	0	0	CCTAG	*ZZ:f:nan	YY:f:20
-nan	16	I	2	1	5M	*	0	0	CCTAG	*ZZ:f:-nan	YY:f:20
ten	16	I	2	1	5M	*	0	0	CCTAG	*ZZ:f:10	YY:f:0

@ seq4d[samtools.../htslib]; ./test/test_view -i filter='exists([ZZ])' _.sam
@SQ	SN:I	LN:1009800
+nan	16	I	2	1	5M	*	0	0	CCTAG	*ZZ:f:nan	YY:f:20
-nan	16	I	2	1	5M	*	0	0	CCTAG	*ZZ:f:nan	YY:f:20
ten	16	I	2	1	5M	*	0	0	CCTAG	*ZZ:f:10	YY:f:0

@ seq4d[samtools.../htslib]; ./test/test_view -i filter='!exists([ZZ])' _.sam
@SQ	SN:I	LN:1009800
+nan	16	I	2	1	5M	*	0	0	CCTAG	*ZZ:f:nan	YY:f:20
-nan	16	I	2	1	5M	*	0	0	CCTAG	*ZZ:f:nan	YY:f:20
```

Here we have two records that get returned with both `exists` and `!exists`, which is plainly wrong.    This is because `exists(null-but-true)` and `!exists(null-but-true)` are both true.

If we change the exists function to always return false for a null value, then exists negation works properly, but tags containing nan are considered to be absent.  This is infact the correct solution as I think not setting `is_true` was a bug, but it shows how reusing nan for two things has caused us a problem in the language.

```
@ seq4d[samtools.../htslib]; ./test/test_view -i filter='exists([ZZ])' _.sam
@SQ	SN:I	LN:1009800
ten	16	I	2	1	5M	*	0	0	CCTAG	*ZZ:f:10	YY:f:0

@ seq4d[samtools.../htslib]; ./test/test_view -i filter='!exists([ZZ])' _.sam
@SQ	SN:I	LN:1009800
+nan	16	I	2	1	5M	*	0	0	CCTAG	*ZZ:f:nan	YY:f:20
-nan	16	I	2	1	5M	*	0	0	CCTAG	*ZZ:f:nan	YY:f:20
```

Put simply, if we want NAN to be a valid value, then we need null as something else, or to use a NAN which doesn't occur in the wild (Danger! see bcftools..)

daviesrob: Actually, when the tag doesn't exist, it's [`is_true=0` and a NULL string pointer](https://github.com/samtools/htslib/blob/148a8c580a/sam.c#L1438).  That part hasn't been changed.  If `[XX]` is taken to be true when the tag exists, then `![XX]` and `!![XX]` both toggle as you would expect on my toy example for non-existent tags, the one with the `NAN` in it, and tags with a non-zero value.  Where it turns out it does go a bit weird is when you have a tag containing 0, e.g. `XN:i:0` or `f0:f:0`:
```
$ ./test/test_view -i filter='[XN]' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800
SRR065390.14978392	16	CHROMOSOME_I	2	1	27M1D73M	CCTAGCCCTAACCCTAACCCTAACCCTAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAA	#############################@B?8B?BA@@DDBCDDCBC@CDCDCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC	XN:i:0	XO:i:1	fN:f:nan	f0:f:0
$ ./test/test_view -i filter='![XN]' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800
SRR065390.14978392	16	CHROMOSOME_I	2	1	27M1D73M	CCTAGCCCTAACCCTAACCCTAACCCTAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAA	#############################@B?8B?BA@@DDBCDDCBC@CDCDCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC	XN:i:0	XO:i:1	fN:f:nan	f0:f:0
$ ./test/test_view -i filter='!![XN]' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800
```

Note that this is a change in behaviour from the current `develop` branch.

jkbonfield: Ah that's an accident then.  When I switched from using null strings to nan-as-null I obviously forgot to change the code that wasn't in hts_expr.c, although the check for being null copes with both deliberately so it works with older versions of code that may be plugging in to this interface (so perhaps it's fine to leave as is?).

The change in behaviour for `![XN]` was accidental it seems.  Initially I misremembered this as the deliberate change in `[XN]!=0` behaviour, but this is different.  It comes from a distinction between null-but-true vs zero-but-true, and is trivial to fix.

I do wonder though whether `!` should solely be working on the `is_true` flag.  Generally we sanitize this at the end of the evaluation function so non-zero is true, but if we verbatim set it everywhere then we could have other scenarios.  I'm not sure it's worth the pain though; eg do we really want "5 but false" for example?  I can't see it.  Zero is a special case and it's fair to keep it that way.

jkbonfield: I take that back, in fixing this I've now broken compatibility with 1.15, so I'll consider reverting this latest commit. 

The problem is the whole '[XX]' as true if it exists with the value as the result was always a bit of an ill-conceived idea as it became somewhat tricky to work out what negation would do.  Is it !value or !exists?  When value is 0 there are two choices.  Samtools 1.15 goes on the value, so `![XX]` for `XX:i:0` is display as !0 is true.  This also meant `[XX] != 0` was broken as null != 0 was true due to the values matching.  In changing the rules so comparisons with null always fail we fixed this bug in the old evaluation.  However the `![XX]` is still strictly a negation of both !null and !0 depending on the tag value.  I think that is correct behaviour, even if confusing.

I copied the tag syntax from [sambamba's filtering language](https://github.com/biod/sambamba/wiki/%5Bsambamba-view%5D-Filter-expression-syntax), but I should have paid more attention to their null detection method.  This is why I subsequently added the explicit `exists()` function, so negation works in a sensible manner.

daviesrob: I'd agree, having an explicit `exists()` makes things a lot easier.  In fact, sambamba seems to have something similar, in the form of explicit [`== null` and `!= null` comparisons](https://github.com/biod/sambamba/blob/f898046c/sambamba/utils/common/filtering.d#L216).

So, given we now have `exists()`, should `[XX]` always go with the value of the tag, including for truthiness?  As of 1feac5e, it still seems to have a foot in both camps:
```
$ ./test/test_view -i filter='[XN]' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800
SRR065390.14978392	16	CHROMOSOME_I	2	1	27M1D73M	CCTAGCCCTAACCCTAACCCTAACCCTAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAA	#############################@B?8B?BA@@DDBCDDCBC@CDCDCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC	XN:i:0	XO:i:1	fN:f:nan	f0:f:0

$ ./test/test_view -i filter='![XN]' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800
SRR065390.14978392	16	CHROMOSOME_I	2	1	27M1D73M	CCTAGCCCTAACCCTAACCCTAACCCTAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAA	#############################@B?8B?BA@@DDBCDDCBC@CDCDCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC	XN:i:0	XO:i:1	fN:f:nan	f0:f:0

$ ./test/test_view -i filter='!![XN]' /tmp/ce#1.sam
@SQ	SN:CHROMOSOME_I	LN:1009800
```

This matches what happened before we started fiddling with the expressions, but current `develop` (ad80f8e) has `[XN]` true, `![XN]` false, `!![XN]` true.  If we went strictly with the value, this would invert to `[XN]` false, `![XN]` true, `!![XN]` false.  Given that the last release was logically inconsistent here both for tags with value 0, and non-existent tags, I guess we should worry too much about the resulting breaking change?

jkbonfield: Indeed we seem to have got back to where we started for `[XN]` syntax.

If I was designing it right now, with what I've learnt, then absolutely I'd require `exists` to check for existance and have `[XN]` purely being true/false based on the content of the tag.  (Obviously when it doesn't exist it should fail regardless.)  The old "[NM]" syntax is short, sweet, and almost certainly in use somewhere.  If we changed that to mean "NM:i:0" now returns false, then we're going to break pipelines.

I also highly doubt anyone really cares about nuances of things like "MD:Z:" (empty string) and "AB:f:nan" handling.  They'll just be dealing with the typical values that occur on a daily basis.  So we shouldn't throw the baby out with the bath-water here and we need to make sure the most basic operations don't change.

I agree the old way of doing things wasn't entirely consistent when it came to negation, so "![NM]" and "[NM] != 0" gave odd results, [but these were actually documented with workarounds](http://www.htslib.org/doc/samtools.html#FILTER_EXPRESSIONS):

>  If no comparison is used with an auxiliary tag it is taken simply to be a test for the existence of that tag. So "[NM]" will return any record containing an NM tag, even if that tag is zero (NM:i:0).
> 
> If you need to check specifically for a non-zero value then use [NM] && [NM]!=0. 

That still works, but it's now been revised so the `[NM] &&` bit is not needed as the expression is only true when the tag exists.  We could improve the documentation further.  I've attempted to do this in https://github.com/samtools/samtools/pull/1687 but if you have further suggestions I'm happy to incorporate them.

daviesrob: The documentation didn't mention the `![NM]` case where the value is 0, probably because no-one had noticed it.  In fact, it's the only anomalous case for unary-not, as far as I can tell.  It can be made consistent with this:
```diff
diff --git a/hts_expr.c b/hts_expr.c
index 9e2454b4..b86803ac 100644
--- a/hts_expr.c
+++ b/hts_expr.c
@@ -380,7 +380,12 @@ static int unary_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
         if (!hts_expr_val_exists(res)) {
             // We can still negate undef values by toggling the
             // is_true override value.
+            res->is_str = 0;
             res->d = res->is_true = !res->is_true;
+        } else if (res->is_true) {
+            // Explicit true becomes false
+            res->is_str = 0;
+            res->d = res->is_true = 0;
         } else if (res->is_str) {
             res->is_str = 0;
             // !null = true, !"foo" = false, NOTE: !"" = false also
```

Note that all the existing tests still pass after applying this change.

jkbonfield: It alluded to the case of `[NM]` and zero values being in the "check for zero" example, but you're correct it didn't explicitly spell out the `![NM]` case.  I'll need to think on the explicit true becoming false thing here a bit and do some experimentation of my own.

jkbonfield: Ok pushed some new changes, including new and updated tests.

I'll update the documentation side in the samtools PR too.

hts_expr.c
```diff
@@ -377,12 +377,15 @@ static int unary_expr(hts_filter_t *filt, void *data, hts_expr_sym_func *fn,
...
+        res->is_str = 0;
+        if (res->is_true) {
+            // Any explicitly true value becomes false
+            res->d = res->is_true = 0;
+        } else if (!hts_expr_val_exists(res)) {
+            // We can also still negate undef values by toggling the
             // is_true override value.
             res->d = res->is_true = !res->is_true;
         } else if (res->is_str) {
-            res->is_str = 0;
```
daviesrob, line 304: It's looking better, but there's a small problem with this change.  Setting `res->is_str = 0;` before all the tests means `res->is_str` here will always be false...

armintoepfer:
Actually @jkbonfield, I care about detecting `nan` ;)

PR #1480, opened 2022-07-21
Minimal support for CRAM files with missing @RG headers.

### LINKED ISSUE #1479: RG tag gets lost in cram format
I was initially thinking this was an issue with `samtools merge -r` as I could successfully merge several bam files into one bam file, with the respective `RG:Z:<name of cell>` appearing fine in the merged bam file, but they wouldn't appear if the final output was cram. Likewise converting with `samtools view` also lost the RG tags (and never were restored converting back, so truly lost).

I believe I narrowed down the cause to [here](https://github.com/samtools/htslib/blob/e97816474e5facdb0ef23df8be7e446eea474f24/cram/cram_encode.c#L2267) from the cram encoding
```
       // RG:Z
        if (aux[0] == 'R' && aux[1] == 'G' && aux[2] == 'Z') {
            rg = &aux[3];
            while (*aux++);
            if (CRAM_MAJOR_VERS(fd->version) >= 4)
                BLOCK_APPEND(td_b, "RG*", 3);
            continue;
        }
```
But even trying version 3.1 and version 4.0 didn't record the RG tag. I commented these lines out, and then it seemingly worked fine with my final cram file correctly having the RG tags. So unless I am missing something, this appears to be a safe/legitimate option (if not highly redundant) until RG tags are correctly stored in cram.

- Alex

jkbonfield:
The SAMtags spec states that RG:Z: lines should point match an RG ID
if RG headers are present, but doesn't explicitly *require* them to be
present.  The SAM spec itself recommends that RG headers are present.
Sadly this means CRAM may need to cope with this semantically
inconsistent edge case.

Given CRAM stores RG as an integer data series as an index into the
corresponding header, in much the same way that BAM stores chromosomes
as numeric "tid" values, this makes things challenging.  However CRAM
can also store text tags, so it's possible to round-trip with missing
headers by claiming RG is -1 (unspecified) and then adding a verbatim
RG:Z string tag.  This is perhaps a bit of a CRAM spec loop hole so
it's questionable if this is the correct solution.

This works and is decodable by both htslib and htsjdk, but it'll break
things like cram_transcode_rg as used by samtools cat.  I think this
is a pretty unlikely combination of events.  Note picard's
SamFormatConverter also drops these RG fields.

This code also whinges, *once for each and every problematic alignment
record*, when RG is absent in the SAM header.  It's considerably more
work to track which ones we've warned about before and to track all
that meta-data across threads in a robust manner, plus this really
could be considered to be a poor SAM file.  Were it not for the SAM
spec explicitly permitting such things (even if recommending against
it) I'd reject it outright.  Instead brow-beating the SAM creators
into fixing the headers could be considered to be a positive outcome.

Fixes #1479


jkbonfield:
Note I'm also open to the simpler strategy of just dropping the RG:Z: tag (as we do now) when they miss a header, but with the addition of a spammy warning message.  It makes it compatible with Picard.

It depends if we wish to be overly precise in accepting things in the specification, even when they're really a bad idea if not outright wrong, or whether we wish to put our own spin on things and enforce good behaviour.

daviesrob:
Seems to work, and probably better than losing the tags completely.  We'll see if anyone complains about the warning messages.

PR #1485, opened 2022-07-29
Report HTS_CFLAGS_AVX2 et al in `make print-config`

jmarshall:
Extend the `print-config` target added in 155e1aacac741febd6fd722cf7e3660da38d6a9b. This target-specific compiler flag information (probed for by _configure_) may also be of use to third parties.

In particular, pysam's current build infrastructure will need this information — cf pysam-developers/pysam#1112.

Pysam might in future switch to using HTSlib's build infrastructure to build _libhts.a_ (rather than using Python setuptools machinery to build the object files listed in `LIBHTS_OBJS` as reported by `print-config`), but this would still be of use to other third parties.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -326,6 +326,10 @@ libhts.a: $(LIBHTS_OBJS)
 	-$(RANLIB) $@
 
 print-config:
+	@echo HTS_CFLAGS_AVX2 = $(HTS_CFLAGS_AVX2)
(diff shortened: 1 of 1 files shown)


jkbonfield:
Thanks - a good idea.

Is there any benefit to extending `samtools version` to report similar information too?  It's a bit niche perhaps.

PR #1487, opened 2022-08-01
Warn if bgzf_getline() returned apparently UTF-16-encoded text

jmarshall:
Text files badly transferred from Windows may occasionally be UTF-16-encoded, and this may not be easily noticed by the user — see for example, [this report of tabix woes](https://www.biostars.org/p/9530170/) (`??#` is as reproduced locally by me, and equally as confusing as the OP's output):

```
$ tabix myVcf.vcf.gz
[E::get_intv] Failed to parse TBX_VCF, was wrong -p [type] used?
The offending line was: "??#"
[E::get_intv] Failed to parse TBX_VCF, was wrong -p [type] used?
The offending line was: ""
[E::get_intv] Failed to parse TBX_VCF, was wrong -p [type] used?
The offending line was: ""
```

It turned out that the text VCF file was UTF-16-encoded, and the lines returned contained alternating NULs when interpreted as ASCII C-strings, hence all lines appearing truncated to 0 or 1 or BOM+1 characters.

HTSlib should not accept such encoding (as other tools surely don't, hence doing so would cause interoperability problems), but it should ideally emit a warning or error message identifying the problem clearly.

Reading text from a `htsFile`/`samFile`/`vcfFile` will already have failed with EFTYPE/ENOEXEC (and printed a suitable `Inappropriate file type or format` message) if the text file is UTF-16-encoded, as the encoding will not have been recognised by `hts_detect_format()`. So no changes are needed here, although this could be extended to add `utf16_text_format` or so to `htsFormatCategory` to aid in reporting this (but I don't think that is really warranted).

OTOH code that uses plain BGZF handles does not return a clear diagnostic, as was seen with `tabix`. `bgzf_getline()` will return a UTF-16-encoded text line successfully, but code not expecting it will misinterpret the resulting string. This PR adds a diagnostic suitable for each context to the BGZF-based `bgzf_getline()` calls in HTSlib:

* in `hts_readlist()`/`hts_readlines()`, emit a warning (once, on the first line);
* in _tbx.c_, emit a more specific error message if `get_intv()` parsing failure is due to UTF-16 encoding.

With this PR, the biostars poster's test case produces the following diagnostics and the root cause is apparent:

```
tabix myVcf.vcf.gz
[E::get_intv] Failed to parse TBX_VCF: offending line appears to be encoded as UTF-16
[E::get_intv] Failed to parse TBX_VCF: offending line appears to be encoded as UTF-16
[E::get_intv] Failed to parse TBX_VCF: was wrong -p [type] used?
The offending line was: ""
```

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -431,6 +431,27 @@ static int is_text_only(const unsigned char *u, const unsigned char *ulim)
     return 1;
 }
 
+static inline int
+alternate_zeros(const unsigned char *u, const unsigned char *ulim)
+{
+    for (; u < ulim; u += 2)
+        if (*u != '\0') return 0;
+    return 1;
+}
+
+static int is_utf16_text(const unsigned char *u, const unsigned char *ulim)
+{
+    if (ulim - u >= 6 &&
+        ((u[0] == 0xfe && u[1] == 0xff && alternate_zeros(u+2, ulim)) ||
+         (u[0] == 0xff && u[1] == 0xfe && alternate_zeros(u+3, ulim))))
+        return 2;
+    else if (ulim - u >= 8 &&
+             (alternate_zeros(u, ulim) || alternate_zeros(u+1, ulim)))
+        return 1;
+    else
+        return 0;
+}
+
 static int is_fastaq(const unsigned char *u, const unsigned char *ulim)
 {
     const unsigned char *eol = memchr(u, '\n', ulim - u);
@@ -1961,6 +1982,12 @@ hFILE *hts_open_tmpfile(const char *fname, const char *mode, kstring_t *tmpname)
     return fp;
 }
 
+int hts_is_utf16_text(const kstring_t *str)
+{
+    const unsigned char *u = (const unsigned char *) (str->s);
+    return (str->l > 0 && str->s)? is_utf16_text(u, u + str->l) : 0;
+}
+
 // For VCF/BCF backward sweeper. Not exposing these functions because their
 // future is uncertain. Things will probably have to change with hFILE...
 BGZF *hts_get_bgzfp(htsFile *fp)
@@ -2030,6 +2057,8 @@ char **hts_readlist(const char *string, int is_file, int *_n)
         while ((ret = bgzf_getline(fp, '\n', &str)) >= 0)
         {
             if (str.l == 0) continue;
+            if (n == 0 && hts_is_utf16_text(&str))
+                hts_log_warning("'%s' appears to be encoded as UTF-16", string);
             if (hts_resize(char*, n + 1, &m, &s, 0) < 0)
                 goto err;
             s[n] = strdup(str.s);
@@ -2089,6 +2118,8 @@ char **hts_readlines(const char *fn, int *_n)
         str.s = 0; str.l = str.m = 0;
         while ((ret = bgzf_getline(fp, '\n', &str)) >= 0) {
             if (str.l == 0) continue;
+            if (n == 0 && hts_is_utf16_text(&str))
+                hts_log_warning("'%s' appears to be encoded as UTF-16", fn);
             if (hts_resize(char *, n + 1, &m, &s, 0) < 0)
                 goto err;
             s[n] = strdup(str.s);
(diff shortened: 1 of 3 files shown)


jmarshall:
The data file in the OP's case was a VCF file, and it is true that VCF 4.3 explicitly and VCF 4.1/4.2 _de facto_ require UTF-8 encoding. However tabix also works with arbitrary position-oriented files that don't have specifications about encodings.

However there is no suggestion (in this PR, at least) that HTSlib or tabix should go out of their way to accept such wackily encoded input files. This PR just improves the error messages so that the user is aided in identifying the underlying cause of the parsing failure.

daviesrob:
This looks useful, but I'm not sure it quite works as intended yet.  If I make myself a UTF-16 encoded vcf and try to index it, I get:
```
$ iconv -t UTF-16 test/index.vcf | ./bgzip > /tmp/i16.vcf.gz
$ ./tabix /tmp/i16.vcf.gz
[E::get_intv] Failed to parse TBX_VCF: offending line appears to be encoded as UTF-16
[E::get_intv] Failed to parse TBX_VCF: offending line appears to be encoded as UTF-16
  [ ... lots of similar messages omitted ...]
[E::get_intv] Failed to parse TBX_VCF: offending line appears to be encoded as UTF-16
[E::hts_idx_push] Unsorted positions on sequence #1: 2 followed by 1
tbx_index_build failed: /tmp/i16.vcf.gz
```

So it's reporting once for each line, because the indexer [skips lines it can't parse](https://github.com/samtools/htslib/blob/1858eed9c9/tbx.c#L333).  I guess that isn't much change from previous behaviour which also printed a message for each bad line, but should we treat this as more fatal?  If the whole file is `UTF-16` then there's no way you're going to get an index for it.

Somewhat more interesting is the last line.  That appeared because `tbx_parse1()` gets to look at the line [before any UTF-16 check happens](https://github.com/samtools/htslib/blob/c8a062b273/tbx.c#L180).  It will also happily [split input lines on NUL bytes](https://github.com/samtools/htslib/blob/c8a062b273/tbx.c#L100), so occasionally it interprets what it gets as VCF even though the returned data is rubbish.  Possibly it should reject lines that appear to have premature NUL bytes so the rest of the detection can work more reliably?

jkbonfield:
I'd be +1 for making it a fatal error instead.

Although as for warnings, I'm quite OK with invalid input producing 1 warning per line for invalid data.  It "encourages" people to fix their inputs and makes spotting the warning significantly more likely. :-)  (Obviously warnings produced from valid data, eg something we don't fully support or another form of warning, shouldn't be quite so spammy.)

jmarshall:
Yes. In general I prefer not to add diagnostic output to library routines, so in the `get_intv()` case I just changed (in some cases) the existing diagnostic being printed rather than altered when a diagnostic might be printed. So it's working as intended by the PR's author :smile: but I suppose I could be convinced to make more drastic changes to `get_intv()`'s / `tbx_parse1()`'s behaviour. (Is it the right time for contemplating further behaviour changes?)

daviesrob:
Well, as no files supported by `tabix` should have NUL bytes in them, I don't think it's too unreasonable to have `tbx_parse1()` reject lines that look obviously wrong.  It would be pretty easy to make it look for something like `len > 6`, a NUL is found in the first four bytes and maybe also there's another NUL two bytes on.  That should make it better at catching the UTF-16 case without getting false positives.

Also, as `hts_is_utf16_text()` is not going to mistake something valid for UTF-16, I think it would be reasonable to make `get_intv()` return -2 if it triggers, turning the error into a fatal one.  Otherwise you could end up with the program apparently working, albeit very noisily, but making an empty index - which actually gets more likely if `tbx_parse1()` is better at detecting bad data.

jmarshall:
Oh dear… I should have mentioned that in the last couple of weeks I've been in the process of reworking this to address your comments in https://github.com/samtools/htslib/pull/1487#issuecomment-1201079314 and add a specific warning about UTF-8-encoded BOMs as well.

Oh well, thanks for merging :smile: and I'll look at reworking those changes as a followup PR…

daviesrob:
I made a note to myself to make tabix reject this sort of input properly.  But I'm happy to wait if you have something brewing.

PR #1491, opened 2022-08-10
Recognise FASTA/Q even if the file has a very long header file

jmarshall:
`secondline_is_bases()` avoids false positives by verifying that the second line contains only base-encoded alphabetic characters, but if the first line is longer than `hts_detect_format2()`'s peek buffer, we fail to recognise FASTA or FASTQ.

Revise (and rename to `is_fastaq()`) so the FASTA/Q check is:
1. Check that the (buffered part of the) first line is entirely textual.
2. Return true if we have none of the second line in our peek buffer.
3. Check that the (buffered part of the) second line contains only base-encoding alphabetic characters.

Fixes samtools/samtools#1689.


jkbonfield:
Thanks

PR #1496, opened 2022-08-15
Fix "make check" on MinGW.

jkbonfield:
If we haven't previously done a "make" or "make all", the .dll file
won't have been created.  The test/with-shlib.sh script will then fail
as the "cp -p ../../hts-*.dll ." command fails.  We no longer even
attempt to execute this test when plugins don't exist (which they
won't on Windows unless we've installed the dlfcn package).

Note the analogous non-mingw commands are "ln -s ../../libhts.so.* ."
and similar, which don't fail in the $? sense, but do fail with
regards to produced a bogus target containing a wild-card pattern.

This implies the error checking in this script is broken for all
platforms bar MinGW, but given the very limited use for this one test
case it's an irrelevance.

(See also the abandoned https://github.com/jkbonfield/htslib/tree/mingw_make_check which adds code so `--enable-plugins` builds on MinGW.  Sadly it doesn't work for gs: and s3:, with cryptic errors, so delaying that to another time.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -572,8 +572,12 @@ check test: $(BUILT_PROGRAMS) $(BUILT_TEST_PROGRAMS) $(BUILT_PLUGINS) $(HTSCODEC
 	test/test_time_funcs
 	test/fieldarith test/fieldarith.sam
 	test/hfile
-	HTS_PATH=. test/with-shlib.sh test/plugins-dlhts -g ./libhts.$(SHLIB_FLAVOUR)
-	HTS_PATH=. test/with-shlib.sh test/plugins-dlhts -l ./libhts.$(SHLIB_FLAVOUR)
+	if test "x$(BUILT_PLUGINS)" != "x"; then \
+	    HTS_PATH=. test/with-shlib.sh test/plugins-dlhts -g ./libhts.$(SHLIB_FLAVOUR); \
+	fi
+	if test "x$(BUILT_PLUGINS)" != "x"; then \
+	  HTS_PATH=. test/with-shlib.sh test/plugins-dlhts -l ./libhts.$(SHLIB_FLAVOUR); \
+	fi
 	test/test_bgzf test/bgziptest.txt
 	test/test-parse-reg -t test/colons.bam
 	cd test/sam_filter && ./filter.sh filter.tst
--- a/Makefile
+++ b/Makefile
@@ -317,6 +317,9 @@ endif
 
 BUILT_PLUGINS = $(PLUGIN_OBJS:.o=$(PLUGIN_EXT))
 
+ifneq "$(BUILT_PLUGINS)" ""
+plugins: lib-shared
+endif
 plugins: $(BUILT_PLUGINS)
 
 
@@ -563,7 +566,7 @@ SRC = $(srcprefix)
 #
 # If using MSYS, avoid poor shell expansion via:
 #    MSYS2_ARG_CONV_EXCL="*" make check
-check test: $(BUILT_PROGRAMS) $(BUILT_TEST_PROGRAMS) $(BUILT_PLUGINS) $(HTSCODECS_TEST_TARGETS)
+check test: all $(HTSCODECS_TEST_TARGETS)
 	test/hts_endian
 	test/test_expr
 	test/test_kfunc


jmarshall:
The `test`/`check` target essentially has a dependency on `all`, except that it spells out what it actually needs explicitly as there's a couple of things in `all` that it doesn't need.

So what this really is is a missing dependency: `test`/`check` should also list a dependency on `libhts.$(SHLIB_FLAVOUR)` (at least when the _test/with-shlib.sh_ lines want to do something non-trivial) as used on the _test/with-shlib.sh_ lines. Except that using `$(SHLIB_FLAVOUR)` directly in filenames like that is actually wrong, so those commands already only work on Linux and macOS, where `$(SHLIB_FLAVOUR)` happens to be exactly the filename extension used.

So probably what this really really wants is these two commands to be pulled out to a separate `test-plugins` target, that listed as a prerequisite of `test`/`check`, and some more variable infrastructure to hold the actual base shared library filename — to be used as a prerequisite of `test-plugins` and nearby.

jkbonfield:
I initially assumed it was missing a dependency on all, but after adding it I realised it doubled the compilation time everywhere as it's building both libraries.  Arguably that's correct - it is test after all - but if we're not using the shared library I had assumed this nuance was a deliberate one.  Hence my fix.  I'm happy if we feel it should do a "make all" as part of check.

I don't see why we'd want to go to the bother of `libhts.$(SHLIB_FLAVOUR) ` for make check (especially given it may not actually be *lib*, so as with hts-3.dll) when simply sticking `all` in the dependency line does the job far simpler and is self explanatory.

I think there is more work to do here, in particular figuring out why all bar 2 plugins work on Windows MINGW.  It's clearly capable of using dlopen and dlsym, but I haven't worked out what's exceptional about s3 and gcs yet.  I don't propose to do that for this release.   Given we're not planning to get plugins working immediately on Windows, I figured the simplest fix was basically to elide the offending tests for now as it's short and sweet, and then work out how best to solve this later on.

daviesrob:
As far as I can tell, the dependencies are working given that `./configure --enable-plugins` causes the shared `libhts` library to be built along with the plugins on UNIX-y and MinGW platforms (but not MacOS).  Without that option you don't get a shared library, and the tests are disabled anyway so skipping them earlier shouldn't be a problem.

However, there's still a problem on MacOS with `--enable-plugins`:
```
HTS_PATH=. test/with-shlib.sh test/plugins-dlhts -g ./libhts.dylib
Can't dlopen "./libhts.dylib": dlopen(./libhts.dylib, 10): image not found
make: *** [test] Error 1
```

As it's `test/plugins-dlhts` that wants to `dlopen("libhts.so")` or whatever, it really needs to depend on `lib-shared`. We then might want to make the dependency of `check / test` on `test/plugins-dlhts` optional depending on the state of `BUILT_PLUGINS` - although we could defer that to future work and just let the main library be built all the time for now.

The other problem with this test is that it uses `libhts.$(SHLIB_FLAVOUR)` for the dll to load, which doesn't work when there is no `lib` prefix.  We might want to add a `SHLIB_NAME` variable, set at the same time as `SHLIB_FLAVOUR` so we know exactly what it's called.

jkbonfield:
That's what my other branch was doing, by conditionally modifying the dependencies.

However I think it's reasonable to say that "make check" perhaps ought to check everything can be built (if it's not been built already), so just having `all` as a dependency seems a reasonable solution.

Regarding the `libhts.` bit, I realised it was the wrong name, but the test is explicitly disabled in the test program for Windows, and it's only windows that doesn't have the lib prefix.  In the fullness of time it wants rewriting, but I didn't think that was appropriate for this release so I went with the smallest acceptable change.

With hindsight smaller still would have been what I originally did which was just stick `all` or `lib-shared` as a check dependency, and ignore the wrong filename issue (for now).

jkbonfield:
Following the meeting discussions, I've lifted over bits of the other branch.

- "make plugins" now has an explicit dependency on "lib-shared".  This was already brought in on linux, but not on some other system types.
- "make check" now depends on "all" so we get the shared library irrespective of whether it's needed.  This is a better workaround than more minutae of dependencies, as knowing we can build everything is a good check in its own right.

PR #1497, opened 2022-08-15
Fix clang builds on Windows/mingw.

jkbonfield:
If we use HTSLIB_EXPORT once, it has to be used everywhere for that symbol.

Also see #1435


jmarshall:
`#include "hts_defs.h"`

Actually the klib headers avoid using other htslib headers — for somewhat historical reasons at this point perhaps. But either this should invent `klib_export` for use inside _ksort.h_ or it should include _hts_defs.h_ and ideally replace `klib_unused` with the _hts_defs.h_ equivalent. (I guess a couple of other _k*.h_ headers are already using _hts_defs.h_.)

PR #1502, opened 2022-08-19
Update man SEE ALSO sections from .BR to .IR so the website uses URLs

jkbonfield:


daviesrob:
This doesn't quite work for `bgzip` and `tabix`.  The program that adds the links [bails out](https://github.com/samtools/www.htslib.org/blob/cd946c0c/add_manpage_links.pl#L92) when it finds the `<p>` between the `SEE ALSO` line and the man pages we want to link.  Removing the `.PP` directly after `.SH SEE ALSO` in those files should fix them.

PR #1504, opened 2022-09-02
Fix error code 0 returned by bcftools after error

janinl:
Without this, bcftools returns error code 0 even when interrupted by these errors visible in stderr:
     - [E::hts_itr_next] Failed to seek to offset 70116896333: Illegal seek
     - [E::bgzf_read_block] Failed to read BGZF block data at offset 29255442 expected 12238 bytes; hread returned -1

This code change is in line with what is written in the next `else` block

### DIFF
--- a/synced_bcf_reader.c
+++ b/synced_bcf_reader.c
@@ -641,7 +641,9 @@ static int _reader_fill_buffer(bcf_srs_t *files, bcf_sr_t *reader)
         }
         else if ( reader->tbx_idx )
         {
-            if ( (ret=tbx_itr_next(reader->file, reader->tbx_idx, reader->itr, &files->tmps)) < 0 ) break;  // no more lines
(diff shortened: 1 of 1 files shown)


janinl:
Note: The same error handling problem may also occur in the streaming case 20 lines above (https://github.com/samtools/htslib/blob/3a374083c11884d22f05c6a095a6b0cc3cc8cdc0/synced_bcf_reader.c#L626 ).
I didn't touch this code as I'm not 100% sure about the meaning of return values < -1 in that case.

janinl:
The second commit addresses the same problem for issues such as
 - [E::bgzf_read_block] Invalid BGZF header at offset xxx
 
Test case: Take a valid vcf.gz file and randomly edit the bytes to make it invalid.
These problems were observed when using HTTPS URL as inputs, possibly receiving invalid data due to some network problems.

daviesrob:
Looks good.  Thanks for the patch.

PR #1507, opened 2022-09-15
Prevent crash when only FASTA entry has no sequence.

jkbonfield:
Eg the minimal file of:

    >foo

Credit to OSS-Fuzz
Fixes oss-fuzz 51378

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -3853,7 +3853,6 @@ static int fastq_parse1(htsFile *fp, bam1_t *b) {
     }
 
     // Name
-
     if (*x->name.s != x->nprefix)
         return -2;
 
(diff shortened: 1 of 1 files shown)


jkbonfield:
Note I did also wonder about changing the `ret = (...) < -1` condition to `< 0` so a blank sequence is an error, but technically we do support empty sequences in FASTA.

Adittedly this is specifically the case of the last sequence being empty (which then triggers a bug if it also happens to be the first so `hts_getline` hasn't yet allocated a buffer), but if we accept empty sequences in FASTA elsewhere then I guess we should accept them at the end too.  Not sure if that's good or bad, but this fix is consistent in that behaviour at least.

PR #1510, opened 2022-10-11
bcf_strerror method to get description of error code

vasudeva8:
Changes to have description of error values through bcf_strerror method.

### DIFF
--- a/htslib/hts.h
+++ b/htslib/hts.h
@@ -1574,6 +1574,17 @@ static inline void *ed_swap_8p(void *x)
     return x;
 }
 
+
+/*
+error description structure definition
+modules can have their own error descriptions until an unified error handling is in place
+*/
+#define MAX_ERR_DESC 255    //max description size
+typedef struct err_desc {
+    int  errorcode;
+    char description[ MAX_ERR_DESC + 1 ];
+}err_desc;
+
 #ifdef __cplusplus
 }
 #endif
--- a/htslib/hts.h
+++ b/htslib/hts.h
@@ -1574,17 +1574,6 @@ static inline void *ed_swap_8p(void *x)
     return x;
 }
 
-
-/*
-error description structure definition
-modules can have their own error descriptions until an unified error handling is in place
-*/
-#define MAX_ERR_DESC 255    //max description size
-typedef struct err_desc {
-    int  errorcode;
-    char description[ MAX_ERR_DESC + 1 ];
-}err_desc;
-
 #ifdef __cplusplus
 }
 #endif
--- a/htslib/vcf.h
+++ b/htslib/vcf.h
@@ -205,6 +205,9 @@ typedef struct bcf_dec_t {
 #define BCF_ERR_CTG_INVALID   32
 #define BCF_ERR_TAG_INVALID   64
 
+// get the error descriptions
+const char* bcf_strerror(int errorcode, char* buffer, size_t maxbuffer);
+
 /*
     The bcf1_t structure corresponds to one VCF/BCF line. Reading from VCF file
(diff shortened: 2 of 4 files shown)


daviesrob (changes requested):
This works OK, but could do with a couple of minor changes.

While testing this PR, I noticed that [`vcf_parse_format()`](https://github.com/samtools/htslib/blob/853b8017/vcf.c#L2490) and [`vcf_parse_info()`](https://github.com/samtools/htslib/blob/853b8017/vcf.c#L2920) both set `v->errcode = BCF_ERR_TAG_UNDEF;` instead of `v->errcode |= BCF_ERR_TAG_UNDEF;`, which I think is an oversight because they wipe out any existing error codes.  It might be an idea to add an extra commit to this PR that changes them to use `|=`.

htslib/vcf.h
```diff
@@ -205,6 +205,10 @@ typedef struct bcf_dec_t {
 #define BCF_ERR_CTG_INVALID   32
 #define BCF_ERR_TAG_INVALID   64
 
+// get the error descriptions
+HTSLIB_EXPORT
+const char* bcf_strerror(int errorcode, char* buffer, size_t maxbuffer);
```
jmarshall, line 210: For consistency with other code nearby, pointer declarations should be in the C style rather than the C++ style:

```suggestion
const char *bcf_strerror(int errorcode, char *buffer, size_t maxbuffer);
```

jkbonfield: I've always wondered where that meme came from.

It never made sense to me, and indeed it's positively dangerous.  To a naive reader the spacing implies that `char* a,b;` is a pair of char pointers, which is plainly not true.  Glad to know I can also blame it on C++ now :-)

jmarshall: The style came hand in hand with another C++ style rule that says “never declare more than one variable in the same declaration” — which avoids the danger (provided you follow it).

htslib/hts.h
```diff
@@ -1574,6 +1574,17 @@ static inline void *ed_swap_8p(void *x)
...
+/*
+error description structure definition
+modules can have their own error descriptions until an unified error handling is in place
+*/
+#define MAX_ERR_DESC 255    //max description size
+typedef struct err_desc {
+    int  errorcode;
+    char description[ MAX_ERR_DESC + 1 ];
+}err_desc;
+
```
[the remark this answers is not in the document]
vasudeva8: the structure is updated and moved to vcf.c as it is used internally.

htslib/vcf.h
```diff
@@ -205,6 +205,10 @@ typedef struct bcf_dec_t {
 #define BCF_ERR_CTG_INVALID   32
 #define BCF_ERR_TAG_INVALID   64
 
+// get the error descriptions
```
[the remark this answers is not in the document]
vasudeva8: comments added as per the reference

htslib/vcf.h
vasudeva8: updated pointer declaration to c style

vasudeva8:
the error reset in vcf_parse_format and vcf_parse_info are modified to retain previous error code

htslib/hts.h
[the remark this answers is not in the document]
vasudeva8: the error reset in vcf_parse_format and vcf_parse_info are modified to retain previous error code

vasudeva8:
Added handling for error codes without description, to show unknown error

daviesrob:
I've added a couple of commits to adjust the documentation slightly, and pull in the libdeflate fix from develop.

I'll squash and merge everything shortly.

daviesrob (approved):

PR #1516, opened 2022-10-20
Add an fai_line_length function.

jkbonfield:
This exposes the input datas line wrap length (minus white-space / newlines).

Also fixed a SEGV in fai_retrieve given incorrect or malicious .fai files with a zero length "blen" figure.

### DIFF
--- a/faidx.c
+++ b/faidx.c
@@ -702,6 +702,12 @@ static char *fai_retrieve(const faidx_t *fai, const faidx1_t *val,
         return NULL;
     }
 
+    if (val->line_blen <= 0) {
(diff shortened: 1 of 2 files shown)


daviesrob:
Looks OK, although I spotted a couple of TODOs for later.
* `fai_get_val()` `fbeg` and `fend` are `hts_pos_t *`, but everything calling it passes `int64_t *`.  This affects more then the new function, so probably best tidied up separately.
* Currently, zero-length sequences get excluded from the index, but if they were present they would presumably have all lengths as zero and would upset the new `blen` check.  I think there may be other issues around including such horrors, so I don't think it's a problem at the moment.

PR #1519, opened 2022-11-02
Add faidx_seq_len64(), fai_adjust_region() interfaces and faidx tests

daviesrob:
Add `faidx_seq_len64()` as a replacement for `faidx_seq_len()` that can return the correct length of sequences longer than `INT_MAX`.

Make `faidx_seq_len()` clamp its output at `INT_MAX`, which is probably slightly less bad than overflowing.

Adds a `fai_adjust_region()` function that can be used to ensure that a given range does not go beyond the end of the requested sequence.  The interface is designed so that the output of `fai_parse_region()` can be passed to it.  This essentially exposes the internal `faidx_adjust_position()` function which is currently used to enforce the same limits in the `faidx_fetch_seq64()` and `faidx_fetch_qual64()` interfaces.  The new function allows callers to get a better idea of what will be retrieved by applying the limits in advance.

While writing this, I discovered that faidx didn't really have any tests of its own, although some were run as side-effects of other tests.  The second (rather bigger) commit adds some dedicated faidx tests.


test/faidx/test-faidx.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+#
+#    Copyright (C) 2017-2018 Genome Research Ltd.
```
whitwham, line 3: Date?

PR #1529, opened 2022-11-23
bgzf_read_block added error return

bir3:


bir3:
fixed memory leak

bir3:
Ready for review, it would be good to get experienced eyes on this.

Not sure about desired semantics of single-thread vs. multi-thread in the presence of a truncated file:  
- One view is to say that it should yield identical results, e.g. valid blocks are returned until we hit a truncated block.  
- One can also say that in the presence of a truncated file it is valid to return error early and not return some valid blocks.  

The fix as written adheres to agreeing with the single-threaded behaviour, e.g. it will return all valid blocks until it hits the truncated block.  The test is however relaxed and allows both cases.  So it could be tightened if the decision is to favor agreeing with single-threaded behaviour.

We might also want to add test for EOF block in the presence of a truncated file.

whitwham:
> Not sure about desired semantics of single-thread vs. multi-thread in the presence of a truncated file

We had a quick discussion here and I think we are happy to have the multi-threaded run return an error early.  We don't really need identical behaviour between single and multi-threaded on a failure.

bir3:
Changed to exit early on error - as the code did before.  Ready for review and merging if deemed acceptable.

whitwham:
Thank you for your fix.

PR #1530, opened 2022-11-24
Ensure strings in config_vars.h are escaped correctly

### LINKED ISSUE #1527: Preprocessor definitions with quotation marks
Hi there,

I'm having a weird issue in the interaction of my GitHub Actions CI environment and htslib.

The htslib `Makefile` defines [some preprocessor values](https://github.com/samtools/htslib/blob/8e43fb0650fabb8ca34775510e1e33ce6e48fc2a/Makefile#L165-L169) using quotation marks such as

    echo '#define HTS_LDFLAGS "$(LDFLAGS)"' >> $@

For my CI, I'm using GitHub Actions with the [setup-cpp](https://github.com/marketplace/actions/setup-cpp-c-c) action to access various compilers. For their clang/llvm setup however, they also [use quotation marks around flags](https://github.com/aminya/setup-cpp/blob/8c9d48b9aa22b181d0ff81b04d3957e19462fb00/src/llvm/llvm.ts#L74-L75) such as

    addEnv("LDFLAGS", `-L"${directory}/lib"`),

which causes the preprocessor value to become for example

    #define HTS_LDFLAGS "-L"/home/runner/llvm/lib" -fvisibility=hidden "

That of course fails to compile, as the path in there is not part of the quoted string. 

I'm currently working around this by having

    export LDFLAGS=`echo ${LDFLAGS} | sed 's/"//g'`
    export CPPFLAGS=`echo ${CPPFLAGS} | sed 's/"//g'`

in my CI setup, but that does not seam ideal.

Honestly, I'm not entirely sure who is right here. On the one hand, putting a path in quotation marks seems okay on the setup-cpp side. On the other hand, having flags in a quoted string seems necessary from the point of view of htslib. My guess is that htslib should escape quotation marks within the `$(LDFLAGS)` when processing the `Makefile`. Still, I've [opened an issue](https://github.com/aminya/setup-cpp/issues/145) on setup-cpp's end as well, as I am not sure that paths should necessarily be quoted.

Cheers and thanks
Lucas

daviesrob:
This is more tricky than it first appears as the strings in question have to make it through make expansion, shell string expansion, interpretation by the command writing the file, and end up making sense to the C compiler.

The values are escaped by converting `\` to `\\`, `"` to `\"` and `'` to `\x27` (as it's impossible to quote a `'` in a single-quoted string).  Writing the file is now done by `printf`, which won't try to interpret any of the strings being printed as an option.  Its arguments are all single-quoted to prevent the shell from trying to expand them in an unwanted manner.

Fixes #1527

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,12 +161,20 @@ show-version:
 	@echo PACKAGE_VERSION = $(PACKAGE_VERSION)
 	@echo NUMERIC_VERSION = $(NUMERIC_VERSION)
 
+config_vars.h: override escape=$(subst ',\x27,$(subst ",\",$(subst \,\\,$(1))))
+config_vars.h: override hts_cc_escaped=$(call escape,$(CC))
+config_vars.h: override hts_cppflags_escaped=$(call escape,$(CPPFLAGS))
+config_vars.h: override hts_cflags_escaped=$(call escape,$(CFLAGS))
+config_vars.h: override hts_ldflags_escaped=$(call escape,$(LDFLAGS))
+config_vars.h: override hts_libs_escaped=$(call escape,$(LIBS))
+
 config_vars.h:
-	echo '#define HTS_CC "$(CC)"' > $@
-	echo '#define HTS_CPPFLAGS "$(CPPFLAGS)"' >> $@
-	echo '#define HTS_CFLAGS "$(CFLAGS)"' >> $@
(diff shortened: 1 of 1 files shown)


jkbonfield:
I'm a bit uneasy about requiring a command-line printf tool, but it seems to be a defined part of the POSIX environment so I assume it's now everywhere that sed, awk, etc are.  Certainly for our own CI.

PR #1532, opened 2022-12-05
Make it easier to modify shared library permissions during install

jkbonfield:
The new LIB_PERM Makefile variable defaults to 644 (as before), but can now be amended to e.g. 755.  Updated the INSTALL file with a section for HP-UX explaining how to do this.

Fixes #1525


jmarshall:
It can already be equivalently overridden with `make install INSTALL_LIB='$(INSTALL)'`, but I guess this avoids the need for quoting there.

jkbonfield:
Indeed, and I think I noted that in the issue.  The comment was to make it "easier", which this most definitely does.  Both easier for users, and inevitably easier for us as (assuming anyone actually uses it) we should get fewer issues caused by incorrect punctuation.

PR #1535, opened 2022-12-08
Fix htslib 1533

pd3:


vcf.c
```diff
@@ -2214,20 +2214,37 @@ char *bcf_hdr_fmt_text(const bcf_hdr_t *hdr, int is_bcf, int *len)
...
+            m = tid + 1;
+        }
         names[tid] = kh_key(d,k);
     }
-    // sanity check: there should be no gaps
-    for (tid=0; tid<m; tid++)
-        assert(names[tid]);
-    *n = m;
+    // ensure there are no gaps
+    for (i=0,tid=0; tid<m; i++,tid++)
```
daviesrob, line 2239: This potentially breaks the link between position in the output array and `bcf1_t::rid` in alignment records.  Of course the original code wouldn't have worked in the case of gaps existing so we're unlikely to break anything by doing this.  But it might be good to document in the header that the output from this function is not suitable for looking up chromosome names, and you should use something like `bcf_seqname()` to do that instead.

vcf.c
```diff
@@ -2214,20 +2214,37 @@ char *bcf_hdr_fmt_text(const bcf_hdr_t *hdr, int is_bcf, int *len)
...
+        if ( !kh_val(d, k).hrec[0] ) continue;  // removed via bcf_hdr_remove
         tid = kh_val(d,k).id;
-        assert( tid<m );
+        if ( tid >= m )
+        {
+            // This can happen after a contig has been removed from BCF header via bcf_hdr_remove()
+            if ( hts_resize(const char*, tid + 1, &m, &names, HTS_RESIZE_CLEAR)<0 )
+            {
+                hts_log_error("Failed to allocate memory");
+                *n = 0;
```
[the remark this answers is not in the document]
pd3: Good point

vcf.c
pd3: I forced pushed the changes

PR #1537, opened 2022-12-14
Cut down the number of embed_ref=2 tests that get run

daviesrob:
Fixes a small testing inefficiency I noticed.

The test only runs on the ce#1000.sam file, so it shouldn't be inside the loop that runs on all sam files in the test directory.


jkbonfield:
Oops.  Indeed that's a silly one.  Thanks :)

Issue #1539, opened 2022-12-16
int overflow for big variant lines in vcf


arnaldurgylfason:
We have a variant with more than 50 alleles for almost 500000 samples. The line in question is over 3 gb in size.
bcftools and our own program that uses htslib can not read it.
After debugging I noticed, at least for .vcf instead of .vcf.gz, that hts_getline seemed to read it but it returned the length read which was over 3gb and since the returned value was captured in an int (4 bytes), (function _reader_fill_buffer in synced_bcf_reader.c)
it resulted in an overflow giving a negative number which was interpreted as failure.
Bigger datasets becoming more common in the future this problem will occur more often.
I´m sure this overflow problem can be found at more places.
I would consider it important to solve this and allow for bigger lines to handle the big data that we have today and even bigger in the near future. Changing to 8 byte integers would solve this. Note, size_t is used at many places, which handles this.

jmarshall:
See also samtools/bcftools#1614 and #1448.

arnaldurgylfason:
Thanks. We will start using FORMAT/LPL and FORMAT/LAA for variants with many alleles.
I still think htslib should be able to handle bigger lines though

daviesrob:
It does look like `bgzf_getline()` has the [same overflow problem](https://github.com/samtools/htslib/blob/d7737aa6/bgzf.c#L2301).  It should be easy to fix in the same way as #1448 did for uncompressed reads.

Note that this may not buy you that much due to other limitations in the library, for example [`bcf_fmt_t::p_len` and `bcf_fmt_t::p_off`](https://github.com/samtools/htslib/blob/d7737aa6/htslib/vcf.h#L160).  As these are part of the public interface we can't easily change them.  I'm also not entirely sure we'd want to as by the time individual records start getting to multi-gigabyte sizes you have to consider if you've reached the limit of what can sensibly be done with flat files, and consider solutions that are more database oriented (for example, [hail](https://github.com/hail-is/hail)).

pd3:
It would be nice to have htslib print a warning and skip such entries. We cannot support such large records fully as that breaks too many things and is not worth it. The feasible solution was proposed above, generate LPL and LAA instead.

PR #1542, opened 2023-01-04
Improve support for very long input lines (> 2Gbyte)

daviesrob:
Some changes to make reading and writing very long lines work better.  It is a partial fix for #1539 - the overflow mentioned in dealt with, but more work will be needed to make the very long VCF records readable.

* Cap the return value of `bgzf_getline()` to avoid integer overflow.
* Change data types in `tbx_parse1()` and `vcf_write()` so they can handle long lines (albeit only on 64-bit platforms).
* Put an extra loop around `read()` and `write()` in `fd_read()`, `fd_write()` to ensure they really have processed the expected amount of data.  This is especially important on Linux, which has a [limit on the amount of data that will be read or written](https://man7.org/linux/man-pages/man2/read.2.html#NOTES) in a single call.

This is mostly useful for `tabix`, which doesn't do much interpretation of its input.  With these changes it will happily index and return long lines if it has enough memory.

This does not change the maximum size of a SAM record, as that is limited by `bam1_t::l_data` which is an `int`.
The situation for VCF is a bit more complicated, and it may be possible to get very slightly over 2Gbytes as various limits apply to different parts of the record.  The size of the sample data, which is likely to be the biggest part, is currently restricted to 2Gbytes by [this check](https://github.com/samtools/htslib/blob/d7737aa694/vcf.c#L2612), and by the size of `bcf_fmt_t::p_off` which is 31 bits.  It might be possible to work around the `p_off` limit by abusing the ability of `bcf_fmt_t` to store edited data in a separate buffer - but doing so would be a bit hairy and would need a lot of thought and testing.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -2298,7 +2298,7 @@ int bgzf_getline(BGZF *fp, int delim, kstring_t *str)
     fp->uncompressed_address += str->l + 1;
     if ( delim=='\n' && str->l>0 && str->s[str->l-1]=='\r' ) str->l--;
     str->s[str->l] = 0;
-    return str->l;
+    return str->l <= INT_MAX ? (int) str->l : INT_MAX;
 }
 
 void bgzf_index_destroy(BGZF *fp)
--- a/htslib/bgzf.h
+++ b/htslib/bgzf.h
@@ -303,7 +303,8 @@ typedef struct BGZF BGZF;
      * @param fp     BGZF file handler
      * @param delim  delimiter
      * @param str    string to write to; must be initialized
-     * @return       length of the string; -1 on end-of-file; <= -2 on error
+     * @return       length of the string (capped at INT_MAX);
+     *               -1 on end-of-file; <= -2 on error
      */
     HTSLIB_EXPORT
     int bgzf_getline(BGZF *fp, int delim, struct kstring_t *str);
--- a/tbx.c
+++ b/tbx.c
@@ -91,9 +91,10 @@ int tbx_name2id(tbx_t *tbx, const char *ss)
     return get_tid(tbx, ss, 0);
 }
 
-int tbx_parse1(const tbx_conf_t *conf, int len, char *line, tbx_intv_t *intv)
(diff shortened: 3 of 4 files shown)


jmarshall:
Re the `fd_read()`/`fd_write()` change: as documented in `struct hFILE_backend` in _hfile_internal.h_, it is expected that these functions might not read or write their entire buffers.

For `fd_write()`, it is always called via `hFILE_backend::write()`. There are two invocations of that in the _hfile.c_ front-end code, and both of them implement such loops to ensure all data is written.

For `fd_read()`, it is always called via `hFILE_backend::read()`. There are three invocations of that: that in `hread2()` implements a loop to ensure the bulk of a large request is read; that in `refill_buffer()` does not have a loop but records the number of bytes read, and **its** callers loop as necessary; the final one is in `multipart_read()`, which is itself an `hFILE_backend::read()` method that will be recalled in a loop as necessary by the front-end functions.

So it should not be necessary to have this sort of loop inside the back-end functions. Did you encounter a bug without this commit? If so, that would seem to signal a bug in the front-end code instead.

Moreover, as mentioned in the `refill_buffer()` doc string, it only calls the back-end reader **once** and may not completely refill the buffer. It's implied :smile: that the back-end function is only going to do a system call **once** too. This was intentional, written this way with interactively typing a SAM file into a terminal in mind. In that scenario, IIRC each `read(2)` system call would return one line after the user pressed Enter. Doing only one `read(2)` system call for each `sam_read1()` call means that `sam_read1()` does not block when reading from an interactive terminal, which seemed desirable at the time. (Similar considerations might apply for reading from the network.)

(I don't recall whether this was described in the commit messages of the time.)

daviesrob:
Yes, you're right.  It does work without the `fd_read()/fd_write()` commit.  I possibly thought they were a problem because I hadn't fixed the `vcf_write()` type at that point, so the whole process still wasn't working and I maybe didn't go far enough in the debugger.

I'll remove that commit.

PR #1545, opened 2023-01-09
Don't error when making an iterator on a tid not in the index

### LINKED ISSUE #1534: hts_itr_query fails on empty BCFs with gapped tid blocks
This is related to https://github.com/samtools/htslib/issues/1533 and concerns BCFs with edited headers and missing data records. Consider this example
```
$ echo -e '##fileformat=VCFv4.2\n##contig=<ID=chr22,length=51304566,IDX=1>\n#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\nchr22\t1\t.\tA\tC\t.\t.\t.' > test.vcf
$ bcftools view test.vcf -o test.bcf
$ bcftools index test.bcf
```
Note the header line contains the field `IDX=1` which makes it behave as if BCF was edited and the first chromosome with `IDX=0` was removed.

Prior to the commit [d64e710](https://github.com/pd3/htslib/commit/d64e710e891be3e5ecd89490d039228075172c9c) this command fails with
```
$ bcftools isec -n 2 test.bcf test.bcf
bcftools: vcf.c:2223: bcf_hdr_seqnames: Assertion `tid<m' failed.
Aborted
```
with the fix applied, it works
```
$ bcftools-d64e710 isec -C test.bcf test.bcf
chr22	1	A	C	11
```

However, if there are no data records, `hts_itr_query` returns an error and the program fails
```
$ echo -e '##fileformat=VCFv4.2\n##contig=<ID=chr22,length=51304566,IDX=1>\n#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO' > test2.vcf
$ bcftools view test2.vcf -o test2.bcf
$ bcftools index test2.bcf
$ bcftools isec -n 2 test2.bcf test2.bcf
[E::_reader_seek] Could not seek: chr22:1-17592186044415
bcftools: synced_bcf_reader.c:532: _reader_seek: Assertion `0' failed.
Aborted
```

This problem does not appear when the chromosome tid block has no gaps, i.e. starts with IDX=0
```
$ echo -e '##fileformat=VCFv4.2\n##contig=<ID=chr22,length=51304566,IDX=0>\n#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO' > test3.vcf
$ bcftools view test3.vcf -o test3.bcf
$ bcftools index test3.bcf
$ bcftools isec -n 2 test3.bcf test3.bcf
```

daviesrob:
Don't error when making an iterator on a tid not in the index, instead return one that instantly finishes.

Fixes #1534, which was an edge case where an index did not include an entry for a chromosome that was mentioned in the file header but had no data records.  Normally these would be present but empty, but it was possible to use the `IDX=` key in a VCF file to make an index where the chromosome simply did not appear.  In this case, rather than an error, we want to return the equivalent of `HTS_IDX_NONE` so the iterator produces no data.

Another scenario where this is useful is if you build an index, and then try to use it immediately without first saving and reading it back in again.  Such an index will have NULL entries in `bidx[]` for any chromosomes with no data.  Again we want to return an `HTS_IDX_NONE` iterator if one of those chromosomes is queried. (This issue didn't usually occur because most programs are loading in an existing index, and `idx_read_core()` makes `bidx[]` entries for everything even if there's nothing in the index for the chromosome.)

Note that this changes `vcf_loop()` in `test_view.c` so that it now treats `bcf_itr_querys()` failures as an error.  The new behaviour matches `sam_loop()` and is needed to detect the problem being fixed here.  All the other tests still work after this change so nothing was relying on the old behaviour of ignoring the errors.


jkbonfield:
Merged as it works and fixes a problem, but I do wonder if there are more things wrong than this in #1534.

It sounds like we have a way to create an index for a BCF file where chromosomes don't appear in the index.  This change makes us no longer fail with such indices, but it doesn't fix the underlying problem that we still have indices with missing chromosomes.  Our code works, but what about other peoples code?

Is it a hard requirement that indices must contain an entry for every chromosome in the header, irrespective of whether they have any coverage?  I looked at the VCF spec but it has a woeful single sentence describing indices! (Basically "same as BAM" which is a bit misleading.)  The BAM spec doesn't explicitly state that all references have BAI entries, but experimentally it does appear to be the case.  If BCF eqvuialent doesn't do this then perhaps it is still in error.

Do we know whether Picard would work on such data without fixing our indices to include them?

daviesrob:
Having index entries with no data is normal (`n_bin` in the entry would be zero; look at the [diagram in the samtools spec.](https://samtools.github.io/hts-specs/SAMv1.pdf#subsection.5.2) for context).  The difference in this case is that the `tid` value was higher than `n_ref`, so it wouldn't appear in the list of indicies.  It's possible that Picard might not like that, but you do have to try quite hard to make an index that does this so it's unlikely many exist in practice - and as the way found involved indexing a file with no variant records such files are unlikely to be much use.

https://github.com/samtools/htslib/issues/1534#issuecomment-1351496991 mentions that we might want to fix the index building side.  I could try to make a PR for that if you'd like a belt-and-braces approach.

jkbonfield:
I know this is old, but in experimenting with fixing test_view on VCF.gz I've concluded this fix is in error for VCF.  (It may help BCF however.)  Specifically https://github.com/samtools/htslib/blob/develop/test/test.pl#L1158-L1161 adds a check to query a chromosome in a VCF file that's empty of records.

VCF.gz.csi builds a CSI index using tabix for chromosome meta-data instead of the more normal CSI/BCF route for chromosome meta-data.  Hence chr1 to chr22 with only chr10 present would store tid=0 and not tid=9 as we'd normally expect with BCF, BAM or SAM indices.  I don't know why we chose to do this, but for whatever reason we did.  I also don't know if we fixed it whether there's even a way to distinguish between CSI-binary and CSI-text index formats, as it's all hidden behind the undocumented uint8_t[l_aux] spec component so I'd need to go grubbing around the code to see if there's a hidden secondary magic number somewhere.

Anyway, with a real VCF example on develop, using bcftools:

```
$ cat _.vcf
##fileformat=VCFv4.2
##contig=<ID=1,length=100>
##contig=<ID=2,length=100>
##contig=<ID=3,length=100>
1	50	.	G	A	0	.	.
3	50	.	A	C	0	.	.
$ bcftools view -Oz -o _.vcf.gz --write-index _.vcf
$ bcftools view -Ob -o _.bcf --write-index _.vcf

$ for i in 1 2 3;do echo === $i ===;bcftools view -H _.vcf.gz $i:50;done
=== 1 ===
1	50	.	G	A	0	.	.
=== 2 ===
=== 3 ===
3	50	.	A	C	0	.	.

$ for i in 1 2 3;do echo === $i ===;bcftools view -H _.bcf $i:50;done
=== 1 ===
1	50	.	G	A	0	.	.
=== 2 ===
=== 3 ===
3	50	.	A	C	0	.	.
```

That works fine.

What about test/test_view?

```
$ for i in 1 2 3;do echo === $i ===;./test/test_view _.bcf $i:50 | grep -v '^#
';done
=== 1 ===
1	50	.	G	A	0	.	.
=== 2 ===
=== 3 ===
3	50	.	A	C	0	.	
.
$ for i in 1 2 3;do echo === $i ===;./test/test_view _.vcf.gz $i:50 | grep -v '^#';done
=== 1 ===
Error reading input.
=== 2 ===
Error reading input.
=== 3 ===
```

It dismally fails!  Why?  Because we're using the wrong API.  Bcftools uses the synced reader.   That uses two different APIs depending on whether it's using a tabix index or a non-tabix index.  Importantly it has to be able to distinguish between indices that count all chromosomes recorded in the header (`bcf_hdr_name2id`) vs one that counts only chromosomes with records in the file (`tbx_name2id`).  So chr "3" would be tid=2 for BCF and tid=1 for VCF.gz in the above example due to CSI-TBI rather than CSI-BCF meta-data.  [See the synced reader](https://github.com/samtools/htslib/blob/develop/synced_bcf_reader.c#L572-L582) for the code that does this.

So what about our test in question?  Well it's a VCF.gz with a CSI-TBI index and we're querying a chromosome that doesn't exist in the index.  It *should* fail, but our test requires that it silently passes.

I fixed test_view so that it could do random access on VCF.gz, but it's a right mess frankly as it needs two different sets of API calls.  Sure enough, it can now handle my example code above, but it fails on the test added in this PR.  My inclination is that it's meant to fail, but I'm not sure therefore what this test was attempting to test and why we ever expected it to succeed.

Maybe it should always just silently ignore the errors, as it used to, given that it's not really obvious whether or not it should have succeeded.  Alternatively it could fall back to using `bcf_hdr_name2id` when `tbx_name2id` fails so we can distinguish between error and exists-but-is-empty cases.  However this breaks the API as they need different data types.

PR #1546, opened 2023-01-11
Expand CRAM API a bit to cope with new samtools cram_size command.

jkbonfield:
- cram_block_get_content_id is updated to return -1 for CORE block, removing the need to add a content_type API and share that enum.

- New cram_block_get_method method

- New cram_codec_get_content_ids, which externalises the existing cram_codec_to_id (with a hopefully better API and name).

- Make cram_decode_compression_header and cram_free_compression_header external.  The struct itself is opaque.

- Added cram_update_cid2ds_map, cram_cid2ds_query and cram_cid2ds_free to produce, query and free a block Content-ID to Data-Series mapping from the meta-data held within a CRAM Compression Header.

- Added a new cram_codec iterator to sequentially step through all cram_codec structures produced by decoding a compression header. This new code is entirely internal to htslib and is solely used by the new cram_update_cid2ds_map function.

- Add a codec->describe method. This produces a text description of codec parameters, such as

    EXTERNAL(id=1)
    HUFFMAN(codes={5,1,3},lengths={1,2,2})
    BYTE_ARRAY_LEN(len_codec={EXTERNAL(id=42)},val_codec={EXTERNAL(id=37)}

  Adds the external API: int cram_codec_describe(cram_codec *c, kstring_t *ks) TODO: Some of the proposed CRAM 4.0 codecs don't yet have descriptions, but these aren't official codecs anyway.

- Add a function to describe all encodings in a compression header block: int cram_describe_encodings(cram_block_compression_hdr *hdr, kstring_t *ks);

- Add cram_container_get_num_records and cram_container_get_num_bases functions. These are two new external APIs for querying sequence meta-data from containers, which in theory could be used for a (non-filtering) "view -c" cmd.

### DIFF
--- a/cram/cram_external.c
+++ b/cram/cram_external.c
@@ -475,9 +475,9 @@ int cram_describe_encodings(cram_block_compression_hdr *hdr, kstring_t *ks) {
         if (key>>16) key_s[key_i++] = key>>16;
         key_s[key_i++] = (key>>8)&0xff;
         key_s[key_i++] = key&0xff;
-        ksprintf(ks, "\t%s\t", key_s);
+        r |= ksprintf(ks, "\t%s\t", key_s) < 0;
         r |= cram_codec_describe(codec, ks) < 0;
-        kputc('\n', ks);
+        r |= kputc('\n', ks) < 0;
     }
 
     return r ? -1 : 0;
@@ -555,31 +555,27 @@ void cram_block_set_offset(cram_block *b, size_t offset) { BLOCK_SIZE(b) = offse
  * Retuns the detected or specified comp method, and fills out *cm
  * if non-NULL.
  */
-enum cram_block_method cram_expand_method(uint8_t *data, int32_t size,
-                                          enum cram_block_method comp,
-                                          cram_method_details *cm) {
-    if (cm) {
-        memset(cm, 0, sizeof(*cm));
-        cm->method = comp;
-    }
+cram_method_details *cram_expand_method(uint8_t *data, int32_t size,
+                                        enum cram_block_method comp) {
+    cram_method_details *cm = calloc(1, sizeof(*cm));
+    if (!cm)
+        return NULL;
 
     const char *xz_header = "\xFD""7zXZ"; // including nul
 
     if (comp == CRAM_COMP_UNKNOWN) {
         // Auto-detect
-        if (size >= 2 && data[0] == 0x1f && data[1] == 0x8b)
+        if (size > 1 && data[0] == 0x1f && data[1] == 0x8b)
             comp = CRAM_COMP_GZIP;
-        else if (size >= 3 && data[1] == 'B' && data[2] == 'Z'
+        else if (size > 3 && data[1] == 'B' && data[2] == 'Z'
                  && data[3] == 'h')
             comp = CRAM_COMP_BZIP2;
-        else if (size >= 6 && memcmp(xz_header, data, 6) == 0)
+        else if (size > 6 && memcmp(xz_header, data, 6) == 0)
             comp = CRAM_COMP_LZMA;
         else
-            return CRAM_COMP_UNKNOWN;
+            comp = CRAM_COMP_UNKNOWN;
     }
-
-    if (!cm)
-        return comp;
+    cm->method = comp;
 
     // Interrogate the compressed data stream to fill out additional fields.
     switch (comp) {
@@ -645,7 +641,7 @@ enum cram_block_method cram_expand_method(uint8_t *data, int32_t size,
         break;
     }
 
-    return comp;
+    return cm;
 }
 
 /*
--- a/cram/cram_external.c
+++ b/cram/cram_external.c
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2015, 2018-2020, 2022 Genome Research Ltd.
+Copyright (c) 2015, 2018-2020, 2022-2023 Genome Research Ltd.
 Author: James Bonfield <jkb@sanger.ac.uk>
 
 Redistribution and use in source and binary forms, with or without
@@ -80,6 +80,14 @@ int32_t cram_container_get_num_blocks(cram_container *c) {
 
 void cram_container_set_num_blocks(cram_container *c, int32_t num_blocks) {
     c->num_blocks = num_blocks;
+}
+
+int32_t cram_container_get_num_records(cram_container *c) {
+    return c->num_records;
+}
+
+int32_t cram_container_get_num_bases(cram_container *c) {
+    return c->num_bases;
 }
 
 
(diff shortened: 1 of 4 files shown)


cram/cram_external.c
```diff
@@ -206,12 +502,15 @@ void cram_slice_hdr_get_coords(cram_block_slice_hdr *h,
...
-int32_t cram_block_get_content_id(cram_block *b)  { return b->content_id; }
+int32_t cram_block_get_content_id(cram_block *b)  {
+    return b->content_type == CORE ? -1 : b->content_id;
+}
 int32_t cram_block_get_comp_size(cram_block *b)   { return b->comp_size; }
 int32_t cram_block_get_uncomp_size(cram_block *b) { return b->uncomp_size; }
 int32_t cram_block_get_crc32(cram_block *b)       { return b->crc32; }
 void *  cram_block_get_data(cram_block *b)        { return BLOCK_DATA(b); }
 int32_t cram_block_get_size(cram_block *b)        { return BLOCK_SIZE(b); }
+int     cram_block_get_method(cram_block *b)      { return b->orig_method; }
```
daviesrob, line 513: If exposing this data, you also need to make (a subset of?) [`enum cram_block_method_int`](https://github.com/samtools/htslib/blob/f56be61b91bf/cram/cram_structs.h#L215) public so the value returned can be interpreted.  Otherwise you end up with something like [this](https://github.com/samtools/samtools/blob/c6a428f1f8/cram_size.c#L54), which adds a hidden dependency between the samtools and htslib code.

htslib/cram.h
```diff
@@ -147,6 +148,10 @@ int32_t *cram_container_get_landmarks(cram_container *c, int32_t *num_landmarks)
 HTSLIB_EXPORT
 void cram_container_set_landmarks(cram_container *c, int32_t num_landmarks,
                                   int32_t *landmarks);
+HTSLIB_EXPORT
+int32_t cram_container_get_num_records(cram_container *c);
+HTSLIB_EXPORT
+int32_t cram_container_get_num_bases(cram_container *c);
```
daviesrob, line 154: `int64_t`

daviesrob (commented):
Some of the functions in the header file could do with a bit more documentation.

Looking at the associated samtools PR samtools/samtools#1777, I wonder if more of the code should be moved here.  A good candidate would be the `cram_block_expand_method()` function as it would allow the messy grubbing around in data structures to be safely encapsulated in HTSlib.  And if we did that, I might also suggest returning a `struct` so that the codec description could be broken out to make it easier to interpret...

cram/cram_external.c
```diff
@@ -82,6 +82,14 @@ void cram_container_set_num_blocks(cram_container *c, int32_t num_blocks) {
     c->num_blocks = num_blocks;
 }
 
+int32_t cram_container_get_num_records(cram_container *c) {
+    return c->num_records;
+}
+
+int32_t cram_container_get_num_bases(cram_container *c) {
```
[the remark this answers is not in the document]
jkbonfield: Good spot.  I naively assumed it'd never be more than 32-bit because that would be an insanely large block (I think CRAM can't cope with that anyway), but CRAM 4 could. :)

cram/cram_external.c
```diff
@@ -180,6 +188,294 @@ int cram_block_compression_hdr_decoder2encoder(cram_fd *fd,
...
 }
 
+typedef struct {
+    cram_block_compression_hdr *hdr;
+    cram_map *curr_map;
+    int idx;
+    int is_tag; // phase 2 using tag_encoding_map
+} cram_codec_iter;
+
+void cram_codec_iter_init(cram_block_compression_hdr *hdr,
```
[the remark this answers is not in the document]
jkbonfield: Yes as it doesn't have an HTSLIB_EXPORT in the public header.  I'm surprised it didn't complain somewhere.  I thought we'd added a check for this in the CI.

cram/cram_external.c
jkbonfield: It's not really a dependency between samtools and htslib as they both have a dependency on the CRAM specification, which is where these numerical values are canonically defined.

However agreed it's cleaner to have a reduced enum.

jkbonfield: The issue of whether `cram_block_expand_method` should be public in htslib is an interesting one.  Maybe, but then again maybe it's an htscodecs thing as it doesn't need to take a cram_block and is a generic utility with just a memory buffer?  It could potentially even just have the original method as some "unknown" to make a best guess effort too (given many formats have a magic number).

I chose a single numeric as I'm using it to look up an array for char and string forms, but agreed maybe it just needs a struct and a couple functions that take that struct and turn it into the char/string forms.  It would be cleaner as we could have eg:

```
typedef struct {
    enum hts_codec_method method;
    int level;    // 1 or 9 for gzip min/max, and 1-9 for bzip2
    int order;    // rans4x8 and Nx16
    int rle;      // rans Nx16 specific
    int pack;
    int stripe;
    int cat;
    int Nway;
} hts_codec_method_t;
```

If we did that though, we'd need to make sure the `method` starts at least with the values used in the CRAM specification, even if it includes additional ones like zstd and bsc (which are definite possibilities for future CRAM, but getting a 3.1 2nd implementation was so hard I rolled back on the idea of including them).

jkbonfield:
I've made a public version of `enum cram_block_method`, avoiding the problems we had before with shortened enums that could clash.  This matches the CRAM specs and doesn't expand upon it to describe any of the specialisations of the codecs.

I'll think on this, but will probably end up with something in htscodecs and a returned structure as you suggest.  I may even add the short and long string form as elements to that struct too, or write dedicated functions.  Undecided atm.

jkbonfield:
I resolved it by backing out the changes to htscodecs and punted them to htslib instead.  Not ideal as it doesn't quite feel right, but it makes things vastly simpler for samtools and I don't think the minimal difference justifies the extra complexity with putting those functions in htscodecs.

htslib/cram.h
```diff
@@ -167,9 +218,15 @@ HTSLIB_EXPORT
...
-
 HTSLIB_EXPORT
 enum cram_content_type cram_block_get_content_type(cram_block *b);
+HTSLIB_EXPORT
+enum cram_block_method cram_block_get_method(cram_block *b);
+
+HTSLIB_EXPORT
+enum cram_block_method cram_expand_method(uint8_t *data, int32_t size,
+                                          enum cram_block_method comp,
+                                          cram_method_details *cm);
```
daviesrob, line 229: It would be useful to design this interface so `cram_method_details` could be expanded if necessary in the future (e.g. to better describe future codecs).  Currently it's difficult because the size of the `cram_method_details` struct is set by the caller, which locks in the ABI.  If instead we made this function allocate `cram_method_details` struct then it would be possible to safely add extra fields on the end without breaking old code.

Alternatively we could wrap up the real function so that we can pass on information about the interface the caller expects, e.g.:
```c
#define cram_method_details_vers 1
static inline enum cram_block_method cram_expand_method(uint8_t *data, int32_t size,
                                          enum cram_block_method comp,
                                          cram_method_details *cm) {
    HTSLIB_EXPORT
    extern enum cram_block_method cram_expand_method_(uint8_t *data, int32_t size,
                                              enum cram_block_method comp,
                                              cram_method_details *cm,
                                              int method_details_version);
    return cram_expand_method_(data, size, comp, cm, cram_method_details_vers);
}
```
or possibly like this:
```c
static inline enum cram_block_method cram_expand_method(uint8_t *data, int32_t size,
                                          enum cram_block_method comp,
                                          cram_method_details *cm) {
    HTSLIB_EXPORT
    extern enum cram_block_method cram_expand_method1(uint8_t *data, int32_t size,
                                              enum cram_block_method comp,
                                              cram_method_details *cm);
    return cram_expand_method1(data, size, comp, cm);
}
```
where we make `cram_expand_method2()` should the struct change, but leave `cram_expand_method1()` exported for the benefit of callers expecting the original version ... or we use symbol versioning (which this sort-of is), but I don't think that works on some platforms we support.

htslib/cram.h
```diff
@@ -200,6 +257,27 @@ void cram_block_set_offset(cram_block *b, size_t offset);
...
+ * -2 is unused.
+ * -1 is CORE
+ * >= 0 is the block with that Content ID
+ */
+HTSLIB_EXPORT
+void cram_codec_get_content_ids(cram_codec *c, int ids[2]);
+
+/*
+ * Produces a human readable description of the codec parameters.
+ * This ia appended to an existing kstring 'ks'.
```
daviesrob, line 273: "This is"

htslib/cram.h
jkbonfield: The versioning here is something that we don't bother with anywhere else.  The traditional htslib way of doing it is simply to add a new function with 2, 3, etc after it.  Eg `hgetc2`, `bcf_index_build3`, `sam_index_load3`, `kgetline2` etc.  We also have structures with numerics such as bam_pileup1_t`, but as yet haven't needed to create a new structure name.

I prefer this as the documentation is then always in sync with the API.  If we silently change the API and use versioning shenanigans to keep the ABI consistent, then there's a risk people use old API documentation that is not in sync with the library.  Explicit versioning via differing function / structure name is much easier to reliably document.

I think just cross the bridge when we come to it by adding a new function / structure name, as it's not likely to happen any time soon given these fields are based on the CRAM specification contents.

Issue #1543, opened 2023-01-05
VCF header parser has O(n^2) complexity in the number of generic header lines


freeseek:
The following bash script:
```
$ (echo "##fileformat=VCFv4.2"
for i in {1..65536}; do echo "##key=$i"; done
echo -e "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO") | \
  bcftools view -Ou -o /dev/null
```
Takes a very long time to run.

But the following bash script:
```
$ (echo "##fileformat=VCFv4.2"
for i in {1..65536}; do echo "##key=0"; done
echo -e "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO") | \
  bcftools view -Ou -o /dev/null
```
And the following bash script:
```
$ (echo "##fileformat=VCFv4.2"
for i in {1..65536}; do echo "##key=<ID=$i>"; done
echo -e "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO") | \
  bcftools view -Ou -o /dev/null
```
Run almost instantaneously.

I believe the problem is in `bcf_hdr_add_hrec()` from `htslib/vcf.c`:
```
int bcf_hdr_add_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec)
{
    int res;
    if ( !hrec ) return 0;

    bcf_hrec_check(hrec);   // todo: check return status and propagate errors up

    res = bcf_hdr_register_hrec(hdr,hrec);
    if (res < 0) return -1;
    if ( !res )
    {
        // If one of the hashed field, then it is already present
        if ( hrec->type != BCF_HL_GEN )
        {
            bcf_hrec_destroy(hrec);
            return 0;
        }

        // Is one of the generic fields and already present?
        int i;
        for (i=0; i<hdr->nhrec; i++)
        {
            if ( hdr->hrec[i]->type!=BCF_HL_GEN ) continue;
            if ( !strcmp(hdr->hrec[i]->key,hrec->key) && !strcmp(hrec->key,"fileformat") ) break;
            if ( !strcmp(hdr->hrec[i]->key,hrec->key) && !strcmp(hdr->hrec[i]->value,hrec->value) ) break;
        }
        if ( i<hdr->nhrec )
        {
            bcf_hrec_destroy(hrec);
            return 0;
        }
    }

    // New record, needs to be added
    int n = hdr->nhrec + 1;
[...]

PR #1548, opened 2023-01-13
Initial test framework for #1543 (to be continued)

pd3:
This is an initial test framework as suggested by https://github.com/samtools/htslib/issues/1543#issuecomment-1373575003

@daviesrob is this what you had in mind?

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -1,7 +1,7 @@
 /*  vcf.c -- VCF/BCF API functions.
 
     Copyright (C) 2012, 2013 Broad Institute.
-    Copyright (C) 2012-2022 Genome Research Ltd.
+    Copyright (C) 2012-2023 Genome Research Ltd.
     Portions copyright (C) 2014 Intel Corporation.
 
     Author: Heng Li <lh3@sanger.ac.uk>
@@ -51,6 +51,10 @@ DEALINGS IN THE SOFTWARE.  */
 KHASH_MAP_INIT_STR(vdict, bcf_idinfo_t)
 typedef khash_t(vdict) vdict_t;
 
+KHASH_MAP_INIT_STR(hdict, bcf_hrec_t*)
+typedef khash_t(hdict) hdict_t;
+
+
 #include "htslib/kseq.h"
 HTSLIB_EXPORT
 uint32_t bcf_float_missing    = 0x7F800001;
@@ -78,6 +82,22 @@ static bcf_idinfo_t bcf_idinfo_def = { .info = { 15, 15, 15 }, .hrec = { NULL, N
 
 #define BCF_IS_64BIT (1<<30)
 
+
+// Opaque structure with auxilary data which allows to extend bcf_hdr_t without breaking ABI.
+// Note that this preserving API and ABI requires that the first element is vdict_t struct
+// rather than a pointer, as user programs may (and in some cases do) access the dictionary
+// directly as (vdict_t*)hdr->dict.
+typedef struct
+{
+    vdict_t dict;   // bcf_hdr_t.dict[0] vdict_t dictionary which keeps bcf_idinfo_t for BCF_HL_FLT,BCF_HL_INFO,BCF_HL_FMT
+    hdict_t *gen;   // hdict_t dictionary which keeps bcf_hrec_t* pointers for generic and structured fields
+}
+bcf_hdr_aux_t;
+
+static inline bcf_hdr_aux_t *get_hdr_aux(const bcf_hdr_t *hdr)
+{
+    return (bcf_hdr_aux_t *)hdr->dict[0];
+}
 
 static char *find_chrom_header_line(char *s)
 {
@@ -866,8 +886,45 @@ static int bcf_hdr_register_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec)
     return 1;
 }
 
+int bcf_hdr_update_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec, const bcf_hrec_t *tmp)
+{
+    // currently only for bcf_hdr_set_version
+    assert( hrec->type==BCF_HL_GEN );
+    int ret;
+    khint_t k;
+    bcf_hdr_aux_t *aux = get_hdr_aux(hdr);
+    for (k=kh_begin(aux->gen); k<kh_end(aux->gen); k++)
+    {
+        if ( !kh_exist(aux->gen,k) ) continue;
+        if ( hrec!=(bcf_hrec_t*)kh_val(aux->gen,k) ) continue;
+        break;
+    }
+    assert( k<kh_end(aux->gen) );   // something went wrong, should never happen
+    free((char*)kh_key(aux->gen,k));
+    kh_del(hdict,aux->gen,k);
+    kstring_t str = {0,0,0};
+    if ( ksprintf(&str, "##%s=%s", tmp->key,tmp->value) < 0 )
+    {
+        free(str.s);
+        return -1;
+    }
+    k = kh_put(hdict, aux->gen, str.s, &ret);
+    if ( ret<0 )
+    {
+        free(str.s);
+        return -1;
+    }
+    free(hrec->value);
+    hrec->value = strdup(tmp->value);
+    if ( !hrec->value ) return -1;
+    return 0;
+}
+
 int bcf_hdr_add_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec)
 {
+    kstring_t str = {0,0,0};
+    bcf_hdr_aux_t *aux = get_hdr_aux(hdr);
+
     int res;
     if ( !hrec ) return 0;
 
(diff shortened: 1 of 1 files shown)


daviesrob:
Yes, something like that, but you currently leak a hash table.  The allocation code should I think look like this:
```c
    h->dict[0] = calloc(1,sizeof(bcf_hdr_aux_t));
    if (!h->dict[0]) goto fail;
    for (i = 1; i < 3; ++i)
        if ((h->dict[i] = kh_init(vdict)) == NULL) goto fail;
```

And we'd need to take care that the code to free it up does the right thing with the franken-structure.

We should also add a static inline function to get the `bcf_hdr_aux_t` out of the header, so we can hide away the grotty details, e.g.:

```c
static inline bcf_hdr_aux_t * get_hdr_aux(bcf_hdr_t *h) {
    return (bcf_hdr_aux_t *) h->dict[0];
}
```

vcf.c
```diff
@@ -1206,6 +1214,12 @@ bcf_hdr_t *bcf_hdr_init(const char *mode)
     if (!h) return NULL;
     for (i = 0; i < 3; ++i)
         if ((h->dict[i] = kh_init(vdict)) == NULL) goto fail;
+
+    bcf_hdr_aux_t *aux = (bcf_hdr_aux_t*)calloc(1,sizeof(bcf_hdr_aux_t));
+    if ( !aux ) goto fail;
+    aux->dict = *((vdict_t*)h->dict[0]);
```
pd3, line 1220: free(h->dict[0]);

pd3:
> Yes, something like that, but you currently leak a hash table. The allocation code should I think look like this:
> 
> ```c
>     h->dict[0] = calloc(1,sizeof(bcf_hdr_aux_t));
>     if (!h->dict[0]) goto fail;
>     for (i = 1; i < 3; ++i)
>         if ((h->dict[i] = kh_init(vdict)) == NULL) goto fail;
> ```
> 
> And we'd need to take care that the code to free it up does the right thing with the franken-structure.

I think it just needs this single line
```
free(h->dict[0]);
```

pd3:
Yes, but does not require knowing what is happening inside `kh_init`, so arguably it's cleaner and easier to understand.

pd3:
I forced pushed an updated version that is resolving the issue #1543 in my tests

vcf.c
```diff
@@ -1206,6 +1299,14 @@ bcf_hdr_t *bcf_hdr_init(const char *mode)
     if (!h) return NULL;
     for (i = 0; i < 3; ++i)
         if ((h->dict[i] = kh_init(vdict)) == NULL) goto fail;
+
+    bcf_hdr_aux_t *aux = (bcf_hdr_aux_t*)calloc(1,sizeof(bcf_hdr_aux_t));
+    if ( !aux ) goto fail;
+    if ( (aux->gen = kh_init(hdict))==NULL ) { free(aux); goto fail; }
+    aux->dict = *((vdict_t*)h->dict[0]);
+    free(h->dict[0]);
+    h->dict[0] = aux;
```
jkbonfield, line 1308: We have:

```
    vdict_t dict;   // bcf_hdr_t.dict[0] vdict_t dictionary which keeps bcf_idinfo_t for BCF_HL_FLT,BCF_HL_INFO,BCF_HL_FMT
    hdict_t *gen;   // hdict_t dictionary which keeps bcf_hrec_t* pointers for generic and structured fields
} bcf_hdr_aux_t;
```

but this just causes me to scratch my head.  Why is one dictionary an actual struct and the other a pointer to one?  What's the significance?  Surely there has to be a meaning? 

Answer - no!  No meaning whatsoever, it's just a byproduct of a strange allocation policy policy that does a structure to structure copy.  Why is it like this?

IMO it would be saner with something like:

```
    vdict_t *dict;   // bcf_hdr_t.dict[0] vdict_t dictionary which keeps bcf_idinfo_t for BCF_HL_FLT,BCF_HL_INFO,BCF_HL_FMT
    hdict_t *gen;   // hdict_t dictionary which keeps bcf_hrec_t* pointers for generic and structured fields
} bcf_hdr_aux_t;
```

and for allocation purposes:

```
bcf_hdr_aux_t *aux = (bcf_hdr_aux_t*)calloc(1,sizeof(bcf_hdr_aux_t));
aux->dict = (vdict_t *)h->dict[0];
aux->gen = kh_init(hdict);
// if (!aux->gen) blah
```

Obviously the code that frees aux->gen would need updating to also destroy aux->dict and free the h->dict pointer as before, but this is all internal so that doesn't change ABI at all.

jkbonfield: Oh god... I see now!

My naive assumption that given hdr->dict is a void pointer and so is an opaque type that is not part of the public API, we do, sadly, have code that understands the private internals of htslib and manipulates it still. :-(

https://github.com/samtools/bcftools/blob/develop/vcfannotate.c#L279

So fair enough I see why it's like it is.  However it needs a comment documenting this.

jkbonfield:
This is a somewhat philosophical question perhaps, but why are we looking for and removing duplicates in the first place?  As far as I can see, the VCF spec does not disallow duplicate unstructured header lines.  I don't think this is an error by omission either, as it does explicitly state structured header lines must have unique IDs.

I don't know what the effect would be though.  FWIW Picard reorders the lines (now alphabetical sorted) and removes duplicates.  I assume that also means the ordering of header lines is irrelevant too?

If duplicates are an error, maybe we should raise an issue with the VCF spec?

pd3:
It is not an error to have duplicate lines but it makes sense to prevent them. VCF does not require a specific order, with the exception of fileformat being first and PASS as the first filter.

jkbonfield:
I was just wondering if there was a simpler alternative, which was enforcing uniqueness checking at all and just regurgitating whatever lines the user had in their input data.  No uniqueness checks on unstructured data does obviously solve the O(N^2) issue, but yes it's probably a thing worth doing.

So if we feel it's something that needs doing, I would recommend adding it to the VCF spec as it's not really good to be enforcing our own undocumented rules on top of the public specification.  (And it appears Picard is already in agreement with our strategy too.)

PS. Hmm, are you sure on PASS being required to be the first filter?  I couldn't find this in the VCF spec, and Picard doesn't honour it.

```
@ seq4c[samtools.../htslib]; head /tmp/b.vcf
##fileformat=VCFv4.1
##fileDate=20150126
##reference=hs37d5
##phasing=partial
##contig=<ID=1,length=123456789>
##FILTER=<ID=PASSS,Description="pass">
##FILTER=<ID=VQSRTrancheSNP99.00to99.90,Description="Truth sensitivity tranche level for SNP model at VQS Lod: -6.6778 <= x < -0.6832">
##FILTER=<ID=INDEL_SPECIFIC_FILTERS,Description="QD < 2.0 || ReadPosRankSum < -20.0 || InbreedingCoeff < -0.8 || FS > 200.0">
##FILTER=<ID=LowQual,Description="Low quality">
##FILTER=<ID=VQSRTrancheSNP99.90to100.00+,Description="Truth sensitivity tranche level for SNP model at VQS Lod < -36469.5723">

@ seq4c[samtools.../htslib]; \java -jar ~/work/picard-2.26.11.jar FilterVcf I=/tmp/b.vcf O=/tmp/_.vcf
INFO	2023-01-30 16:51:30	FilterVcf	
...

@ seq4c[samtools.../htslib]; head /tmp/_.vcf
##fileformat=VCFv4.2
##FILTER=<ID=AllGtsFiltered,Description="Site filtered out because all genotypes are filtered out.">
##FILTER=<ID=AlleleBalance,Description="Heterozygote allele balance below required threshold.">
##FILTER=<ID=INDEL_SPECIFIC_FILTERS,Description="QD < 2.0 || ReadPosRankSum < -20.0 || InbreedingCoeff < -0.8 || FS > 200.0">
##FILTER=<ID=LowQD,Description="Site exhibits QD value below a hard limit.">
##FILTER=<ID=LowQual,Description="Low quality">
##FILTER=<ID=PASSS,Description="pass">
##FILTER=<ID=StrandBias,Description="Site exhibits excessive allele/strand correlation.">
##FILTER=<ID=VQSRTrancheSNP99.00to99.90,Description="Truth sensitivity tranche level for SNP model at VQS Lod: -6.6778 <= x < -0.6832">
##FILTER=<ID=VQSRTrancheSNP99.90to100.00+,Description="Truth sensitivity tranche level for SNP model at VQS Lod < -36469.5723">
```

However if it's an htslib bug then it's a diversion from this PR and should be opened as a new issue.

vcf.c
```diff
@@ -866,8 +883,40 @@ static int bcf_hdr_register_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec)
...
+    kstring_t str = {0,0,0};
+    ksprintf(&str, "##%s=%s", tmp->key,tmp->value);
+    k = kh_put(hdict, aux->gen, str.s, &ret);
+    if ( ret<0 )
+    {
+        free(str.s);
+        return -1;
+    }
+    free(hrec->value);
+    hrec->value = strdup(tmp->value);
```
jkbonfield, line 911: `bcf_hdr_update_hrec` is only called in one place and it's an internal function.  The next thing that immediately happens is `bcf_hrec_destroy(tmp)`, so we could simplify this (and avoid needing to check if `strdup` failed) with:

```
hrec->value = tmp->value;
tmp->value = NULL;
```

vcf.c
```diff
@@ -911,29 +982,47 @@ int bcf_hdr_add_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec)
...
+        if ( value )
+        {
+            kstring_t str = {0,0,0};
+            ksprintf(&str, "##%s=%s", key,value);
+            bcf_hdr_aux_t *aux = get_hdr_aux(hdr);
+            khint_t k = kh_get(hdict, aux->gen, str.s);
+            free(str.s);
+            if ( k == kh_end(aux->gen) ) return NULL;
+            return kh_val(aux->gen, k);
+        }
```
jkbonfield, line 1002: Obviously one benefit of storing all the unstructured keys in a hash table is fast searching.  However I'm struggling to see the use case here given, for uniqueness purposes, we hash on both key and value.

If I already know "source" is "GenomicsDBImport" then I can query it and get back that source is GenomicsDBImport.  Not particularly useful!  About all it adds is a way to probe a list of known values and see what comes back present.  So we could for example test if fileformat is 4.1, 4.2, 4.3 or 4.4, but we couldn't ask the basic question of "what is the file format version?".

Are there plans to extend the hash to have both keys and key=value components, so we could pass in NULL and return the first matching key?

Bonus points if the keys can form an internal linked list, but I'm not sure how much scope we have for doing that without breaking the ABI.  `bcf_hrec_t` has nkeys and pointers to arrays of key/val for the structured data, so there's room to store the result.    To make this transparent though it'd either need some abuse/hack of those array elements to add the internal linked list of hrecs, or use an internal struct for the values in hdict instead of hrec directly (the cleaner option).

Note: This isn't a reason to not merge this PR though.  Very much something to consider for the "to be continued" part.

vcf.c
```diff
@@ -1191,8 +1281,11 @@ int bcf_hdr_set_version(bcf_hdr_t *hdr, const char *version)
     }
     else
     {
-        free(hrec->value);
-        hrec->value = strdup(version);
+        bcf_hrec_t *tmp = bcf_hrec_dup(hrec);
+        free(tmp->value);
+        tmp->value = strdup(version);
```
jkbonfield, line 1286: Needs checks for failure in `bcf_hdr_dup`, `strdup` and `bcf_hdr_update_hrec`.

PR #1558, opened 2023-01-27
Embed ref2 unsorted

jkbonfield:
- Fix crash when `cram_encode_container` during multi-threaded CRAM encoding.

- Prevent embed_ref=2 on name sorted data from being a hard failure for cram encoding.  It now switches to non-ref mode instead, so we always encode something, even if it's suboptimal.

### DIFF
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -3829,7 +3829,7 @@ int cram_put_bam_seq(cram_fd *fd, bam_seq_t *b) {
             if (NULL == (c = cram_next_container(fd, b))) {
                 if (fd->ctr) {
                     // prevent cram_close attempting to flush
-                    cram_free_container(fd->ctr);
+                    fd->ctr_mt = fd->ctr; // delay free when threading
                     fd->ctr = NULL;
                 }
                 return -1;
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -61,7 +61,7 @@ KHASH_MAP_INIT_STR(m_s2u64, uint64_t)
 static int process_one_read(cram_fd *fd, cram_container *c,
                             cram_slice *s, cram_record *cr,
                             bam_seq_t *b, int rnum, kstring_t *MD,
-                            int embed_ref);
+                            int embed_ref, int no_ref);
 
 /*
  * Returns index of val into key.
@@ -86,6 +86,8 @@ cram_block *cram_encode_compression_header(cram_fd *fd, cram_container *c,
     cram_block *cb  = cram_new_block(COMPRESSION_HEADER, 0);
     cram_block *map = cram_new_block(COMPRESSION_HEADER, 0);
     int i, mc, r = 0;
+
+    int no_ref = c->no_ref;
 
     if (!cb || !map)
         return NULL;
@@ -162,7 +164,7 @@ cram_block *cram_encode_compression_header(cram_fd *fd, cram_container *c,
                 kh_val(h->preservation_map, k).i = h->qs_seq_orient;
             }
 
-            if (fd->no_ref || embed_ref>0) {
+            if (no_ref || embed_ref>0) {
                 // Reference Required == No
                 k = kh_put(map, h->preservation_map, "RR", &r);
                 if (-1 == r) return NULL;
@@ -1746,7 +1748,7 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
     cram_block_compression_hdr *h = c->comp_hdr;
     cram_block *c_hdr;
     int multi_ref = 0;
-    int r1, r2, sn, nref, embed_ref;
+    int r1, r2, sn, nref, embed_ref, no_ref;
     spare_bams *spares;
 
     if (CRAM_MAJOR_VERS(fd->version) == 1)
@@ -1755,25 +1757,17 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
 //#define goto_err {fprintf(stderr, "ERR at %s:%d\n", __FILE__, __LINE__);goto err;}
 #define goto_err goto err
 
+ restart:
     /* Cache references up-front if we have unsorted access patterns */
     pthread_mutex_lock(&fd->ref_lock);
     nref = fd->refs->nref;
-    embed_ref = fd->embed_ref;
     pthread_mutex_unlock(&fd->ref_lock);
-
-    if (!fd->no_ref && c->refs_used) {
-        for (i = 0; i < nref; i++) {
-            if (c->refs_used[i]) {
-                cram_get_ref(fd, i, 1, 0);
-                if (validate_md5(fd, i) < 0)
-                    goto_err;
-            }
-        }
-    }
+    embed_ref = c->embed_ref;
+    no_ref = c->no_ref;
 
     /* To create M5 strings */
     /* Fetch reference sequence */
-    if (!fd->no_ref) {
+    if (!no_ref) {
         if (!c->bams || !c->bams[0])
             goto_err;
         bam_seq_t *b = c->bams[0];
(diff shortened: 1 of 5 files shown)


daviesrob:
This seems OK in the case where you have no references at all, but I managed to break it by allowing it to find a subset of the references.  This may be a different, but related, bug.

To reproduce, make a name-sorted file that has no `@SQ UR:` tags but does have `@SQ M5:`.  Then find the reference for the first read it would try to load, and make an MD5 reference directory with just that one in it.  Then try to convert the file to CRAM.

The problem is that getting into multi_seq mode also [allocates `c->refs_used[]`](https://github.com/samtools/htslib/blob/1a222ef448/cram/cram_encode.c#L3857), which triggers the [MD5 validation code here](https://github.com/samtools/htslib/blob/1a222ef448/cram/cram_encode.c#L1766).  The validation code fails when it can't load one of the used references.

It also looks like there are possible race conditions on `fd->no_ref` as after this PR it can change state in another thread.

jkbonfield:
I think I've fixed the new (or quite possibly old) problem you found, as well as fixed the threading issues caused by changing no_ref from within cram threads.  That is particularly insideous as there's so many places it's looked at.  In the end I took the same approach we did for some other fields, which is migrate them from fd into the container at the time of creation, and make everything underneath that (eg slices, records, aux fields, etc) all relate to the container variable which doesn't change from that point onwards.  It may not be necessary everywhere, but it's safe (and failed attempts to fix this showed at least some of it was necessary).

Also fixed the CRAM decoder to be more robust so it doesn't access uninitialised md5sum buffers when we have no reference loaded.  Such CRAM files are malformed, and only temporarily created by the earlier commit (now fixed), but improving the robustness of the decoder is good.

cram/cram_encode.c
```diff
@@ -87,6 +87,10 @@ cram_block *cram_encode_compression_header(cram_fd *fd, cram_container *c,
     cram_block *map = cram_new_block(COMPRESSION_HEADER, 0);
     int i, mc, r = 0;
 
+    pthread_mutex_lock(&fd->metrics_lock);
```
daviesrob, line 90: Do you need to lock this (as you're accessing `c->no_ref` rather that `fd->no_ref`)?

cram/cram_encode.c
```diff
@@ -3852,6 +3904,28 @@ int cram_put_bam_seq(cram_fd *fd, bam_seq_t *b) {
...
 
+            // Cram_next_container may end up flushing an existing one and
+            // triggering fd->embed_ref=2 if no reference is found.
+            // Embedded refs are incompatible with multi-seq, so we bail
+            // out and switch to no_ref in this scenario.  We do this
+            // within the container only, as multi_seq may be temporary
+            // and we switch back away from it again.
+            pthread_mutex_lock(&fd->ref_lock);
+            if (embed_ref + c->embed_ref + fd->embed_ref != 0)
+            if (fd->embed_ref > 0 && c->curr_rec == 0 && c->curr_slice == 0) {
```
daviesrob, line 3915: The indentation of these `if` statements looks dubious.  Are they both supposed to be here?

cram/cram_encode.c
jkbonfield: Haha indeed not!  I expect this was originally fd->no_ref and I blithly converted it to `c->` instead.

daviesrob: You'll also need to check other places where it was changed, as there's at least one other case where you grab a lock before accessing `c->no_ref`.

cram/cram_encode.c
jkbonfield: A bit odd I'll agree and definitely not intended as-is.

When `c->curr_rec == 0 && c->curr_slice == 0` at the start of a slice, `embed_ref`, `c->embed_ref` and `fd->embed_ref` are all the same thing.  So the former statement is redundant I think.  Maybe it was an alternative statement from an earlier incarnation.  I sometimes had two variants and comment one or the other out to compare how they work, so it's probably a failure to tidy up properly.

It does make me wonder whether this works correctly when swapping from one slice to another in a multi-slice containers, but you've got to really go through hoops to explore this scenario as it can never happen automatically.  It's complex because the reference ID is stored in the container header, but the local ref MD5sum and/or embedded ref block number are stored in the slice header.  The whole unsorted detection though is done at the container level as it's that which stores ref start/end.

The whole multi-slice thing was probably a mistake in CRAM tbh, and it dates back to an earlier design in CRAM 1.x.  Even Vadim had stopped using it by the time I came along, so now it's just baggage!  (Unless we redesign the codecs to put more frequency meta-data into the container compression header.)

cram/cram_encode.c
```diff
@@ -1978,12 +2015,14 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
 
 
     /* Compute MD5s */
+    no_ref = c->no_ref;
```
daviesrob, line 2018: I don't think this line is needed.

cram/cram_encode.c
```diff
@@ -1822,6 +1829,16 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
         c->ref_seq_id = c->ref_id;
     }
 
+    if (!no_ref && c->refs_used) {
+        for (i = 0; i < nref; i++) {
+            if (c->refs_used[i]) {
+                cram_get_ref(fd, i, 1, 0);
```
daviesrob, line 1835: It was going well until I enabled multi-threading:
```
REF_PATH=/tmp/cc/%s ./test/test_view -C -@ 8 -p /tmp/t8.cram /tmp/9305_4#1n_nour.bam
[W::cram_get_ref] Failed to populate reference for id 5
[W::cram_encode_container] Failed to load reference #5
[W::cram_encode_container] Enabling embed_ref=2 mode to auto-generate reference
[W::cram_encode_container] NOTE: the CRAM file will be bigger than using an external reference
[W::cram_get_ref] Failed to populate reference for id 16
[W::cram_encode_container] Failed to load reference #16
[W::cram_encode_container] Switching to non-ref mode
[W::cram_get_ref] Failed to populate reference for id 1
[E::validate_md5] SQ header M5 tag discrepancy for reference '10'
[E::validate_md5] Please use the correct reference, or consider using embed_ref=2
[E::cram_flush_thread] Call to cram_encode_container failed
Error writing output.
```

Note that you might have to run it a few times to get this to happen.  I think it depends on the main thread managing to queue up enough containers before one of the workers switches it to `no_ref` mode.

I suspect the easiest solution might be to check for reference loading failures here, and switch the container to `no_ref` mode if it happens.

jkbonfield: > Note that you might have to run it a few times to get this to happen. I think it depends on the main thread managing to queue up enough containers before one of the workers switches it to `no_ref` mode.
> 
> I suspect the easiest solution might be to check for reference loading failures here, and switch the container to `no_ref` mode if it happens.

More than a few times for me - it was failing about 1 in 100 on my smaller test set (with approx 10 containers worth and 3 threads).  Anyway, it's now doing as suggested and I can't break it any more (several thousand runs done).

daviesrob:
This looks better, albeit a bit chatty when running multi-threaded and it discovers it needs no-ref mode.  I also see that the stray `no_ref = c->no_ref;` mentioned above is still there.

Could you squash this into its final form please?

cram/cram_encode.c
jkbonfield: The function has grown over the years and is overly long, so it really needs a refactor.  However `no_ref` is used a few lines lower down and above it there are several places that modify `c->no_ref`.  It's possible the logic is such that they never have any effect on the cached `no_ref`, but I wouldn't like to say with 100% certainty and for clarity alone it's worth keeping this line IMO.

jkbonfield:
Chatty is deliberate, as it has an impact on the CRAM size.  If someone accidentally forgets to specify the reference I'd rather they get a few warning messages, possibly more than once if threaded, than a silent degregadation in compression performance.  I don't think it's worth putting lots of extra thread locks and a shared variable to track how many times we've reported the warnings, so I think the current warnings represent a good middle ground unless you can think of a simple fix.

I rebased and squashed it.

PR #1567, opened 2023-02-17
Fix issue 1566 - basemod API and hardclips

### LINKED ISSUE #1566: basemod API and hardclipped alignments
The `MM` tag is intended to be with respect to the instrument's basecalled sequence and hence can be copied as-is so aligners (and other tools) do not need to be aware of base modifications. In the case where the aligner hard clips the sequence and copies the MM tag however, the coordinates may no longer be valid. The current implementation of the API will process the tag for hardclipped alignments returning (potentially) invalid coordinates. Here is a small example sam:

```
@SQ     SN:chr1 LN:248956422
@CO     Basemod reference to hardclipped base
test-read       0       chr1    19931836        60      20H36M  *       0       0       AGGATCTCTAGCGGATCGGCGGGGGATATGCCATAT    *       Mm:Z:C+m,1;     Ml:B:C,12
```

Using `bam_parse_basemod` followed by `bam_next_mod` will result in a single `hts_base_mod` at position 7 of SEQ. I think it is preferable in this case for `bam_parse_basemod` to return -1.

Thanks,
Jared

jts:
This PR checks whether an aligned read has hard clipped bases, and if so returns -1 from bam_parse_basemod. Fixes #1566

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -6220,9 +6220,6 @@ int bam_parse_basemod(const bam1_t *b, hts_base_mod_state *state) {
     // so such records are invalid
     if( !(b->core.flag & BAM_FUNMAP) && b->core.n_cigar > 0 ) {
         const uint32_t *cigar = bam_get_cigar(b);
-
-        // bam_cigar_type returns 0 when neither the query nor reference
-        // are consumed. Whenever this is true we can't parse MM
         if( bam_cigar_op(cigar[0]) == BAM_CHARD_CLIP ||
             bam_cigar_op(cigar[b->core.n_cigar - 1]) == BAM_CHARD_CLIP ) {
             return -1;
--- a/sam.c
+++ b/sam.c
@@ -6215,6 +6215,19 @@ int bam_parse_basemod(const bam1_t *b, hts_base_mod_state *state) {
         return -1;
     }
 
+    // check whether the read is aligned and hardclipped
+    // hardclipping breaks the correspondance between SEQ and MM
+    // so such records are invalid
+    if( !(b->core.flag & BAM_FUNMAP) && b->core.n_cigar > 0 ) {
+        const uint32_t *cigar = bam_get_cigar(b);
+
+        // bam_cigar_type returns 0 when neither the query nor reference
+        // are consumed. Whenever this is true we can't parse MM
+        if( bam_cigar_type(bam_cigar_op(cigar[0])) == 0 ||
+            bam_cigar_type(bam_cigar_op(cigar[b->core.n_cigar - 1])) == 0)
+            return -1;
+    }
+
     uint8_t *ml = bam_aux_get(b, "ML");
     if (!ml) ml = bam_aux_get(b, "Ml");
     if (ml && (ml[0] != 'B' || ml[1] != 'C')) {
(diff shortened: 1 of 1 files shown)


jts:
Thanks @jkbonfield, fine to squash and rebase in any way that's convenient for you. I'll come back with some thoughts for the points you raised a little later.

jts:
I'm back now with some comments. I do think hardclips should be handled somehow but this isn't an urgent problem since users that care about modifications at supplementary alignments can tell minimap2 to use softclips instead of hardclips with the `-Y` option. 

> We could say that tools that do the hard-clipping also do the trimming of the MM/ML tags, but how would you be able to distinguish between tools which have been updated to cope with this tag, and those which haven't? You simply can't. 

I agree we can't ask the SAM producers to update the MM tags but is it worth considering a tool that fixes the MM tag for hardclipped reads? There's precedent with `samtools fixmate` and `samtools calmd`. We need a another flag though to indicate whether MM accounts for hardclips (I guess something like a mode bit to say whether the offsets refer to SEQ or whatever the basecalled reads was). At this point it might be best to switch to a bitwise flag instead of the symbolic flags (this has the benefit of reserving bits for future cases that come up).

I'm not sure how I feel about switching everything to `N+m` notation. I think here explicit mode would be required since nearly all of the skipped bases were not called.

@cjw85 thoughts?

cjw85:
I like the idea of a samtools command that fixes up "broken" MM tags. It does immediately raise the need to then encode some extra state as you say. What a nightmare!

This would then I think necessitate a need to beef up the API in order to serve clients with the correct interpretation of all the information.

I'm in two minds about `N+m`. On the one hand it just feels wrong. On the other before the MM tags were created I was just storing modification information in dense array-valued tags.

jkbonfield:
Indeed.  One reason I haven't started working on the API again is I'm not 100% sure I know the final requirements yet!  All help and suggestions welcomed here.

It's hard to know what fix can be applied to a read with hard clipping and MM, unless we extend the MM syntax and the fixmates equiv command turn adds the additional flags as appropriate.  So that needs both spec update and API updates.  Or it uses N+m style syntax and we just define what that means with hardclips.

jkbonfield:
There is ongoing discussion about this on twitter:

https://twitter.com/Psy_Fer_/status/1632943849386692608
https://twitter.com/DrKangabu/status/1632934256019402752

This appears to show that some tools are doing hard-clipping and creating MM tags in the same pipeline (I assume just the two steps in that order).  So this PR would make that correct looking data fail.

We're kind of stuck, but I wonder whether the right thing is just to sress that the MM tags count along the sequence as recorded in the SEQ record (albeit in original orientation) - this is already stated in the spec - and for the implementations to have sanity checks to ensure that out of bounds data is reported.

So rather than just rejecting the data, for reads with hard-clipping we could check if Nth C base is beyond the end of the recorded seq.  It's not ideal from a performance viewpoint, and we may erroneously accept some duff data, but given a sufficient number of sequences it seems highly unlikely that they'd all accidentally pass the checks.

Long term we need to figure out a way of recording the difference between MM-corrected and MM-uncorrected (and now broken).  I can't think how to do that unless it's a third party piece of data.  Eg embedding the sequence length as an error check.  If we modify SEQ len in any way and don't make a corresponding edit to MM, then MM is now invalid.  It's disappointing as it'd be yet another format change, but I cannot see any other robust way to do it unless it's an additional tag.  Eg MZ:i:<seq_len> for Modification-siZe check.

Thoughts?

jkbonfield:
I'm going to close this as we know it will affect some pipelines which already do the right thing.  So stuck between a rock and a hard place of detecting erroneous data vs breaking corrected data.

I'll work on an MZ:i alternative tag instead to act as a validation.

PR #1572, opened 2023-02-28
Switch to CURLINFO_CONTENT_LENGTH_DOWNLOAD_T for newer libcurl

daviesrob:
The older CURLINFO_CONTENT_LENGTH_DOWNLOAD has been deprecated, and is now causing warnings to be printed on distributions that have moved on to the latest libcurl versions.  CURLINFO_CONTENT_LENGTH_DOWNLOAD_T was [officially added in curl 7.55.0](https://curl.se/libcurl/c/CURLINFO_CONTENT_LENGTH_DOWNLOAD_T.html).

### DIFF
--- a/hfile_libcurl.c
+++ b/hfile_libcurl.c
@@ -1330,13 +1330,20 @@ libcurl_open(const char *url, const char *modes, http_headers *headers)
     }
 
     if (mode == 'r') {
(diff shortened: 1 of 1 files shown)


daviesrob:
Updated to use LIBCURL_VERSION_NUM.  Sadly CURL_AT_LEAST_VERSION() doesn't exist in old curl include files (like those in CentOS 7).

PR #1574, opened 2023-03-01
Fix cram_index_query_last function

### LINKED ISSUE #1569: sam_itr_regarray not reporting reads
In the file s3://1000genomes/1000G_2504_high_coverage/additional_698_related/data/ERR3988780/HG00512.final.cram the region chr7:106092666-106093543 contains 286 reads.

The following code, correctly outputs all of the starting positions of those reads when argv[1] is the path of HG00512.final.cram:

```
samFile* fp = sam_open(argv[1], "r");
bam_hdr_t* hdr = sam_hdr_read(fp);
hts_idx_t* idx = sam_index_load(fp, argv[1]);

bam1_t* read = bam_init1();
hts_itr_t* iter = sam_itr_querys(idx, hdr, "chr7:106092666-106093543");
while (sam_itr_next(fp, iter, read) >= 0) {
    std::cout << read->core.pos << std::endl;
}
```

However, the following code, which I expect to be equivalent, only outputs one read for me
```
samFile* fp = sam_open(argv[1], "r");
bam_hdr_t* hdr = sam_hdr_read(fp);
hts_idx_t* idx = sam_index_load(fp, argv[1]);

bam1_t* read = bam_init1();
char* region[] = { "chr7:106092666-106093543" };
hts_itr_t* iter = sam_itr_regarray(idx, hdr, region, 1);
while (sam_itr_next(fp, iter, read) >= 0) {
    std::cout << read->core.pos << std::endl;
}
```

jkbonfield:
The cram_index_query_last used by sam_itr_regarray had problems when dealing with slices whose region is contained within another region.

Fundamentally this is due to the cram_index arrays being contiguous in memory until a containment is found, at which point the pointers will be to an entirely different array.  This breaks naive pointer comparisons.

The cram_index struct already had a "next" field holding the file offset of the next container.  This has been replaced by e_next pointing to the next cram_entry struct in file ordering, and e_next->offset is equivalent to the old "next".  This allows consumption of the index either as the original nested containment list or as a traditional linked list.

Also fixed cram_index_query with from != NULL, which similarly was incorrect before.  We never used this function and it's not public, but we now use it within the rewrite of cram_index_query_last.

Fixes #1569

Some testing on a local CRAM file, with nested containers (nested containment list in index) and some multi-ref containers present too.

```
                Real(s)   User(s)       No.Seqs
1.7:
BAM @8          0m12.972s 0m54.406s     1101651 WRONG
CRAM-1k @8      theaded fails with CRC err
" no threads    0m31.657s 0m28.115s     1100721 WRONG

1.9:
BAM @8          0m10.607s 0m33.546s     1101651 WRONG
CRAM-1k @8      1m13.605s 7m5.137s      17031   VERY WRONG!
" no threads    0m42.706s 0m40.298s     1100731 WRONG

1.10.2:
BAM @8          0m11.040s 0m34.387s     1101852 (first BAM to be correct)
CRAM-1k @8      1m38.255s 7m44.389s     1100922 WRONG
" no threads    0m28.594s 0m26.180s     1100922 WRONG

1.11:
CRAM-1k @8      0m24.957s 0m46.595s     1586311 VERY WRONG
" no threads    0m39.835s 0m34.317s     1586311 VERY WRONG

1.17:
CRAM-1k @8      0m26.114s 0m52.064s     1586311 VERY WRONG
" no threads    0m42.399s 0m36.827s     1586311 VERY WRONG

this PR:
BAM @8          0m9.148s  0m40.372s     1101852
" no threads    0m9.414s  0m9.201s      1101852
CRAM-1k @8      0m23.206s 0m43.078s     1101852
" no threads    0m25.381s 0m23.155s     1101852
```

It's surprising to see the earlier releases given very many more results back from this bunch of queries.

For reference for future code archaeologists, my test was:

```
time for r in `seq 1 10`;do echo -n "$r ";samtools view -H 17537_1#4.bam|perl -lane 'BEGIN{srand('$r')} next unless /^@SQ/;$F[1]=~s/^SN://;$F[2]=~s/^LN://;$l1 = int(rand($F[2]/4));$l2 = int(rand(10000));for ($i=rand()<0.2?$l2:$l1;$i<$F[2];$i+=rand()<0.7?$l2:$l1) {$en=$i+int(rand(1000));$en=$F[2] if $en>$F[2];print "$F[1]:$i-$en";$l1 = int(rand($F[2]/10)+50000);$l2 = int(rand(1000));}' > _reg;./test/test_view -M 17537_1#4.s1k.cram `cat _reg` | egrep -cv '^@';done | awk '{a+=$2;print} END {print "total",a}'
```

The .s1k.cram was built with `-O cram,seqs_per_slice=1000` to force more container junctions.  I also tested 10k and 100 seq per slice plus 3 slices per container, which stressed the previous version even greater.

### DIFF
--- a/cram/cram_index.c
+++ b/cram/cram_index.c
@@ -306,7 +306,8 @@ int cram_index_load(cram_fd *fd, const char *fn, const char *fn_idx) {
             idx_stack[(idx_stack_ptr = 0)] = idx;
         }
 
-        while (!(e.start >= idx->start && e.end <= idx->end) || idx->end == 0) {
+        while (!(e.start >= idx->start && e.end <= idx->end) ||
+               (idx->start == 0 && idx->refid == -1)) {
             idx = idx_stack[--idx_stack_ptr];
         }
 
--- a/cram/cram_index.c
+++ b/cram/cram_index.c
@@ -72,7 +72,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 static void dump_index_(cram_index *e, int level) {
     int i, n;
     n = printf("%*s%d / %d .. %d, ", level*4, "", e->refid, e->start, e->end);
-    printf("%*soffset %"PRId64"\n", MAX(0,50-n), "", e->offset);
+    printf("%*soffset %"PRId64" %p %p\n", MAX(0,50-n), "", e->offset, e, e->e_next);
     for (i = 0; i < e->nslice; i++) {
         dump_index_(&e->e[i], level+1);
     }
@@ -85,6 +85,37 @@ static void dump_index(cram_fd *fd) {
     }
 }
 #endif
+
+// Thread a linked list through the nested containment list.
+// This makes navigating it and finding the "next" index entry
+// trivial.
+static cram_index *link_index_(cram_index *e, cram_index *e_last) {
+    int i;
+    if (e_last)
+        e_last->e_next = e;
+
+    // We don't want to link in the top-level cram_index with
+    // offset=0 and start/end = INT_MIN/INT_MAX.
+    if (e->offset)
+        e_last = e;
+
+    for (i = 0; i < e->nslice; i++)
+        e_last = link_index_(&e->e[i], e_last);
+
+    return e_last;
+}
+
+static void link_index(cram_fd *fd) {
+    int i;
+    cram_index *e_last = NULL;
+
+    for (i = 0; i < fd->index_sz; i++) {
+        e_last = link_index_(&fd->index[i], e_last);
+    }
+
+    if (e_last)
+        e_last->e_next = NULL;
+}
 
 static int kget_int32(kstring_t *k, size_t *pos, int32_t *val_p) {
     int sign = 1;
@@ -313,7 +344,10 @@ int cram_index_load(cram_fd *fd, const char *fn, const char *fn_idx) {
     free(kstr.s);
     free(tfn_idx);
 
-    // dump_index(fd);
+    // Convert NCList to linear linked list
+    link_index(fd);
+
+    //dump_index(fd);
 
     return 0;
 
@@ -356,7 +390,7 @@ void cram_index_free(cram_fd *fd) {
  * entries, but we require at least one per reference.)
  *
  * If the index finds multiple slices overlapping this position we
- * return the first one only. Subsequent calls should specifying
+ * return the first one only. Subsequent calls should specify
  * "from" as the last slice we checked to find the next one. Otherwise
  * set "from" to be NULL to find the first one.
  *
@@ -370,6 +404,17 @@ cram_index *cram_index_query(cram_fd *fd, int refid, hts_pos_t pos,
                              cram_index *from) {
     int i, j, k;
     cram_index *e;
+
+    if (from) {
+        // Continue from a previous search.
+        // We switch to just scanning the linked list, as the nested
+        // lists are typically short.
+        e = from->e_next;
+        if (e && e->refid == refid && e->start <= pos)
+            return e;
+        else
+            return NULL;
+    }
 
     switch(refid) {
     case HTS_IDX_NONE:
@@ -400,8 +445,7 @@ cram_index *cram_index_query(cram_fd *fd, int refid, hts_pos_t pos,
             return NULL;
     }
 
-    if (!from)
-        from = &fd->index[refid+1];
+    from = &fd->index[refid+1];
 
     // Ref with nothing aligned against it.
     if (!from->e)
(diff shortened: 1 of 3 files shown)


jkbonfield:
See also https://github.com/samtools/htslib/pull/1064 which was a previous attempt to fix the few missing records comparing BAM to CRAM.  This is what caused the big spike of extra records for 1.11 onwards, while apparently not fixing all missing ones.

That was despite very extensive testing.  So please also do your own testing!

I'm trying to construct some synthetic stress tests, but so far everything works on them which has me scratching my head...

daviesrob:
Oops, posted this on the wrong tab...

This looks OK so far, although enabling the `dump_index()` code to see what's going on shows some rather deep nesting on the unmapped reads:
```
-1 / -2147483648 .. 2147483647,                   offset 0 0x27755d0 (nil)
    -1 / 0 .. 149,                                offset 15726314885 0x27523b0 0x2752740
        -1 / 0 .. 149,                            offset 15727086877 0x2752740 0x2752ad0
            -1 / 0 .. 149,                        offset 15727730588 0x2752ad0 0x2752e60
                -1 / 0 .. 149,                    offset 15728372279 0x2752e60 0x27531f0
                    -1 / 0 .. 149,                offset 15729011232 0x27531f0 0x2753580
                        -1 / 0 .. 149,            offset 15729651684 0x2753580 0x2753910
                            -1 / 0 .. 149,        offset 15730292336 0x2753910 0x2753ca0
                                -1 / 0 .. 149,    offset 15730940888 0x2753ca0 0x2754030
                                    -1 / 0 .. 149, offset 15731618936 0x2754030 0x27543c0
                                        -1 / 0 .. 149, offset 15732293423 0x27543c0 0x2754750
                                            -1 / 0 .. 149, offset 15732964238 0x2754750 0x2754ae0
                                                -1 / 0 .. 149, offset 15733646463 0x2754ae0 0x2754e70
                                                    -1 / 0 .. 149, offset 15734330256 0x2754e70 0x2755200
                                                        -1 / 0 .. 149, offset 15734980719 0x2755200 0x2755590
                                                            -1 / 0 .. 149, offset 15735564531 0x2755590 0x2755920
                                                                -1 / 0 .. 149, offset 15736146386 0x2755920 0x2755cb0
                                                                    -1 / 0 .. 149, offset 15736729893 0x2755cb0 0x2756040
                                                                        -1 / 0 .. 149, offset 15737333674 0x2756040 0x1ee46e0

```

This isn't a problem in this case, and I guess it would be OK for `cram_index_query()` as you can only ask to go to the start of the unmapped data.  But if you accidentally made a file with a lot of unmapped data I guess you could end up getting some deep recursion in `link_index_()`, not to mention a lot of allocated `cram_index::e` structs.  Do you think this could be a problem, and is there anything that could easily be done about it?

cram/cram_index.c
```diff
@@ -86,6 +86,37 @@ static void dump_index(cram_fd *fd) {
...
+// trivial.
+static cram_index *link_index_(cram_index *e, cram_index *e_last) {
+    int i;
+    if (e_last)
+        e_last->e_next = e;
+
+    // We don't want to link in the top-level cram_index with
+    // offset=0 and start/end = INT_MIN/INT_MAX.
+    if (e->offset)
+        e_last = e;
```
daviesrob, line 100: It is possible to avoid the recursion, by spotting the ladder and walking along it...
```suggestion
    if (e->offset) {
        e_last = e;

        // Prevent excess recursion on very deep single-child nestings.
        // This can especially happen on unmapped, unplaced reads, which all
        // pile up at position 0.
        while (e->nslice == 1) {
            e = &e->e[0];
            e_last->e_next = e;
            e_last = e;
        }
```

jkbonfield: I'm trying to recall why the nested containment list was used in the first place.  I think treating the index as an array that we can binary search on only really works if we ensure the list of start positions are always sorted, and similarly for the end positions.  This is ensured by using NCList here, or the recursive binning in BAM (which could be viewed as a nested list too).  I think we should permit single-child nesting, but instead avoid making containments where start and end are equal (as that's still sorted by both ends).

Unlike BAM where we produce a hit list of multiple bins to scan, with seeking between them.

I think the intention here with containment lists was mixing very long reads with very short reads would return the first large outer container, and then subsequent offsets within a containment list would be yielded by calling the query again using the "from" parameter.  (That's rewritten and fixed here, but not in the way originally intended of recursing down and doing a second level of binary searching).  Instead we just found the first container and then did reading and/or seeking to consume or skip containers until we're past the end of the range.  It's almost always the same amount of I/O, but in the extreme case it's possible we could be seeking past numerous containers.  In practice though it's sufficient and fast enough to not worry about it.  CRAM can get away with it due to the container structure permitting seekage, unlike BAM.

So actually it may not matter if both start/end are sorted (start alone is sufficient if we then skip from that point on), but I'd still like to keep that as a possibility.

jkbonfield: Ah the lack of context in the quoted code above confused me.  You're not changing the index loading to avoid deeply nesting the list, but simply avoiding recursion in the link_index function.  Yeah that's safer, but I think unnecessary too.

We should just change the index loading code to work better and not create containments for exact matching start/end on unmapped data.

jkbonfield:
Corrected the index loading code to detect indices with invalid end coords for unmapped data.  This avoids the deep nesting reported above.

I cannot work out what produced that broken index though.  None of the tools I've tested (going back into ancient history) are doing so.  There's no PG meta-data in the file either, so it's a mystery.

PR #1587, opened 2023-03-22
Make SIMD configure tests work for MacOS multiarch builds

daviesrob:
MacOS multiarch binaries compile source code for each architecture and then join them together using 'lipo'.  This means architecture specific code both in the actual source and configure tests need to be compilable on both architectures.

This pulls in htscodecs changes needed, and makes changes to HTSlib's Makefile, configure and hts_probe_cc.sh scripts so that multiarch compilation works.  It also makes the MacOS CI tests build a multiarch library.

See also samtools/htscodecs#76 and samtools/htscodecs#78.


daviesrob:
MacOS CI tests have been updated to add the `-arch` options to `LDFLAGS`.  This seems to be necessary to ensure you've really made a multiarch library / executable.

PR #1589, opened 2023-03-28
Demo

vasudeva8:
This pull request adds a number of sample programs to demonstrate the usage of HTSLib apis.


samples/split.c
```diff
@@ -0,0 +1,182 @@
...
+        goto end;
+    }
+    //infile name
+    if (!inname && ((argc - optind) == 1))
+        inname = argv[optind];
+    if (!inname)
+        inname = "-";
+    //output directory
+    if (!outdir) {
+        fprintf(stderr, "Output directory not specified\n");
```
jkbonfield, line 92: One for @daviesrob to consider maybe, but we never use `hts_log` functions from samtools and I don't know why.  We could have instead:

```
hts_log_warning("no output directory specified");
```

or 

```
hts_log(HTS_LOG_WARNING, "split","no output directory specified");
```

I'm not sure why we always use fprintf here, but we even simulate the same style of output:

https://github.com/samtools/samtools/blob/develop/sam_view.c#L1164

I suspect it's just history.

That said, it can be useful to not start with this most demonstration programs as we don't want to bamboozle people with a large API to start with.

samples/split.c
```diff
@@ -0,0 +1,182 @@
...
+    }
+    //write header
+    if (-1 == sam_hdr_write(outfile1, in_samhdr) || -1 == sam_hdr_write(outfile2, in_samhdr)) {
+        goto end;
+        fprintf(stderr, "Failed to write header\n");
+    }
+
+    //check flags, write and update count
+    while (0 <= (c = sam_read1(infile, in_samhdr, bamdata))) {
+        cnt++;
```
jkbonfield, line 143: cnt is an unused variable.

jkbonfield (commented):
Just a cursory glance through flags_demo.c and  split.c so far.  More tomorrow.

We need to also decide on a house style for bracing.  We have multiple ones in use in this code, but we should be consistent.  Personally my preference is this as it's fewer lines of code.

```
if (expr)
    statement;
else
    statement;
    
if (multi_line_expr) {
    do1;
    do2;
} else {
    do3;
}
```

I know others have open braces in the same column as the closing one (but as I say I see both here).

samples/split.c
```diff
@@ -0,0 +1,182 @@
...
+    //bam data storage
+    bamdata = bam_init1();
+    if (!bamdata) {
+        fprintf(stderr, "Failed to initialize bamdata\n");
+        goto end;
+    }
+    //open input file
+    infile = sam_open(inname, "r");
+    if (!infile) {
+        fprintf(stderr, "Could not open %s\n", inname);
```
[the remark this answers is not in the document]
jmarshall: All samtools code that invokes `sam_open()` writes its own error message itself.

Personally I think it's a bug that `sam_open` prints its own diagnostic. It is a library routine similar to `fopen` so should be quiet by default, leaving it to the invoking code to print a diagnostic — which can print a better diagnostic because it knows what the purpose of the file is (e.g. it can write “can't open reference file” or similar to distinguish further which of several input files is missing).

Some htslib library functions print their own diagnostics because it's likely that invoking code won't print its own diagnostic. However file opening functions like `sam_open` are the one case where HTSlib can depend on the invoking application code universally printing the diagnostic itself!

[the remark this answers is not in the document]
jkbonfield: Agreed on the library not reporting errors this way, but right now it does so.  Ideally it needs an hts_strerror type of interface, maybe optionally taking a file pointer (like ferror), or we split those into two APIs.    (See also comments in https://github.com/samtools/htslib/pull/1049).  A long standing wish-list!

This code serves as two purposes really.  One is standalone basic starting points that people may wish to copy and build on top of - a simple template design.  In that things like error checking and reporting are important to put in as they foster good design principles.

The second is as text quoted in a tutorial or how-to guide on using htslib.  This isn't just doxygen type manuals, which don't easily lend themselves to guidance.  For those I think we could probably do some liberal pruning of error reporting (but not detection) for the sake of brevity.  We already have some of this with liberal "..." to jump to the relevant code blocks.  It may be better to avoid putting too much in main so we can document entire functions as isolated code fragments. That also lends itself better to just returning error codes instead of printing up messages.  That depends on the application though.

samples/read_header.c
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    if (header == &types[0]) {
+        //no header specified, reset position and tag fields as it would be display of full line
+        tag = NULL;
+        pos = 0;
+    }
+
+    //iterate and show 1st of each type or requested one
+    for (c = 0; headercnt && (c < (sizeof(types) / sizeof(types[0]))); ++c) {
```
jkbonfield, line 135: This is wrong.  You need `c < headercnt` where `headercnt` defaults to `sizeof(types) / sizeof(*types)`.

As it stands it buffer overflows if I try to view `CO` or even something that doesn't exist such as `XX`.

samples/read_header.c
```diff
@@ -0,0 +1,165 @@
...
+        //get count of given header type
+        linecnt = sam_hdr_count_lines(in_samhdr, *(header + c));
```
jkbonfield, line 137: `*(header + c)` is better just written as `header[c]` and makes it clearer we're just using an array of header types.

samples/read_header.c
```diff
@@ -0,0 +1,165 @@
...
+            opt = optarg;
+            header = &opt;
+            headercnt = 1;
+            //find the id_key field for the requested header type - uses NULL for HD
+            for (c = 0; c < (sizeof(types) / sizeof(types[0])); ++c) {
+                if (!strcmp(types[c], optarg)) {
+                    idf = &idfield[c];
+                    break;
+                }
+            }
```
jkbonfield, line 84: This comment is confusing.

jkbonfield (commented):
General comments for header.c.

The header API is complex and has a lot of features.  We probably want to focus on the most common and basic operations people may want to do.  That's things like checking the length of a reference sequence, or a general purpose find value X for key Y in header type `@ZZ`.

This code specifies the `-n`th line, and it then converts that to a string form using `sam_hdr_line_name` and the `idf` ("SN" for SQ and "ID" for PG/RG) to a string, which is then looked up using `sam_hdr_find_tag_id`.  This doesn't really feel like something the user is ever going to want to do.  We're unlikely to want to know about the 85th reference, but we may want to know about ref "hs37d5".

Furthermore there's no specific block on  `sam_hdr_find_tag_id` working with any key type.  We can query SQ by SN (the most likely) but also by M5 if we wished to.  The only real difference is that SN has an internal index (a hash table) for fast lookup while the others will do a full scan to find the line.  However that's something we could describe in text if needs be.

Also as a demonstrationtool it feels a little muddled by trying to do many things.  I don't think it needs the whole complexity of reporting all header lines types unless the user specifies one, which then gives rise to the whole headercnt and line types arrays etc.  Viewing the entire header is easier via other mechanisms, so as demonstration of the API we should focus on the querying specific components aspect.  This will make the code much simpler to follow, and hence focus on the key points we want to explain which is our API, rather than the program flow and logic.

Eg:
```
$ read_header
Usage: read_header file line-type key value [tag]

# Uses sam_hdr_find_line_id
$ read_header in.bam SQ SN hs37d5
@SQ	SN:hs37d5	LN:35477943	M5:5b6a4b3a81a2d3c134b7d14bf6ad39f1	UR:/nfs/srpipe_references/references/Human/1000Genomes_hs37d5/all/fasta/hs37d5.fa

# Uses sam_hdr_find_tag_id
$ read_header in.bam SQ SN hs37d5 LN
35477943
```

We can do that already (`-h SQ -n 85 -t LN ~/scratch/data/novaseq.10m.cram`) but with a -n option instead.

samples/read_bam.c
```diff
@@ -0,0 +1,169 @@
...
+    while (0 <= (ret_r = sam_read1(infile, in_samhdr, bamdata)))
+    {
+        if (idx != cnt++)
+            continue;
+
+        //QNAME FLAG RNAME POS MAPQ CIGAR RNEXT PNEXT TLEN SEQ QUAL [TAG:TYPE:VALUE]…
+        fprintf(stdout, "%s", bam_get_qname(bamdata));                                  //get the query name using the macro
+        fprintf(stdout, "\t%d", bamdata->core.flag);                                    //flag is available in core structure
+        fprintf(stdout, "\t%s", sam_hdr_tid2name(in_samhdr, bamdata->core.tid));        //retrieves the target name using the value in bam and by referring the header
+        fprintf(stdout, "\t%d", bamdata->core.pos + 1);                                 //internally position is 0 based and on text output / SAM it is 1 based
```
jkbonfield, line 124: %d here is too small.  It's a 64-bit number usually, but we normally use `%"PRIhts_pos"` instead of `%d` which expands up to %ld or %lld as required by the system.

samples/read_bam.c
```diff
@@ -0,0 +1,169 @@
...
+    while ((c = getopt_long(argc, argv, "i:n:", lopts, NULL)) >= 0)
+    {
+        switch (c)
+        {
+        case 'i':                           //in file name
+            inname = optarg;
+            break;
+        case 'n':                           //position
+            idx = atoi(optarg);
+            break;
```
jkbonfield, line 75: This is both misleading and also not something we ever want to do.

Position is not chr:pos, or even pos, but the Nth record.   This will be confusing to most readers as they think of position in terms of genomic coordinates, doubly emphasised by calling it "idx" implying indices are invovled.  Also I cannot think of any reason to scan through a file to pull out the Nth record, so it doesn't make for a good example.

I think the best thing is to ignore this field entirely and just have a loop to read a file and dump out the contents.

Or perhaps better still to report the first N records and then exit.  So it's more like a "head" command.

samples/read_bam.c
```diff
@@ -0,0 +1,169 @@
...
+        goto end;
+    }
+
+    while (0 <= (ret_r = sam_read1(infile, in_samhdr, bamdata)))
+    {
+        if (idx != cnt++)
+            continue;
+
+        //QNAME FLAG RNAME POS MAPQ CIGAR RNEXT PNEXT TLEN SEQ QUAL [TAG:TYPE:VALUE]…
+        fprintf(stdout, "%s", bam_get_qname(bamdata));                                  //get the query name using the macro
```
jkbonfield, line 121: While printing out the contents is a good way of demonstrating the API (or lack of API via bam structure access), I think it is a mistake to print it up in a deliberate SAM looking format incase people mistakenly believe this is an example of BAM to SAM conversion.

Rather I think something like:

```
Read 1:
     Name\tFoo
     Flag\t0x41
     Chr\t1
     Pos\t1000
     ...
Read 2:
    ...
```

It's very clearly not SAM format, but its own thing. Maybe document it as a tool to read a BAM and report or dump the contents to stdout.


Also no need for `fprintf(stdout,...)`.  That's just `printf()`.

samples/read_bam.c
```diff
@@ -0,0 +1,169 @@
...
+        fprintf(stdout, "\t%s", sam_hdr_tid2name(in_samhdr, bamdata->core.tid));        //retrieves the target name using the value in bam and by referring the header
+        fprintf(stdout, "\t%d", bamdata->core.pos + 1);                                 //internally position is 0 based and on text output / SAM it is 1 based
+        fprintf(stdout, "\t%d", bamdata->core.qual);                                    //map quality value
+
+        cigar = bam_get_cigar(bamdata);                                                 //retrieves the cigar data
+        fprintf(stdout, "\t");
+        for (i = 0; i < bamdata->core.n_cigar; ++i) {                                   //no. of cigar data entries
+            fprintf(stdout, "%d%c", bam_cigar_oplen(cigar[i]), bam_cigar_opchr(cigar[i]));  //the macros gives the count of operation and the symbol of operation for given cigar entry
+        }
+        fprintf(stdout, "\t%s", bamdata->core.mtid == bamdata->core.tid ? "=" : sam_hdr_tid2name(in_samhdr, bamdata->core.mtid));   // = if rnext and pnext are same and the target name otherwise
```
jkbonfield, line 132: Good demonstration of the API for querying reference names.  Maybe print it up in both numeric and string form.  Eg:

```
printf("Mate-ref:\t#%d %s\n", b->core.mtid, sam_hdr_tid2name(hdr, b->core.tid));
```

We don't need to try and recreate SAM's "=" notation here either, as we shouldn't attempt to recreate a format ourselves.  We just want to demomstrate how you access the data fields.

samples/read_aux.c
```diff
@@ -0,0 +1,210 @@
...
+    static const struct option lopts[] = {
+        {NULL, 0, NULL, 0}
+    };
+
+    const char *inname = NULL;
+    int c = 0, ret = EXIT_FAILURE, idx = -1;
+    sam_hdr_t *in_samhdr = NULL;
+    samFile *infile = NULL;
+
+    int ret_r = 0, i = 0, auxcnt = 0, auxBcnt = 0, cnt = 0;
```
jkbonfield, line 120: auxcnt and auxBcnt are unused variables.

samples/read_aux.c
```diff
@@ -0,0 +1,210 @@
...
+
+    //read_aux [-n position] [-i infile] [<infile>]
+    while ((c = getopt_long(argc, argv, "i:n:", lopts, NULL)) >= 0)
+    {
+        switch (c)
+        {
+        case 'i':                           //in file name
+            inname = optarg;
+            break;
+        case 'n':                           //position
```
jkbonfield, line 132: The comments for read_bam.c apply here too.  It's not position, but the Nth record.  Again consider reporting the first N instead of Nth.

samples/read_aux.c
```diff
@@ -0,0 +1,210 @@
...
+
+/// printauxdata - prints aux data
+/** @param fp - file to which it to be printed - stdout or null
+ *  @param type - aux type
+ *  @param idx - index in array, -1 when not an array type
+ *  @param data - data
+ *  recurses when the data is array type
+returns 1 on failure 0 on success
+*/
+int printauxdata(FILE *fp, char type, int32_t idx, const uint8_t *data)
```
jkbonfield, line 51: See `bam_aux_get_str` and `sam_format_aux1` APIs in htslib/sam.h.

These basically replace this with code that converts and aux tag to a printable kstring.

Although we may wish to demonstrate somewhere how to extract the value numerically and do something with it.  bam_aux2i etc.

jkbonfield (commented):
I think both read_bam and read_aux would benefit from deliberately not trying to look like SAM.  Maybe it more of a debug or "dump" tool.

Also for the aux handling, demonstrating the aux iterator is nice, but we should also demonstrate querying with eg

```
char *c = bam_aux_get(b, "NM");
int NM = c ? bam_aux2i(c) : 0;
```

So maybe the tool can print up entire lines or just report on specific tag types.

samples/read_aux.c
vasudeva8: The intent was the later, to demonstrate the raw access of data, based on aux type - basic and array of different data types.

jkbonfield:
I still have specifics to go through with the demos, but I decided to take a more high level overview of things.  I should have started with this before, sorry.

Basically we have a few categories of tools, and their closest matching samtools equivalent where possible:

- Samtools flagstat:  flags_demo / flags_htsopt_field / flags_tpool.
- Samtools view read/write loop:  read_bam / index_read / index_reg_read / index_multi_read / index_multireg_read / index_write.
- Samtools index: index_create
- Pileup API: pileup.c / mpileup.c
- SAM aux tag manipulation: read_aux.c
- Header API: read_header / update_header
- split.c ?

So we have a lot of tools, but not too many themes.  That's good as it permits a better narrative in a tutorial or sets of related worked examples.  Some of the APIs though I think are in the wrong place.

Flagstat is a read-only tool, which is good for demonstrating decoding of the basic sequence API.  This may mean read_bam is irrelevant.  Or perhaps read_bam belongs in the same chapter / topic as the flagstat one.  Flagstat is a perfect example of the basic `hts_set_opt` API with `HTS_OPT_NTHREADS` and `CRAM_OPT_REQUIRED_FIELDS`.  It's not appropriate for the thread pool, as we only have one thing to do.

The read/write loops could may be considered as a file format conversion tool, so we can introduce filename handling and setting format based on suffixes.  All tools currently just use `sam_open(..., "w")` to write SAM, so we're missing useful functionality here.  A read/write loop is the perfect place to bring in using a shared thread pool, instead of explicitly controlling a per-file-descriptor thread count.

Pileup is its own thing, and definitely needs its own tutorial section.  I'm happy with pileup vs mpileup, but the constructor/destructor interface is an extra complexity that has no purpose in the examples.  We probably need something more advanced, such as optimising away an overhead from being computed per-column to per-read.  There are choices there too - caching numerics in pileup1_t->cd.i vs caching complex things needing memory allocation, which is where the constructor/destructor come in.  I'd need to think of a concise example, but test/pileup_mod.c is one such use where `bam_parse_basemod` is called during the constructor.

The index querying can come in at any point I guess, as it applies equallty to both a read-only scenario and a read-write scenario.  I'm not sure which topic is the best one to add it to, but it's OK in the read/write code.

The header API feels a bit disjoint, as well as split.c which I can't follow the intention of.  Maybe these two can be merged into the same overall topic.  Eg split by read-group requires processing of the header, and maybe we'd want to show how we can add new PG entries or remove other RG lines when splitting.  Again this is a perfect example of where the thread pool can be useful as we're writing out several files at once.

Aux tag manipulation could do with extending.  Right now it's just regurgitating the aux tag data, which isn't so helpful (and also not the ideal way to do it).  Maybe we need something to summarise tag utilisation (tagstats), or to add a new tag computed from the read data.  Eg adding a cg:f tag to report the CG %age for each read.  Maybe also for fun gather statistics on CG content vs MQUAL.  Infact the latter half of that could be a read-BAM-only while the former is a read-write loop, so perhaps it's better than the flagstat and view alternatives?  Thoughts?

samples/flags_demo.c
```diff
@@ -0,0 +1,182 @@
...
+    };
+    flags flgcnt = {0};                     //count for each flag
+    const char *inname = NULL;              //input file name
+    int c = 0, ret = EXIT_FAILURE;
+    samFile *infile = NULL;                 //sam file
+    sam_hdr_t *in_samhdr = NULL;            //header of file
+    bam1_t *bamdata = NULL;                 //to hold the read data
+    size_t cnt = 0;
+
+    //[-i infile] [<infile>]
```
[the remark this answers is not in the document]
vasudeva8: Modified.

samples/flags_demo.c
```diff
@@ -0,0 +1,182 @@
...
+    //read header
+    in_samhdr = sam_hdr_read(infile);
+    if (!in_samhdr)
+    {
+        fprintf(stderr, "Failed to read header from file!\n");
+        goto end;
+    }
+
+    //read data, check flags and update count
+    while (0 <= (c = sam_read1(infile, in_samhdr, bamdata))) {
```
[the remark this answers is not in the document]
vasudeva8: Updated

samples/flags_demo.c
```diff
@@ -0,0 +1,182 @@
...
+        %d unmapped and %d mate unmapped\n \
+        %d reversed and %d mate reversed\n \
+        %d from read1 and %d read2\n \
+        %d secondary and %d supplementary\n \
+        %d duplicates and %d qc failures\n",
+        inname, cnt, flgcnt.paired, flgcnt.proppaired, flgcnt.unmapped, flgcnt.mateunmapped, flgcnt.reveresed, flgcnt.matereversed, flgcnt.read1, flgcnt.read2,
+        flgcnt.secondary, flgcnt.supplementary, flgcnt.duplicate, flgcnt.qcfailed);
+
+        ret = EXIT_SUCCESS;
+    }
```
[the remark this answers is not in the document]
vasudeva8: Updated.

samples/flags_demo.c
```diff
@@ -0,0 +1,182 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE
+
+*/
+
+/* The pupose of this code is to demontrate the library apis and need proper error handling and optimization */
```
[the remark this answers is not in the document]
vasudeva8: Updated

samples/split.c
vasudeva8: Modified.

samples/read_header.c
```diff
@@ -0,0 +1,165 @@
...
+            id = sam_hdr_line_name(in_samhdr, *(header + c), pos);
+            if (tag) {
+                //get value of given tag for given unique identifier
+                sam_hdr_find_tag_id(in_samhdr, *(header + c), *(idf + c), id, tag, &data);
+            }
+            else {
+                //get line at given position
+                sam_hdr_find_line_pos(in_samhdr, *(header + c), pos, &data);
+            }
+            printf("\n%s\n", data.s);
```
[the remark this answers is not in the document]
vasudeva8: Updated.

samples/read_header.c
```diff
@@ -0,0 +1,165 @@
...
+        if (linecnt) {
+            //get unique identifier field's value for given type
+            id = sam_hdr_line_name(in_samhdr, *(header + c), pos);
+            if (tag) {
+                //get value of given tag for given unique identifier
+                sam_hdr_find_tag_id(in_samhdr, *(header + c), *(idf + c), id, tag, &data);
+            }
+            else {
+                //get line at given position
+                sam_hdr_find_line_pos(in_samhdr, *(header + c), pos, &data);
```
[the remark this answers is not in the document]
vasudeva8: Added.

samples/read_header.c
vasudeva8: Updated.

samples/read_bam.c
vasudeva8: Updated

samples/read_bam.c
vasudeva8: Updated.

samples/read_bam.c
vasudeva8: Updated.

samples/read_bam.c
vasudeva8: Updated.

samples/read_aux.c
vasudeva8: Updated.

samples/read_aux.c
vasudeva8: updated.

PR #1594, opened 2023-03-30
Remove use of sprintf() from HTSlib source

### LINKED ISSUE #1586: snprintf instead of unsafe sprintf
`HTSLib` is a dependency for >20 different R packages including the core ones of Bioconductor. Next R release (v4.3) [will issue a warning](https://cran.r-project.org/doc/manuals/r-devel/NEWS.html#:~:text=The%20use%20of%20sprintf) upon use of `sprintf` in C/C++ code, which is then treated as an error in CRAN's `R CMD check` (routine check for all R packages). Besides, this warning is simply annoying and indeed flags a security risk.
Suggested solution for this is to substitute all `[v]sprintf`s with `[v]snprintf`s. This probably should be done in `HTSlib` as well as in `samtools` code. For an experienced C/C++ (`HTSlib`) developer it is very easy, so I'm not sure it is a good idea to open a PR myself. But please let me know if it is of interest.
Thanks for looking into this!

daviesrob:
None of these instances were really a problem, but using it upsets some downstream packagers (notably R).  The easiest way to keep them happy is to stop using it and (mostly) switch to snprintf() instead.

Fixes #1586

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -2531,7 +2531,7 @@ static refs_t *refs_load_fai(refs_t *r_orig, const char *fn, int is_err) {
         /* Only the reference file provided. Get the index file name from it */
             if (!(r->fn = string_dup(r->pool, fn)))
                 goto err;
-            sprintf(fai_fn, "%.*s.fai", PATH_MAX-5, fn);
+            snprintf(fai_fn, PATH_MAX, "%.*s.fai", PATH_MAX-5, fn);
         }
     }
 
@@ -4816,7 +4816,7 @@ static void full_path(char *out, char *in) {
         strncpy(out, in, PATH_MAX-1);
         out[PATH_MAX-1] = 0;
     } else {
-        int len;
+        size_t len;
 
         // unable to get dir or out+in is too long
         if (!getcwd(out, PATH_MAX) ||
@@ -4826,7 +4826,7 @@ static void full_path(char *out, char *in) {
             return;
         }
 
-        sprintf(out+len, "/%.*s", PATH_MAX - 2 - len, in);
+        snprintf(out+len, PATH_MAX - len, "/%s", in);
 
         // FIXME: cope with `pwd`/../../../foo.fa ?
     }
--- a/hfile_s3.c
+++ b/hfile_s3.c
@@ -451,12 +451,12 @@ static int auth_header_callback(void *ctx, char ***hdrs) {
 
 /* like a escape path but for query strings '=' and '&' are untouched */
 static char *escape_query(const char *qs) {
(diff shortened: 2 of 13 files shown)


jmarshall:
Code in _test/*.c_ doesn't go into the library, so you could consider leaving those ones as is.

hfile_s3.c
```diff
@@ -467,29 +467,25 @@ static char *escape_query(const char *qs) {
...
             j += 3;
         }
     }
 
-    if (i != length) {
-        // in the case of a '?' copy the rest of the qs across unchanged
-        strcpy(escaped + j, qs + i);
-    } else {
-        escaped[j] = '\0';
-    }
```
jkbonfield, line 480: I noticed this while looking at it a few days ago.  It's an incorrect copy from escape_path, irrelevant here because it doesn't have the `if (c == '?') break` clause.

However it should probably get a mention in the commit message.

plugin.c
```diff
@@ -210,7 +210,7 @@ const char *hts_plugin_path(void) {
     }
 
     static char s_path[1024];
-    sprintf(s_path, "%.1023s", ks.s ? ks.s : "");
+    snprintf(s_path, sizeof(s_path), "%.1023s", ks.s ? ks.s : "");
```
jkbonfield, line 213: Again we may as well change this to `%s` now we don't need the length protection.

cram/cram_io.c
```diff
@@ -2531,7 +2531,7 @@ static refs_t *refs_load_fai(refs_t *r_orig, const char *fn, int is_err) {
         /* Only the reference file provided. Get the index file name from it */
             if (!(r->fn = string_dup(r->pool, fn)))
                 goto err;
-            sprintf(fai_fn, "%.*s.fai", PATH_MAX-5, fn);
+            snprintf(fai_fn, PATH_MAX, "%.*s.fai", PATH_MAX-5, fn);
```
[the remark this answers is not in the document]
daviesrob: It depends if you want to keep the `.fai` on the end...

As this involves `PATH_MAX` which may be removed later, I'm inclined to leave this as-is for now.

hfile_s3.c
daviesrob: Now mentioned.

plugin.c
daviesrob: Updated.

cram/cram_io.c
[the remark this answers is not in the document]
jkbonfield: Good point.  I guess it's better if it truncates the pathname leaving .fai, as it prevents something catastrophic such as overwriting the main file.

PR #1595, opened 2023-03-30
Fix bug where bin number could overflow when looking for max_off

daviesrob:
When searching for `max_off`, `hts_itr_query()` and `hts_itr_multi_bam()` look for a bin to the right of the end of the region.  For whole chromosomes, this would be HTS_POS_MAX, which is far beyond the maximum bin position supported.  The `bin` calculation overflowed leading to either a negative bin number or an incorrect positive one, depending on the number of levels in the index.  Negative bin numbers simply caused time to be wasted as the search loop eventually counted up to zero, but incorrect positive ones could cause the iterator to finish too early.

Fix by catching the out-of-bounds case and setting max_off to UINT64_MAX, which should be used for bins beyond the end of the indexable range.

Luckily in practice this didn't cause too much harm, at least for the default `min_shift` value.  Indexes with up to six levels overflowed to negative bin numbers.  For seven you got one referencing a region starting at about 17Gbases, and 257Gbases for eight, so it's unlikely searches on real data were affected.  The fix is trivial though and avoids some negative value shifts so worth doing.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -3145,17 +3145,24 @@ hts_itr_t *hts_itr_query(const hts_idx_t *idx, int tid, hts_pos_t beg, hts_pos_t
             }
 
             // compute max_off: a virtual offset from a bin to the right of end
-            bin = hts_bin_first(idx->n_lvls) + ((end-1) >> idx->min_shift) + 1;
-            if (bin >= idx->n_bins) bin = 0;
-            while (1) {
-                // search for an extant bin by moving right, but moving up to the
-                // parent whenever we get to a first child (which also covers falling
-                // off the RHS, which wraps around and immediately goes up to bin 0)
-                while (bin % 8 == 1) bin = hts_bin_parent(bin);
-                if (bin == 0) { max_off = (uint64_t)-1; break; }
-                k = kh_get(bin, bidx, bin);
-                if (k != kh_end(bidx) && kh_val(bidx, k).n > 0) { max_off = kh_val(bidx, k).list[0].u; break; }
-                bin++;
+            // First check if end lies within the range of the index (it won't
+            // if it's HTS_POS_MAX)
+            if (end < 1LL << (idx->min_shift + 3 * idx->n_lvls)) {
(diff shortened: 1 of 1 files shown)


hts.c
```diff
@@ -3145,17 +3145,24 @@ hts_itr_t *hts_itr_query(const hts_idx_t *idx, int tid, hts_pos_t beg, hts_pos_t
...
-                // parent whenever we get to a first child (which also covers falling
-                // off the RHS, which wraps around and immediately goes up to bin 0)
-                while (bin % 8 == 1) bin = hts_bin_parent(bin);
-                if (bin == 0) { max_off = (uint64_t)-1; break; }
-                k = kh_get(bin, bidx, bin);
-                if (k != kh_end(bidx) && kh_val(bidx, k).n > 0) { max_off = kh_val(bidx, k).list[0].u; break; }
-                bin++;
+            // First check if end lies within the range of the index (it won't
+            // if it's HTS_POS_MAX)
+            if (end < 1LL << (idx->min_shift + 3 * idx->n_lvls)) {
```
jkbonfield, line 3150: It's a useful shortcut I think, but it looks like using uint64_t for bin would have made the old code work too, as `hts_bin_first` is then huge and bin gets reset to 0, falling out of the while loop with `max_off = UINT64_MAX`.

Otherwise we're still using a negative bin below, eg:

```
            for (i = n_off = 0; i < iter->bins.n; ++i)
                if ((k = kh_get(bin, bidx, iter->bins.a[i])) != kh_end(bidx))
                    n_off += kh_value(bidx, k).n;
```

jkbonfield: Ah I now see how to get a positive legal bin and still cause mayhem, so agreed your solution of checking for the maximum value is definitely the best way of handling it.

```
$ tabix -m10 -f /tmp/_.bed.gz;./tabix /tmp/_.bed.gz a
```

That takes 10 sec to query my 5 entry bed file.  (Admittedly with debugging on)

PR #1619, opened 2023-05-18
Add new annot-tsv program

pd3:
See also
    https://raw.githubusercontent.com/pd3/utils/master/annot-regs/doc.annot-regs.pdf
    https://github.com/pd3/utils/tree/master/annot-regs


pd3:
Mmm, I am confused why the tests complain about `strdup` being implicitly declared after `#include <string.h>` has been added while tabix code gets away with it...?

pd3:
> In subsequent emails I think I decided on better terminology of "input" and "annotation". "Source" still feels nebulous and not very descriptive of the function.

Unless there is a danger of confusion, let's preserve the existing naming please. It is to avoid conflicting short names, I already had to make sacrifices by abandoning the term "destination" which you disliked.

The naming was intended to imply the information flow and for that reason I liked best the original "source" and "destination". Using that naming would also prevent the next question:



> This is a bit ambiguous as it sort of implies it can work in either direction.
> 
> I'm still struggling to work out if it's completely symmetric and we basically have file1 file2 that are merged (akin to unix `join`), or whether we have a primary input file and an annotation file that is used to modify the input.

It is not symmetrical and is not intended to imply a change in direction. It aims to explain that the argument to `--transfer` are  column names, in the source and the target file, and what happens if such column does or does not exist in the source and the target file: if the column exists in both, values in the target file are overwritten by values from the source file on match or left as is when there is no match; if such column does not exist in the target file, a new column with that name will be created in the target file; if such column does not exist in the source file, the name is interpreted as a string to be filled in the target file. For example, if the column `MATCH` does not exist in either file, `--transfer MATCH:MATCH` (which can be abbreviated to `--transfer MATCH`) will create a new column `MATCH` in the target file and put the value `MATCH` whenever there is a match or a dot when there isn't.

@daviesrob, @whitwham could you please help to improve the wording of the man page and the usage text to make it clearer? Despite my best effort I am not succeeding in  making this understandable.

> I don't understand the ordering though. What's the use case for it working on unordered data? What's the benefit here? It just feels like it'd make it far slower than necessary. Or the half way house: is there a benefit to one order (primary) and one unordered (annotation) file?

I am confused by the question. There are two types of ordering you might be asking, by column and by row. You seem to be asking about column ordering, which is the only type of ordering mentioned throughout the man page and the code. 

To state the obvious, the benefit of supporting tab-delimited files with columns in arbitrary order is that one can use the program on files with columns in arbitrary order. For example, on files that were exported from spreadsheets, people tend to use them a lot. Regarding ordering by row, even though that is not mentioned anywhere, the program can work with data unordered by row as well. That's thanks to the regidx library.

As a side note, comparing the program to `awk`, `join` etc is misleading. It is also a bit like `grep` and `bedtools`, but mostly provides a unique functionality that is not found elsewhere.

jkbonfield:
I meant row ordering, so classic sorted vs unsorted (or non-position sorted) genomic data.

We have lots of algorithms that are efficient because they require sorted data.  Eg samtools merge.  Here you seem to be stating that both files can be unsorted.  This obviously means it is using a less efficient strategy, which is fine if that is the user requirement.

I'm just asking where that requirement for operating on unsorted data comes from.  What inputs are we dealing with that aren't sorted?  (Even unix "join" requires sorted inputs)

jkbonfield:
What specifies the ordering?

My source file has columns "tag3" and "m" in that order, but I always get them in the opposite order regardless of what I specify in the `-f` option.

```
@ seq4d[samtools.../htslib]148; cat /tmp/src.1.txt
#chr	beg	end	tag3	m
1	14	15	a	1
1	35	35	b	01
@ seq4d[samtools.../htslib]; ./annot-tsv -s /tmp/src.1.txt -t /tmp/dst.1.txt -f m,tag3
#[1]chr	[2]beg	[3]end	[4]tag1	[5]tag2	[6]m	[7]tag3
1	14	15	x1	x2	1	a
1	35	35	y1	y2	01	b
@ seq4d[samtools.../htslib]; ./annot-tsv -s /tmp/src.1.txt -t /tmp/dst.1.txt -f tag3,m
#[1]chr	[2]beg	[3]end	[4]tag1	[5]tag2	[6]m	[7]tag3
1	14	15	x1	x2	1	a
1	35	35	y1	y2	01	b
```

Also, please consider making the usage appear to stdout unless an error has occurred.  This is already a cause of a lot of "known failures" in the Samtools test harness, and we shouldn't be adding more here.  Specifically an error is an error (eg `anno-tsv --unknown-opt`), but a request for usage (`anno-tsv -h`) is not.  Same with exit codes.  `-h` should have exit code 255.  (Nor should errors really - you probably wanted `exit(1)` instead of `exit(-1)` there)

pd3:
> I can do a nop function of specifying nothing to transfer over:

In this mode it functions as `grep` would.

pd3:
> I notice a few things. Firstly, the header line has changed format. Is that deliberate or just debugging left over? It's undocumented.

That's a deliberate feature. If it is bothersome to anyone, we can add an option to not modify the header. I would leave this for future improvements though.

pd3:
> Also what you label as the dst file in your tests (old destination) is the default output, and not the source as I had expected! If we use `-f` to transfer tags, then these come from source and not the target file. This is still confusingly named therefore. PLEASE consider just renaming the input to the "input" as it's then crystal clear what it is.

I really don't understand the problem. The documentation is clear on that, no? The program transfers columns from the file with source annotations (`-s`) to the destination file, newly "target" file (`-t`).
```
  -s, --source-file FILE          Source file to take annotations from
  -t, --target-file FILE          Target file to be extend with annotations from -s
```

To turn this around: both files are technically inputs and typically both have annotations. I can be equally confused here.

pd3:
> ```
> ./annot-tsv -s /tmp/src.1.txt -t /tmp/dst.1.txt -f tag3
> AddressSanitizer:DEADLYSIGNAL
> =================================================================
> ==8978==ERROR: AddressSanitizer: SEGV on unknown address (pc 0x0000004d3015 bp 0x7fff4ad0c9d0 sp 0x7fff4ad0c760 T0)
> ==8978==The signal is caused by a READ memory access.
> ==8978==Hint: this fault was caused by a dereference of a high value address (see register values below).  Disassemble the provided pc to learn which register was used.
>     #0 0x4d3015 in __ac_X31_hash_string /nfs/users/nfs_j/jkb/work/samtools_master/htslib/./htslib/khash.h:401
>     #1 0x4d3015 in kh_get_str2int /nfs/users/nfs_j/jkb/work/samtools_master/htslib/./htslib/khash_str2int.h:30
>     #2 0x4d3015 in khash_str2int_has_key /nfs/users/nfs_j/jkb/work/samtools_master/htslib/./htslib/khash_str2int.h:69
>     #3 0x4d3015 in process_line /nfs/users/nfs_j/jkb/work/samtools_master/htslib/annot-tsv.c:685
>     #4 0x4d596a in main /nfs/users/nfs_j/jkb/work/samtools_master/htslib/annot-tsv.c:875
>     #5 0x7f8fdbee3c86 in __libc_start_main /build/glibc-CVJwZb/glibc-2.27/csu/../csu/libc-start.c:310
>     #6 0x41d939 in _start ??:?
> 
> AddressSanitizer can not provide additional info.
> SUMMARY: AddressSanitizer: SEGV /nfs/users/nfs_j/jkb/work/samtools_master/htslib/./htslib/khash.h:401 in __ac_X31_hash_string
> ==8978==ABORTING
> Aborted
> ```

Thank you, added a commit to address issues like this https://github.com/samtools/htslib/pull/1619/commits/ee299e355665ee5443449d5cc26ea714f19ea90e. More breaking test cases will be much appreciated

pd3:
> What specifies the ordering?
> 
> My source file has columns "tag3" and "m" in that order, but I always get them in the opposite order regardless of what I specify in the `-f` option.

New columns are only appended, never added among existing ones. I suspect the dst.1.txt file in this example contains the column `m` already? Reordering of existing columns in the target/destination file is not considered in this version, perhaps it could be addressed in a future enhancement request.

annot-tsv.txt
daviesrob, line 1: I'd prefer to only have one copy of the manual page checked in - and preferably the nroff one so HTSlib doesn't gain a build dependency on asciidoc.

annot-tsv.1
```diff
@@ -0,0 +1,216 @@
...
+.fi
+.if n \{\
+.RE
+.\}
+.SH "AUTHORS"
+.sp
+The program was written by Petr Danecek and was originally published on github as annot\-regs
+.SH "COPYING"
+.sp
+The MIT/Expat License or GPL License, see the LICENSE document for details\&. Copyright (c) Genome Research Ltd\&.
```
daviesrob, line 216: MIT/Expat only.  GPL is not currently mentioned in the LICENSE document, and I don't see why anyone would want to use it if you can go for MIT/Expat instead.

annot-tsv.c
```diff
@@ -0,0 +1,886 @@
...
+// coordinates are converted to beg<<1 and end coordinates
+// to (end<<1)+1.
+#define NBP_SET_BEG(x) ((x)<<1)
+#define NBP_SET_END(x) (((x)<<1)+1)
+#define NBP_GET(x)     ((x)>>1)
+#define NBP_IS_BEG(x)  (((x)&1)==0)
+#define NBP_IS_END(x)  (((x)&1)==1)
+typedef struct
+{
+    int n,m;            // n is a multiple of two: breakpoints are stored in regs, not regions
```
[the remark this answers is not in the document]
pd3: `int` or `uint` should be sufficient for any normal use. But ok, no harm in using `size_t`.

annot-tsv.c
```diff
@@ -0,0 +1,886 @@
...
+#include "htslib/bgzf.h"
+#include "htslib/regidx.h"
+
+#define ANN_NBP     1
+#define ANN_FRAC    2
+#define ANN_CNT     4
+
+typedef struct
+{
+    uint32_t n,m;
```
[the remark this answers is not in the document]
pd3: This would mean a file with more than 4,294,967,295 columns

annot-tsv.txt
pd3: I agree with the reserveration to asciidoc though, asciidoctor is what we've been using in bcftools

jkbonfield: I think I said it's fine in principle, but if we use it it should be universal and not just one thing.  Adding dependencies willy nilly isn't pleasant for users so we have to be considered and not just drift into decisions like this.

Bcftools aleady uses this, so it's fine there.  Htslib right now is `nroff -man` so we should stick to that unless we decide to move enmasse to a new regime.

jkbonfield: Also, can we please cease taking discussions offline.  It's a pain to keep switching back between github and private emails.  So much of the original discussion on this was in email threads, which take ages to sift through and search (thanks to our glacial email server).

test/test.pl
```diff
@@ -154,6 +210,37 @@ sub _cmd
...
+{
+    my ($cmd) = @_;
+
+    my $tmp = "$$opts{tmp}/tmp";
+    my $pid = fork();
+    if ( !$pid )
+    {
+        exec('bash', '-o','pipefail','-c', "($cmd) 2>$tmp.e >$tmp.o");
+    }
+    waitpid($pid,0);
```
daviesrob, line 223: This could just be replaced by
```
system('bash', '-o','pipefail','-c', "($cmd) 2>$tmp.e >$tmp.o");
```

Compared with `_cmd()`, the `$ENV{TEST_PRECMD}` functionality has been lost (although we don't use it that much these days, anyway).

test/test.pl
```diff
@@ -154,6 +210,37 @@ sub _cmd
...
+
+    my $status  = $? >> 8;
+    my $signal  = $? & 127;
+
+    my (@out,@err);
+    if ( open(my $fh,'<',"$tmp.o") )
+    {
+        @out = <$fh>;
```
daviesrob, line 231: I'm not sure why we use an array for this (it's in other code too).  It would be more efficient to just read everything into a string:
```
local $/; # Read whole file
$out = <$fh>;
```

test/test.pl
```diff
@@ -216,7 +304,16 @@ sub test_cmd
         }
         else
         {
-            failed($opts,$test,"The outputs differ:\n\t\t$$opts{path}/$args{out}\n\t\t$$opts{path}/$args{out}.new");
+            if ( exists($args{exp}) && !-e "$$opts{path}/$args{out}" )
```
daviesrob, line 307: The second part of this will always be false due to the condition for the surrounding `if` statement.

test/test.pl
```diff
@@ -175,8 +262,9 @@ sub test_cmd
     print "$test:\n";
     print "\t$args{cmd}\n";
 
-    my ($ret,$out) = _cmd("$args{cmd}");
-    if ( $ret ) { failed($opts,$test); return; }
+    my ($ret,$out,$err) = _cmd3("$args{cmd}");
+    if ( length($err) ) { $err =~ s/\n/\n\t\t/gs; $err = "\n\n\t\t$err\n"; }
```
daviesrob, line 266: Can be done more easily with `$err =~ s/^/\t/mg;`

test/test.pl
```diff
@@ -216,7 +304,16 @@ sub test_cmd
...
         else
         {
-            failed($opts,$test,"The outputs differ:\n\t\t$$opts{path}/$args{out}\n\t\t$$opts{path}/$args{out}.new");
+            if ( exists($args{exp}) && !-e "$$opts{path}/$args{out}" )
+            {
+                open(my $fh,'>',"$$opts{path}/$args{out}") or error("$$opts{path}/$args{out}");
+                print $fh $exp;
+                close($fh);
+            }
+            my @diff = `diff $$opts{path}/$args{out} $$opts{path}/$args{out}.new`;
```
daviesrob, line 313: Easier to capture into a string, then use `s/^/\t/mg`.

test/test.pl
```diff
@@ -154,6 +210,37 @@ sub _cmd
...
+    my $tmp = "$$opts{tmp}/tmp";
+    my $pid = fork();
+    if ( !$pid )
+    {
+        exec('bash', '-o','pipefail','-c', "($cmd) 2>$tmp.e >$tmp.o");
+    }
+    waitpid($pid,0);
+
+    my $status  = $? >> 8;
+    my $signal  = $? & 127;
```
daviesrob, line 226: This value is never used.

test/test.pl
```diff
@@ -34,33 +34,61 @@
...
-test_vcf_various($opts);
-test_bcf_sr_sort($opts);
-test_command($opts,cmd=>'test-bcf-translate -',out=>'test-bcf-translate.out');
-test_convert_padded_header($opts);
-test_rebgzip($opts);
-test_logging($opts);
-test_plugin_loading($opts);
-test_realn($opts);
-test_bcf_set_variant_type($opts);
+run_test(\&test_bgzip,$opts, 0);
```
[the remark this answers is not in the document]
pd3: I know it is not bullet proof for all tests, but it works if you want to rerun and debug the ones you're currently developing. Saves lot of time and does not have to be used, I'd advocate for keeping it

test/test.pl
pd3: OK

test/test.pl
pd3: This is just out of habit, usually I'd do something with the output per-line. I don't think this is a performance issue in these tests, but okay.

daviesrob:
Rebased, with some squashing of a couple of trivial bug-fix commits, and some adjustments from me...

PR #1630, opened 2023-06-05
Add support for non-standard chromosome names containing [:-] characters

### LINKED ISSUE #1620: Synced reader does not support regions with contigs including colons
The synced_bcf_reader currently (as of version 1.17) fails with an error when it is initialised via `bcf_sr_regions_init` using region strings containing contig names with colons, despite the fact that colons are [no longer disallowed characters in contig names since VCFv4.3](https://samtools.github.io/hts-specs/VCFv4.3.pdf):

> The VCF specification previously disallowed colons (‘:’) in contig names to avoid confusion when parsing
breakends, but this was unnecessary. Even with contig names containing colons, the breakend mate position
notation can be unambiguously parsed because the “:pos” part is always present.

The underlying `_regions_init_string` function will likely need to be updated to scan for colons from the back instead of the front to accommodate this change of spec.

pd3:
Note hts_parse_region() cannot be used because it requires the header and without the header the caller does not learn the contig name.

Resolves #1620

### DIFF
--- a/htslib/synced_bcf_reader.h
+++ b/htslib/synced_bcf_reader.h
@@ -336,6 +336,8 @@ int bcf_sr_set_regions(bcf_srs_t *readers, const char *regions, int is_file);
  *              supply 'from' in place of 'to'. When 'to' is negative, first
  *              abs(to) will be attempted and if that fails, 'from' will be used
  *              instead.
+ *              If chromosome name contains the characters ':' or '-', it should
+ *              be put in curly brackets, for example as "{weird-chr-name:1-2}:1000-2000"
  *
  *  The bcf_sr_regions_t struct returned by a successful call should be freed
  *  via bcf_sr_regions_destroy() when it is no longer needed.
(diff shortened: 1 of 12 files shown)


jkbonfield:
Apparently I had misremembered the public `hts_parse_region` API as having the header as an optional field.   Pity :/

I'm inclined to think that if we have a requirement to parse regions without a header being present, then perhaps this API should be modified to cope with it.  My preference would be to avoid code duplication where possible, but I haven't really looked in detail to know how much work is involved and whether we just replace duplication in one place with it in another (eg a separate code-path for headerless parsing).

pd3:
I just checked the code and can confirm that `test-bcf-sr` does not test indexes at all. That's done properly only through the `bcftools` test suite, and there are hundreds of them.  If you want to test it natively in htslib, best might be to write a simplified version of `vcfview.c`. Tests for this pull requests are https://github.com/samtools/bcftools/pull/1938/files.

daviesrob:
Merge commit removed...

Issue #1617, opened 2023-05-05
Segfault when writing uncompressed BAM


tijyojwad:
Hi,

I am running into a segfault when I attempt to write a header into an uncompressed BAM. I'm opening the file using `hts_open(fname, "wbu")`. I tried with the commit `6143086502567c5c4bb5cacb2951f664ba28ed6e` (release 1.17) but still see that issue.

The same code runs if I just change the mode to "wb".

Could that be looked into?
Thanks!

tijyojwad:
thanks to a tip from @cjw85 I learnt that `wb0` did the trick for generating uncompressed BAMs. Could the API documentation be updated for that?

daviesrob:
`u` should work, and certainly shouldn't segfault.  We'll look into it.

PR #1632, opened 2023-06-05
changes to avoid segfault with uncompressed bam

vasudeva8:
Fix for #1617
when bam write is attempted with 'u', the bgzf compression/decompression buffers are not allocated as it denotes uncompressed output. bgzf_flush, called at end, accesses the compression buffers and causes segfault in such case.

Binary data files are implicitly expected to be bgzf compressed and this 'u' flag is not valid / supported.

The fix made is to remove the 'u' flag on write of bam/bcf that the output is written as bgzf compressed with default compression level.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -871,9 +871,9 @@ htsFile *hts_open_format(const char *fn, const char *mode, const htsFormat *fmt)
         *mode_c = format_to_mode[fmt->format];
     }
 
-    // Uncompressed bam/bcf is not supported, remove 'u' flag on write
+    // Uncompressed bam/bcf is not supported, change 'u' to '0' on write
     if (uncomp && *mode_c == 'b' && (strchr(smode, 'w') || strchr(smode, 'a'))) {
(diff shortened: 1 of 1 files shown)


hts.c
```diff
@@ -866,6 +871,11 @@ htsFile *hts_open_format(const char *fn, const char *mode, const htsFormat *fmt)
         *mode_c = format_to_mode[fmt->format];
     }
 
+    // Uncompressed bam/bcf is not supported, remove 'u' flag on write
+    if (uncomp && *mode_c == 'b' && (strchr(smode, 'w') || strchr(smode, 'a'))) {
+        memcpy(uncomp, uncomp + 1, cp2 - uncomp - 1);
```
daviesrob, line 876: You should use `memmove()` here because the source and destination buffers overlap, and [`memcpy()` disallows that](https://pubs.opengroup.org/onlinepubs/9699919799/functions/memcpy.html).  But better still would be to turn the 'u' into a '0', as the result will then be uncompressed while still BGZF wrapped.

vasudeva8: Updated accordingly.

PR #1637, opened 2023-06-30
Fix possible double frees in bcf_hdr_add_hrec() error handling

daviesrob:
Spotted while looking at the VCF parser.  `bcf_hdr_add_hrec()` should neither call `bcf_hrec_destroy(hrec)` nor store any pointers to `hrec` when it returns -1, otherwise double frees or stale pointer dereferences may result.

Remove `bcf_hrec_destroy(hrec)` call that was incorrectly made when handling a hash table insert failure, and move `hdr->hrec` reallocation so that all possible failures occur before `hrec` is added into the header.

Add `bcf_hdr_add_hrec()` documentation, including a warning that the caller should not touch `hrec` after a successful return.

It looks like the bug appeared in 3fe2a59034.  You'd have to be very unlucky to actually trigger it as it requires a particular allocation to fail, but it's best to ensure that the problem can't occur.

### DIFF
--- a/htslib/vcf.h
+++ b/htslib/vcf.h
@@ -691,6 +691,22 @@ set to one of BCF_ERR* codes and must be checked before calling bcf_write().
     HTSLIB_EXPORT
     int bcf_hrec_format(const bcf_hrec_t *hrec, kstring_t *str);
 
+    /// Add a header record into a header
+    /**
+     *  @param hdr  Destination header
+     *  @param hrec Header record
+     *  @return 0 on success, -1 on failure
+     *
+     *  If this function returns success, ownership of @p hrec will have
+     *  been transferred to the header structure.  It may also have been
+     *  freed if it was a duplicate of a record already in the header.
+     *  Therefore the @p hrec pointer should not be used after a successful
+     *  return from this function.
+     *
+     *  If this function returns failure, ownership will not have been taken
+     *  and the caller is responsible for cleaning up @p hrec.
+     */
+
     HTSLIB_EXPORT
     int bcf_hdr_add_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec);
 
--- a/vcf.c
+++ b/vcf.c
@@ -974,23 +974,27 @@ int bcf_hdr_add_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec)
             return 0;
         }
     }
+
+    // New record, needs to be added
+    int n = hdr->nhrec + 1;
+    bcf_hrec_t **new_hrec = realloc(hdr->hrec, n*sizeof(bcf_hrec_t*));
+    if (!new_hrec) {
+        free(str.s);
+        return -1;
+    }
+    hdr->hrec = new_hrec;
+
     if ( str.s )
     {
         khint_t k = kh_put(hdict, aux->gen, str.s, &res);
         if ( res<0 )
         {
-            bcf_hrec_destroy(hrec);
             free(str.s);
             return -1;
         }
         kh_val(aux->gen,k) = hrec;
     }
 
-    // New record, needs to be added
-    int n = hdr->nhrec + 1;
-    bcf_hrec_t **new_hrec = realloc(hdr->hrec, n*sizeof(bcf_hrec_t*));
-    if (!new_hrec) return -1;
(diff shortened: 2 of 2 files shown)


jkbonfield:
Definitely an improvement, but I'm wondering now about the `bcf_hdr_register_hrec` call at the start of the function.  This takes a copy of hrec and adds it to the header dictionaries even when it subsequently fails later on.

I can't find a `bcf_hdr_unregister_hrec` to remove it again if it fails.  What is the consequence of having an extra dict item which isn't referred to by the header?  Even if it's ignored, it does mean there is an extra pointer to it cached.  Although maybe that's not an issue if the header freeing doesn't free via the `hdr->dict` pointers but instead frees via `get_hdr_aux()` function.

jkbonfield:
Ah wait, I'm wrong.  `hdr->dict` *is* `get_hdr_aux()`, but for a specific element. :/

```
static inline bcf_hdr_aux_t *get_hdr_aux(const bcf_hdr_t *hdr)
{
    return (bcf_hdr_aux_t *)hdr->dict[0];
}
```

That `[0]` ought to be `BCF_DT_ID` so it's not obscure, given that `bcf_hdr_register_hrec` uses `hdr->dict[BCF_DT_CTG]` meaning I had to go hunting for the defines to work out if it was manipulating the same dictionary.  I'm also unsure why "aux" when it's "id"?  There's some confusing terminology in this code.

Anyway, later on there is also use of `BCF_DT_IT` in `bcf_hdr_register_hrec`:

```
    vdict_t *d = (vdict_t*)hdr->dict[BCF_DT_ID];
    k = kh_get(vdict, d, str);
    if ( k != kh_end(d) )
    {
        // already present
        free(str);
        if ( kh_val(d, k).hrec[info&0xf] ) return 0;
        kh_val(d, k).info[info&0xf] = info;
        kh_val(d, k).hrec[info&0xf] = hrec;
        if ( idx==-1 ) {
            if (hrec_add_idx(hrec, kh_val(d, k).id) < 0) {
                return -1;
            }
        }
        return 1;
    }
    k = kh_put(vdict, d, str, &ret);
```

So this function has clearly taken a copy of `hrec`, irrespective of whether the function call fails, meaning the comment you added is incorrect.  The user should probably just never touch `hrec` again after calling our API, and maybe we should make it free `hrec` itself given it mostly will already do this, even after an error.

daviesrob:
Yes, leaving a dangling pointer in `bcf_hdr_t::dict[].hrec[]` probably isn't good.  I don't think it's a problem for cleaning up though - `bcf_hdr_destroy()` removes `hrec`s [via `bcf_hdr_t::hrec[]`](https://github.com/samtools/htslib/blob/e86126bc45f6cf6b1c3d67e13de96aeaa5e58805/vcf.c#L1363).  `bcf_hdr_remove()` either [iterates through `bcf_hdr_t::hrec[]`](https://github.com/samtools/htslib/blob/e86126bc45f6cf6b1c3d67e13de96aeaa5e58805/vcf.c#L1179) or [checks that the pointer it gets from the dictionary is in there](https://github.com/samtools/htslib/blob/e86126bc45f6cf6b1c3d67e13de96aeaa5e58805/vcf.c#L1210) although the latter is a bit dodgy as it might throw an assertion (or do nothing if assertions were disabled).

So I think the main risk of a dangling pointer in `bcf_hdr_t::dict[].hrec[]` would come from failing to check that `bcf_hdr_add_hrec()` worked, managing to obtain a copy via `bcf_hdr_get_hrec()` and then using it.  Apart maybe from the issue with `bcf_hdr_remove()`, nothing should try to delete an `hrec` that way.

daviesrob:
I've added a commit that should prevent dangling pointers in `bcf_hdr_t::dict[].hrec[]`.

While working on that, I also noticed that `bcf_hdr_remove()` wasn't removing items from the `bcf_hdr_aux_t::gen` dictionary.  This is possibly a worse bug, as trying to retrieve the deleted item with `bcf_hdr_get_hrec()` would return the stale pointer.  A second commit fixes that problem, and adds some tests to ensure removal works correctly.

jkbonfield:
Scary stuff, but looks good.  I checked it doesn't trigger issues with bcftools too.

PR #1643, opened 2023-07-10
Fix decompress_peek_gz to cope with files starting on empty gzip blocks.

jkbonfield:
Demonstration:

```
$ echo -e "@seq\nAAAAA\n+\nABCDE" | gzip > _normal.gz                           
$ (gzip < /dev/null;cat _normal.gz) > _empty.gz                                 
                                                                                
$ ./htsfile _normal.gz                                                          
_normal.gz:     FASTQ gzip-compressed sequence data                             
                                                                                
$ ./htsfile _empty.gz                                                           
_empty.gz:      empty gzip-compressed data                                      
$ ./test/test_view _empty.gz                                                    
Unsupported or unknown category of data in input file                           
```

Zlib's `inflate` ends with `Z_STREAM_END` and then repeated calls to get more won't give any difference, so we need the explicit reset to move forwards with the next block.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -324,10 +324,24 @@ decompress_peek_gz(hFILE *fp, unsigned char *dest, size_t destsize)
     zs.avail_out = destsize;
     if (inflateInit2(&zs, 31) != Z_OK) return -1;
 
-    while (zs.total_out < destsize)
-        if (inflate(&zs, Z_SYNC_FLUSH) != Z_OK) break;
+    int ret;
+    unsigned char *last_in = buffer;
+    while (zs.total_out < destsize) {
+        ret = inflate(&zs, Z_SYNC_FLUSH);
+        if (ret == Z_STREAM_END && zs.avail_in && zs.total_out < destsize) {
+            if (last_in == zs.next_in)
+                break; // paranoia to avoid potential looping
+            else
+                last_in = zs.next_in;
+            inflateReset(&zs);
+            continue;
+        }
+        if (ret != Z_OK)
+            break;
+    }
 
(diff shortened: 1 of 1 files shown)


jkbonfield:
Now with added paranoia for loop detection.  Just incase...

hts.c
```diff
@@ -324,10 +324,24 @@ decompress_peek_gz(hFILE *fp, unsigned char *dest, size_t destsize)
...
+                last_in = zs.next_in;
+            inflateReset(&zs);
+            continue;
+        }
+        if (ret != Z_OK)
+            break;
+    }
 
-    destsize = zs.total_out;
+    // zs.total_out can sometimes be wrong as inflateReset resets it
```
jmarshall, line 343: If `total_out` may have been reset by `inflateReset`, is it still appropriate to use it in the loop above?

jkbonfield: Doh!  Good point.  No (although in practice avail_in seems to catch it).

I'll make a follow up fix with the similar `next_out - dest` logic instead.

jkbonfield: It appears I was being overly paranoid (although I left in the loop detection bit, even though I can't actually trigger zlib to consume no input without producing an error).

`avail_out` would have been the correct variable, as unlike `total_out` this isn't reset.  However it turns out neither are necessary.  Calling `inflate` with one of those as zero returns an error (Z_BUF_ERROR) so it promptly bails out the loop anyway, and even if it just returned Z_STREAM_END again it'd be caught by the other checks.

So it's now much simpler.  See #1646 for 2nd try at fixing this :)

PR #1645, opened 2023-07-13
Add fai_thread_pool interface.

### LINKED ISSUE #1638: Decompression threads for `bgzip`ped FASTA files
Hi,
I am using `faidx.h` functions to load compressed human genome into memory. Currently, it does not seem possible to bind decompression threads to file pointer of `faidx_t` structure, as it's not exposed in header file. For example, this code won't compile:

```
char fn[] = "/path/to/bgzipped.fasta.fa.gz";
int nthreads = 1;

faidx_t *faidx = fai_load(fn);
hts_tpool *tpool = hts_tpool_init(nthreads);
bgzf_thread_pool(faidx->bgzf, tpool, 0);
```

Unsavoury hack `bgzf_thread_pool(*(BGZF **)faidx, tpool, 0);` results in 30-40% faster reading using a single additional thread (please check [this discussion](https://www.biostars.org/p/9567843/)).

I don't know if gain is large enough... But would you consider opening `faidx_t` structure to allow quicker decompression?

jkbonfield:
This is a simple skim around bgzf_thread_pool.

Also added a missing `@param qsize` to the bgzf_thread_pool documentation.

Fixes #1638

### DIFF
--- a/htslib/faidx.h
+++ b/htslib/faidx.h
@@ -70,6 +70,9 @@ struct faidx_t;
 /// Opaque structure representing FASTA index
 typedef struct faidx_t faidx_t;
 
+/// Opaque structure; sole item needed from htslib/thread_pool.h
+struct hts_tpool;
+
 /// File format to be dealing with.
 enum fai_format_options {
     FAI_NONE,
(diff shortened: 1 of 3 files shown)


htslib/faidx.h
```diff
@@ -357,6 +357,15 @@ int fai_adjust_region(const faidx_t *fai, int tid,
...
 void fai_set_cache_size(faidx_t *fai, int cache_size);
 
+/// Adds a thread pool to the underlying BGZF layer.
+/** @param fai         FAI file handler
+ *  @param pool        The thread pool (see hts_create_threads)
+ *  @param qsize       The size of the job queue.  If 0 this is twice the
+ *                     number of threads in the pool.
+ */
+HTSLIB_EXPORT
+int fai_thread_pool(faidx_t *fai, struct hts_tpool *pool, int qsize);
```
jmarshall, line 367: It currently gets it as a byproduct of _hts.h_, but it would probably be worth adding a `struct hts_tpool;` declaration above (beside the other struct declarations).

jkbonfield: Easy enough to do. Thanks

PR #1646, opened 2023-07-13
Fix to 2e672f33 decompress_peek_gz change.

jkbonfield:
The "zs.total_out < destsize" should have been "zs.avail_out" to be more robust to total_out being reset by inflateReset.  However looking again neither the avail_in or avail_out checks are necessary, as once we hit the end of either input or output buffer the next cycle triggers ret == Z_BUF_ERROR and we drop out as normal.

Thanks to John Marshall for the spot.

See #1643

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -326,21 +326,21 @@ decompress_peek_gz(hFILE *fp, unsigned char *dest, size_t destsize)
 
     int ret;
     unsigned char *last_in = buffer;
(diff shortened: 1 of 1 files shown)


daviesrob:
Isn't the same problem still present in the main `while` loop?
I.e.:
```
    while (zs.total_out < destsize) {
```
would be better done as:
```
    while (zs.avail_out > 0) {
```

PR #1647, opened 2023-07-13
Ensure number of modifications is always set in bam_parse_basemod2()

daviesrob:
Set `state->nmods` to 0 at the start, in case the function returns early.  This ensures calls to `bam_mods_at_next_pos()` won't try to replay the mods from an earlier line.


jkbonfield:
Thanks.  Kind of suprising we haven't noticed this before.

PR #1649, opened 2023-07-14
NEWS updates for pending release

daviesrob:
All the NEWS so far...

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,170 @@
 Noteworthy changes in release a.b
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Updates
+-------
+
+* Using CRAM 3.1 no longer gives a warning about the specification
+  being draft.  Note CRAM 3.0 is still the default output format.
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -1,6 +1,160 @@
...
+Bug fixes
+---------
+
+* Fix a major CRAM index querying bug where one container has start
+  and end coordinates entirely contained within a previous container.
+  This would occasionally miss data, and sometimes return much more
+  than required.  The bug affected versions 1.11 to 1.17, although the
+  change in 1.11 was fixing multi-threaded index queries.
+  (PR#1574, PR#1640. Fixes #1569, #1639, samtools/samtools#1808,
+  samtools/samtools#1819.  Reported by xuxif, Jens Reeder and Jared Simpson)
```
jmarshall, line 90: This uses the word “querying” which is probably a clue, but it would be good to be explicit as to whether CRAM indexes need to be regenerated to resolve this issue.

It would also be good to mention this major bug in SAMtools's _NEWS_ as well.

PR #1653, opened 2023-07-19
Add missing Makefile dependencies and other minor fixes

jmarshall:
* A few missed _Makefile_ changes from recent PRs.

* Add `const` to avoid a warning when zlib is configured (via `-DZLIB_CONST`) to be more const-correct than is the default:

```
hts.c:335:25: warning: assigning to 'unsigned char *' from 'const Bytef *' (aka 'const unsigned char *') discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
                last_in = zs.next_in;
```


jkbonfield:
Thanks for this and the Samtools patch John.

Issue #1642, opened 2023-07-10
Request to support multiple input files.  Currently, bgzip <file1> <file2> only compresses <file1>. 


bw2:
bgzip's command-line interface is similar to gzip, so I was surprised to see that 

```
bgzip file1   file2   file3  ...
```
only compressed the first file1 (while gzip would compress each of the files).

I think `bgzip` would be more user-friendly if it behaved like gzip in this case, or at least threw an error to say that the extra files were unrecognized arguments instead of silently ignoring them.

jmarshall:
See also #391.

daviesrob:
Yes, it's about time we made this work.  We'll look into it.

PR #1658, opened 2023-08-02
changes to support multiple files

vasudeva8:
This PR contains the changes for #1642.

Multiple files are supported and given operation is performed on all inputs, except with rebgzip.
Returns failure when multiple inputs are given with -I option.
Returns failure with rebgzip and index used together.
-b / -s options, which requires index file, work with single input when index is explicitly providing using -I.
-b / -s can work with multiple inputs if implicit indexes, i.e. \<compressed filename.gzi\>, are in use (i.e. no -I option
and all having their implicit index file along with it).

Corrected the error message with rebgzip invocation without -I option.

hfile.c/h, bgzf.c updated to add new methods which doesn't close the stdin/out, to support multi file processing.

### DIFF
--- a/bgzip.1
+++ b/bgzip.1
@@ -4,7 +4,7 @@
 bgzip \- Block compression/decompression utility
 .\"
 .\" Copyright (C) 2009-2011 Broad Institute.
-.\" Copyright (C) 2018, 2021-2022 Genome Research Limited.
+.\" Copyright (C) 2018, 2021-2023 Genome Research Limited.
 .\"
 .\" Author: Heng Li <lh3@sanger.ac.uk>
 .\"
@@ -54,7 +54,7 @@ bgzip \- Block compression/decompression utility
 .IR size ]
 .RB [ -@
 .IR threads ]
-.RI [ file ]
+.RI [ file " ...]"
 .PP
 .SH DESCRIPTION
 .PP
@@ -70,7 +70,8 @@ If the -c option is used, the result will be written to standard output,
 otherwise when compressing bgzip will write to a new file with a .gz
 suffix and remove the original.  When decompressing the input file must
 have a .gz suffix, which will be removed to make the output name.  Again
-after decompression completes the input file will be removed.
+after decompression completes the input file will be removed. When multiple
+files are given as input, the operation is performed on all of them.
 
 .SH OPTIONS
 .TP 10
--- a/bgzip.c
+++ b/bgzip.c
@@ -1,7 +1,7 @@
 /* bgzip.c -- Block compression/decompression utility.
 
    Copyright (C) 2008, 2009 Broad Institute / Massachusetts Institute of Technology
-   Copyright (C) 2010, 2013-2019, 2021-2022 Genome Research Ltd.
+   Copyright (C) 2010, 2013-2019, 2021-2023 Genome Research Ltd.
 
    Permission is hereby granted, free of charge, to any person obtaining a copy
    of this software and associated documentation files (the "Software"), to deal
@@ -134,7 +134,7 @@ int main(int argc, char **argv)
     char *buffer;
     long start, end, size;
     char *index_fname = NULL;
-    int threads = 1;
+    int threads = 1, isstdin = 0, usedstdout = 0, ret = 0;
 
     static const struct option loptions[] =
     {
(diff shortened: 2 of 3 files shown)


hfile.c
```diff
@@ -527,6 +526,49 @@ typedef struct {
     unsigned is_socket:1;
 } hFILE_fd;
 
+/// hclose2 - closes the file using hclose when it is not stdin or stdout
+/** @param fp pointer to hFile
+ * returns 0 on success, -1 on failure
+*/
+int hclose2(hFILE *fp)
+{
+    hFILE_fd *hp = (hFILE_fd*)fp;
```
jmarshall, line 535: This will absolutely not work. A public API function taking an `hFILE` may be called with any kind of hFILE (and will be called thus given e.g. `bgzip -dc s3://foo/bar.gz`), not just a local file that is represented as an `hFILE_fd`. This cast is invalid and in general there is no way to tell what flavour of `hFILE` you are being given, and most of them won't return anything meaningful for `hp->fd`.

Instead you will need to track or test in _bgzip.c_ whether the input/output file is stdin/out and avoid calling `bgzf_close`/`hclose` appropriately. Then you won't need to touch _bgzf.c_ or _hfile.[ch]_ at all.

daviesrob: Yes, I think the solution for `stdout` would be to open it if needed outside the main processing loop,  and pass it into `operate()`.  Then you can keep writing to the same file handle.

For `stdin`, it's arguable that it shouldn't be allowed when specifying file names unless you're using `-c` as it would write to `-.gz` which is probably not what was intended.  (Although testing the existing `bgzip` it does do that...)

jmarshall: It is admittedly annoying when you accidentally close your only connection to stdin/out. 😄 

However IMHO it is more important for `hclose()` (and `fd_close()`) to reliably actually close things.

In particular, there are some I/O errors that are not reported until you call `close(2)`. So to fully check for errors in `some-htslib-using-command args > output.bam` you need to check the return value of `hclose()` and it needs to reflect the return value of `close(2)` even for stdout.

jkbonfield: Just to avoid confusion, I assume you mean `close(2)` as in the system call `close` in section 2 of the man pages, rather than fd 2 (traditionally stderr).    Agreed though, we want to close stdout and if that produces an error report it to stderr via the usual channels.

I think this PR shouldn't need any htslib API changes.

When operating on a single file, it worked as before barring maybe some idioscyncracies like -.gz that ought to be tidied - gzip input stdin means output to stdout (but that's really an orthogonal issue).  When operating on multiple files it's really just a loop over the arguments with existing semantics assuming command line options are compatible.  Generally that means multiple files, not stdin/stdout.  The only slight fly in the ointment is attempt to specify "-" or "/dev/stdout" of one of a series of files.  I think that's such a weird corner case we probably don't want to let the tail wag the dog and just accept we may have closed stdin/stdout early and live with it. 

Curiously, this actually works though:

```
$ gzip - - > _.gz
foo
^D
bar
^D
$ zcat _.gz
foo
bar
```

That surprised me.  It's such an esoteric case though I really don't think it's worth trying to handle it.

jmarshall: I am sure @jkbonfield is well aware that I'm old school enough that I'm using `manpage(section)` notation here despite it being annoyingly ambiguous for a function with one integer argument. :smile:

Seems to me (not having thought about it enormously much) one of the few use cases that would require keeping stdin or stdout open that's actually useful is `bgzip -dc foo.gz bar.gz baz.gz`.

jkbonfield: Obviously - it was more for @vasudeva8's benefit who may not be aware of the long-standarding unix notation.

Good point on the `bgzip -dc foo.gz bar.gz baz.gz` case.  Not closing stdout is far more important here than not closing stdin, which is a very esoteric terminal-only situation (I think).  This still doesn't need API changes though. We can just delay the htsclose of stdout until the last file has been written.

jkbonfield:
This feels overly complicated, but some of the changes are also down to improved error detection and identification of clashing options, such as region querying or rebgzip.

However the bulk of the functionality could be as simple as a do while loop on optind:

```
diff --git a/bgzip.c b/bgzip.c
index 589f79f6..fe6eacb0 100644
--- a/bgzip.c
+++ b/bgzip.c
@@ -188,6 +188,7 @@ int main(int argc, char **argv)
         fprintf(stderr, "[bgzip] Illegal region: [%ld, %ld]\n", start, end);
         return 1;
     }
+    do {
     if (compress == 1) {
         hFILE* f_src = NULL;
         char out_mode[3] = "w\0";
@@ -369,7 +370,6 @@ int main(int argc, char **argv)
             error("Input close failed\n");
         if (argc > optind && !pstdout && !keep) unlink(argv[optind]);
         free(buffer);
-        return 0;
     }
     else if ( reindex )
     {
@@ -401,7 +401,6 @@ int main(int argc, char **argv)
         }
 
         if ( bgzf_close(fp)<0 ) error("Close failed: Error %d\n",fp->errcode);
-        return 0;
     }
     else
     {
@@ -494,6 +493,7 @@ int main(int argc, char **argv)
 #ifdef _WIN32
         _setmode(f_dst, O_BINARY);
 #endif
+        long start_reg = start, end_reg = end;
         while (1) {
             if (end < 0) c = bgzf_read(fp, buffer, WINDOW_SIZE);
             else c = bgzf_read(fp, buffer, (end - start > WINDOW_SIZE)? WINDOW_SIZE:(end - start));
@@ -508,9 +508,12 @@ int main(int argc, char **argv)
             }
             if (end >= 0 && start >= end) break;
         }
+        start = start_reg;
+        end = end_reg;
         free(buffer);
         if (bgzf_close(fp) < 0) error("Close failed: Error %d\n",fp->errcode);
         if (argc > optind && !pstdout && !test && !keep) unlink(argv[optind]);
-        return 0;
     }
+    } while (++optind < argc);
+    return 0;
 }
```

Obviously that needs better handling of the rebgzip case and it's probably better to have a pair of do-while loops handling compress and decompress separately, but I'm not entirely sure the large scale changes are justified by this feature request.  That said, some code restructuring isn't a bad idea given the length of main.  It just makes it a little trickier to work out what's changed and whether all of the functionality has been kept.

Specifically the stdin/stdout code feels inherently wrong and I don't entirely understand why it's not just dealt with automatically as any other filename.  (We can add an implicit "-" in the case of no args and stdin isn't a tty.)    Opening it in main and having custom code for it feels a bit strange and very tricky to follow.  What am I missing here?

Is it the `./bgzip -cd _a.gz _b.gz` case where we wish to output two files to the same stream (which my naive do-while didn't handle I expect)?  If so we can probably resolve this by only closing stdout hts streams at the end of the program, so we can concatenate outputs.

jmarshall:
> Specifically the stdin/stdout code feels inherently wrong and I don't entirely understand why it's not just dealt with automatically as any other filename.

`hopen_fd_stdinout()` just creates an `hFILE_fd` with an `fd` of 0 or 1. Hence `in = hopen("-", "r") … hclose(in) … in2 = hopen("-", "r")` would fail because fd 0 is no longer open or no longer refers to the original stdin. This could be complexified by having it use `dup()` instead of literally STDIN_FILENO (0) or STDOUT_FILENO (1) so `hclose()` wouldn't close the only reference to the original stdin/stdout stream, but really this has caused no significant problems in all the years since it was implemented so far — largely because reading two BAM files from stdin or writing two BAM files to stdout just isn't a useful operation.

So if there are useful bgzip operations involving reading multiple streams from stdin (probably not) or writing multiple streams to stdout (maybe), they are currently going to have to do something special for stdin/stdout. Even if that is only not actually calling `hclose()` except the last time.

The code is now more complicated than I was expecting too. Perhaps all the casting could be avoided by separating `operate()` into separate functions for the different modes?

hfile.c
vasudeva8: updated based on comments

vasudeva8:
Updated to avoid the complexities, uses the do..while framework and uses dup to open/close stdin/out as required.
-I is supported with multiple inputs as well - as we do rebgzip!

jmarshall:
> Specifically the stdin/stdout code feels inherently wrong and I don't entirely understand why it's not just dealt with automatically as any other filename.

Upon further reflection, I think stdin/stdout should just be dealt with by `hts_open`/`hts_close` automatically just as any other filename, and @vasudeva8 should be able to repeatedly `hts_open` and `hts_close` stdin and stdout just as for any other filename.

PR #1665 modifies `hopen_fd_stdinout()` to enable that. If that approach is accepted by the maintainers, this PR can be rebased onto that once merged, and all the `dup()` gymnastics can be removed — which should enable some simplification of the code.

vasudeva8:
Updated and uses new shared stdin/out hfile. Uses the do/while as earlier and explicitly closes stdout at the end.
Moved a few validations to ensure they are made before output file update.

jkbonfield (commented):
It seems to work bar one corner case.

Create two multi-block bgzip files:

```
head -10000 /usr/share/dict/words > _w1                                                             
tail -10000 /usr/share/dict/words > _w2                                                             
./bgzip -i -I _w1.gz.gzi -c _w1 > _w1.gz
./bgzip -i -I _w2.gz.gzi -c _w2 > _w2.gz
md5sum _w[12].gz.gzi
2f17308a50726f862258407527d50dad  _w1.gz.gzi
c47ffdd4c264f34200370048ae62ebbc  _w2.gz.gzi
```

Compare this to compression both together.

```
./bgzip -i -I _w12.gz.gzi -c _w1 _w2 > _w12.gz                                                      
md5sum _w12.gz.gzi                                                                                  
c47ffdd4c264f34200370048ae62ebbc  _w12.gz.gzi                                                       
```

The compression worked, with `bgzip _w12.gz` matching `cat _w1 _w2`, however the index file _w12.gz.gzi is the same as the index for the 2nd file. 

Note we don't need `-c` and `> file.gz` for this to fail.  Just compressing a whole bunch of files when using a named index (`-I`) will also fail with the same logic.  This is just nonsensical.  Technoically it's done the right thing - build an index for each file we asked for and write it to the named file, overwriting the previous index.  That's a case of asking for something stupid gets you something stupid.

However the file concatenation to stdout where this could actually be used.  As the concatenation works and we could then do `bgzip -r _w12.gz` to index it.  Fixing it though is hard as our main loop opens and closes the output file each time, in this case stdout.  We could get it to not close when using stdout, so we only have one file and the index then works, but it would make the code more complex and it also changes functionality - we're not getting the same output as concatenating _w1.gz _w2.gz together.

My personal preference here is the simplest solution which is simply to spot poor usage and complain.  If someone later comes along with a clear use case showing this behaviour must work, then we can reconsider.

Ie:

```
diff --git a/bgzip.c b/bgzip.c
index 829ebe8a..4b85272e 100644
--- a/bgzip.c
+++ b/bgzip.c
@@ -199,6 +199,10 @@ int main(int argc, char **argv)
         fprintf(stderr, "[bgzip] Index file name expected with rebgzip.  See -I option.\n");
         return 1;
     }
+    if (index_fname && argc - optind > 1) {
+        fprintf(stderr, "[bgzip] Cannot specify index filename for more than one data file.\n");
+        return 1;
+    }
 
     do {
         isstdin = optind >= argc ? 1 : !strcmp("-", argv[optind]);          //using stdin or not?
```

Other than that it passes my tests, and actually fixes a whole bunch of long-standanding bgzip bugs.  Eg a lot of the checks for stdin weren't working if the user specified stdin as "-" (giving us "-.gz.gzi" filenames for example).

jkbonfield:
Also, this sounded on paper like a trivial issue to resolve, but the myriad of interaction between all the command line arguments shows it's far trickier than I expected!  Good going on navigating the minefield, especially fixing the long-standing issues caused by said minefield. :)

bgzip.c
```diff
@@ -188,329 +188,374 @@ int main(int argc, char **argv)
...
+            if (compress_level >= 0) {
+                out_mode[1] = compress_level + '0';
+                out_mode_exclusive[2] = compress_level + '0';
+            }
+            if (!(f_src = hopen(!isstdin ? argv[optind] : "-", "r"))) {
+                fprintf(stderr, "[bgzip] %s: %s\n", strerror(errno), isstdin ? "stdin" : argv[optind]);
+                return 1;
+            }
+
+            if ( argc>optind && !isstdin )
```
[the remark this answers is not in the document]
vasudeva8: done.

bgzip.1
```diff
@@ -54,7 +54,9 @@ bgzip \- Block compression/decompression utility
 .IR size ]
 .RB [ -@
 .IR threads ]
-.RI [ file ]
+.RI [ file1 ]
+.RI [ file2 ]
+.RI [ ... ]
```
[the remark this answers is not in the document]
vasudeva8: done

vasudeva8:
Updated to avoid explicit index file with multiple input files, during index and reindex.
These options are not considered during decompress/read operation.

PR #1663, opened 2023-08-18
Vcf write speed

jkbonfield:
Speeds up the VCF writer.  The main changes are:

- VCF writer can handle packed BCF objects, meaning we don't have to unpack before writing as VCF.
- Improve the GT formatting loop so once we've found it (often early on) we don't continue the penalty of looking again.  This assumes it only ever turns up once, but the VCF specification forbids duplicate keys.
- Optimise `bcf_fmt_array` on character arrays, treating them only as strings.  This is a bit vague in the spec, but I believe this code to be valid.
- Cache header key lengths to avoid repeated calls to strlen everywhere (`kputsn` over `kputs`).

Benchmarks showing 3 trials of "perf stat" cycle counts, in millions to make it easier to read, for develop (1st) vs this PR (2nd) on a variety of data sets.  The third number in parentheses is the number of cycles spent during the read portion of test_view, so for the true write speed up that could be subtracted from both the Dev vs PR values.  Host was an old Intel box, using the system gcc (7).

```
=== bcftools ===                                                                
  24,756   22,725    (3,992)     M.cycles  #    2.983 GHz                       
  25,892   22,445    (3,961)     M.cycles  #    2.983 GHz                       
  25,039   22,741    (4,037)     M.cycles  #    2.983 GHz                       
=== freebayes ===                                                               
  75,838   59,067   (11,068)     M.cycles  #    2.982 GHz                       
  75,342   59,437   (10,359)     M.cycles  #    2.983 GHz                       
  78,406   60,511   (10,309)     M.cycles  #    2.983 GHz                       
=== gatk ===                                                                    
   2,823    2,703      (387)     M.cycles  #    2.983 GHz                       
   2,680    2,248      (387)     M.cycles  #    2.983 GHz                       
   2,662    2,245      (386)     M.cycles  #    2.981 GHz                       
=== info ===                                                                    
  12,555    7,693    (1,724)     M.cycles  #    2.983 GHz                       
  12,562    7,947    (1,724)     M.cycles  #    2.983 GHz                       
  12,365    7,970    (1,734)     M.cycles  #    2.983 GHz                       
=== fmt ===                                                                     
  25,125   22,917    (3,395)     M.cycles  #    2.977 GHz                       
  25,074   22,604    (3,421)     M.cycles  #    2.981 GHz                       
  25,228   22,450    (3,408)     M.cycles  #    2.967 GHz                       
=== GIAB ===                                                                    
  43,056   26,989    (7,488)     M.cycles  #    2.972 GHz                       
  43,959   27,237    (7,409)     M.cycles  #    2.975 GHz                       
  47,072   28,113    (7,475)     M.cycles  #    2.979 GHz                       

```

"GIAB" is the GIAB HG002 truth set. "info" and "fmt" are INFO heavy and FORMAT 1000 genomes heavy files from GNOMAD.  Bcftools, Freebayes and GATK are calls made by the repsective tools on SynDip (CHM1/CHM13), so represent real-world single-sample tool outputs.

My test timings were `./test/test_view /tmp/_$i.bcf -p /tmp/_.vcf` where the BCF file is an uncompressed BCF, so causing the minimum of read/decode time.  The read-portion in the above chart came from `test_view -B /tmp/_$i.bcf`.

Most files are 10-20% faster at writing, with the GIAB data being 40% faster for some reason (probably due to being dominated by very long strings in INFO).  All files were md5sumed to validate the output matched.

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -2722,11 +2722,22 @@ int bcf_fmt_array(kstring_t *s, int n, int type, void *data)
 
     if (type == BCF_BT_CHAR)
     {
-        char *p = (char*)data;
-        for (j = 0; j < n && *p; ++j, ++p)
-        {
-            if ( *p==bcf_str_missing ) e |= kputc_('.', s) < 0;
-            else e |= kputc_(*p, s) < 0;
+        char *p = (char *)data;
+
+        // Can bcf_str_missing only occur at the start of a CHAR array?
+        // We use these as strings only I believe, and INT8 for numerics.
+        // Checking further however, I'm not sure bcf_str_missing can occur
+        // even there.  I think it's likely an unimplemented feature, whose
+        // only existance in the specification is a single line mention in
+        // an example (for the "ID" field).
+        if (n >= 8 && *p != bcf_str_missing) {
+            char *p_end = memchr(p, 0, n);
+            e |= kputsn(p, p_end ? p_end-p : n, s) < 0;
+        } else {
+            for (j = 0; j < n && *p; ++j, ++p) {
+                if ( *p==bcf_str_missing ) e |= kputc_('.', s) < 0;
+                else e |= kputc_(*p, s) < 0;
+            }
         }
     }
     else
@@ -2738,7 +2749,7 @@ int bcf_fmt_array(kstring_t *s, int n, int type, void *data)
                 type_t v = convert(p); \
                 if ( is_vector_end ) break; \
                 if ( j ) e |= kputc_(',', s) < 0; \
-                e |= (is_missing ? kputc('.', s) : kprint) < 0; \
+                e |= (is_missing ? kputc_('.', s) : kprint) < 0; \
             } \
         }
         switch (type) {
@@ -3846,7 +3857,7 @@ int bcf_unpack(bcf1_t *b, int which)
         ptr_ori = ptr;
         ptr = bcf_fmt_sized_array(&tmp, ptr);
         b->unpack_size[0] = ptr - ptr_ori;
-        kputc('\0', &tmp);
+        kputc_('\0', &tmp);
         d->id = tmp.s; d->m_id = tmp.m;
 
         // REF and ALT are in a single block (d->als) and d->alleles are pointers into this block
@@ -3857,7 +3868,7 @@ int bcf_unpack(bcf1_t *b, int which)
             // Use offset within tmp.s as realloc may change pointer
             d->allele[i] = (char *)(intptr_t)tmp.l;
             ptr = bcf_fmt_sized_array(&tmp, ptr);
-            kputc('\0', &tmp);
+            kputc_('\0', &tmp);
         }
         b->unpack_size[1] = ptr - ptr_ori;
         d->als = tmp.s; d->m_als = tmp.m;
@@ -3911,7 +3922,7 @@ int vcf_format(const bcf_hdr_t *h, const bcf1_t *v, kstring_t *s)
         return -1;
     }
 
-    bcf_unpack((bcf1_t*)v, BCF_UN_ALL);
+    bcf_unpack((bcf1_t*)v, BCF_UN_ALL & ~(BCF_UN_INFO|BCF_UN_FMT));
 
     // Cache of key lengths so we don't keep repeatedly using them.
     // This assumes we're not modifying the header between successive calls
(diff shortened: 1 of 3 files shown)


jkbonfield:
Checking it again I just spotted another slow down.  I obviously hadn't finished my repeated rounds of optimisation and forgot where I was at, but this is still an improvement.  I have a tiny new commit to add later though.

jkbonfield:
Latest benchmarks, in cycle counts for the same files

```
=== bcftools ===                                                                
       19743638639      cycles                    #    2.983 GHz                
       19489797455      cycles                    #    2.973 GHz                
       19395872945      cycles                    #    2.971 GHz                
300f6c790bc5bc725005a2a36f8d67fd  -                                             
=== freebayes ===                                                               
       49194527204      cycles                    #    2.983 GHz                
       50985255922      cycles                    #    2.982 GHz                
       51973400318      cycles                    #    2.984 GHz                
3ab21fcda2287e0cb012c99337010be3  -                                             
=== gatk ===                                                                    
        2401848476      cycles                    #    2.931 GHz                
        1927701416      cycles                    #    2.915 GHz                
        1924751331      cycles                    #    2.937 GHz                
968b6446197e4d132ae382acc24cee71  -                                             
=== info ===                                                                    
        7494270098      cycles                    #    2.984 GHz                
        7702558176      cycles                    #    2.983 GHz                
        7703784698      cycles                    #    2.969 GHz                
690a78c76697c04eeb9faf564a8c78d6  -                                             
=== fmt ===                                                                     
       22428582248      cycles                    #    2.982 GHz                
       22596287183      cycles                    #    2.976 GHz                
       22498706702      cycles                    #    2.984 GHz                
898432fcce471fbfa89065b6befcd459  -                                             
=== GIAB ===                                                                    
       24562360470      cycles                    #    2.977 GHz                
       25045186584      cycles                    #    2.983 GHz                
       25870306585      cycles                    #    2.980 GHz                
de4cb618fbe55bad8268a3e510a28384  -                                             
```

jkbonfield:
I'll invesitgate the test failures on Monday.  Bizarrely it's some index difference - I assume because the VCF is somehow different?  All my VCFs I produced matched perfectly, and the locally run test harness also works fine.

Anyway, I can check which commit it is easily by reverting and pushing over this to validate.  Would be easier if I could reproduce it locally though so I'll try other systems to start with.  Most perplexing.

jkbonfield:
Note: the cause of the test failure was using `builtin_clz` to check the binary size of the double, after conversion to integer.  This function is undefined when the value is zero, meaning `d < 1`.  A simple +1 would have sufficed, but benchmarking it again I only see a couple percent difference so decided it's sufficiently minimal gains to not be worth while, so culled that part of code.

jkbonfield:
I've rebased and squashed it back down again.  A few small additional tweaks to the vcf_format code, but the main change was more major improvements to kputd.  The final loops were culled as we already know the end point (within 1 at least, due to the slight rounding problem of eg 9999.999 to 10000).

Benchmarks on `sprintf(buf, "%g", d)`, `kputd` in devel and `kputd` in this PR for values generated by `for (d = 0.0001; d <= 999999; d *= 1.0000001)` are, in billions of CPU cycles:

sprintf: 252.2
kputd (devel): 37.2
kputd (this): 16.0

sprintg

jkbonfield:
Added `rint` as suggested by Rob.  A note to others and self, values like 1.015625 were rounded up instead of down.  This was something the original got wrong too.

Also adjusted the layout slightly, so it's more a thing of beauty. :)

vcf.c
```diff
@@ -2626,20 +2637,91 @@ int bcf_enc_vchar(kstring_t *s, int l, const char *a)
...
     if (type == BCF_BT_CHAR)
     {
-        char *p = (char*)data;
-        for (j = 0; j < n && *p; ++j, ++p)
-        {
-            if ( *p==bcf_str_missing ) e |= kputc('.', s) < 0;
-            else e |= kputc(*p, s) < 0;
+        char *p = (char *)data;
+
+        // Can bcf_str_missing only occur at the start of a CHAR array?
```
daviesrob, line 2711: Looking at the spec, I think `bcf_str_missing` (0x07) comes from the [type encoding](https://samtools.github.io/hts-specs/VCFv4.4.pdf#subsubsection.6.3.3) of an empty string (i.e. type 7, length 0).  So the `bcf_str_missing` value will actually occur before the start of the string, and will be handled by the `n == 0` case above.

I therefore suspect the original version of this code was wrong, and note that the [equivalent in htsjdk](https://github.com/samtools/htsjdk/blob/38dbd0e6b87a2bd5791bb81399f2b5ec1008773e/src/main/java/htsjdk/variant/bcf2/BCF2Decoder.java#L179) doesn't go looking for 0x07 characters in the actual string.

vcf.c
```diff
@@ -2626,20 +2637,91 @@ int bcf_enc_vchar(kstring_t *s, int l, const char *a)
...
+    return e == 0 ? 0 : -1;
+}
+
 int bcf_fmt_array(kstring_t *s, int n, int type, void *data)
 {
     int j = 0;
     uint32_t e = 0;
     if (n == 0) {
-        return kputc('.', s) >= 0 ? 0 : -1;
+        return kputc_('.', s) >= 0 ? 0 : -1;
```
daviesrob, line 2704: As this function is part of the API, this should remain as `kputc()` so the caller gets a properly terminated string.

vcf.c
```diff
@@ -2650,9 +2732,8 @@ int bcf_fmt_array(kstring_t *s, int n, int type, void *data)
             { \
                 type_t v = convert(p); \
                 if ( is_vector_end ) break; \
-                if ( j ) kputc(',', s); \
-                if ( is_missing ) kputc('.', s); \
-                else e |= kprint < 0; \
+                if ( j ) e |= kputc_(',', s) < 0; \
+                e |= (is_missing ? kputc_('.', s) : kprint) < 0; \
```
daviesrob, line 2736: And we probably want `kputc()` here too, unless we ensure the kstring is properly terminated before the function returns.

vcf.c
```diff
@@ -2626,20 +2637,84 @@ int bcf_enc_vchar(kstring_t *s, int l, const char *a)
...
-        for (j = 0; j < n && *p; ++j, ++p)
-        {
-            if ( *p==bcf_str_missing ) e |= kputc('.', s) < 0;
-            else e |= kputc(*p, s) < 0;
+        char *p = (char *)data;
+
+        // Note bcf_str_missing is already accounted for in n==0 above.
+        if (n >= 8) {
+            char *p_end = memchr(p, 0, n);
+            kputsn(p, p_end ? p_end-p : n, s);
```
daviesrob, line 2714:
```suggestion
            e |= kputsn(p, p_end ? p_end-p : n, s) < 0;
```

daviesrob:
Squashed and rebased...

PR #1674, opened 2023-09-21
Apply seqtk PR to improve kseq.h parsing performance

jmarshall:
Port of @kloetzl's lh3/seqtk#123 and attractivechaos/klib#173 to HTSlib's copy of _kseq.h_.

Per the original PR: “Doubles throughput; can now parse FASTA at 2GB/s”.

For one large data file on my machine, this reduces klib's _kseq_bench2_'s `kstream` time from ~1.65s to ~0.92s.

(I have not heavily tested it otherwise.)

As [noted here](https://twitter.com/jomarnz/status/1661586585769611270), HTSlib doesn't use `kseq` itself though a few samtools commands do use it to read a few small BED files etc.

### DIFF
--- a/htslib/kseq.h
+++ b/htslib/kseq.h
@@ -110,8 +110,8 @@
 				} else break; \
 			} \
 			if (delimiter == KS_SEP_LINE) {  \
-				for (i = ks->begin; i < ks->end; ++i)  \
-					if (ks->buf[i] == '\n') break; \
+				unsigned char *sep = memchr(ks->buf + ks->begin, '\n', ks->end - ks->begin); \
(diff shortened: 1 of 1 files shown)


jkbonfield:
Thanks both.  A nice simple but effective change.

Sad face over kstream using int everywhere, giving a 2GB limit on fasta sequence length, but that's a totally different PR. :/

kloetzl:
This is missing a cast to support C++ with its stricter typing rules. https://github.com/lh3/seqtk/commit/7fe58c806c80b6634e96ab752ffdcb286e44c0bc

Issue #1625, opened 2023-05-26
Calling tabix on S3 files via jupyter notebook fails


agilly:
I am experiencing a very peculiar issue. I am using jupyter notebook to test some R code that calls tabix to query some files stored in S3 buckets.

The following command is an example of what I would be running on the bash command line:

```bash
AWS_PROFILE=tabix tabix s3://path/to.tsv.gz 2:1234-1234
```
The above command works. Now wrapping it in R (4.1.2):
```R
out=system("AWS_PROFILE=tabix tabix s3://path/to.tsv.gz 2:1234-1234", intern=T)
```
This also works.

But now, using the exact same code in a jupyter notebook running IRKernel crashes. The log of the kernel reports the following:

```
[E::bgzf_read_block] Invalid BGZF header at offset 356561
Reading "s3://path/to.tsv.gz" failed: Resource temporarily unavailable
```

I am querying 100s of files but this happens only for a handful of them, systematically, for seemingly any region. I tried reindexing the file, the bug still occurs. I tried with htslib 1.12 and the latest 1.17, same thing.

This is probably going to be difficult to debug because there are many moving parts: tabix, R, and jupyter. Just checking if you have an idea why specifically calling something from jupyter would cause the error above? I cannot send the tsv files as they are confidential (and very large) but I am happy to do more exploration/debugging if you tell me what to do.

may be related to https://github.com/samtools/htslib/issues/1037

TIA!

daviesrob:
The "Resource temporarily unavailable" message comes from `errno` getting set to `EAGAIN` or `EWOULDBLOCK`, which indicates that a system call was interrupted.  This may be harmless, if the call was restarted, or it may indicate that something isn't handling such interruptions correctly.

You could try running `tabix --verbosity=8` to increase the amount of debugging output printed (be warned this could be quite a lot).  If we're lucky, that might give some clues as to what's going wrong.

Otherwise, if I were debugging this I'd attempt to find out what's going on at the point where `bgzf_read_block` reports its failure.  As you're running lots of `tabix` processes, this might involve adding something before [this `return` statement](https://github.com/samtools/htslib/blob/878cff4a443fe5b3b49aae61953de1bbddeddc8a/bgzf.c#L1147) to either print more information about what's going on, or allow a debugger to be attached (either by dumping a core file with `abort()` and inspecting the results, or making it sleep for long enough for a debugger to be manually attached using `gdb --pid` assuming this is possible).

daviesrob:
[...]

PR #1676, opened 2023-09-25
Fix hfile_libcurl small seek bug

daviesrob:
Ensure that `to_skip` is updated correctly when handling a small seek by skipping data in libcurl_read().  In particular, if the number of bytes returned by wait_perform() exactly matches the number left to skip, `to_skip` needs to be set to zero, and wait_perform() needs to be called again to get some data for the function to return.

Fixes samtools/samtools#1918; possibly also #1037, #1625 and samtools/samtools#1622

@jmarshall I've made you author of ths commit, as it was your patch.  Let me know if you'd like to change anything before it gets merged.

### DIFF
--- a/hfile_libcurl.c
+++ b/hfile_libcurl.c
@@ -837,7 +837,7 @@ static ssize_t libcurl_read(hFILE *fpv, void *bufferv, size_t nbytes)
         got = fp->buffer.ptr.rd - buffer;
 
         if (to_skip >= 0) { // Skipping over a small seek
-            if (got < to_skip) { // Need to skip more data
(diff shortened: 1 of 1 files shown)


jmarshall:
The commit and the explanation in the commit message look good.

IMHO you should add a `Co-Authored-By: RMD` — I just tried a `<=` tweak and verified that it fixed Alex's problem, but didn't look into how `got` and `to_skip` work. You've added the understanding and explanation that this is indeed the right fix in general.

daviesrob:
Cheers.  I've updated the commit message as suggested.

PR #1677, opened 2023-09-26
Faster nibble parsing and a SSSE3 routine for even faster nibble parsing

rhpvorderman:
Nibble parsing was written as a for loop. This leads to an extra loop control variable being used. By using pointer arithmetic, the loop control variable can be parsed out. This leads to faster code.

I also wrote a routine to utilise the pshufb instruction in order to get even faster translation.

Speed before:
```
Benchmark 1: ./test/bin/htsfile ~/test/nanopore_100000reads.u.bam  --view > /dev/null
  Time (mean ± σ):     893.6 ms ±   9.7 ms    [User: 770.5 ms, System: 122.9 ms]
  Range (min … max):   878.6 ms … 913.4 ms    10 runs
```

Speed after:
```
Benchmark 1: ./test/bin/htsfile ~/test/nanopore_100000reads.u.bam  --view > /dev/null
  Time (mean ± σ):     694.4 ms ±   6.7 ms    [User: 583.3 ms, System: 110.6 ms]
  Range (min … max):   682.1 ms … 703.0 ms    10 runs
```

Speed after, using SSSE3
```
Benchmark 1: ./test/bin/htsfile ~/test/nanopore_100000reads.u.bam  --view > /dev/null
  Time (mean ± σ):     524.7 ms ±  11.9 ms    [User: 400.3 ms, System: 124.3 ms]
  Range (min … max):   510.0 ms … 552.7 ms    10 runs
```

Both the non-ssse3 and ssse3 function deliver the same htsfile output for all the 100.000 reads tested. (Tested using a checksum.)

### DIFF
--- a/sam_internal.h
+++ b/sam_internal.h
@@ -26,7 +26,9 @@ DEALINGS IN THE SOFTWARE.  */
 #include <errno.h>
 #include <stdint.h>
 #include "htslib/sam.h"
-
+#ifdef __SSSE3__
+#include "tmmintrin.h"
+#endif
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -87,15 +89,71 @@ static inline void nibble2base(uint8_t *nib, char *seq, int len) {
         "B=BABCBMBGBRBSBVBTBWBYBHBKBDBBBN"
         "N=NANCNMNGNRNSNVNTNWNYNHNKNDNBNN";
 
-    int i, len2 = len/2;
     seq[0] = 0;
+    const char *seq_end_ptr = seq + len;
+    char *seq_cursor = seq;
+    const uint8_t *nibble_cursor = nib;
+    const char *seq_end_ptr_twoatatime = seq + (len & (~1ULL));
+    #ifdef __SSSE3__
+    const char *seq_vec_end_ptr = seq_end_ptr - (2 * sizeof(__m128i));
+    __m128i first_upper_shuffle = _mm_setr_epi8(
+        0, 0xff, 1, 0xff, 2, 0xff, 3, 0xff, 4, 0xff, 5, 0xff, 6, 0xff, 7, 0xff);
+    __m128i first_lower_shuffle = _mm_setr_epi8(
+        0xff, 0, 0xff, 1, 0xff, 2, 0xff, 3, 0xff, 4, 0xff, 5, 0xff, 6, 0xff, 7);
+    __m128i second_upper_shuffle = _mm_setr_epi8(
+        8, 0xff, 9, 0xff, 10, 0xff, 11, 0xff, 12, 0xff, 13, 0xff, 14, 0xff, 15, 0xff);
+    __m128i second_lower_shuffle = _mm_setr_epi8(
+        0xff, 8, 0xff, 9, 0xff, 10, 0xff, 11, 0xff, 12, 0xff, 13, 0xff, 14, 0xff, 15);
+    __m128i nuc_lookup_vec = _mm_lddqu_si128((__m128i *)seq_nt16_str);
+    /* Work on 16 encoded characters at the time resulting in 32 decoded characters
+       Examples are given for 8 encoded characters A until H to keep it readable.
+        Encoded stored as |AB|CD|EF|GH|
+        Shuffle into |AB|00|CD|00|EF|00|GH|00| and
+                     |00|AB|00|CD|00|EF|00|GH|
+        Shift upper to the right resulting into
+                     |0A|B0|0C|D0|0E|F0|0G|H0| and
+                     |00|AB|00|CD|00|EF|00|GH|
+        Merge with or resulting into (X stands for garbage)
+                     |0A|XB|0C|XD|0E|XF|0G|XH|
+        Bitwise and with 0b1111 leads to:
+                     |0A|0B|0C|0D|0E|0F|0G|0H|
+        We can use the resulting 4-bit integers as indexes for the shuffle of
+        the nucleotide lookup. */
+    while (seq_cursor < seq_vec_end_ptr) {
+        __m128i encoded = _mm_lddqu_si128((__m128i *)nibble_cursor);
 
-    for (i = 0; i < len2; i++)
+        __m128i first_upper = _mm_shuffle_epi8(encoded, first_upper_shuffle);
+        __m128i first_lower = _mm_shuffle_epi8(encoded, first_lower_shuffle);
+        __m128i shifted_first_upper = _mm_srli_epi64(first_upper, 4);
+        __m128i first_merged = _mm_or_si128(shifted_first_upper, first_lower);
+        __m128i first_indexes = _mm_and_si128(first_merged, _mm_set1_epi8(0b1111));
+        __m128i first_nucleotides = _mm_shuffle_epi8(nuc_lookup_vec, first_indexes);
+        _mm_storeu_si128((__m128i *)seq_cursor, first_nucleotides);
+
+        __m128i second_upper = _mm_shuffle_epi8(encoded, second_upper_shuffle);
+        __m128i second_lower = _mm_shuffle_epi8(encoded, second_lower_shuffle);
+        __m128i shifted_second_upper = _mm_srli_epi64(second_upper, 4);
+        __m128i second_merged = _mm_or_si128(shifted_second_upper, second_lower);
+        __m128i second_indexes = _mm_and_si128(second_merged, _mm_set1_epi8(0b1111));
+        __m128i second_nucleotides = _mm_shuffle_epi8(nuc_lookup_vec, second_indexes);
+        _mm_storeu_si128((__m128i *)(seq_cursor + 16), second_nucleotides);
+
+        nibble_cursor += sizeof(__m128i);
+        seq_cursor += 2 * sizeof(__m128i);
+    }
+    #endif
+    while (seq_cursor < seq_end_ptr_twoatatime) {
         // Note size_t cast helps gcc optimiser.
-        memcpy(&seq[i*2], &code2base[(size_t)nib[i]*2], 2);
-
-    if ((i *= 2) < len)
-        seq[i] = seq_nt16_str[bam_seqi(nib, i)];
+        memcpy(seq_cursor, code2base + ((size_t)*nibble_cursor * 2), 2);
+        seq_cursor += 2;
+        nibble_cursor += 1;
+    }
+    if (seq_cursor != seq_end_ptr) {
+        /* There is a single encoded nuc left */
+        uint8_t nibble_c = *nibble_cursor;
+        uint8_t upper_nuc_index = nibble_c >> 4;
+        seq_cursor[0] = seq_nt16_str[upper_nuc_index];
+    }
 }
 
 #ifdef __cplusplus


daviesrob:
This looks interesting, but it would need some form of dynamic dispatch.  While SSSE3 may be widespread now, we can't make any assumptions about the hardware in use.

rhpvorderman:
@daviesrob Sure. I have been meaning to practice with dynamic dispatch for sometime. At least it is also quite a lot faster without the SSSE3 routine. Would it be an idea to remove the SSSE3 code from this PR and do that later in a separate PR?

rhpvorderman:
Strangely enough I retested on a intel core i7 -6700 system and the result was slower than on my ryzen 5 5650 system. So it seems to differ per CPU strangely enough whether it is faster or slower using the while loop vs the for loop.

Here is the loop in assembly for this PR
```Assembly
.L3:
        movzx   esi, BYTE PTR [rcx]
        add     rax, 2
        add     rcx, 1
        movzx   esi, WORD PTR code2base.0[rsi+rsi]
        mov     WORD PTR [rax-2], si
        cmp     rax, r8
```
Develop branch
```
.L3:
        movzx   esi, BYTE PTR [rdi+rax]
        movzx   esi, WORD PTR code2base.0[rsi+rsi]
        mov     WORD PTR [rbp+0+rax*2], si
        add     rax, 1
        cmp     ecx, eax
        jg      .L3
 ```
 I can't see why one branch would be faster than the other.

The SSSE3 routine is always faster, but indeed that requires some extra work.

**Comment by jkbonfield** [comment]
I was testing on Intel and found it's also highly dependent on compiler, and probably version.

Original scalar code, for 100k ONT reads (NB this nibble function only, perf counts):  clan16 543, gcc13 413
Your pointer based scalar code: clang16 314, gcc13 553.
An unrolled alternative: clang16 325, gcc13 368.

The SSSE3 implementation of this function however was around 110 perf hits/counts for both compilers.
However I'm unsure if the code is worth it as for me at least it's only looking like ~4% speed gain to the overall test_view program.  What was in your BAM file?  Is it just sequence, or also having quality values, alignments, aux tags, etc?  Also what compiler and version are you using?  What CPU?

The unrolled alternative looks better than both orig and the new variant as far as scalar code goes.  It's a work-in-progress, but I have atm:

```
    const char *seq_end_ptr_8atatime = seq + (len & (~7ULL));                                        
    uint16_t *c16 = (uint16_t *)code2base;                                                           
    while (seq_cursor < seq_end_ptr_8atatime) {                                                      
        memcpy(seq_cursor+0*2, &c16[nibble_cursor[0]], 2);                                           
        memcpy(seq_cursor+1*2, &c16[nibble_cursor[1]], 2);                                           
        memcpy(seq_cursor+2*2, &c16[nibble_cursor[2]], 2);                                           
        memcpy(seq_cursor+3*2, &c16[nibble_cursor[3]], 2);                                           
        memcpy(seq_cursor+4*2, &c16[nibble_cursor[4]], 2);                                           
        memcpy(seq_cursor+5*2, &c16[nibble_cursor[5]], 2);                                           
        memcpy(seq_cursor+6*2, &c16[nibble_cursor[6]], 2);                                           
        memcpy(seq_cursor+7*2, &c16[nibble_cursor[7]], 2);                                           
                                                                                        
        seq_cursor += 16;                                                                            
        nibble_cursor += 8;                                                                          
    }                                                                                                
```

I can tidy that up a bit more as it has remnant of earlier experiments in there too with uint16_ts.

I can test AMD too though as it sounds like that's different again.

**Comment by rhpvorderman** [comment]
The reads in this set are very long. They are from the GM24385_1.fastq.gz file (from the GIAB data indices, extra long ONT reads). The average read length was longer than 7000. This data was converted to BAM using Picard FastqToSam, in order to simulate uBAM. Because of the very long reads I think the algorithm has a chance to truly shine.

I wrote a little uBAM-only parser for dnaio (alignments are ignored) and found that the SSSE3 routine improved parsing speed about 20%, so that is concordant with your result. If I cheat and do not store the BAM in bgzip blocks (just decompress the bam with gzip), then the speedup is 33%. In my uBAM parser I also used SSE2 for adding +33 to all the qualities so that is less of a bottleneck. 

How should I proceed from this point? It is clear that the scalar code in this PR is mixed depending on hardware (so not worth it) and the SSSE3 code is phenomally better, but also much harder to integrate. 10-15% overall application performance by improving a single function is quite nice. So it would feel a bit wasteful not to do anything with it.

**Comment by jkbonfield** [comment]
My ONT file had an average read length of about 4.6kb, so similar ballpark.  You probably gained more because you didn't have any alignment fields or aux tags to print up, meaning sequence printing becomes more significant.

Agreed a 10-15% speed gain for this case is worth having, even more so if the circumstances make that a larger percentage.  I'm not sure it's worth it if it's only included if the user builds using `-march=native` or `-mssse3` though.

I haven't tried it, but there is this: https://gcc.gnu.org/wiki/FunctionMultiVersioning and more blurb at https://johnnysswlab.com/cpu-dispatching-make-your-code-both-portable-and-fast/

It'd need appropriate compiler guards to ensure it's only there on compilers that support it (see htslib/hts_defs.h for examples).

For htscodecs [I went with the more complex dispatching route](https://github.com/samtools/htscodecs/blob/master/htscodecs/rANS_static4x16pr.c#L833-L1124), but primarily because when I started it I wasn't aware of the builtin methods.   Since then though it's turned out to be helpful because I have multiple versions and it's not always best to use the highest SIMD implementation on all platforms.  That's over the top though for this function I think.

Regarding the qual+33, I'm surprised it's not auto-vectorised.  If not then maybe this is a case where we ought to be using `restrict`.

**Comment by jkbonfield** [comment]
> Regarding the qual+33, I'm surprised it's not auto-vectorised. If not then maybe this is a case where we ought to be using `restrict`.

It turns out that gcc *really* hates using SIMD in loops.  I've absolutely no idea why.

[code]
@ seq4d[samtools.../htslib]; cat _.c
void add33(char *restrict a, char *restrict b, int len) {
    for (int i = 0; i < len; i++)
        a[i] = b[i] + 33;
}

@ seq4d[samtools.../htslib]; gcc13 -fopt-info-loop -O2 -fopt-info-vec-missed -Drestrict="" -c _.c
_.c:2:23: missed: couldn't vectorize loop
_.c:2:23: missed: would need a runtime alias check

@ seq4d[samtools.../htslib]; gcc13 -fopt-info-loop -O2 -fopt-info-vec-missed -c _.c
_.c:2:23: missed: couldn't vectorize loop
_.c:2:23: missed: Loop costings not worthwhile.

@ seq4d[samtools.../htslib]1; gcc13 -fopt-info-loop -O2 -fopt-info-vec-missed -ftree-loop-vectorize -c _.c
_.c:2:23: optimized: loop vectorized using 16 byte vectors
_.c:2:23: optimized: loop vectorized using 8 byte vectors
_.c:1:6: optimized: loop turned into non-loop; it never loops
[/code]

The first compilation is disabling restrict and asking what gcc thinks.  It notices an alias issue and refuses the vectorise.  Fine - that's what restrict is for.  (Although I'll note clang builds vectorised and non-vectorised versions with a run-time check for aliasing.)

The second compilation line is the one that astounded me.  It just doesn't believe vectorising is worth while.  Whatever costing model it uses is clearly not fit for purpose here.  Explicitly telling it to use the tree-loop-vectorize option fixes this (it's also part of -O3), but it does amaze me that it's not an automatic thing.

Adding `__attribute__((optimize("O3")))` around that function also rescues this so the default optimisation levels we use build sensible code.

Using this, with and without the -O3 attribute, on my `test_view` command to turn uncompressed BAM into SAM changed from 3.9 to 3.2s.  So a full 20% speed gain across the entire program.

I rebuilt both versions (with and without -O3) with -fno-inline so I could report just that one function.  Auto-vectorisation of this code is massively beneficial, as we'd expect.

```
$ for i in `seq 1 5`;do taskset -c 20 perf record ./test/test_view -N 100000 /tmp/_.bam -p /dev/null 2>/dev/null;perf report -n 2>/dev/null| grep add33;done
    27.04%          1302  test_view  test_view           [.] add33
    27.80%          1268  test_view  test_view              [.] add33
    27.86%          1261  test_view  test_view           [.] add33
    27.07%          1222  test_view  test_view           [.] add33
    27.09%          1225  test_view  test_view              [.] add33

$ for i in `seq 1 5`;do taskset -c 20 perf record ./test/test_view -N 100000 /tmp/_.bam -p /dev/null 2>/dev/null;perf report -n 2>/dev/null| grep add33;done
     2.26%            80  test_view  test_view              [.] add33
     2.33%            84  test_view  test_view           [.] add33
     2.56%            91  test_view  test_view           [.] add33
     2.25%            80  test_view  test_view           [.] add33
     2.23%            80  test_view  test_view              [.] add33
```

That's not even remotely close!  I don't want to add explicit vectorisation code, but clearly supplying some hints to the compiler are appropriate. `#pragma GCC ivdep` isn't enough either, as that's much the same as restrict.  Maybe gcc just doesn't know how many loop counts exist?

An alternative implementation, rather brain dead, does get vectorised, but only with gcc 12 onwards.  Otherwise it's still rubbish:

```
void add33(uint8_t *restrict a, uint8_t *restrict b, int len) {
    int i, j;
#define NSTEP 16

    // Unrolled inner loop with fixed size to aid gcc -O2
    for (i = 0; i < len-NSTEP; i+=NSTEP)
        for (j = 0; j < NSTEP; j++)
            a[i+j] = b[i+j] + 33;

    // remainder
    for (; i < len; i++)
        a[i] = b[i] + 33;
}
```

**Comment by rhpvorderman** [comment]
Yeah I don't usually bother with the compiler optimizations and go straight for the SIMD. This is my function:
```C

static void
decode_bam_qualities(uint8_t *dest, const uint8_t *encoded_qualities, size_t length)
{
    const uint8_t *end_ptr = encoded_qualities + length;
    const uint8_t *cursor = encoded_qualities;
    uint8_t *dest_cursor = dest;
    #ifdef __SSE2__
    const uint8_t *vec_end_ptr = end_ptr - sizeof(__m128i);
    while (cursor < vec_end_ptr) {
        __m128i quals = _mm_loadu_si128((__m128i *)cursor);
        __m128i phreds = _mm_add_epi8(quals, _mm_set1_epi8(33));
        _mm_storeu_si128((__m128i *)dest_cursor, phreds);
        cursor += sizeof(__m128i);
        dest_cursor += sizeof(__m128i);
    }
    #endif
    while (cursor < end_ptr) {
        *dest_cursor = *cursor + 33;
        cursor += 1;
        dest_cursor += 1;
    }
}
```
Because of the compile guards this will compile to slow but working code. Since SSE2 is part of the X86_64 spec, it is automatically compiled in on that platform. (Which is 90%+ of the target audience anyway so quite happy with that).
I am very thankful for your digressions into making the compiler behave. That is very instructive to me. I am quite new to this area of programming for performance. Especially when it comes to dynamic dispatch and making the compiler behave.

jkbonfield:
My own preference is where possible to avoid explicit intrinsics, as it doesn't help ARM for example. There are generalised vectorisation libraries, but then we get additional code complexities (although some are as simple as a header file you copy which isn't so bad).  Best though is working out how to get the compiler to do the work for us.

I added https://github.com/samtools/htslib/pull/1679 to beat the qual+33 problems into submission, but the nibble-to-ASCII sequence conversion is your PR here.  I'll let you progress with that one.

I think it would be good though to improve both the scalar version (eg my unrolled form) as well as to incorporate the SSSE3 variant.

rhpvorderman:
> My own preference is where possible to avoid explicit intrinsics, as it doesn't help ARM for example. There are generalised vectorisation libraries, but then we get additional code complexities (although some are as simple as a header file you copy which isn't so bad). Best though is working out how to get the compiler to do the work for us.

Yes, but it only works well for trivial problems. Writing the nibble decoding in such a way that the compiler automatically comes up with the sequence presented here is not feasible. Luckily most problems are trivial.

For this function I will try out dynamic dispatch. I think it is a good fit for this problem as the dispatch overhead won't be that bad given that sequences are usually at least 150bp. So compiler inlining won't save the day here. Also it is a nice opportunity for me to learn about dynamic dispatch and having someone already experienced review the code. 

It might take me a few day to get to this though, if you don't mind.

jkbonfield:
Is this still being worked on.

We cannot accept this as it currently stands as it'd need CPU detection or some auto-dispatch mechanism so the binaries are compatible across hardware platforms.   At the very least, we could consider a non SSSE3 variant if there is better auto-vectorised or optimisable code.

rhpvorderman:
Hi, this is still on my todo list, but work has been tremendously busy at the moment. Feel free to close this and I can make another attempt using dynamic dispatch at a later date. Unless you want to implement that yourself of course.

daviesrob:
Ok, I'll close this for now.  Feel free to open a new pull request if you come up with a new version.

PR #1683, opened 2023-10-03
Add C++ casts for external headers.

jkbonfield:
Klist.h and kseq.h use calloc, realloc and memchr in static inline code, but if we wish to permit our external headers to be used from a C++ include then these all need explicit casts as "void *" doesn't work the same in C++ as in C.

It's tempting to use `extern "C"` in an `#ifdef __cplusplus` guard, but the nature of these pseudo-template klib headers is such that the code will be expanded up later inside a C++ file so the extern "C" doesn't solve it.

See #1674 and #1682

### DIFF
--- a/htslib/klist.h
+++ b/htslib/klist.h
@@ -43,7 +43,7 @@
 		kmptype_t **buf;												\
 	} kmp_##name##_t;													\
 	SCOPE kmp_##name##_t *kmp_init_##name(void) {						\
-		return calloc(1, sizeof(kmp_##name##_t));						\
+		return (kmp_##name##_t *)calloc(1, sizeof(kmp_##name##_t));		\
 	}																	\
 	SCOPE void kmp_destroy_##name(kmp_##name##_t *mp) {					\
 		size_t k;														\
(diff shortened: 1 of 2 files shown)


htslib/klist.h
```diff
@@ -88,9 +88,9 @@
 		size_t size;													\
 	} kl_##name##_t;													\
 	SCOPE kl_##name##_t *kl_init_##name(void) {							\
-		kl_##name##_t *kl = calloc(1, sizeof(kl_##name##_t));			\
+		kl_##name##_t *kl = (kl_##name##_t *)calloc(1, sizeof(kl_##name##_t));			\
 		kl->mp = kmp_init(name);										\
-		kl->head = kl->tail = kmp_alloc(name, kl->mp);					\
+		kl->head = kl->tail = kmp_alloc(name, kl->mp);		\
```
jmarshall, line 93: Was touching this line an accident? OTOH the lines with the new casts could benefit from adjusting their backslashes.

jkbonfield: Sort of.  I started updating the backslashes to be neat and then just realised the true scale of the horror so backed out of it, mostly...  I spotted it after submitting the PR but thought it's largely irrelevant given how messy the rest are!  I can undo that line too, or add another separate commit that fixes the hideous whitespace in this file.

It's a bit problematic as it can make comparing upstream trickier, but `-w` should suffice.

jmarshall: Apart from the newly lengthened lines, the file lines up nicely with 4-space hard tabs. (Which is an abomination in itself, but at least it's upstream's conventional abomination…)

jkbonfield: Ugh!  It never dawned on me it'd look OK with a non-standard tab width!

Most of the changed lines are now too long, but the one that isn't I've got indenting the same as the others (when using said abomination).  Also removed the accidental white space change.

Thanks, even though I now feel slightly ill.

jmarshall: Turns out Heng's down with the kids 😛

Issue #1699, opened 2023-11-16
Allocation size too big/invalid memory access during `extend_ref`/`cram_add_to_ref`


OctavioGalland:
# Summary

Allocation size too big during `extend_ref` during parsing of a crafted SAM/FASTA file pair. If this error is ignored, it leads to an invalid memory access during `cram_add_to_ref`.

# Environment

Built using LLVM 14 with ASAN on Ubuntu 22.04

# How to reproduce

Build with ASAN on latest commit like so:

```
git clone --recursive https://github.com/samtools/htslib
cd htslib
autoreconf -i
CC=clang-14 CXX=clang++-14 CFLAGS="-fsanitize=address -g" CXXFLAGS="-fsanitize=address -g" LDFLAGS="-fsanitize=address -g" ./configure
make -j$(nproc)

git clone --recursive https://github.com/samtools/samtools
cd samtools
autoheader
autoconf -Wno-syntax
CC=clang-14 CXX=clang++-14 CFLAGS="-fsanitize=address -g -I$(pwd)/../htslib" CXXFLAGS="-fsanitize=address -g -I$(pwd)/../htslib" LDFLAGS="-fsanitize=address -g -L$(pwd)/../htslib" ./configure
make -j$(nproc)
```

Within the samtools folder, get poc file and reproduce with:

```
echo -ne "QFNRCVNOOmMxCUxOOjFATDIyAAACACoKczEJMAljMQkyCTAJMTBJCSoJMAkwCUFDQ0fDUUdUVEEJ
KglDWlpaWloqCnMxCTAJYzEJMTExMTExMTExMTExMTExMTEyCTAJMTBNCSoJMAkwCUFDQ0fDUUdU
VEEJKglDWlpaWlpaCSoqKkkyKzIyMjIyCUhEOmMyMjIgIglDWlpaWlpaCU1EQ1oAKgpzMQkwCWMx
CTIJMAkxMEkJKgkwCTAJQUNDR8NRR1RUQQkqCUNaWlpaWioKczEJMAljMQkyCTAJMTBJCSoJMAkw
CUFDQ0fDUUdUVEEJbGxsbGxsbGxsbGxsbGxsbGxsbGxsKglDWlpaWlpaCSoyMjIyMjIyMjIyMjIy
MjIaMjIyMjIyMjIyMjIyMjIyOzIyMjIyMjIyMjIyMjIyMjIyMjIyKioqSTIrMjIJTURDWloyMjIg
[...]

Issue #1697, opened 2023-11-16
Assertion violation during `cram_beta_encode_init`


OctavioGalland:
# Summary

Assertion violation in `cram_beta_encode_init` during parsing of a crafted SAM/FASTA file pair.

# Environment

Built using LLVM 14 with ASAN on Ubuntu 22.04

# How to reproduce

Build with ASAN on latest commit like so:

```
git clone --recursive https://github.com/samtools/htslib
cd htslib
autoreconf -i
CC=clang-14 CXX=clang++-14 CFLAGS="-fsanitize=address -g" CXXFLAGS="-fsanitize=address -g" LDFLAGS="-fsanitize=address -g" ./configure
make -j$(nproc)

git clone --recursive https://github.com/samtools/samtools
cd samtools
autoheader
autoconf -Wno-syntax
CC=clang-14 CXX=clang++-14 CFLAGS="-fsanitize=address -g -I$(pwd)/../htslib" CXXFLAGS="-fsanitize=address -g -I$(pwd)/../htslib" LDFLAGS="-fsanitize=address -g -L$(pwd)/../htslib" ./configure
make -j$(nproc)
```

Within the samtools folder, get poc file and reproduce with:

```
echo -ne "QFNRCVNOOmMyCUxOOis3Nzc3Nzc3Nzc3KysrKysrKysgAAAAMApAU1Gbm5ubm5ubm4+bmwkqCTAJ
MAlBQ0NHQzAJMAlBQ0NHQ0dHm5ubm5u0m5ubm5ubm5ubm5ubm5ubm5ubm5ubm5ubm5ubm5ubm5ub
m5ubm5ubm5ubm5ubCUxOOis3OkkqKpubm5ubm5ubm5ubm5ubm5ubm5ubm5ubm5ubm5ubm5ubm5ub
m5ubCUxOOis3CnMxCTMJYzEJMgkwCTEwTQkqCTAJMAlBJkNHQ0dHVFRDCSoJQU46STEtMTAKczAJ
MAljMQkxCTAJMTBNCSoJMAkwCUFBQzBHQ0dHVFQJKioqKioqKioqKgpzMQkxMTExMTExMTExMTEx
MAljMQkxMTExMTExMTExMTExCTAJMTBNCSoJMAkwCUFBQ0NHQ0dHVFQJKioqKioqKioqKgpzMQkw
[...]

PR #1701, opened 2023-11-16
Cram fuzzing improvements.

jkbonfield:
Fixes a raft of CRAM errors, with thanks to Octavio Galland for finding and reporting.

Draft as I have a few memory leaks when erroring out after failure to encode.

### DIFF
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -3558,7 +3558,7 @@ static int process_one_read(cram_fd *fd, cram_container *c,
         cr->cigar  = 0;
         cr->ncigar = 0;
         cr->nfeature = 0;
-        cr->aend = cr->apos;
+        cr->aend = MIN(cr->apos, c->ref_end);
         for (i = 0; i < cr->len; i++)
             if (cram_stats_add(c->stats[DS_BA], seq[i]) < 0)
                 goto block_err;
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -1454,6 +1454,12 @@ static inline int extend_ref(char **ref, uint32_t (**hist)[5], hts_pos_t pos,
     if (pos < *ref_end)
         return 0;
 
+    // Refuse to work on excessively large blocks.
+    // We'll just switch to referenceless encoding, which is probably better
+    // here as this must be very sparse data anyway.
+    if (*ref_end - ref_start > UINT_MAX/sizeof(**hist))
+        return -1;
+
     // realloc
     hts_pos_t old_end = *ref_end ? *ref_end : ref_start;
     hts_pos_t new_end = *ref_end = ref_start + 1000 + (pos-ref_start)*1.5;
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -1760,6 +1760,9 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
     int r1, r2, sn, nref, embed_ref, no_ref;
     spare_bams *spares;
 
+    if (!c->bams)
+        goto err;
+
     if (CRAM_MAJOR_VERS(fd->version) == 1)
         goto err;
 
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -1696,6 +1696,7 @@ static int cram_generate_reference(cram_container *c, cram_slice *s, int r1) {
     c->ref       = ref;
     c->ref_start = ref_start+1;
     c->ref_end   = ref_end+1;
+    c->ref_free  = 1;
 
     return 0;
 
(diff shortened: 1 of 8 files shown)


cram/cram_encode.c
```diff
@@ -1825,8 +1843,11 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
             // This starts as 'N' and is amended on-the-fly as we go
             // based on MD:Z tags.
             if ((c->ref_id = bam_ref(b)) >= 0) {
-                c->ref_free = 1;
                 c->ref = NULL;
+                // c->ref_ref is boolean; whether to free c->ref.  In this
```
jmarshall, line 1847: I'm confused by this comment, as there is no `ref_ref` :smile: Should be `c->ref_free`?

cram/cram_encode.c
```diff
@@ -1825,8 +1843,11 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
...
+                // case c->ref will be our auto-embedded sequence.
+                // Do not confuse with fd->ref_free which is a ptr to free
```
jmarshall, line 1849: Slightly ambiguous — is this “a ptr to be freed” or ”`fd->ref_free` points to [something named] `free`”?

cram/cram_encode.c
jkbonfield: Doh!  Yes.  Not a helpful comment afterall. :)

Thanks

cram/cram_encode.c
jkbonfield: I'll edit.  I was I admit being somewhat deliberately terse to avoid it going over 80 columns :)

cram/cram_io.c
```diff
@@ -5564,7 +5571,7 @@ int cram_close(cram_fd *fd) {
     if (fd->mode == 'w') {
         /* Write EOF block */
         if (0 != cram_write_eof_block(fd))
-            return -1;
+            ret = -1;
```
daviesrob, line 5574: It might be an idea to avoid writing the eof block if an error has already been detected.

cram/cram_encode.c
jmarshall: A very valid viewpoint 😄 though I have to admit sometimes I've restandardised on 96.

jkbonfield: Indeed sometimes 80 is a bit restrictive.  It does need a limit somewhere though.  Sometimes it gets rediculous.  Htslib's longest line is 245 (an error string which could be broken up with `"blah""blah"` syntax),   Similarly all of samtools longest lines are printfs in stats.c, and in bcftools it's VCF printing in csq.c.  Htslib's average line length is < samtools < bcftools too, as I expected.

The longest lines without quotes though are
https://github.com/samtools/samtools/blob/develop/reset.c#L364
https://github.com/samtools/htslib/blob/develop/samples/mpileup.c#L152
https://github.com/samtools/samtools/blob/develop/stats.c#L1518
https://github.com/samtools/htslib/blob/develop/hts.c#L4035

I find such source just hard to read as they break the indentation once things get line-wrapped in the terminal, or without line-wrapping (eg github) they're equally a pain as you have to keep scrolling back and forth.  Doubly bad in github due to having the horizontal scrollbar needing vertical scrolling to get to!  (Can use arrow keys instead)

jkbonfield:
Slowly beating this into shape.  The good news is that I can confirm cherry-picking the last commit here (hts_open_fuzzer) onto develop quite quickly detected one of the recently reported fuzzing issues, so I have confidence that this is exercising a much better portion of our code base.

With this PR, it's not (YET!) failed though, so it's looking good for us to be able to detect more problems. (I'm not naive enough to assume they've all been discovered.)

cram/cram_io.c
```diff
@@ -4412,7 +4431,8 @@ void cram_free_slice(cram_slice *s) {
         int i;
 
         if (s->hdr) {
-            for (i = 0; i < s->hdr->num_blocks; i++) {
+            // minus naux_blocks as we free elsewhere?
```
daviesrob, line 4434: As far as I can tell, aux blocks are not freed elsewhere.  Although this does raise the question of if the data could be leaked between the pointers being [moved into `s->aux_block[]`](https://github.com/samtools/htslib/blob/be1630260dc374b5a82495d9394ac46149b2547c/cram/cram_encode.c#L2032) and then [copied into `s->block[]`](https://github.com/samtools/htslib/blob/be1630260dc374b5a82495d9394ac46149b2547c/cram/cram_encode.c#L1167) (called from [here](https://github.com/samtools/htslib/blob/be1630260dc374b5a82495d9394ac46149b2547c/cram/cram_encode.c#L2367))

daviesrob: When reading the proof-of-concept files in #1697 and #1699, it looks like data is being leaked in this way.  It happens after failing to initialise a BETA codec [here](https://github.com/samtools/htslib/blob/be1630260dc374b5a82495d9394ac46149b2547c/cram/cram_encode.c#L2135).

jkbonfield: I did find the issue with BETA codec and fixed it, and the new fuzzer verifies this.  However oddly samtools still leaks.  Maybe we're still doing something different then.  I'll investigate.

daviesrob:
This is better resisting fuzzing since the last update.  I will merge the progress so far.

jkbonfield:
Thanks.  I think that's the right decision as it's all a series of incremental steps and I don't think this introduces new bugs (I hope not anyway).  If we find a bunch more, I'll start a cram-fuzz-2 PR :)   Let's hope it'll be a while coming!

PR #1705, opened 2023-11-24
Miscellaneous trivia and minor fixes

jmarshall:
Several minor fixes that you may wish to select from and squash:

* Remove remnants of now unused `HTS_HAVE_NEON`. This has already been cherry-picked into pysam, as we print it out in the logs so these remnants could have caused confusion.

* Remove remnants of `NUMERIC_VERSION`, unused since #1226. Eventually with a soversion bump you may revisit how `MACH_O_CURRENT_VERSION` is set and perhaps set it from `version.sh numeric`, so that code in _version.sh_ has been left. (We will soon not ship _htslib/version.sh_ in pysam so will need to apply something like this to our _htslib/Makefile_ whether htslib does or not.)

* Document primarily `MM`/`ML` tags — rather than the draft `Mm`/`Ml` — and fix base-mod-related typos.

* Avoid a “`PRId64` used with `size_t`” warning. This change adds a cast to the always-run `if()` condition as well as to the almost-never-run diagnostic printing, so deserves more review than the rest of this PR.

* Install _annot-tsv.1_ man page and alphabetise `annot-tsv` rules. I thought perhaps the maintainers would rename this to something more appealing — e.g. `tabanno`, à la `tabix` — before merging this new tool :smile: but in lieu of that this restores the obvious alphabetical ordering of `bgzip`/`htsfile`/`tabix` et al.


daviesrob:
Thanks, these all look good.

PR #1714, opened 2023-12-04
Annottsv

vasudeva8:
updated to limit usage text to 80 char per line.
updated version display to have copyright.
returns success/0 on help and failure code/1 on invalid option.


daviesrob:
Adjusted line wrapping slightly, and I've also made a small change so that the output goes to stderr if you use an unrecognised option (-h,--help still go to stdout).

PR #1716, opened 2023-12-06
man update for -a option

vasudeva8:
Man update for change 1709. Updated with a sample usage.


daviesrob:
I've added a bit of extra text so it's easier to see what should be in the `tgt_with_header.txt.gz` file.

Issue #1723, opened 2023-12-19
Passing missing --reference incorrectly reports cram file missing as well


ASLeonard:
Hi,
This is more of a user error, but maybe a helpful bug to fix. This is on
```
samtools 1.19-1-ga4c73e4-dirty
Using htslib 1.19-1-g61b037bb-dirty
```

If I try to do something with a cram, but pass a non-existant reference (in this case a typo) `samtools stats --reference <ref_with_typo>  sample.cram`, I get the following output that misleadingly suggests the cram file itself doesn't exist.

```
[E::refs_load_fai] Failed to open reference file '<ref_with_typo>'
[E::hts_open_format] Failed to open file "sample.cram" : No such file or directory
samtools stats: failed to open "sample.cram": No such file or directory
```

Oddly, in the case of `stats` I get a segfault, but in other operations like `view` I just get the errors that the reference can't be opened **and** the cram doesn't exist. I would guess the expected behaviour would be error out when the reference couldn't be loaded and avoid misleading the user the cram file doesn't exist (and potentially segfault).

Best,
Alex

vasudeva8:
samtools and htslib are updated to avoid the segfault.
The error display is updated to show file not found on reference file open in faibuild3core and invalid argument on hts_open_format on cram file.

PR #1724, opened 2023-12-21
fix for incorrect failure message with invalid reference file and crash fix

vasudeva8:
This change is for #1723 where misleading failure message is displayed.
The error number is reset if it is due to reference file access before reporting the error, to avoid misleading message.
Another change is made to avoid crash/invalid access in hts_close.
A change in samtools will also be raised to do null check before invoking hts_close.

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -938,10 +938,18 @@ htsFile *hts_open_format(const char *fn, const char *mode, const htsFormat *fmt)
          fmt->format == fastq_format))
         fp->format.format = fmt->format;
 
-    if (fmt && fmt->specific)
-        if (hts_opt_apply(fp, fmt->specific) != 0)
+    if (fmt && fmt->specific) {
+        if (hts_opt_apply(fp, fmt->specific) != 0) {
+            if (((hts_opt*)fmt->specific)->opt == CRAM_OPT_REFERENCE &&
+                (errno == ENOENT || errno == EIO || errno == EBADF ||
+                  errno == EACCES || errno == EISDIR)) {
+                /* error during reference file operation
+                 for these specific errors, set the error as EINVAL */
+                errno = EINVAL;
+            }
             goto error;
-
+        }
+    }
     if ( rmme ) free(rmme);
     return fp;
 
@@ -1595,6 +1603,10 @@ error:
 int hts_close(htsFile *fp)
 {
     int ret = 0, save;
+    if (!fp) {
+        errno = EINVAL;
+        return -1;
+    }
 
     switch (fp->format.format) {
     case binary_format:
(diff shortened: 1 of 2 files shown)


hts.c
```diff
@@ -938,10 +938,15 @@ htsFile *hts_open_format(const char *fn, const char *mode, const htsFormat *fmt)
          fmt->format == fastq_format))
         fp->format.format = fmt->format;
 
-    if (fmt && fmt->specific)
-        if (hts_opt_apply(fp, fmt->specific) != 0)
+    if (fmt && fmt->specific) {
+        if (hts_opt_apply(fp, fmt->specific) != 0) {
+            if (((hts_opt*)fmt->specific)->opt == CRAM_OPT_REFERENCE) {
+                //reset error on reference file to avoid incorrect error report
+                errno = 0;
```
daviesrob, line 945: Rather than reset `errno` unconditionally, it would be better to take a copy before calling `hts_opt_apply()` and then use that to restore the old value if necessary.  It might also be good to only do this if `errrno` is `ENOENT`, so we only suppress the "No such file or directory" errors and leave others alone.

E.g. something like this:
```
    if (fmt && fmt->specific) {
        int save_errno = errno;
        if (hts_opt_apply(fp, fmt->specific) != 0) {
            if (((hts_opt*)fmt->specific)->opt == CRAM_OPT_REFERENCE  && errno == ENOENT) {
                //reset error on reference file to avoid incorrect error report
                errno = save_errno;
```

vasudeva8: made changes as discussed.

hts.c
```diff
@@ -938,10 +939,19 @@ htsFile *hts_open_format(const char *fn, const char *mode, const htsFormat *fmt)
...
         fp->format.format = fmt->format;
 
-    if (fmt && fmt->specific)
-        if (hts_opt_apply(fp, fmt->specific) != 0)
+    if (fmt && fmt->specific) {
+        int save_errno = errno;
+        if (hts_opt_apply(fp, fmt->specific) != 0) {
+            if (((hts_opt*)fmt->specific)->opt == CRAM_OPT_REFERENCE) {
+                const char *ref = ((hts_opt*)fmt->specific)->val.s;
+                hts_log_error("Couldn't load reference \"%s\" : %s",
```
daviesrob, line 947: Having tested this, it looks like `refs_load_fai` already reports this error, so we probably don't need to do it again here after all.  Although making `refs_load_fai` print the error message might be useful:

```
samtools stats --reference /foo 9305_4#1c.cram
[E::fai_build3_core] Failed to open the file /foo
[E::refs_load_fai] Failed to open reference file '/foo'
[E::hts_open_format] Couldn't load reference "/foo" : No such file or directory
[E::hts_open_format] Failed to open file "9305_4#1c.cram"
samtools stats: failed to open "9305_4#1c.cram"
```

PR #1727, opened 2024-01-16
bgzip - modified and access time as in source file

### LINKED ISSUE #1718: bgzip: keep original modification time when compressing a file.
It would be nice if bgzip had an option to keep the original modification date of the file like gzip does:

```bash
$ cp -a original.tsv aa.tsv

# Compressing with gzip keeps original file modification time.
$ gzip aa.tsv

$ ls -l aa.tsv.gz 
-rw-rw-r--+ 1 ghuls ghuls 97776 May 22  2018 aa.tsv.gz

$ rm aa.tsv.gz 

$ cp -a original.tsv aa.tsv

# Compressing with bgzip does not kees original file modification time.
$ bgzip aa.tsv

$ ls -l aa.tsv.gz 
-rw-rw-r--+ 1 ghuls ghuls 94512 Dec  8 14:56 aa.tsv.gz
```

vasudeva8:
Fixes #1718 
Updated bgzip to retain the file modification and access time during compression and decompression.
In Mac/linux it retains upto microsecond and in windows it does with second precision.

### DIFF
--- a/bgzip.c
+++ b/bgzip.c
@@ -34,6 +34,8 @@
 #include <stdarg.h>
 #include <getopt.h>
 #include <inttypes.h>
+#include <sys/stat.h>
+#include <sys/time.h>
 #include "htslib/bgzf.h"
 #include "htslib/hts.h"
 #include "htslib/hfile.h"
@@ -41,6 +43,7 @@
 #ifdef _WIN32
 #  define WIN32_LEAN_AND_MEAN
 #  include <windows.h>
+#  include <sys/utime.h>
 #endif
 
 static const int WINDOW_SIZE = BGZF_BLOCK_SIZE;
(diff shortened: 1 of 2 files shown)


bgzip.c
```diff
@@ -103,6 +106,83 @@ static int confirm_filename(int *is_forced, const char *name, const char *ext)
...
+#else
+    //time upto usec - access time
+    tval[0].tv_sec = status->st_atime;
+    tval[0].tv_usec = status->st_atimensec / 1000;
+    //time upto usec - modification time
+    tval[1].tv_sec = status->st_mtime;
+    tval[1].tv_usec = status->st_mtimensec / 1000;
+    if (utimes(path, &tval[0]) < 0) {
+        fprintf(stderr, "[bgzip] Failed to set file specifications.\n");
+        return -1;
```
jkbonfield, line 180: Do we need nanosec accuracy?

It's thwart with problems.  See the myriad of code in https://github.com/kunpengcompute/gzip/blob/6a3ebab7e475fb5ca9b01c344eb45c27ea81ef89/lib/stat-time.h#L66

It also doesn't build on linux even unless we define `_XOPEN_SOURCE` or similar.  For example:

```
$ cat _.c
#include <sys/types.h>
#include <sys/stat.h>
#include <sys/time.h>
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>

  struct stat sb;
  if (stat(argv[1], &sb) == 0)
    printf("%ld %ld\n", sb.st_mtime, sb.st_mtimensec);

  return 0;
}

$ cc _.c
_.c: In function 'main':
_.c:11:41: error: 'struct stat' has no member named 'st_mtimensec'; did you mean 'st_mtim'?
     printf("%ld %ld\n", sb.st_mtime, sb.st_mtimensec);
                                         ^~~~~~~~~~~~
                                         st_mtim

```

While we could do nanosecond precise, frankly it's probably sufficient to just squash that to zero and just do second precision.  That will make things far more portable without the need for lots of ifdefs.

jkbonfield: Also, I note that `cc -D_XOPEN_SOURCE=600` works, but `cc -D_XOPEN_SOURCE=700` does not.  We get away with it in htslib because we add 600 to the config.h file.  However I'm not sure we want to rely on something which is already known to have been removed in newer XOpen specifications.

jkbonfield: Thinking more, if we did want to keep nanoseconds, then we could do a `#if defined(_XOPEN_SOURCE) && _XOPEN_SOURCE == 600` ... or similar.  However I'm still not convnced of the necessity of it being this fine grained.

bgzip.c
```diff
@@ -103,6 +106,83 @@ static int confirm_filename(int *is_forced, const char *name, const char *ext)
...
+   filestatus  - pointer to status structure in which the data to be stored
+   returns 0 on success and -1 on failure
+*/
+static int getfilespec(const char *path, void *filestatus)
+{
+#ifdef _WIN32
+    struct _stat *status = (struct _stat*)filestatus;
+#else
+    struct stat *status = (struct stat*)filestatus;
+#endif //_WIN32
```
jkbonfield, line 120: This shouldn't be necessary.

We use MinGW which supports this, but even MSVC appears to have "struct stat" as a native type.  Eg see a draft MSVC native port here: https://github.com/refresh-bio-sandbox/htslib-msvc/blob/develop/hts.c#L4600

If we needed to do OS specific shenanigans, it's better hidden away in os/* and hts_os.c.

bgzip.c
vasudeva8: changed to use the same stat structure for all OS.
changed to set time value with second precision on all cases.

bgzip.c
vasudeva8: Not expecting this API to be used anywhere else and is for internal use only, so not adding to hts_os.
updated to avoid _stat and uses stat in all cases.
Note: to set time, uses _utime in windows and utimes with micro sec part, if any, set to 0 in linux and Mac.

jkbonfield:
It appears to work fine, which is good, but it could do with some documentation to explain it.

Gzip isn't that clear either until you carefully read the manual.  Basically gzip and `--name` and `--no-name` options which claim to be "(do not)save or restore the original name and time stamp".  I naively assumed with `--no-name` that compressing and decompressing would change the date, but no it does not.

What it's referring to is whether the time stamp (and filename) are stored within the gzip header meta-data.  Normally we can compress, touch the .gz file, and then decompress and we get the touched timestamp used for the uncompressed data as it's simply keeping the time stamp in the filesystem meta-data.  However with `--name` you can copy the file around, touch the time stamp, etc, and on decompression it'll restore the time back to the timestamp of the original input file.

We don't have that form encoded here, and it's probably not needed either.  (And may trip up some lazy programs that detect BGZF format by exact bit-string matching.)  Instead we're simply preserving the filesystem time stamps, albeit with 1 second granularity.  How this works probably needs a line in the manual, especially given it has subtle differences to command line gzip (or at least how command-line gzip *can* work given the correct options).

vasudeva8:
man page updated with timestamp being used from filesystem and possible overwrite on access time by system.

PR #1736, opened 2024-02-01
Drop duplicate FORMAT tags and fix an error with dropping large tags

### LINKED ISSUE #1733: bcftools corrupts duplicate GT format fields
Running `bcftools` (here `filter` but seems to affect other commands as well) on a VCF with duplicate GT (genotype) FORMAT fields seems to change all but the last GT value. Looks like `./.`, `0/0`, `0/1`, `1/1` get converted to `0,0`, `2,2`, `2,4`, and `4,4`, respectively. I'm not 100% sure if duplicate GT values are legal, but I would expect an error instead of invalid data. Non-GT fields don't seem to have the problem. I'm using bcftools 1.19, but this can also be reproduced in bcftools 1.12.

Minimized test case follows. I would expect the payload to match that of the input.

```
$ cat foo.vcf 
##fileformat=VCFv4.1
##FORMAT=<ID=GT,Number=1,Type=String>
##FORMAT=<ID=X,Number=1,Type=Integer>
##contig=<ID=chr1>
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	A
chr1	1	.	A	C	.	.	.	GT:X:GT:X	0/1:9:0/1:9
$ bcftools filter foo.vcf | sed '/^#/d'
chr1	1	.	A	C	.	PASS	.	GT:X:GT:X	2,4:9:0/1:9
```

pd3:
The removal of large tags introduced by b49eea4 and 9db565de could not work correctly, the memmove pointers were wrong!

Resolves #1733

### DIFF
--- a/bgzip.1
+++ b/bgzip.1
@@ -71,7 +71,9 @@ otherwise when compressing bgzip will write to a new file with a .gz
 suffix and remove the original.  When decompressing the input file must
 have a .gz suffix, which will be removed to make the output name.  Again
 after decompression completes the input file will be removed. When multiple
-files are given as input, the operation is performed on all of them.
+files are given as input, the operation is performed on all of them. Access
+and modification time of input file from filesystem is set to output file. 
+Note, access time may get updated by system when it deems appropriate.
 
 .SH OPTIONS
 .TP 10
--- a/bgzip.c
+++ b/bgzip.c
@@ -34,6 +34,8 @@
 #include <stdarg.h>
 #include <getopt.h>
 #include <inttypes.h>
+#include <sys/stat.h>
+#include <sys/time.h>
 #include "htslib/bgzf.h"
 #include "htslib/hts.h"
 #include "htslib/hfile.h"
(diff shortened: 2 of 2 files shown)


jkbonfield:
What does "pd3 force-pushed the develop branch from 6fd3fde to dd9c561 26 minutes ago" mean?  Normally we see merge commits (which are fast-forward only).

Eg "@[vasudeva8](https://github.com/vasudeva8) vasudeva8 merged commit [a71551b](https://github.com/samtools/htslib/commit/a71551bd56efafecd1f00e7092ae89f74e69eaf6) into samtools:develop [Jan 18, 2024](https://github.com/samtools/htslib/pull/1731#event-11527250346)
9 checks passed" from the last NEWS updates.

Unfortunately I just did a git pull and my develop didn't have 6fd3fde in it anyway, so I can't get a sensible network graph out to know what happened.

Generally however, we shouldn't be force pushing over the top of develop as it can cause breakage.  Was 6fd3fde on the main line before this happened?

Was this an accidental push to samtools:develop instead of pd3:develop?

pd3:
I sincerely hope the force push was on pd3:develop after samtools:develop was merged to it, that was the intention :-(
Not sure what happen, probably the open pull request. Uh.

PR #1738, opened 2024-02-06
Fix arithmetic overflow in load_ref_portion() on very long refs

### LINKED ISSUE #1734: CRAM load_ref_portion() fails on some Mistletoe references
Even though the Mistletoe genome has been chopped into references less than 2^31 bases long, it seems they're still long enough to trigger an arithmetic overflow in [`load_ref_portion()`](https://github.com/samtools/htslib/blob/65ae5744347c9403c061585fa2fc9f5262f2f977/cram/cram_io.c#L3244).

The fai entry for the first reference is:
```
OY728119.1      2143528264      58      80      81
```

which gives these values for `offest` and `len` in the code linked above:
```
(gdb) p e->line_length ? e->offset + (start-1)/e->bases_per_line * e->line_length + (start-1) % e->bases_per_line : start-1
$7 = 58

(gdb) p (e->line_length ? e->offset + (end-1)/e->bases_per_line * e->line_length + (end-1) % e->bases_per_line : end-1) - offset + 1
$9 = -2124644929
```

The `len` value goes on to cause a `malloc()` failure, which is happily caught but causes the CRAM writer to drop into embed_ref mode when it shouldn't have to.

daviesrob:
While the Mistletoe reference has been chopped into segments of less than 2^31 bases, they were still long enough to cause an overflow in the load_ref_portion() `len` calculation.  This was due to the line endings taking the total over INT_MAX.  Fix be changing the data type of `start` and `end` to `hts_pos_t` so the entire calculation is done on 64-bit values.

Callers to load_ref_portion() are also updated where necessary to use `hts_pos_t`, making long references more likely to work should they ever be supported in CRAM.  All callers of callers were already using 64-bit values due to earlier upgrades.

There's a potential complication over calculating the MD5 checksums as the exported hts_md5_update() function takes an unsigned long for the length.  On 64-bit platforms with 32-bit unsigned long (i.e. [Windows](https://learn.microsoft.com/en-us/cpp/cpp/data-type-ranges?view=msvc-170)) it is necessary to add a loop if the reference is a very long one.  On platforms with 64-bit long a single call is still used, and the loop should be optimised out.

Fixes #1734 (CRAM load_ref_portion() fails on some Mistletoe references)

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -3221,7 +3221,8 @@ void cram_ref_decr(refs_t *r, int id) {
  * Returns all or part of a reference sequence on success (malloced);
  *         NULL on failure.
  */
-static char *load_ref_portion(BGZF *fp, ref_entry *e, int start, int end) {
+static char *load_ref_portion(BGZF *fp, ref_entry *e,
+                              hts_pos_t start, hts_pos_t end) {
     off_t offset, len;
     char *seq;
 
@@ -3317,7 +3318,7 @@ static char *load_ref_portion(BGZF *fp, ref_entry *e, int start, int end) {
  */
 ref_entry *cram_ref_load(refs_t *r, int id, int is_md5) {
     ref_entry *e = r->ref_id[id];
-    int start = 1, end = e->length;
+    hts_pos_t start = 1, end = e->length;
     char *seq;
 
     if (e->seq) {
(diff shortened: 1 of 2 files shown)


daviesrob:
The test file I have is fairly shallow data, and the saving from having a reference is significant:
```
$ ls -1sh mistletoe/
total 163G
45G drVisAlbu1.merge.embed1.cram
34G drVisAlbu1.merge.embed2.cram
25G drVisAlbu1.merge.fasta_refs.cram
25G drVisAlbu1.merge.md5_refs.cram
37G drVisAlbu1.merge.no_ref.cram
```

PR #1741, opened 2024-02-07
Speed up sam_parse_B_vals

jkbonfield:
There are a few issues with this.

hts_str2int and hts_str2uint are slightly inefficient.

- The use of unsigned char for "d" causes type conversions which are unnecessary.  All we care about is wrap around to positive values.

- The check on !fast normally passes for values that aren't huge, as we nearly always end before hitting that loop.  Hence we then pointlessly do devisions and a while loop with zero cycles, so we can lift that check up one left.

- Simplification of the overflow conditions.  This is not necessary as every possible positive number can be negated to negative without overflow.  (Overflow can only possibly occur when doing the reverse.)

The sam_parse_B_vals function is more tricky to nail. In attempting to speed up the str2int functions I found huge variation in speeds based on compiler (gcc, clang), version, and optimisation levels.  Sometimes wildly different.  The most minor changes would flip speed.

After a lot of trial and error I realised gcc -falign-loops=5 also wildly changes things, and that most speed changes were down to instruction cache line boundaries and their interplay with the uOPS caches.  A lot of "perf" analysis later showed the main culprit to be the skip_to_comma_ loop, so the code was restructed so this isn't normally executed.

In the process I also discovered and fixed a bug.

    Bi:B:I,-2147483648,-2147483647,0,2147483647

is meant to detect overlaps and rewrite it as

    Bi:B:i,-2147483648,-2147483647,0,2147483647

However the detection of '-' sign for unsigned values is incorrect as it's pointing to ',' instead, so an entire chunk of logic was never executed and those out-of-bounds negatives got modified to zero.

Note this does change behaviour.  Previously an aux tag of XX:B:C,1,2#3,4 would silently return 1,2,4.  It now errors on the 2#3 instead.  I believe this is an improvement over silently changing broken inputs.

--------------------------------------------------

Perf stat cycles on an ONT sam file with 4685 records of :B:C, and :B:c, using "test_view -B ont.sam" as a benchmark.

    gcc7    -O2: 5565824443
    gcc7    -O3: 5779736469
    gcc13   -O2: 5565893109
    gcc13   -O3: 5392426978
    clang10 -O2: 5344657729
    clang10 -O3: 5348030140
    clang16 -O2: 4563321159
    clang16 -O3: 4575986193

For reference, some stats on the develop branch on the same file:

    gcc13   -O3: 9724589000
    clang16 -O3: 6398268577

So the effective speed of samtools is 80% and 40% slower respectively! That's total time spent, including everything else being executed too, so the sam_parse_B_vals speed up is considerably higher.

Investigating with specific counters for the gcc13 case we see this.

develop:

     9724589000      cycles
     3308765106      stalled-cycles-frontend   #   34.02% frontend cycles idle
     1898104974      stalled-cycles-backend    #   19.52% backend cycles idle
    17596774767      instructions              #    1.81  insn per cycle
                                               #    0.19  stalled cycles/insn
     1629365133      idq.empty
    13489362500      idq_uops_not_delivered.core

    3.259651973 seconds time elapsed

This PR:

     5397541312      cycles
     1438049271      stalled-cycles-frontend   #   26.64% frontend cycles idle
      514817530      stalled-cycles-backend    #    9.54% backend cycles idle
    15125819343      instructions              #    2.80  insn per cycle
                                               #    0.10  stalled cycles/insn
      164253475      idq.empty
     2653272955      idq_uops_not_delivered.core

    1.809554598 seconds time elapsed

We can clearly see the frontend was being bottlenecked on on the instruction decode queue, which is caused by the aforementioned spanning of instruction cache lines.

Different architectures will be affected in different ways, and newer ones probably less badly affected.  I haven't benchmarked on other hardware.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -2383,6 +2383,14 @@ int sam_hdr_change_HD(sam_hdr_t *h, const char *key, const char *val)
  *** SAM record I/O ***
  **********************/
 
+// Future experiments to try for improving the robustness of compilation.
+// If parsing suddenly changes speed again, we may wish to force specific
+// alignment within this code so that it is immune to location changes
+// caused by modifications elsewhere.  For now however it appears to
+// have consistent behaviour.
+//
+//__attribute__((aligned(32)))
+//__attribute__((optimize("align-loops=5")))
 static int sam_parse_B_vals(char type, uint32_t n, char *in, char **end,
                             char *r, bam1_t *b)
 {
@@ -2410,66 +2418,69 @@ static int sam_parse_B_vals(char type, uint32_t n, char *in, char **end,
     b->data[b->l_data++] = type;
     i32_to_le(n, b->data + b->l_data);
     b->l_data += sizeof(uint32_t);
+
     // This ensures that q always ends up at the next comma after
     // reading a number even if it's followed by junk.  It
     // prevents the possibility of trying to read more than n items.
 #define skip_to_comma_(q) do { while (*(q) > '\t' && *(q) != ',') (q)++; } while (0)
+
     if (type == 'c') {
-        while (q < r) {
+        while (q < r && *q == ',') {
             *(b->data + b->l_data) = hts_str2int(q + 1, &q, 8, &overflow);
             b->l_data++;
-            skip_to_comma_(q);
         }
     } else if (type == 'C') {
-        while (q < r) {
-            if (*q != '-') {
+        while (q < r && *q == ',') {
+            if (q[1] != '-') {
                 *(b->data + b->l_data) = hts_str2uint(q + 1, &q, 8, &overflow);
                 b->l_data++;
             } else {
                 overflow = 1;
+                q++;
+                skip_to_comma_(q);
             }
-            skip_to_comma_(q);
         }
     } else if (type == 's') {
-        while (q < r) {
+        while (q < r && *q == ',') {
             i16_to_le(hts_str2int(q + 1, &q, 16, &overflow), b->data + b->l_data);
             b->l_data += 2;
-            skip_to_comma_(q);
         }
     } else if (type == 'S') {
-        while (q < r) {
-            if (*q != '-') {
+        while (q < r && *q == ',') {
+            if (q[1] != '-') {
                 u16_to_le(hts_str2uint(q + 1, &q, 16, &overflow), b->data + b->l_data);
                 b->l_data += 2;
             } else {
                 overflow = 1;
+                q++;
+                skip_to_comma_(q);
             }
-            skip_to_comma_(q);
         }
     } else if (type == 'i') {
-        while (q < r) {
+        while (q < r && *q == ',') {
             i32_to_le(hts_str2int(q + 1, &q, 32, &overflow), b->data + b->l_data);
             b->l_data += 4;
-            skip_to_comma_(q);
         }
     } else if (type == 'I') {
-        while (q < r) {
-            if (*q != '-') {
+        while (q < r && *q == ',') {
+            if (q[1] != '-') {
                 u32_to_le(hts_str2uint(q + 1, &q, 32, &overflow), b->data + b->l_data);
                 b->l_data += 4;
             } else {
                 overflow = 1;
+                q++;
+                skip_to_comma_(q);
             }
-            skip_to_comma_(q);
         }
     } else if (type == 'f') {
-        while (q < r) {
+        while (q < r && *q == ',') {
             float_to_le(strtod(q + 1, &q), b->data + b->l_data);
             b->l_data += 4;
-            skip_to_comma_(q);
         }
-    } else {
-        hts_log_error("Unrecognized type B:%c", type);
+    }
+    if (q < r) {
+        // Unknown B array type or junk in the numbers
+        hts_log_error("Malformed B:%c", type);
         return -1;
     }
 
--- a/sam.c
+++ b/sam.c
@@ -2383,17 +2383,175 @@ int sam_hdr_change_HD(sam_hdr_t *h, const char *key, const char *val)
  *** SAM record I/O ***
  **********************/
 
-// Future experiments to try for improving the robustness of compilation.
-// If parsing suddenly changes speed again, we may wish to force specific
-// alignment within this code so that it is immune to location changes
-// caused by modifications elsewhere.  For now however it appears to
-// have consistent behaviour.
-//
-//__attribute__((aligned(32)))
-//__attribute__((optimize("align-loops=5")))
-static int sam_parse_B_vals(char type, uint32_t n, char *in, char **end,
-                            char *r, bam1_t *b)
-{
+// The speed of this code can vary considerably depending on minor code
+// changes elsewhere as some of the tight loops are particularly prone to
+// speed changes when the instruction blocks are split over a 32-byte
+// boundary.  To protect against this, we explicitly specify an alignment
+// for this function.  If this is insufficient, we may also wish to
+// consider alignment of blocks within this function via
+// __attribute__((optimize("align-loops=5"))) (gcc) or clang equivalents.
+// However it's not very portable.
+// Instead we break into separate functions so we can explicitly specify
+// use __attribute__((aligned(32))) instead and force consistent loop
+// alignment.
+static inline int64_t grow_B_array(bam1_t *b, uint32_t *n, size_t size) {
+    // Avoid overflow on 32-bit platforms, but it breaks BAM anyway
+    if (*n > INT32_MAX*0.666) {
+        errno = ENOMEM;
+        return -1;
+    }
+
+    size_t bytes = (size_t)size * (size_t)(*n>>1);
+    if (possibly_expand_bam_data(b, bytes) < 0) {
+        hts_log_error("Out of memory");
+        return -1;
+    }
+
+    (*n)+=*n>>1;
+    return 0;
+}
+
+
+// This ensures that q always ends up at the next comma after
+// reading a number even if it's followed by junk.  It
+// prevents the possibility of trying to read more than n items.
+#define skip_to_comma_(q) do { while (*(q) > '\t' && *(q) != ',') (q)++; } while (0)
+
+HTS_ALIGN32
+static char *sam_parse_Bc_vals(bam1_t *b, char *q, uint32_t *nused,
+                               uint32_t *nalloc, int *overflow) {
+    while (*q == ',') {
+        if ((*nused)++ >= (*nalloc)) {
+            if (grow_B_array(b, nalloc, 1) < 0)
+                return NULL;
+        }
+        *(b->data + b->l_data) = hts_str2int(q + 1, &q, 8, overflow);
+        b->l_data++;
+    }
+    return q;
+}
+
+HTS_ALIGN32
+static char *sam_parse_BC_vals(bam1_t *b, char *q, uint32_t *nused,
+                               uint32_t *nalloc, int *overflow) {
+    while (*q == ',') {
+        if ((*nused)++ >= (*nalloc)) {
+            if (grow_B_array(b, nalloc, 1) < 0)
+                return NULL;
+        }
+        if (q[1] != '-') {
+            *(b->data + b->l_data) = hts_str2uint(q + 1, &q, 8, overflow);
+            b->l_data++;
+        } else {
+            *overflow = 1;
+            q++;
+            skip_to_comma_(q);
+        }
+    }
+    return q;
+}
+
+HTS_ALIGN32
+static char *sam_parse_Bs_vals(bam1_t *b, char *q, uint32_t *nused,
+                               uint32_t *nalloc, int *overflow) {
+    while (*q == ',') {
+        if ((*nused)++ >= (*nalloc)) {
+            if (grow_B_array(b, nalloc, 2) < 0)
+                return NULL;
+        }
+        i16_to_le(hts_str2int(q + 1, &q, 16, overflow),
+                  b->data + b->l_data);
+        b->l_data += 2;
+    }
+    return q;
+}
+
+HTS_ALIGN32
+static char *sam_parse_BS_vals(bam1_t *b, char *q, uint32_t *nused,
+                               uint32_t *nalloc, int *overflow) {
+    while (*q == ',') {
+        if ((*nused)++ >= (*nalloc)) {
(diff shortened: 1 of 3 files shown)


jkbonfield:
Hmm, maybe we want more work, but it's scarier.

There are two scans through the B string.  One to count commas and look for the end of the string `r`.  The other to parse the values.  We can pass in 0 for an unknown quantity and just get it to count on-the-fly as it goes (dynamically regrowing the buffer intermittently), stopping when we see a non-comma and erroring if that is then neither nul nor tab.  We already returned `*end` which was the same as the outer functions `r` so we have the end already.

My quick hack of this is between 3.3 and 4.4 billion cycles, compared to usually over 5.  So it's definitely a win despite the more complex loop structure.

jkbonfield:
After testing my idea of a single pass instead of two passes, I added another commit to do this.  It's typically a 80% speed improvement over the previous commit, making it up to 200% over develop.  (On that platform.  Newer platforms are likely less affected.)

I validated there are no major instruction alignment issues left in that function by rerunning my tests with 16 `asm("nop")` to shift all instructions by half a cache line and the differences in speed were typically 1-3%.

jkbonfield:
Some benchmarks on a modern Intel machine (same ONT file).  These show develop followed by this PR.  Here plain "gcc" is the system gcc (7.5.0)

```
=== test_view.clang10.O2 ===

real	0m1.622s
user	0m1.428s
sys	0m0.184s

real	0m1.256s
user	0m1.047s
sys	0m0.208s
=== test_view.clang10.O3 ===

real	0m1.444s
user	0m1.307s
sys	0m0.136s

real	0m1.014s
user	0m0.813s
sys	0m0.200s
=== test_view.clang16.O2 ===

real	0m1.783s
user	0m1.616s
sys	0m0.165s

real	0m0.877s
user	0m0.679s
sys	0m0.196s
=== test_view.clang16.O3 ===

real	0m1.524s
user	0m1.377s
sys	0m0.144s

real	0m0.830s
user	0m0.678s
sys	0m0.151s
=== test_view.gcc.O2 ===

real	0m2.654s
user	0m2.469s
sys	0m0.184s

real	0m0.840s
user	0m0.642s
sys	0m0.197s
=== test_view.gcc.O3 ===

real	0m2.550s
user	0m2.316s
sys	0m0.232s

real	0m1.162s
user	0m0.913s
sys	0m0.246s
=== test_view.gcc13.O2 ===

real	0m2.503s
user	0m2.281s
sys	0m0.220s

real	0m0.951s
user	0m0.733s
sys	0m0.217s
=== test_view.gcc13.O3 ===

real	0m2.478s
user	0m2.316s
sys	0m0.160s

real	0m0.990s
user	0m0.806s
sys	0m0.184s
```

And on an AMD EPYC 7713

```
=== test_view.clang10.O2 ===

real	0m1.224s
user	0m1.095s
sys	0m0.128s

real	0m0.808s
user	0m0.711s
sys	0m0.096s
=== test_view.clang10.O3 ===

real	0m1.410s
user	0m1.346s
sys	0m0.064s

real	0m0.814s
user	0m0.721s
sys	0m0.092s
=== test_view.clang16.O2 ===

real	0m1.288s
user	0m1.192s
sys	0m0.096s

real	0m0.906s
user	0m0.813s
sys	0m0.092s
=== test_view.clang16.O3 ===

real	0m1.339s
user	0m1.290s
sys	0m0.048s

real	0m0.806s
user	0m0.719s
sys	0m0.084s
=== test_view.gcc.O2 ===

real	0m2.114s
user	0m2.014s
sys	0m0.100s

real	0m0.798s
user	0m0.656s
sys	0m0.139s
=== test_view.gcc.O3 ===

real	0m2.159s
user	0m2.074s
sys	0m0.084s

real	0m0.797s
user	0m0.716s
sys	0m0.080s
=== test_view.gcc13.O2 ===

real	0m2.022s
user	0m1.953s
sys	0m0.068s

real	0m0.793s
user	0m0.725s
sys	0m0.068s
=== test_view.gcc13.O3 ===

real	0m2.152s
user	0m2.075s
sys	0m0.076s

real	0m0.793s
user	0m0.705s
sys	0m0.088s
```

Clang was consistently better, but both clang and gcc have sped up and are now a similar time to do "test_view -B ont.sam".

textutils_internal.h
```diff
@@ -197,51 +197,43 @@ const char *hts_strprint(char *buf, size_t buflen, char quote,
 // Faster replacements for strtol, for use when parsing lots of numbers.
 // Note that these only handle base 10 and do not skip leading whitespace
 
-/// Convert a string to a signed integer, with overflow detection
```
daviesrob, line 200: Is there a good reason why these functions swapped round?  It makes spotting the differences much harder than if they stayed in the original order.

jkbonfield: There was, but it's no longer relevant so I can swap them back.

Originally I implemented the signed conversion as a call out to the unsigned one for the non-negative case.  This had a huge speed increase.  However with hindsight that was purely because the few extra bytes of code shuffled subsequent functions down and changed the behaviour of the instruction case alignment.  Once I realised that was the cause of the wild differences, the other changes were mostly undone.

jkbonfield:
For evaluation purposes, I've been doing the following:

Build multiple test_view binaries for gcc and clang:

```
mkdir _pr3
for c in gcc gcc13 clang10 clang16;do for o in O2 O3;do make clean;make CC=$c CFLAGS="-g -$o" -j test/test_view;cp test/test_view _pr3/test_view.$c.$o;done;done

# Similarly for develop branch in a _dev subdirectory
```

And benchmark develop vs this PR on a specific file:

```
cd ../_dev
for i in *;do echo === $i ===;time ./$i -B /tmp/_.sam;time ../_pr3/$i -B /tmp/_.sam;done | tee _out
```

Note this is sufficient still to demonstrate there's a speed up by code duplicating in the `hts_str2int` switch statement.  I haven't explained why, but it's there and measureable!

This PR with and without that code-dup:

```
=== test_view.gcc13.O2 ===

real	0m1.012s
user	0m0.919s
sys	0m0.092s

real	0m2.178s
user	0m1.960s
sys	0m0.212s
=== test_view.gcc13.O3 ===

real	0m1.143s
user	0m0.978s
sys	0m0.164s

real	0m2.443s
user	0m2.275s
sys	0m0.161s
```

sam.c
```diff
@@ -2383,8 +2383,155 @@ int sam_hdr_change_HD(sam_hdr_t *h, const char *key, const char *val)
...
+// speed changes when the instruction blocks are split over a 32-byte
+// boundary.  To protect against this, we explicitly specify an alignment
+// for this function.  If this is insufficient, we may also wish to
+// consider alignment of blocks within this function via
+// __attribute__((optimize("align-loops=5"))) (gcc) or clang equivalents.
+// However it's not very portable.
+// Instead we break into separate functions so we can explicitly specify
+// use __attribute__((aligned(32))) instead and force consistent loop
+// alignment.
+static inline int64_t grow_B_array(bam1_t *b, int64_t n, size_t size) {
```
daviesrob, line 2397: This interface could be simplified a bit by passing `n` as a pointer (and possibly also as a `size_t`), returning 0 on success.  The callers already take a pointer to the value, so wouldn't need to dereference it.

There's no check for overflow when calculating `bytes`.  It might be possible to get this to happen on 32-bit platforms by creating a tag of the form `bb:B:I:1000000,,,,,,` with a lot of commas.  It will expand to 4x the size of the textual version, which might be enough to do the trick before you run out of virtual address space.  The best place for such a test might be in `sam_parse_Bi_vals()` and `sam_parse_BI_vals()` where it can be specialised for a 4-byte `size`.

I don't think adding `sizeof(uint32_t)` to `bytes` is necessary here.  It was done in the original code to account for the length bytes, but they've already been factored in as this point.

sam.c
```diff
@@ -2399,6 +2546,10 @@ static int sam_parse_B_vals(char type, uint32_t n, char *in, char **end,
     }
 
     // Ensure space for type + values
+    if (!n) n=7;
+    uint32_t nused = 0;
+    int64_t nalloc = n;
```
daviesrob, line 2551: Instead of taking a copy of `n`, you could just make `nalloc` the input parameter.

sam.c
```diff
@@ -2493,19 +2614,19 @@ static int sam_parse_B_vals(char type, uint32_t n, char *in, char **end,
         if (!overflow) {
             if (min < 0) {
                 if (min >= INT8_MIN && max <= INT8_MAX) {
-                    return sam_parse_B_vals('c', n, in, end, r, b);
+                    return sam_parse_B_vals('c', n, in, end, b);
                 } else if (min >= INT16_MIN && max <= INT16_MAX) {
-                    return sam_parse_B_vals('s', n, in, end, r, b);
+                    return sam_parse_B_vals('s', n, in, end, b);
```
daviesrob, line 2619: It would be more efficient to pass `nused` for the second attempt.  Increasing the data size may trigger a reallocation, but at this point we know how many slots are really needed so we can ask for the exact size required.

sam.c
```diff
@@ -2493,19 +2614,19 @@ static int sam_parse_B_vals(char type, uint32_t n, char *in, char **end,
...
                 } else if (min >= INT32_MIN && max <= INT32_MAX) {
-                    return sam_parse_B_vals('i', n, in, end, r, b);
+                    return sam_parse_B_vals('i', n, in, end, b);
                 }
             } else {
                 if (max < UINT8_MAX) {
-                    return sam_parse_B_vals('C', n, in, end, r, b);
+                    return sam_parse_B_vals('C', n, in, end, b);
```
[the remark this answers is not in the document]
jkbonfield: Indeed, an infinite recursion!  I'm surprised it didn't happen before as I haven't changed that logic, but I guess it was rescued by the attempt to fix up the type not actually working before.  Will fix it anyway, once I figure out the best route.

sam.c
jkbonfield: The sizes are different, but I assign from n to nalloc and back again so that wasn't valid anyway.  BAM can't cope with more than 32-bit, so I'll make nalloc uint32_t too.

sam.c
jkbonfield: It shouldn't matter, but actually I see both are wrong.  Our `possibly_expand_bam_data` is always growing, unlike the kstring equivalent which is asking for a size capable of holding at least 'x' bytes.  Hence whatever we pass in to the next loop when we recurse will be what we have plus a load more.  That's wrong, as we don't need any more at all.

The whole recursion thing is a bit of a mess frankly.  I need to rethink it.

sam.c
```diff
@@ -2398,87 +2564,67 @@ static int sam_parse_B_vals(char type, uint32_t n, char *in, char **end,
...
+    // The first pass through here we don't know the number of entries and
+    // nalloc == 0.  We start with a small working set and then parse the
+    // data, growing as needed.
+    //
+    // If we have a second pass through we do know the number of entries
+    // and nalloc is already known.  We have no need to expand the bam data.
+    if (!nalloc) {
+         nalloc=7;
+
+         bytes = (size_t) nalloc * (size_t) size;
```
daviesrob, line 2577: This was correct before.  If recursing, `size` may have changed from the value used in the original allocation.  If it has, you need to get some extra space.  `possibly_expand_bam_data()` will not do anything if the space already available is adequate.

You can test this using a tag like `bb:B:c,1000000,,,,,,,,,,,,,,,,,,,,,,,,,,`, making it long enough to need a reallocation.

jkbonfield: Ah yes I forgot about the growth case.  However it's still not correct before.  Eg it could grow by Nx1 (N "c" records), and then grow by a further Nx4 when recoding to int.  It probably doesn't matter much though as growing too much isn't a major concern given the number of records it affects *should* be tiny and we can only do 1 extra pass.

jkbonfield: Plus, tbh, we should probably have always just flat out rejected such data as an error.  I don't know why we're bending over backwards to fix broken SAM records really.

daviesrob: It was correct, because `b->l_data` gets reset, and `possibly_expand_bam_data()` measures the expansion requested from there.  If `size` hasn't changed, it will do nothing.

jkbonfield: Ah I see.  Cryptic!  I hadn't noticed the reset of `b->l_data` in the recursion section.  I do question the whole ethos of fixing up broken data instead of simply rejecting it as it just encourages people to write bad code given we correct their mistakes for them, but we are where we are so reverted that change back again.

sam.c
```diff
@@ -2383,9 +2383,175 @@ int sam_hdr_change_HD(sam_hdr_t *h, const char *key, const char *val)
...
+// for this function.  If this is insufficient, we may also wish to
+// consider alignment of blocks within this function via
+// __attribute__((optimize("align-loops=5"))) (gcc) or clang equivalents.
+// However it's not very portable.
+// Instead we break into separate functions so we can explicitly specify
+// use __attribute__((aligned(32))) instead and force consistent loop
+// alignment.
+static inline int64_t grow_B_array(bam1_t *b, uint32_t *n, size_t size) {
+    // Avoid overflow on 32-bit platforms, but it breaks BAM anyway
+    if (size > INT32_MAX*0.666) {
```
daviesrob, line 2399: I just noticed that this test does nothing, as `size` is never greater than 4.  It probably need to be something like this to trap the wrap-around case:

```suggestion
    if (size > 2 && (*n) > SIZE_MAX/2) {
```

Anything else will be dealt with by `possibly_expand_bam_data()`.

jkbonfield: Oops, indeed.   Confusing myself with the meaning of variables.

jkbonfield:
I squashed the review changes and tweaked the grow_B_array logic as commented above.

sam.c
jkbonfield: I changed this to `if (*n > INT32_MAX*0.666)` as the BAM limit in the B array is the 32-bit number of elements and this is independent on the size of elements.  I just used the wrong field.  `SIZE_MAX` is also the wrong type as I'm not worried about overflowing size_t, but the BAM 32-bit limitation.

The *0.666 is because we basically multiply n by 1.5 each time we grow it. Ie n += (n>>1).

daviesrob: I was more worried about the line:
```
size_t bytes = (size_t)size * (size_t)(*n>>1);
```
as if `size == 4`, `bytes` is set to `2 * (*n)`.  I'll grant that the number of elements in the B array is also a problem though.  Luckily you compare to `INT32_MAX` so it works for both.

PR #1747, opened 2024-02-13
Allow a specified output name in bgzip

### LINKED ISSUE #1726: bgzip: allow specification of output name.
It would be nice to be able to specify an output name when using bgzip>
Especially when the input comes from `stdin` (e.g output from some preprocessing of an input file).

```bash
$ echo test | bgzip /dev/stdin
[bgzip] can't create /dev/stdin.gz: Permission denied

# Now this is required.
$ echo test | bgzip -c /dev/stdin > test.gz

# This would be useful to have.
$ echo test | bgzip -o test.gz /dev/stdin
```

whitwham:
Choose a specific output name for a bgzipped file. e.g.

`bgzip file.txt -o myfile.txt.gz`

This will overwrite files without warning but will not change the existing source files and it will not conserve file modification dates.

Fixes  #1726.

### DIFF
--- a/bgzip.c
+++ b/bgzip.c
@@ -297,9 +297,15 @@ int main(int argc, char **argv)
         return 1;
     }
 
-    if (write_fname && pstdout) {
-        fprintf(stderr, "[bgzip] Cannot write to %s and stdout at the same time.\n", write_fname);
-        return 1;
+    if (write_fname) {
+        if (pstdout) {
+            fprintf(stderr, "[bgzip] Cannot write to %s and stdout at the same time.\n", write_fname);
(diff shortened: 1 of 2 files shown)


jkbonfield:
More complex than I expected indeed.

It passed all the things I could think of bar one.  "-" as a synonym for stdin/stdout works for `-o` when compressing but not when decompressing.  `bgzip -d -o - -k foo.gz` writes to a file named "-".

whitwham:
There is some special code to handle stdout already present.  I've changed it so "-o -" is treated as -c instead.

PR #1750, opened 2024-02-16
Support building with htscodecs not in a submodule.

teo-tsirpanis:
This PR changes the generation logic of `htscodecs/htscodecs/version.h` in the Makefile to run `git describe` only if htscodecs is a submodule, otherwise the version will be extracted exclusively from `htscodecs/configure.ac`.

This upstreams a patch to build htslib with [vcpkg](https://vcpkg.io).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -548,8 +548,10 @@ htscodecs/htscodecs:
 
 # Build the htscodecs/htscodecs/version.h file if necessary
 htscodecs/htscodecs/version.h: force
-	@if test -e $(srcdir)/htscodecs/.git && test -e $(srcdir)/htscodecs/configure.ac ; then \
-	  vers=`cd $(srcdir)/htscodecs && git describe --always --dirty --match 'v[0-9]\.[0-9]*'` && \
+	@if test -e $(srcdir)/htscodecs/configure.ac ; then \
+	  if test -e $(srcdir)/htscodecs/.git ; then \
+	    vers=`cd $(srcdir)/htscodecs && git describe --always --dirty --match 'v[0-9]\.[0-9]*'` ; \
+	  fi ; \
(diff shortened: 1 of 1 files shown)


daviesrob:
This is incorrect.  If the `htscodecs/.git` directory is not present, `$vers` is not set and whatever value is used will not be the correct version number for htscodecs.

Having found [your proposed patch](https://github.com/teo-tsirpanis/vcpkg/commit/6ce798163853788f31faf8fb7110006776b3db8d), it looks like you're trying to build the package in a way we do not support.  You should either:

1. Use a git checkout with htscodecs as a submodule
2. Use one of our [release tarballs](https://github.com/samtools/htslib/releases), which come with a pre-made `htscodecs/htscodecs/version.h` file
3. Make htscodecs a dependency of htslib in the package manager, build and install it separately, and tell htslib where to find it when running `./configure`.

Number 3 appears to be [recommended by vcpkg](https://learn.microsoft.com/en-gb/vcpkg/contributing/maintainer-guide#do-not-use-vendored-dependencies).

Your patch also has the wrong homepage for HTSlib, and the wrong license.  Out of interest, who would be maintaining this port?  Conda has [a bot](https://github.com/bioconda/bioconda-recipes/commits/master/recipes/htslib/meta.yaml) that keeps HTSlib up to date, but I don't see anything similar for vcpkg.

teo-tsirpanis:
Thanks for the feedback, I will try moving htscodecs to a separate port. My port is still WIP and was originally copied from another one, hence the wrong values which I fixed, thanks for noticing.

> Out of interest, who would be maintaining this port?

Vcpkg ports are mostly updated by those that need them updated. I might subscribe to new releases of htslib or htscodecs for some time after the port gets published.

PR #1752, opened 2024-02-22
 Drop duplicate FORMAT tags and fix an error with dropping large tags

pd3:
The removal of large tags introduced by https://github.com/samtools/htslib/commit/b49eea478114759efe19e718c38e7deb00614b21 and https://github.com/samtools/htslib/commit/9db565de3d89235f625de07363f1cd726c43f8f2 could not work correctly, the memmove pointers were wrong!

Resolves https://github.com/samtools/htslib/issues/1733

This is a fix of the botched pull request https://github.com/samtools/htslib/pull/1736


jkbonfield (changes requested):

jkbonfield:
I'll try rebasing and pushing back.  Some of these build errors are due to you basing this on an old htslib, which used htscodecs 1.4 instead of the current release (1.6).

PR #1755, opened 2024-02-29
added thread pool to tabix operations

### LINKED ISSUE #1735: feature request: parallelize tabix
I'd like to index a bed file (tbi) using bcftools (because of the ability to use multiple threads).

Is this a feature that you could easily support?

Many thanks,
Dan.

vasudeva8:
Fixes #1735.
Added option to provide number threads to use, disabled by default.
Uses the thread pool in operations except listing of sequence names (-l).

### DIFF
--- a/tabix.1
+++ b/tabix.1
@@ -4,7 +4,7 @@
 tabix \- Generic indexer for TAB-delimited genome position files
 .\"
 .\" Copyright (C) 2009-2011 Broad Institute.
-.\" Copyright (C) 2014, 2016, 2018, 2020, 2022 Genome Research Ltd.
+.\" Copyright (C) 2014, 2016, 2018, 2020, 2022, 2024 Genome Research Ltd.
 .\"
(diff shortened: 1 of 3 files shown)


vasudeva8:
it showed ~40% improvement in indexing and data retrieval on bed/sam files and ~15% on vcf files.

tabix.1
```diff
@@ -78,6 +78,8 @@ and BAI index formats can handle individual chromosomes up to 512 Mbp
 If your input file might contain data lines with begin or end positions
 greater than that, you will need to use a CSI index.
 
+A thread pool can be used for operations except listing sequence names.
```
whitwham, line 81: Suggestion, change "thread pool" to "threads".  "thread pool" is more of an internal name.

whitwham: Also, do the same in the other places in the man file where "thread pool" is used.

vasudeva8: Updated and rephrased.

vasudeva8:
Updated tests as well.

PR #1756, opened 2024-02-29
Allowing S3 virtual paths and S3 aliases as data source

mapk-amazon:
Related issue: https://github.com/samtools/samtools/issues/1984

The two main changes are:
*  In `s3_open_v4` an additional `hopen` is called if a redirect (status code between 300 and 399) is detected. 
*  In `redirect_endpoint_callback` the url is parsed and a new one constructed respecting the virtual path style in Amazon S3 links.

Some minor changes:
* Saving the address style of S3 web links in `s3_auth_data` to allow the use in `redirect_endpoint_callback`
* Replacing the hardcoded initial value `{0, 0, NULL}` by `KS_INITIALIZE` for various variables.


hfile_s3.c
```diff
@@ -1298,6 +1311,20 @@ static hFILE *s3_open_v4(const char *s3url, const char *mode, va_list *argsp) {
 
         if (fp == NULL) goto error;
 
+        if (http_response >= 300 && http_response < 400) {
```
daviesrob, line 1314: I think doing this for all 300 codes might be a bit over-tolerant.  It would be safer to check for the specific codes we expect for this sort of redirection.

Also, you need an `hclose_abruptly(fp);` here to avoid leakage of the old connection.

daviesrob (commented):
At a first glance, this mostly looks OK apart from the file handle leakage.  I have still to check it against the scenario described in #1984, though.

hfile_s3.c
mapk-amazon: Hi. Thank you for the review!

I specified in the latest commit the `http_response` to just check for `307`. Also, I (hopefully) fixed the leakage.

PR #1759, opened 2024-03-18
Allow tabix indexing of GAF graph alignment files

adamnovak:
This pulls in @jmonlong's implementation of Tabix indexing for [GAF files](https://github.com/lh3/gfatools/blob/master/doc/rGFA.md#the-graph-alignment-format-gaf), in graph node ID space.

This is to allow random access to bgzip-compressed GAF files in vgteam/vg using a tabix index.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -3,7 +3,7 @@ according to the terms of the following MIT/Expat license.]
 
 The MIT/Expat License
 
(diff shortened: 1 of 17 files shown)


whitwham:
Thanks for the PR.  You will need to rebase the commits to the latest version of htslib.  I cannot do it because I do not have the permissions to make changes to your commits.

PR #1763, opened 2024-03-22
Allow tabix indexing of GAF graph alignment files

adamnovak:
This pulls in @jmonlong's implementation of Tabix indexing for [GAF files](https://github.com/lh3/gfatools/blob/master/doc/rGFA.md#the-graph-alignment-format-gaf), in graph node ID space.

This is to allow random access to bgzip-compressed GAF files in vgteam/vg using a tabix index.

### DIFF
--- a/htslib/tbx.h
+++ b/htslib/tbx.h
@@ -38,6 +38,7 @@ extern "C" {
 #define TBX_GENERIC 0
 #define TBX_SAM     1
 #define TBX_VCF     2
+#define TBX_GAF     3
 #define TBX_UCSC    0x10000
 
 typedef struct tbx_conf_t {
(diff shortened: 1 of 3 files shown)


adamnovak:
@whitwham Github closed #1759 without permission when I pushed the wrong thing trying to do the rebase. This is the rebased version you requested.

htslib/tbx.h
```diff
@@ -54,6 +55,7 @@ typedef struct tbx_t {
 
 HTSLIB_EXPORT
 extern const tbx_conf_t tbx_conf_gff, tbx_conf_bed, tbx_conf_psltbl, tbx_conf_sam, tbx_conf_vcf;
```
whitwham, line 57: Looks like this lines has been left in and needs to be removed.

adamnovak:
@whitwham I've fixed the duplicate line.

whitwham:
Merged.  Thank you.

PR #1764, opened 2024-04-02
Make nibble2base faster using x86-64 pshufb instruction (SSSE3) and using dynamic dispatch.

rhpvorderman:
See https://github.com/samtools/htslib/pull/1677 for prior discussion.

I made the PR such that nibble2base is dynamically dispatched on x86-64 cpus with SSSE3 instructions. 

No build options need to be changed and nibble2base gets a nice speed up on the majority of the install base.

### DIFF
--- a/htslib/hts_defs.h
+++ b/htslib/hts_defs.h
@@ -34,6 +34,10 @@ DEALINGS IN THE SOFTWARE.  */
 #define HTS_COMPILER_HAS(attribute) __has_attribute(attribute)
 #endif
 
+#ifdef __has_builtin
+#define HTS_COMPILER_HAS_BUILTIN(function) __has_builtin(function)
+#endif
+
 #elif defined __GNUC__
 #define HTS_GCC_AT_LEAST(major, minor) \
     (__GNUC__ > (major) || (__GNUC__ == (major) && __GNUC_MINOR__ >= (minor)))
@@ -42,6 +46,10 @@ DEALINGS IN THE SOFTWARE.  */
 #ifndef HTS_COMPILER_HAS
 #define HTS_COMPILER_HAS(attribute) 0
 #endif
+#ifndef HTS_COMPILER_HAS_BUILTIN
+#define HTS_COMPILER_HAS_BUILTIN(function) 0
+#endif
+
 #ifndef HTS_GCC_AT_LEAST
 #define HTS_GCC_AT_LEAST(major, minor) 0
 #endif
@@ -116,6 +124,16 @@ DEALINGS IN THE SOFTWARE.  */
 #define HTS_FORMAT(type, idx, first) __attribute__((__format__ (type, idx, first)))
 #else
 #define HTS_FORMAT(type, idx, first)
+#endif
+
+#define HTS_COMPILER_HAS_TARGET_AND_BUILTIN_CPU_SUPPORTS \
+ ((HTS_COMPILER_HAS(target) && HTS_COMPILER_HAS_BUILTIN(__builtin_cpu_supports)) \
+ || HTS_GCC_AT_LEAST(4, 8))
+
+#if (defined(__x86_64__) || defined(_M_X64))
+#define HTS_BUILD_IS_X86_64 1
+#else
+#define HTS_BUILD_IS_X86_64 0
 #endif
 
 #if defined(_WIN32) || defined(__CYGWIN__)
--- a/sam_internal.h
+++ b/sam_internal.h
@@ -99,7 +99,8 @@ static inline void nibble2base_default(uint8_t *nib, char *seq, int len) {
         seq[i] = seq_nt16_str[bam_seqi(nib, i)];
 }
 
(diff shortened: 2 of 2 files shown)


rhpvorderman:
@jkbonfield Thank you for the feedback. As you mention, this does only significantly speed up some use cases (uBAM + long reads) and only in the lower double digit percentages. So it is not as critical to get this out ASAP. In fact, I have already implemented this in my own code that only parses uBAM. I figured it would be nice to contribute it back to htslib so that all users may enjoy it eventually. 

> I didn't test AMD either

I did. It works. This should be faster on any ssse3 enabled platform simply due to the fact that memory lookup is almost always going to be slower than register action. In this case the shufb instruction has 1 cycle of latency, so memory can never beat that. L1 cache latency is typically 3 cycles.

Thank you for all your feedback on the previous PR as well. This urged me to learn dynamic dispatch and it is great! It enabled an AVX2 routine for one of my projects.

rhpvorderman:
> I can confirm the code works and this function is considerably faster. Curiously despite using ssse3 intrinsics, that code uses different instructions if I enable -mavx2.

This is expected. It will probably use ymm registers rather than xmm registers due to these being practically the same. 
Technically AVX2 machines can use _mm256_shuffle_epi8 and this might code into something even more compact. 

> However it's clumsy with portability.

I now have done the appropriate clang checks as well as according to https://clang.llvm.org/docs/LanguageExtensions.html#builtin-cpu-supports

`target` + `__builtin_cpu_supports` is very powerful. I am glad that clang also has it.

rhpvorderman:
I had a look at the code. Using a broadcast instruction the 16 bytes can be loaded into two 128bit vectors within in a 256-bit vector. Instead of two sequences of operations, only one sequence of operations is needed. 

The code will look like cleaner at the cost of supporting less computers. In theory it should be faster, but I think the boost in application performance compared to what is there now is negligible. For now I think it is better to merge the code as is. I may try it out on some other projects I work on before I update the htslib code in a new PR if it turns out to be worth it.

jkbonfield:
Thanks.  I'll take another look at this soon.

rhpvorderman:
It is never fast enough. As long as bioinformatics pipelines take longer than my coffee breaks, the performance is still subpar as far as I am concerned. Having said that, I don't mind taking a 20ms longer coffee break. ;-)

jkbonfield:
Thanks.  I've had a look at the algorithm which is pretty simple and well documented but very effective.  The CPU detection is a bit of a change on how we've done it before, but I like the compiler auto-detection methods and automatic dispatch.  As you say they're very powerful and maybe we can use these for other bits of code in the future where custom logic will help.

Thanks.

rhpvorderman:
Thank you for not merging the initial PR, which encouraged me to look into dynamic dispatching. I have used that a few times now in other projects. It is especially useful to enable the pshufb instructions. These basically allow 16-entry lookup tables. Since a lot of code contains 256-entry or 128-entry (for ASCII text) lookup tables these can be many times be written with shuffle instructions. For example, dna only contains A,C,G,T. So rather than using some sort of lookup, there can be some quick checking using vector compare instructions with A,C,G,T. Then the resulting masks can be merged with bitwise OR, so we have an ACGT-only mask. If everything is ACGT, which we just checked with only a few instructions, we can do a bitwise AND with 0b1111, resulting in 4-bit indices that range from 0-15. Since A,C,G,T are distinct in the last few bits, and we verified our vector to be only be A,C,G,T we can then apply the shuffle. 

A bit verbose, but in terms of CPU instructions it saves a lot of work. I used that for converting DNA to twobit representation in Sequali. See the code here: https://github.com/rhpvorderman/sequali/blob/a39ceee8a5b25668e068c0812719429386673dc3/src/sequali/function_dispatch.h#L259

jmarshall:
Interesting discussion in the last couple of comments… 🤔 

It is probably fine and the assignment to `nibble2base` is probably “atomic enough” on the platforms for which the code is activated, but is there a multithreading issue here?

jkbonfield:
A very good point and generally something I've thought of too (hence the plethora of pthread_once things around).  Given the way this is implemented `nibble2base_dispatch` can do locking and doesn't need to be super efficient as it's a once-only affair by design.  

Yes it ideally does need locking, although as you say it's probably "atomic enough" to never cause an issue in practice.  On x86-64 that is, which in this case is what matters due to the ifdefs.  In general though, it is not acceptable to assume aligned 64-bit writes on a 64-bit system will be atomic.  Eg:

https://godbolt.org/z/vso1K7b64

rhpvorderman:
Ah interesting. I'd like to add that while x86-64 compilers do compile a pointer write to one store instruction, that does not guarantee that the memory is written in one go by the CPU. If it writes byte by byte, and reads byte by byte on the backend then an issue might still occur. if I understood correctly the smallest memory unit that is retrieved and written to memory is the cacheline, which is 64 bytes, so there might not be a problem at all. 

Adding mutexes won't hurt however.

PR #1768, opened 2024-04-09
Increase the input block size for bgzip.

### LINKED ISSUE #1767: bgzip performance drop from v1.16
Hi,

I stumbled across an issue when testing multiple versions of htslib/bgzip. With release of a blasting fast decompression algorithm in 2023 [https://github.com/mxmlnkn/rapidgzip](https://github.com/mxmlnkn/rapidgzip), which achives a throughput of 1.5GB/s on my machine, I observed a massive performance difference in data compression for bgzip versions <= 1.16 and >= 1.17, not attributable to the change of default compression level (from v1.16) or cold cache effects.

v1.13 - v1.15
```
rapidgzip -P 8 -kdc --import-index fastq.gzri fastq.gz | bgzip -l 6 -kc -@ 128 | pv > /dev/null
4.40GiB 0:00:12 [ 355MiB/s]
```
v1.16 (with adapted compression level)
```
rapidgzip -P 8 -kdc --import-index fastq.gzri fastq.gz | bgzip -l 5 -kc -@ 128 | pv > /dev/null
4.40GiB 0:00:12 [ 351MiB/s]
```
v1.17 - v1.19 (with adapted compression level)
```
rapidgzip -P 8 -kdc --import-index fastq.gzri fastq.gz | bgzip -l 5 -kc -@ 128 | pv > /dev/null
4.40GiB 0:00:17 [ 260MiB/s]
```

jkbonfield:
Commit e495718 changed bgzip from unix raw POSIX read() calls to hread().  Unfortunately hread gets its buffer size from stat of the input file descriptor, which can be 4kb for a pipe.  We're reading 0xff00 bytes, so this ends up being split over two reads mostly, with one or both involving additional memcpys.  This makes the buffered I/O worse performing than non-buffered.  In the most extreme cases (cat data | bgzip -l0 > /dev/null) this is a two fold slow down.

The easy solution is just to increase the buffer size to something sensible.  It's a little messy as we have to use hfile_internal.h to get hfile_set_blksize, but it works.  I'm not sure why we didn't elect to make that API more public.  Probably simply out of caution.

Fixes #1767

### DIFF
--- a/hfile.c
+++ b/hfile.c
@@ -107,9 +107,11 @@ hFILE *hfile_init(size_t struct_size, const char *mode, size_t capacity)
     hFILE *fp = (hFILE *) malloc(struct_size);
     if (fp == NULL) goto error;
 
-    if (capacity == 0) capacity = 32768;
+    const int maxcap = 128*1024;
+
+    if (capacity == 0) capacity = maxcap;
     // FIXME For now, clamp input buffer sizes so mpileup doesn't eat memory
-    if (strchr(mode, 'r') && capacity > 32768) capacity = 32768;
+    if (strchr(mode, 'r') && capacity > maxcap) capacity = maxcap;
 
     fp->buffer = (char *) malloc(capacity);
     if (fp->buffer == NULL) goto error;
@@ -629,7 +631,12 @@ static size_t blksize(int fd)
 #ifdef HAVE_STRUCT_STAT_ST_BLKSIZE
     struct stat sbuf;
     if (fstat(fd, &sbuf) != 0) return 0;
-    return sbuf.st_blksize;
+
+    // Pipes/FIFOs on linux return 4Kb here often, but it's much too small
+    // for performant I/O.
+    return S_ISFIFO(sbuf.st_mode)
+        ? 128*1024
+        : sbuf.st_blksize;
 #else
     return 0;
 #endif
--- a/hfile.c
+++ b/hfile.c
@@ -113,8 +113,14 @@ hFILE *hfile_init(size_t struct_size, const char *mode, size_t capacity)
     // FIXME For now, clamp input buffer sizes so mpileup doesn't eat memory
     if (strchr(mode, 'r') && capacity > maxcap) capacity = maxcap;
 
+#ifdef HAVE_POSIX_MEMALIGN
+    fp->buffer = NULL;
+    if (posix_memalign((void **)&fp->buffer, 256, capacity) < 0)
+        goto error;
+#else
     fp->buffer = (char *) malloc(capacity);
     if (fp->buffer == NULL) goto error;
+#endif
 
     fp->begin = fp->end = fp->buffer;
     fp->limit = &fp->buffer[capacity];
--- a/configure.ac
+++ b/configure.ac
@@ -326,7 +326,7 @@ HTS_HIDE_DYNAMIC_SYMBOLS
 
 dnl FIXME This pulls in dozens of standard header checks
 AC_FUNC_MMAP
-AC_CHECK_FUNCS([gmtime_r fsync drand48 srand48_deterministic getauxval elf_aux_info])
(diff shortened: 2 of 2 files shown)


jkbonfield:
Edited the buffer down to 256Kb instead of 1Mb as it seems to be sufficient (tested on tmpfs, fast NFS and lustre).

jkbonfield:
Hmm, it's still variable!  The effect of a bigger block size is more memcpy as we have fewer direct reads (as @daviesrob points out a readv solves that, but it doesn't fit with the backend semantics), but it also reduces the impact of many small reads due to small pipe size.

Hence some machines it's slower to have a bigger block size, and it can have weird interactions with CPU load too which I cannot explain.  Eg it's a win at -l0, but a penalty at -l1.  It needs more head scratching probably.

jkbonfield:
I retested this with different block sizes on a few systems.  This time it was aggregate wall clock time on 100 trials of the same file, so cached and a lot of reproducability.  Bgzip -l5, but our systems all have CPU frequency scaling on and the filesystems are sometimes shared so there are sources of error and randomness too.  Even so, the charts are interesting.

image

image

So 128k is enough on the Intel and 256k is enough on the AMD.  I also tried an older Intel machine, but it just flatlined the CPU and made very little difference in buffer sizes.  Trying -l1 instead changed it a little, but not significantly.

I also tested reading an actual file rather than a pipe from Lustre on the AMD system and basically it was identical performance regardless of buffer size.  This backs up my findings in the linked issue.  Looking at CPU instead of elapsed time shows a small drop at around 128KB onwards, but not huge and not worth taking into account I suspect.

TODO: re test NFS on the AMD Epyc.  It may have been having a bad period for that stage of test.  As mentioned above, it's a shared resource that I have no control over.  I also have no control over dropping cache between tests to validate that either.

TODO: try level 1 on the other hosts.

TODO: test non-threaded throughput.  Is there a CPU cost?  It seems unlikely given the real vs user benchmarks mentioned above.

However I think the 256k here is fine so we should consider this PR again.

daviesrob:
I think 128kB blocks are better, but would it be a good idea to put this in hFILE so all pipe users could benefit?  I think it should probably only be for FIFOs though to avoid potential issues with over-reading on index-based jobs.  The current [limit of 32k](https://github.com/samtools/htslib/blob/cf0e7568e155f1d8e194938ed3b97fb544bc3a25/hfile.c#L112) in `hfile_init` would need to be increased, but I doubt making it 128k instead would be a major issue.

jkbonfield:
It's not possible to distinguish between normal pipes and named pipes (FIFOs), but I don't think that matters much and it still works OK on named pipes.

The real problem is the buffer size is pathetic at 4096 according to stat.  This is actually smaller than the standard linux pipe size.  It's still internally cachine at 64kb, but we're reading it in small chunks.  I suspece that's because it's the size the OS supports for atomic writes, but we're optimise for speed here.

I've moved the change to hfile instead.

hfile.c
```diff
@@ -107,12 +107,19 @@ hFILE *hfile_init(size_t struct_size, const char *mode, size_t capacity)
...
-    if (capacity == 0) capacity = 32768;
+    const int maxcap = 128*1024;
+
+    if (capacity == 0) capacity = maxcap;
     // FIXME For now, clamp input buffer sizes so mpileup doesn't eat memory
-    if (strchr(mode, 'r') && capacity > 32768) capacity = 32768;
+    if (strchr(mode, 'r') && capacity > maxcap) capacity = maxcap;
 
+#ifdef HAVE_POSIX_MEMALIGN
+    if (posix_memalign((void **)&fp->buffer, 256, capacity) < 0)
```
daviesrob, line 117: One small issue.  It seems `posix_memalign()` doesn't set the buffer pointer if it fails, so it needs to be set to `NULL` on entry to avoid problems when cleaning up.

See the example on the [Open Group posix_memalign() man page](https://pubs.opengroup.org/onlinepubs/9799919799/functions/posix_memalign.html).

PR #1775, opened 2024-04-22
fix fuzz integer overflow in cram encoder.

jkbonfield:
Input files with very long CIGAR strings and consensus generated embedded reference can lead to exceptionally long CRAM blocks which overflow the check for large size fluctuations (to trigger new compression metric assessments).

Reformulated the expression to avoid scaling up values.

Credit to OSS-Fuzz
Fixes oss-fuzz 68225

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -1984,11 +1984,15 @@ int cram_compress_block2(cram_fd *fd, cram_slice *s,
         // We also get large fluctuations based on genome coordinate for
         // e.g. SA:Z and SC series, but we consider the typical scale of
         // delta between blocks and use this to look for abnormality.
+
+        // Equivalent to (but minus possible integer overflow)
+        //   (b->uncomp_size + 1000)/4 > metrics->input_avg_sz+1000 ||
+        //    b->uncomp_size + 1000    < (metrics->input_avg_sz+1000)/4)
         if (metrics->input_avg_sz &&
-            (b->uncomp_size + 1000 > 4*(metrics->input_avg_sz+1000) ||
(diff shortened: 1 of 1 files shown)


cram/cram_io.c
```diff
@@ -1985,10 +1985,10 @@ int cram_compress_block2(cram_fd *fd, cram_slice *s,
         // e.g. SA:Z and SC series, but we consider the typical scale of
         // delta between blocks and use this to look for abnormality.
         if (metrics->input_avg_sz &&
-            (b->uncomp_size + 1000 > 4*(metrics->input_avg_sz+1000) ||
+            ((b->uncomp_size + 1000)/4 > metrics->input_avg_sz+1000 ||
              b->uncomp_size + 1000 < (metrics->input_avg_sz+1000)/4) &&
```
daviesrob, line 1989: Strictly, I think this can still overflow due to the `+ 1000` on both sides.  However, by expanding out the brackets and subtracting 1000 from both sides it's possible to get something that behaves itself:
```suggestion
            (b->uncomp_size/4 - 750 > metrics->input_avg_sz ||
             b->uncomp_size < metrics->input_avg_sz/4 - 750) &&
```
Note that this relies on both `b->uncomp_size` and `metrics->input_avg_sz` being signed values.  As the next part also has this requirement I don't think it's a major problem, but we might want to bear it in mind if we ever wanted to change either to be unsigned (e.g. `size_t`).

jkbonfield: I've worked through the expansion and I agree with the same thing.  It's a bit obscure, so I'll add a comment.

daviesrob:
Rebased to pick up the recent test fixes...

PR #1779, opened 2024-04-30
Extend annot-tsv, adding several new options

pd3:
-d, --delim
    .. add support for other delimiters, not just tab

-h, --headers
    .. allow header lines in different formats. This is motivated
       by the common use of files with the first row containing
       column names but without the leading # character.

-I, --no-header-idx
    .. suppress column index numbers in the printed header. When
       given twice, drops the entire header

--help
    .. changed from `-h, --help` to just `--help`. Note this option
       was not advertised in the man page nor the usage page,
       therefore it should be okay to reuse it for `-h, --headers`.


annot-tsv.c
```diff
@@ -315,47 +316,90 @@ void free_payload(void *payload)
...
         }
+        if ( nth_row>0 )
+        {
+            // N>1 .. regardless of this being a comment or data line, read until Nth line
+            if ( ++irow < nth_row ) continue;
+            break;
+        }
+        // N<0 .. keep abs(N) comment lines in a sliding buffer
+        if ( dat->line.s[0]!='#' ) break;   // data line
+        if ( nbuf == -nth_row )
```
whitwham, line 365: With a negative in the nth_row the first line of the actual data is missing in the output and memory is leaked from the strdup.

whitwham:
Thanks for the memory leak change but the negative header values are still not working.

So (from the tests):
`annot-tsv -s src.11.txt -t dst.11.txt -c chr1,beg1,end1:chr,beg,end -f smpl1:src_smpl -h 2:2 -II` 

gives:

> 1	10	20	A	A
1	30	40	A	B
1	50	60	A	.

Which is correct, but using the same files:
`annot-tsv -s src.11.txt -t dst.11.txt -c chr1,beg1,end1:chr,beg,end -f smpl1:src_smpl -h 2:-1 -II`

gives:

> 1	30	40	A	B
1	50	60	A	.

So unless I'm using it wrong the first line is going missing.

pd3:
Duh, it should work now, sorry

PR #1792, opened 2024-06-20
Replace assert(0) with abort()

### LINKED ISSUE #1791: assert(0) is bad practice
if you compile with NDEBUG asserts go away 
https://github.com/samtools/htslib/blob/5f205334cd04bcdac74b1b6d46f88d57a05eaf1c/synced_bcf_reader.c#L545
suggest replacing with abort();

mp15:
Replaces assert(0) with abort(). This ensures no behaviour change except the check is now not skipped when NDEBUG is defined.
Fixes #1791

### DIFF
--- a/synced_bcf_reader.c
+++ b/synced_bcf_reader.c
@@ -25,6 +25,7 @@ DEALINGS IN THE SOFTWARE.  */
 #define HTS_BUILDING_LIBRARY // Enables HTSLIB_EXPORT, see htslib/hts_defs.h
 #include <config.h>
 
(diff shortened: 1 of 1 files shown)


jkbonfield:
I see it calls `exit(1)` earlier.  It ought to return rather than exit, but I haven't looked into what uses it and whether any code checks for errors.

pd3:
This would benefit from a review. It might be better to propagate some of these errors to the caller rather than exit. However, this is certainly better than what was there before. Thank you

PR #1796, opened 2024-06-26
Windows builds via GitHub Actions

jkbonfield:
The AppVeyor builds have become slower to launch and they're also a bit slow on execution.  GitHub Actions runs in about half the time.

Note this needs https://github.com/samtools/htscodecs/pull/123 merging first, but for purposes of testing I just incorporated it in this PR via an extra commit to change the submodule commit.  If the htscodecs PR gets modified in any way, changing the hash, then I'll update this appropriately.

Also, if you wish to see these tests in-situ here, we could copy this PR branch to samtools/htslib from jkbonfield/htslib as otherwise it won't enable the workflow.  (As I did for htscodecs)

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -1,0 +1,40 @@
+name: Windows/MinGW-W64 CI
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: windows-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        submodules: true
+    - name: Set up MSYS2 MinGW-W64
+      uses: msys2/setup-msys2@v2
+      with:
+        msystem: mingw64
+        update: false
+        install: >-
+          git
+          zlib-devel
+          libbz2-devel
+          liblzma-devel
+          mingw-w64-x86_64-toolchain
+          mingw-w64-x86_64-autotools
+          mingw-w64-x86_64-tools-git
+          mingw-w64-x86_64-libdeflate
+    - name: Compile htslib
+      shell: msys2 {0}
+      run: |
+        export PATH=/mingw64/bin:$PATH
+        export MSYSTEM=MINGW64
+        autoreconf -i
+        ./configure
+        make -j6
+    - name: Check Htslib
+      shell: msys2 {0}
+      run: |
+        export PATH=/mingw64/bin:$PATH
+        export MSYSTEM=MINGW64
+        make test-shlib-exports && make check
+
--- a/.gitattributes
+++ b/.gitattributes
@@ -24,3 +24,14 @@ test/index_dos.sam -text
 # Remove the text attribute from various faidx test files
 test/faidx/faidx*.fa* -text
 test/faidx/fastqs*.fq* -text
+test/fastq/*.fa -text
+test/fastq/*.fq -text
+*.tst -text
+*.out -text
+*.crai    -text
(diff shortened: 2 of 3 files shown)


jmarshall:
I was initially going to suggest a `git config core.autocrlf-core.eol-etc` step before the checkout step (on the basis that it would be better to affect only the action runner rather than all Windows users as adding to _.gitattributes_ does), but after reading through actions/checkout#135 I have changed my mind. Fixing this for the test files for all Windows users is actually the right thing to do!

That's a long thread, but https://github.com/actions/checkout/issues/135#issuecomment-613342603 makes a very good point that just doing a blanket `** -text` that would also affect _*.c_ files would be the **wrong** thing to do. So I'm glad to see this PR just does `-text` more specifically for a bunch of test file extensions.

jkbonfield:
I confess for the Samtools one I gave up worrying about every little extension and just went for `test/**`, but I do accept there are some C files in there.  I wonder if we can then do `text/*.c +text` or similar to reenable it as an exception to the blanket rule?  I doubt it matters though tbh as I'm not convinced we'll be getting many, if any, PRs from windows developers, and even if so the code tools typically all support nl anyway.

.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,40 @@
...
+        export MSYSTEM=MINGW64
+        autoreconf -i
+        ./configure
+        make -j6
+    - name: Check Htslib
+      shell: msys2 {0}
+      run: |
+        export PATH=/mingw64/bin:$PATH
+        export MSYSTEM=MINGW64
+        make check
```
daviesrob, line 39: Compared with the old script, we've lost `make test-shlib-exports`.  Assuming it does actually work on Windows, it's probably useful to have.

jkbonfield: Good spot.  I'm assuming it does work as it was in the appveyor config.  I'll put it back and see what happens.

jkbonfield:
Hmm, disappointingly adding back the test-shlib-exports now fails:

```
Checking shared library exports
kh_begin
kh_clear_##name
kh_del_##name
kh_destroy_##name
kh_get_##name
kh_init_##name
kh_put_##name
kh_resize_##name
ks_destroy
ks_getuntil2
ks_init
kseq_destroy
kseq_init
kseq_read
Error: Found unexported symbols (listed above)
```

I initially thought they're all erroneous things from the templates-in-C style of klib, but `kseq_read` is more normal, although it's still done as a macro with a `SCOPE` define.  I'll need to dig deeper into what it's attempting to do and why we're no longer passing this script.

jkbonfield:
Fixing the header checking perl script to cope with windows line endings worked.  So I've patched that and done some rebasing and squashing of commits.  Assuming it now passes this is ready for re-review.

daviesrob:
As [foretold](https://github.com/samtools/htslib/pull/1796#issue-2374817629), the htscodecs submodule commit needs to be adjusted.  It also might be good to leave the AppVeyor configuration in place for now, for a less messy transition.  Once the couple of recent pull requests that straddle the change have been committed, (#1797, #1799) we can turn it off in the settings and push a commit to remove the configuration.

jkbonfield:
Ah yes I meant to update the htscodecs submodule and forgot.

I was assuming we'd just turn off the appveyor, rerun tests, and if it passed merge it.  But we can do it in two separate PRs if really needed.  It's minor fluff tbh.

Issue #1798, opened 2024-06-28
Using `bgzf_read_small()` breaks subsequent `bgzf_useek()` calls


jmarshall:
PR #1772 added a simplified inline version of `bgzf_read()` that uses inline code when the request can be satisfied directly from the buffer, otherwise punts to the real `bgzf_read()`. Very laudable. However I encountered seek problems when using it.

I rederived the inline function by starting with the code of the full `bgzf_read()`, assuming the invariant that `length < fp‑>block_length - fp‑>block_offset`, and simplifying the code accordingly. I ended up with a function that is fairly similar to `bgzf_read_small()` as added to _htslib/bgzf.h_ but with some additions:

```diff
static inline ssize_t bgzf_read_small(BGZF *fp, void *data, size_t length) {
     // A block length of 0 implies current block isn't loaded (see
     // bgzf_seek_common).  That gives negative available so careful on types
     if ((ssize_t)length < fp->block_length - fp->block_offset) {
         // Short cut the common and easy mode
         memcpy((uint8_t *)data,
                (uint8_t *)fp->uncompressed_block + fp->block_offset,
                length);
         fp->block_offset += length;
+
+        // For raw gzip streams this avoids short reads.
+        if (fp->block_offset == fp->block_length) {
+            fp->block_address = bgzf_htell(fp);
+            fp->block_offset = fp->block_length = 0;
+        }
+
+        fp->uncompressed_address += length;
+
         return length;
     } else {
         return bgzf_read(fp, data, length);
     }
 }
```

Because `bgzf_read_small()` does not currently update `fp‑>uncompressed_address`, subsequent `bgzf_useek()` calls may jump to the wrong location. And probably other functions use `fp‑>uncompressed_address` too and are affected.

The `if` block looks harder to deal with, because it calls `bgzf_htell()` which is private within _bgzf.c_. However it turns out that the invariant implies that this `if` will never be true, so in fact I should have simplified it away to nothing too. Phew.

So `bgzf_read_small()` just needs ` fp->uncompressed_address += length;` added to it to make it equivalent to `bgzf_read()`.

(I have not analysed `bgzf_write_small()` to see if it has any similar infelicities.)

daviesrob:
I'd checked that the `fp->block_address` update wasn't needed while going over the PR, but evidently missed `fp->uncompressed_address`.  What did you do to spot this?  As the test harness currently passes with the incorrect change, it could do with extending to cover this problem.

jkbonfield:
[...]

PR #1799, opened 2024-07-01
Speed up fai_retrieve() by reading whole lines at once

jmarshall:
While looking around for functions that could benefit from SIMD techniques a few weeks ago, I looked at `fai_retrieve()`'s search for (uncommon) non-graphic characters. However I soon realised that it already knows exactly where these characters are, so doesn't need to search or check for them at all!

Because `fai_retrieve()` is given only well-formatted input containing lines of the same length, it already knows exactly where the base and non-graphic characters are. So in general the interval to be read will look like

```
    ......ATGCAT    (read last six bases and line terminator)
    ATGCATGCATGC    (read complete line including line terminator)
    ATGCATGCATGC    (read complete line including line terminator)
    ATGC........    (read first four base characters)
```

and can be read a line at a time instead of a character at a time, with special handling for the partial first and last lines, and discarding the terminator characters at the end of each line read.

Timings in seconds for queries for regions of various lengths, with the iteration counts chosen so that **develop** took about a minute on each “repeatedly querying the same locus” test:

```
 len 6     len 41    len 100   len 400   len 4000
(on one FASTA line)  (spread over two / several / lots of FASTA lines)

Uncompressed FASTA, reading the same region repeatedly

 63.092    56.103    64.216    57.681    57.825    develop branch
 47.779    39.477    46.256    41.923    41.989    PR #1797: inline isgraph and bgzf_getc
 46.992    12.043     7.168     2.589     1.690    This PR using bgzf_read()
 42.097    10.667     5.844     2.054     1.070    This PR using inlined bgzf_read_small()

Uncompressed FASTA, jumping around to actually different regions

296.450   119.696    96.292    68.044    57.622    develop branch
283.962    99.184    76.719    52.291    41.920    PR #1797: inline isgraph and bgzf_getc
280.806    73.992    39.230    12.152     1.825    This PR using bgzf_read()
276.599    72.327    38.039    11.542     1.197    This PR using inlined bgzf_read_small()
```

~~Opened as a draft PR because this will cause test failures until #1798 is fixed.~~

### DIFF
--- a/faidx.c
+++ b/faidx.c
@@ -715,9 +715,8 @@ faidx_t *fai_load_format(const char *fn, enum fai_format_options format) {
 
 static char *fai_retrieve(const faidx_t *fai, const faidx1_t *val,
                           uint64_t offset, hts_pos_t beg, hts_pos_t end, hts_pos_t *len) {
-    char *s;
-    size_t l;
-    int c = 0;
+    char *buffer, *s;
+    ssize_t nread, remaining, firstline_len, firstline_blen;
     int ret;
 
     if ((uint64_t) end - (uint64_t) beg >= SIZE_MAX - 2) {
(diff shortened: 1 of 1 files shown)


jkbonfield:
I thought about this when doing my trivial edit before, but for some reason was misremembering my code.  I used to use code that dealt with the EMBL sequence formats which had blocks of text with spacing in them (eg a white space every 10 chars for human readability when counting) and so discounted whole line reads as too fiddly.

However it doesn't take much thinking to realise this isn't compatible with the fai index format and it simply can't work for random access (although it *is* incorrectly parseable with `samtools faidx` and yields incorrect results, which is probably a minor bug, albeit "garbage in, garbage out").

Thanks for the update.  It's a significant speed gain.  I don' t know how much this actually impacts on real world pipelines, but I'm sure there are a few pathological cases out there.

jmarshall:
> I used to use code that dealt with the EMBL sequence formats which had blocks of text with spacing in them (eg a white space every 10 chars for human readability when counting) and so discounted whole line reads as too fiddly.

Hilariously that could work with faidx if the file has Unix line terminators; e.g.

```
>funky
ATGCATGCAT AATTGGCCAT AAAATTTTGG GGCCCCATGC
ATGCGTCAAT AATT
```

If this was indexed as blen 10 len 11, faidx would retrieve the right bases without trouble! I thought this would Just Work, but unfortunately faidx indexing looks for \\n and \\r specifically rather than `!isgraph`…

PR #1808, opened 2024-07-16
Adjust GitHub actions windows build package list

daviesrob:
Use the mingw-w64-x86_64 flavour of all packages.  Add curl, so hfile_libcurl etc. build.  Remove the mingw git package as git for windows is already present.  Add the location of git for windows to `$PATH` (note this must be at the end, else the wrong shell gets picked up when running tests leading to backslash conversion issues).

Switch badge on README.md from AppVeyor to GitHub.

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -15,18 +15,18 @@ jobs:
         msystem: mingw64
         update: false
         install: >-
-          git
-          zlib-devel
-          libbz2-devel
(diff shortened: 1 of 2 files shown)


jkbonfield:
Adding curl is important, thanks.  I don't get the git thing though.  Adding to path only changes builds, not clones, and the only bit of the build process using git is probably a `git describe` for the version.  Still, it's not going to harm.

PR #1807, opened 2024-07-16
Miscellaneous minor build infrastructure and code fixes

jmarshall:
Accumulated extremely squashable minor build infrastructure and code fixes:

* Ignore and clean `test/*/FAIL*` for six subdirectories — these files can appear in base_mods, fastq, mpileup, and sam_filter as well as faidx and tabix.

* Fix _MM-explicit.sam_ comment header to use `@CO\t` as per other comment headers.

*    Remove extraneous inclusion and add missing dependency.

*    Remove last traces of previously deleted `bgzf_idx_amend_last()` — as noted in #1722, this function was removed in PR #1672.

*    Use `isspace_c()` et al in _annot-tsv.c_.

PR #1809, opened 2024-07-17
added hts thread pool header to tabix build dependency

vasudeva8:
Added build dependency on thread pool header


daviesrob:
Closing, as the same update appeared in #1807

PR #1812, opened 2024-07-30
Stricter limit on POS/MPOS/TLEN in sam_parse1()

daviesrob:
Help avoid overflow on arithmetic involving POS, MPOS and TLEN by limiting values in the SAM parser to fit in 62 bits (or 63 for TLEN as it's signed).  The new limit is still massively bigger than any known reference so it should not cause any problems in practice.

The actual problem reported was a signed value overflow [in `process_one_read()`](https://github.com/samtools/htslib/blob/fbe5ff6c/cram/cram_encode.c#L3851).  Getting this calculation to work correctly for the entire range of `int64_t` would be complicated.  It's easier just to ensure the inputs are within a range where overflow cannot happen.

Credit to OSS-Fuzz
Fixes oss-fuzz 68750

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -2947,7 +2947,7 @@ int sam_parse1(kstring_t *s, sam_hdr_t *h, bam1_t *b)
     } else c->tid = -1;
 
     // pos
-    c->pos = hts_str2uint(p, &p, 63, &overflow) - 1;
+    c->pos = hts_str2uint(p, &p, 62, &overflow) - 1;
     if (*p++ != '\t') goto err_ret;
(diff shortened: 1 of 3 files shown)


whitwham:
Does `v->pos = hts_str2uint(p, &p, 63, &overflow);` in vcf.c at line 3706 also need changing?

daviesrob:
There's probably less urgency on that one, as I think there's fewer troublesome calculations done on positions in VCF files.  But the change doesn't appear to cause any problems in HTSlib or BCFtools, so I guess it's worth doing.

PR #1828, opened 2024-08-22
Update htscodecs and add more warnings

jkbonfield:
The extra warnings do spot a few buglets, such as in hts_parse_format().

Also resurrect `-Werror` which was absent from some builds.  (It's still absent from the configure-less macos builds, but the recommended configure route has it which is sufficient.)

### DIFF
--- a/test/test-bcf-sr.c
+++ b/test/test-bcf-sr.c
@@ -179,6 +179,7 @@ int main(int argc, char *argv[])
                 break;
             case 'h':
                 usage(EXIT_SUCCESS);
+                // fall-through
             default: usage(EXIT_FAILURE);
         }
     }
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -72,6 +72,7 @@ gcc_task:
        DO_MAINTAINER_CHECKS: yes
        DO_UNTRACKED_FILE_CHECK: yes
        USE_CONFIG: no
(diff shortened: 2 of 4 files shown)


test/test-bcf-sr.c
```diff
@@ -179,6 +179,7 @@ int main(int argc, char *argv[])
                 break;
             case 'h':
                 usage(EXIT_SUCCESS);
+                // fall-through
```
jmarshall, line 182: This isn't really a fall-through situation. `usage()` (and `error()`) wants `HTS_NORETURN`.

daviesrob:
I've pushed a few adjustments:

* Fixed the only instance of an empty body, so `-Wno-empty-body` can be removed.
* Fixed a signed compare in `htslib/bgzf.h` which triggered a warning from `g++`.
* Adjust setting of `CFLAGS` in `.cirrus.yml` so they're applied more consistently.
* Add `HTS_NORETURN` to `test-bcf-sr` usage and error functions, as suggested by @jmarshall 
* Remove `-lz` from the link lines of some test programs, as it was already in `$(LIBS)` and the duplication annoyed the MacOS linker.

Additionally, the MacOS builds were getting a banner saying "Only ghcr.io/cirruslabs/macos-runner:sonoma is allowed. Automatically upgraded." so the MacOS image has been changed to the Sonoma one.

PR #1831, opened 2024-09-05
Fix gawk-ism in Makefile

daviesrob:
The awk script used to extract htscodec's version string (if `git describe` fails or returns a hash) used a three-argument form of `match`, which is a GNU extension.  Rewrite for better compatibility with other awk implementations, notably `mawk`, which supplies awk in some Cirrus-CI images.

This doesn't appear to show up on HTSlib's tests, which get a suitable version number from `git describe`, but [does on SAMtools'](https://cirrus-ci.com/task/6105782926704640?logs=htslib_compile#L96) where HTSlib (and so htscodecs) is cloned separately.

Also fix the cirrus-ci badge, which linked to the wrong location.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -573,7 +573,7 @@ htscodecs/htscodecs/version.h: force
 	  vers=`cd $(srcdir)/htscodecs && git describe --always --dirty --match 'v[0-9]\.[0-9]*'` && \
 	  case "$$vers" in \
 	    v*) vers=$${vers#v} ;; \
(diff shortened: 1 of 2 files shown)


jkbonfield:
The awkism change works, and is simpler.

However in review I've noticed the regexp has always been wrong as it doesn't permit more than one digit.  At the risk of falsely accepting `12.34..56` we could just replace the regexp with `/[0-9.]+/` which is much easier to understand.

daviesrob:
Multiple digits should now be accepted.

PR #1834, opened 2024-09-10
Add support for elf_aux_info() on OpenBSD

brad0:


jmarshall:
It appears that `elf_aux_info()` has existed on OpenBSD for [about two months](https://cvsweb.openbsd.org/src/lib/libc/gen/elf_aux_info.c) and is not yet present in any OpenBSD release. Under the circumstances, it would be premature to add this without a configure test or other version check.

brad0:
Something like this is what I use most elsewhere.

jkbonfield:
Using `AC_CHECK_FUNCS` seems like a good approach, although I question why we still need the `__linux__` ifdef if we're detecting it via autoconf.  Ie why remove `defined __FreeBSD__` but not `defined __linux__`?

I guess this is to support non-configure build environments (but personally I think having two build configurations just leads to more problems than it solves and weird code like this where we have a mixture of ways of detecting things).  If that is the logic, then we should be keeping all the existing system ifdefs and *adding* the `HAVE_ELF_AUX_INFO` bit and not replacing FreeBSD with it.

brad0:
> Using `AC_CHECK_FUNCS` seems like a good approach, although I question why we still need the `__linux__` ifdef if we're detecting it via autoconf. Ie why remove `defined __FreeBSD__` but not `defined __linux__`?
> 
> I guess this is to support non-configure build environments (but personally I think having two build configurations just leads to more problems than it solves and weird code like this where we have a mixture of ways of detecting things). If that is the logic, then we should be keeping all the existing system ifdefs and _adding_ the `HAVE_ELF_AUX_INFO` bit and not replacing FreeBSD with it.

Well it would be better to have autoconf detect getauxval as well. I can do so.

jmarshall:
We previously discussed this in samtools/htscodecs#63 but punted on detecting during configure instead. It's easier here in htslib because the htscodecs build system vs htslib build system issue (mentioned in https://github.com/samtools/htscodecs/pull/63#issuecomment-1278827624) does not arise. But it would be unfortunate for people doing non-configure builds if SIMD were now deactivated for them.

Personally from reading the tea leaves of the maintainers' activity, I would have guessed that now is not the time for this sort of change… :smile:

brad0:
I made another PR.

https://github.com/samtools/htslib/pull/1835

jmarshall:
If it was considered sufficiently important (for the benefit of bioinformatics tools users on future OpenBSD) to activate this Neon code on OpenBSD using a system call that does not exist on current OpenBSD deployments, the following would suffice:

```diff
 #if defined __linux__ || defined __FreeBSD__
 #include <sys/auxv.h>
+#elif defined __OpenBSD__
+#include <sys/param.h>
+#if OpenBSD >= 202409
+#include <sys/auxv.h>
+#endif
 #elif defined __APPLE__
…
-#elif defined __FreeBSD__ && defined __arm__ && defined HWCAP_NEON
+#elif (defined __FreeBSD__ || (defined __OpenBSD__ && OpenBSD >= 202409)) && defined __arm__ && defined HWCAP_NEON
… etc …
```

while being a smaller change for HTSlib — one that, in particular, would not affect other platforms.

If activating SIMD code on future OpenBSD installations is your goal, you should really focus on htscodecs, where activating SIMD implementations would make a larger difference to overall run time.

daviesrob:
Closing as replaced by #1835

PR #1835, opened 2024-09-10
Detect the presence of getauxval and elf_aux_info

brad0:


daviesrob:
I've added a commit which should ensure `configure`-less builds (which are supported by HTSlib for reasons of History) still enable the NEON code as before, resolving the issue raised in [ths comment](https://github.com/samtools/htslib/pull/1834#issuecomment-2340057186) by @jmarshall.

I've tested this both with and without `configure` on arm64 Linux, FreeBSD, NetBSD, and OpenBSD (7.5 and 7.6).  In all cases it works as expected (i.e. the NEON code is built, apart from on OpenBSD 7.5, which lacks the `elf_aux_info()` interface).

brad0:
@daviesrob Thanks. That looks reasonable.

PR #1839, opened 2024-09-18
Make --write-index work for concat with vcf

whitwham:
Add a new function to work like _vcf_write_line_ but with index handling code.

To make _vcf_write_line_ work to produce an index I would of had to change its arguments as more information was needed.  Instead I added a new function.

In conjunction with the BCFtools change (samtools/bcftools#2283) this should fix samtools/bcftools#2246.

### DIFF
--- a/htslib/vcf.h
+++ b/htslib/vcf.h
@@ -516,6 +516,9 @@ set to one of BCF_ERR* codes and must be checked before calling bcf_write().
     */
     HTSLIB_EXPORT
     int vcf_write_line(htsFile *fp, kstring_t *line);
+    
+    HTSLIB_EXPORT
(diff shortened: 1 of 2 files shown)


htslib/vcf.h
```diff
@@ -517,6 +517,25 @@ set to one of BCF_ERR* codes and must be checked before calling bcf_write().
...
+
+
+    /// Write a line to a VCF file for indexing
+    /** @param line   Line to write
+        @param fp     File to write it to
+        @param h      The header for the vcf file
+        @param chr_id Chromosome id
+        @param name   Chromosome name
+        @param beg    Beginning position
+        @param end    End position
```
jkbonfield, line 529: The parameters do not match the function declaration.  It has pos & len while this has beg & end.  Also we should probably list them in the same order, so move "line" down after the header.

whitwham:
I'm going to go with a different option so this PR is no longer needed.

PR #1843, opened 2024-09-26
Add instructions to INSTALL for FreeBSD, NetBSD and OpenBSD.

daviesrob:


whitwham:
That all worked nicely.  For FreeBSD the package seemed to be called perl5 rather than perl.  Since Perl was a dependency of another package it installed it anyway.

PR #1844, opened 2024-10-02
Add support for matching VCF lines by ID

pd3:
See https://github.com/samtools/bcftools/issues/1739

This is accompanied by https://github.com/samtools/bcftools/tree/bcf-pair-id

### DIFF
--- a/bcf_sr_sort.c
+++ b/bcf_sr_sort.c
@@ -1,5 +1,5 @@
 /*
-    Copyright (C) 2017-2021 Genome Research Ltd.
+    Copyright (C) 2017-2024 Genome Research Ltd.
 
     Author: Petr Danecek <pd3@sanger.ac.uk>
 
@@ -32,6 +32,7 @@
 #include "htslib/khash_str2int.h"
 #include "htslib/kbitset.h"
 
+// Variant types and pair-wise compatibility of their combinations, see bcf_sr_init_scores()
 #define SR_REF   1
 #define SR_SNP   2
 #define SR_INDEL 4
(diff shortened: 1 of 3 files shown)


bcf_sr_sort.c
```diff
@@ -384,6 +385,11 @@ static int bcf_sr_sort_set(bcf_srs_t *readers, sr_sort_t *srt, const char *chr,
             srt->off[srt->noff++] = srt->str.l;
             size_t beg  = srt->str.l;
             int end_pos = -1;
+            if ( srt->pair & BCF_SR_PAIR_ID )
```
daviesrob, line 388: This looks OK, and seems to work as advertised.

bcf_sr_sort.c
```diff
@@ -417,7 +423,8 @@ static int bcf_sr_sort_set(bcf_srs_t *readers, sr_sort_t *srt, const char *chr,
             }
 
             // Create new variant or attach to existing one. But careful, there can be duplicate
-            // records with the same POS,REF,ALT (e.g. in dbSNP-b142)
+            // records with the same POS,REF,ALT (e.g. in dbSNP-b142). In such case, append a
+            // numeric index (var_idx)
```
daviesrob, line 427: I think this could do with a bit more explanation, as it took a while to work out what's going on here.  How about:
```suggestion
            // records with the same POS,REF,ALT (e.g. in dbSNP-b142). In such case, use a
            // hash table (srt->var_str2int) and a counter (var_idx) to ensure they are
            // treated as separate variants, while still allowing them to be matched
            // between readers.
```

pd3 (commented):
@daviesrob I included your suggested commit, thank you. I am happy for this to be merged now.

PR #1849, opened 2024-10-14
Fix issue created in PR#1848

jkbonfield:
We shouldn't be attempting to create embedded reference sequences for CRAM containers with reads mapped to chr -1 (ie unmapped).  We don't permit embed_ref in multi-ref mode and it's nonsensical for entirely unmapped data.

The only real fix needed here is `&& c->ref_id >= 0` just before calling cram_generate_reference(), but checking elsewhere can sidestep other potential issues and we have safety in checking in more than one place.

Credit to OSS_Fuzz
Fixes oss-fuzz issue 372547397

Note the following test triggered it, but it's only for that most recent commit and 1.21 was fine so this is a brief regression.

```
@SQ	SN:foo	LN:999
x	4	*	1	0	*	*	0	0	ACGT	PQRS
```

### DIFF
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -1842,7 +1842,7 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
     // Don't try embed ref if we repeatedly fail
     pthread_mutex_lock(&fd->ref_lock);
     int failed_embed = (fd->no_ref_counter >= 5); // maximum 5 tries
-    if (!failed_embed && c->embed_ref == -2) {
+    if (!failed_embed && c->embed_ref == -2 && c->ref_id >= 0) {
         hts_log_warning("Retrying embed_ref=2 mode for #%d/5", fd->no_ref_counter);
         fd->no_ref = c->no_ref = 0;
         fd->embed_ref = c->embed_ref = 2;
@@ -1921,6 +1921,12 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
                 // Do not confuse with fd->ref_free which is a pointer to a
                 // reference string to free.
                 c->ref_free = 1;
+            } else {
+                // Double check for broken input.  We shouldn't have
+                // embedded references enabled for unmapped data, but our
+                // data could be broken.
+                embed_ref = 0;
+                no_ref = c->no_ref = 1;
             }
         }
         c->ref_seq_id = c->ref_id;
@@ -1967,7 +1973,7 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
 
         // Embed consensus / MD-generated ref
         if (embed_ref == 2) {
-            if (cram_generate_reference(c, s, r1) < 0) {
+            if (c->ref_id < 0 || cram_generate_reference(c, s, r1) < 0) {
                 // Should this be a permanent thing via fd->no_ref?
                 // Doing so means we cannot easily switch back again should
                 // things fix themselves later on.  This is likely not a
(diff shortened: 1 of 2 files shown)


cram/cram_encode.c
```diff
@@ -1921,6 +1921,12 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
                 // Do not confuse with fd->ref_free which is a pointer to a
                 // reference string to free.
                 c->ref_free = 1;
+            } else {
+                // Double check for broken input.  We shouldn't have
+                // embedded references enabled for unmapped data, but our
+                // data could be broken.
+                embed_ref = 0;
+                no_ref = 1;
```
daviesrob, line 1929: Should this also set `c->no_ref` [as done elsewhere](https://github.com/samtools/htslib/blob/b3a4c0a62240faf437105d7604897cf7046e3474/cram/cram_encode.c#L1948)?  `no_ref` is reset from `c->no_ref` later.  `c->no_ref` is also used in [`cram_encode_compression_header()`](https://github.com/samtools/htslib/blob/b3a4c0a62240faf437105d7604897cf7046e3474/cram/cram_encode.c#L90), which is called from this function.

cram/cram_encode.c
```diff
@@ -1966,7 +1972,7 @@ int cram_encode_container(cram_fd *fd, cram_container *c) {
         kstring_t MD = {0};
 
         // Embed consensus / MD-generated ref
-        if (embed_ref == 2) {
+        if (embed_ref == 2 && c->ref_id >= 0) {
```
daviesrob, line 1975: I don't think `c->ref_id >= 0` can now happen due to the other changes, but if it did, could it cause problems elsewhere?  It looks like you might end up calling `process_one_read()` with `embed_ref = 2` but `c->ref = NULL`, which I'm not sure it can handle...  It might be safer to treat `c->ref_id < 0` like a `cram_generate_reference()` failure, so you switch firmly into non-ref mode.

daviesrob (commented):
I've managed to come up with a better input to reproduce the original problem:
```
@SQ	SN:1	LN:100
I	4	*	1	0	*	*	0	0	CCTAGCCCTAACCCTAACCCTAACCCTAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAA	#############################@B?8B?BA@@DDBCDDCBC@CDCDCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC
II	0	1	1	1	100M	*	0	0	CCTAGCCCTAACCCTAACCCTAACCCTAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAAGCCTAA	#############################@B?8B?BA@@DDBCDDCBC@CDCDCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC
```

As it makes no claim as to version or sort order, this may even be a valid SAM file.

cram/cram_encode.c
jkbonfield: I was wary of changing anything outside of the immediate code it was calling due to potential threading issues, but the container is OK as it's never accessed by more than one thread once we're at the encode stage.  It looks safe to do this.  I'm not sure if it matters though for us, but there are comments about using no_ref as a flag during decode to avoid java implementation validation errors.

cram/cram_encode.c
jkbonfield: That's a simple enough change and safe to do I think.

PR #1850, opened 2024-10-15
Add an hts_crc32 function to use zlib or libdeflate.

jkbonfield:
This follows on from the hts_md5* functions which wrap up either OpenSSL or our own implementation.  Libdeflate's crc32 function is considerably faster than the native zlib, so we want to use it in (for example) the new "samtools checksum" code, but we do not wish to add baggage of looking for libdeflate in the configure script.

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -548,6 +548,10 @@ BGZF *bgzf_hopen(hFILE *hfp, const char *mode)
 }
 
 #ifdef HAVE_LIBDEFLATE
+uint32_t hts_crc32(uint32_t crc, const void *buf, size_t len) {
+    return libdeflate_crc32(crc, buf, len);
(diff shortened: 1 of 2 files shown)


jkbonfield:
Updated bgzf_encode_level0_func to use hts_crc32 to remove an ifdef.  (Thanks to @vasudeva8 for idea.)

We should probably consider at some stage exposing `bgzf_compress` and `bgzf_uncompress` as these too could very well be useful elsewhere for people who want to link against htslib but don't want the faff of doing their own libdeflate checking and if/else type clauses.  Plus it's a MUCH simpler interface than zlib's due to the non-streaming nature.

However that's a different PR IMO and more nuanced with BGZF vs generic Deflate.

PR #1853, opened 2024-10-23
BUGFIX header.c sam_hdr_remove_line_pos()

7PintsOfCherryGarcia:
Parameter check on argument 'position' was not accounting for 0 based indices as per header description:

/* sam.h
/// Remove nth line of a given type from a header
/*!
 * @param type     Type of the searched line. Eg. "SQ"
 * @param position Index in lines of this type (zero-based). E.g. 3
 * @return         0 on success, -1 on error
 *
 * Remove a line from the header by specifying the position in the type
 * group, i.e. 3rd @SQ line.
 */
HTSLIB_EXPORT
int sam_hdr_remove_line_pos(sam_hdr_t *h, const char *type, int
position);
*/


jkbonfield:
Thanks

PR #1857, opened 2024-11-06
hfile_dx : Support for the UK Biobank API (DNA Nexus "DX")

lindenb:
Hi all,

This is a preliminary PR, it is not ready for commit.

I'm using the UK biobank data https://ukbiobank.dnanexus.com . As far as I understand, all the CRAM/BAM are in the cloud (backed by amazon ?) and every time you need a CRAM/BAM, it needs to be [downloaded using their tool](https://documentation.dnanexus.com/user/objects/uploading-and-downloading-files/small-sets-of-files/downloading-using-dx) `dx download`.

Hence, even if you just want to just get the region of a BAM, the whole file must be downloaded ( ~ 3 min for 18G )

```
real    3m3.116s
user    0m45.935s
sys     1m0.066s
```

In the current PR, I used the DNA nexus C++ library : https://github.com/dnanexus/dx-toolkit to create an specialized `hfile_dx` similar to hfile_s3, hfile_gs etc... and now using the url scheme `dx:` I can use samtools with ukbb bams using `samtools view` and, of course, it's much faster.

```
time samtools view -O BAM -o out.bam -T ref/ref.fa --customized-index dx:file-G70qYb8JPKk6B36gKZK413fQ dx:file-G70qYbQJPKk7x1zQPBP94bpj "chrY:56673926-56673985"

real	0m10.317s
user	0m0.210s
sys	0m0.062s
```

What's next ?

- is it ok to include this PR in htslib ?
- is it the right way to include this dx 'plugin' in the current code ?
- I don't know enough the GNU automake, autoconf ecosystem, I tried to add some code in the configure.ac file but , for example , i don't know how to re-use `--dx-include-path`, how to prevent some .o and .pico to be included in the libraries, and at the end I modified the `Makefile`.
- It could be even faster if the callback used by htslib for reading used larger buffer, is it possible to control that size ?
- etc...
-

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@
 # DEALINGS IN THE SOFTWARE.
 
 CC     = gcc
+CPP    = g++
 AR     = ar
 RANLIB = ranlib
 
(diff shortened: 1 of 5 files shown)


lindenb:
(Later) Oh I see there is a plugin API, I'll try this first.

PR #1859, opened 2024-11-19
sam_internal.h: code2base[512 -> 513] because the byte for end-of-string is missing.

lindenb:
when compiling with g++ and including sam_internal.h (I need nibble2base) there is an error

```
../../htslib/sam_internal.h: In function ‘void nibble2base(uint8_t*, char*, int)’:
../../htslib/sam_internal.h:73:9: error: initializer-string for ‘const char [512]’ is too long [-fpermissive]
```

because `code2base[512]` is missing one byte for the end-of-string character.

this PR just set `code2base[513]`.

### DIFF
--- a/sam_internal.h
+++ b/sam_internal.h
@@ -70,7 +70,7 @@ static inline int possibly_expand_bam_data(bam1_t *b, size_t bytes) {
  *    seq[i] = seq_nt16_str[bam_seqi(nib, i)];
  */
 static inline void nibble2base_default(uint8_t *nib, char *seq, int len) {
-    static const char code2base[512] =
+    static const char code2base[513] = /* 512 + 1 for end-of-string */
         "===A=C=M=G=R=S=V=T=W=Y=H=K=D=B=N"
         "A=AAACAMAGARASAVATAWAYAHAKADABAN"
         "C=CACCCMCGCRCSCVCTCWCYCHCKCDCBCN"


daviesrob:
Looking at the C99 specification, I don't think this is necessary because it explicitly mentiones the case `t[3] = "abc"` as valid.  Also, `sam_internal.h` is not intended to be used outside HTSlib, and particularly not with a c++ compiler.  The only supported public interfaces are those in the `htslib/*.h` files.

There is however an argument to be made that it would be useful to expose something like `nibble2base()` in our public API.  If we do that, it would probably be best done by adding an exported function that can be called using an interface in `htslib/sam.h`.

jkbonfield:
For an assignment, I believe it copies as many as are needed.  By the sounds of it Rob's looked into the specification text which makes this more explicit.

However it wouldn't be a detrimental change so I'm amenable to changing it if it's tripping up some code sanity checking tool (with a suitable comment).  How did you spot this?  Was it something automated, or just eye-balling things?

lindenb:
@jkbonfield I was creating a C++ class wrapping the sequence of a read and I wanted to use nibble2base to get the read string

https://gitlab.univ-nantes.fr/pierre.lindenbaum/htsplusplus/-/blob/master/src/AbstractSamRecord.cpp?ref_type=heads#L166

 g++ complained with the error 

```
 initializer-string for ‘const char [512]’ is too long
```

Jut after this PR (there must be a Murphy's law for this) I just found that I can just use 

  return seq_nt16_str[bam_seqi(s, pos)];

So feel free to close this PR if you think that this '512' is safe for C code.

daviesrob:
Yes, that should work.

PR #1863, opened 2024-12-09
Fix cram_encode fuzzer issue caused by negative reference lengths.

jkbonfield:
Introduced in 24e4e31 with the cache of the LN_length field.  We now validate this as positive.  The check in cram_encode.c should now not be necessary, but it's fixed in two places incase we ever get this cropping up via another route.

Credit to OSS_Fuzz
Fixes oss-fuzz issue 382922241

### DIFF
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -3247,6 +3247,10 @@ cram_slice *cram_next_slice(cram_fd *fd, cram_container **cp) {
                 if (fd->ooc)
                     break;
 
+//                printf("%p %d:%ld-%ld vs %d:%ld-%ld\n", fd,
+//                       c_next->ref_seq_id, c_next->ref_seq_start, c_next->ref_seq_start+c_next->ref_seq_span-1,
+//                       fd->range.refid, fd->range.start, fd->range.end);
+
                 /* Skip containers not yet spanning our range */
                 if (fd->range.refid != -2 && c_next->ref_seq_id != -2) {
                     // ref_id beyond end of range; bail out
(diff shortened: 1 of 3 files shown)


daviesrob:
The fuzz artifact shows a discrepancy on how we're parsing `@SQ` lines with more than one `LN:` tag.  Where this happens, `sam_hdr_create()` records the value for the [last](https://github.com/samtools/htslib/blob/a662866807407021a0bebec995b10ae0e593c37a/sam.c#L1967) tag, while `refs_from_header()` uses [the result from `sam_hrecs_find_key()`](https://github.com/samtools/htslib/blob/a662866807407021a0bebec995b10ae0e593c37a/cram/cram_io.c#L2793) which is the [first](https://github.com/samtools/htslib/blob/a662866807407021a0bebec995b10ae0e593c37a/header.c#L2646) one.  That's how the negative value appeared even though `sam_hdr_create()` tries to filter them out - from its interpretation the value was zero.

Given the risk of discrepant lengths, do you think it would be a good idea to tighten up `sam_hdr_create()` to completely reject headers with multiple `LN` tags on a line?  It's illegal SAM, and I think the behaviour is only there because the original reader didn't have error checking for the problem.  (The same argument could also be made for `SN` where we actually jump through hoops to support such broken inputs.)

jkbonfield:
I hadn't spotted that discrepancy in checking.  Yes we'd be perfectly justified to just discard them.  Although I do have to wonder *why* we jump through hoops to support multiple SN.  Was there some pipeline which was generating duplicate SN fields at some stage that we were working around?  If so was it also duplicating the LN too?

Note this PR is still worth having though just incase we get some other mechanism for calling it that hasn't been via `sam_hdr_create`.

daviesrob:
I suspect it was done in case there were files like that.  I don't know of any examples, and if there are any then they're not compliant with the specification.

daviesrob:
As it would be a bit behaviour changing it might be best done separately.  So I'll merge this as-is, and we can look at the header API in another PR.

PR #1868, opened 2024-12-23
update to use htsFile in synced reader (for #1862)

### LINKED ISSUE #1862: feat: request, make synced_bcf_reader::bcf_sr_add_reader equivalent that accepts htsFile
The header has `int bcf_sr_add_reader(bcf_srs_t *files, const char *fname)`

It would be nice to have something like: `int bcf_sr_add_hreader(bcf_srs_t *files, htsFile *file_ptr)`

My used case is code that is detecting the file type (BAM/CRAM/BCF/etc) using an htsFile and then I want to still use bcf_synced_reader if it is a BCF/VCF.

vasudeva8:
Fixes #1862 
Added bcf_sr_add_hreader to add already opened htsfile to reader list.
  The file added can be set to close along with reader or closed explicitly by user. Aux data in synced_bcf_reader.c is updated to hold the file closure status.
  When the file name is passed along with file pointer, the index can be anywhere and mentioned using ##idx## method with file name.  As file name in htsfile does not maintain index information, the index has to be present in default location with default name (i.e. same location as file with file name.<index extension>) when no filename is passed to this method.

bcf_sr_add_reader updated to open the file and passes the file to bcf_sr_add_hreader with autoclose enabled.

Test utility and script updated to use the new interface. With -u argument, it uses the new interface where file pointer is made and used.

### DIFF
--- a/htslib/synced_bcf_reader.h
+++ b/htslib/synced_bcf_reader.h
@@ -1,7 +1,7 @@
 /// @file htslib/synced_bcf_reader.h
 /// Stream through multiple VCF files.
 /*
-    Copyright (C) 2012-2017, 2019-2024 Genome Research Ltd.
+    Copyright (C) 2012-2017, 2019-2025 Genome Research Ltd.
 
     Author: Petr Danecek <pd3@sanger.ac.uk>
 
@@ -233,9 +233,29 @@ void bcf_sr_destroy_threads(bcf_srs_t *files);
  *
  *  See also the bcf_srs_t data structure for parameters controlling
  *  the reader's logic.
+ *  Invokes bcf_sr_add_hreader with opened file
  */
 HTSLIB_EXPORT
 int bcf_sr_add_reader(bcf_srs_t *readers, const char *fname);
+
+/**
+ *  bcf_sr_add_hreader() - open new reader using htsfile
+ *  @readers: holder of the open readers
+ *  @file_ptr: htsfile already opened
+ *  @autoclose: close file along with reader or not, 1 - close, 0 - do not close
+ *  @idxname: index file name for file in @file_ptr
+ *
+ *  Returns 1 if the call succeeded, or 0 on error.
+ *
+ *  See also the bcf_srs_t data structure for parameters controlling
+ *  the reader's logic.
+ *  If idxname is NULL, uses file_ptr->fn to find index file.
+ *  With idxname as NULL, index file must be present along with the file with
+ *  default name
+ */
+HTSLIB_EXPORT
+int bcf_sr_add_hreader(bcf_srs_t *readers, htsFile *file_ptr, int autoclose,
+    const char *idxname);
 
 HTSLIB_EXPORT
 void bcf_sr_remove_reader(bcf_srs_t *files, int i);
--- a/test/test-bcf-sr.c
+++ b/test/test-bcf-sr.c
@@ -1,5 +1,5 @@
 /*
-    Copyright (C) 2017, 2020, 2023 Genome Research Ltd.
+    Copyright (C) 2017, 2020, 2023, 2025 Genome Research Ltd.
 
     Author: Petr Danecek <pd3@sanger.ac.uk>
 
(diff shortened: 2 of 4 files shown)


htslib/synced_bcf_reader.h
```diff
@@ -233,10 +233,30 @@ void bcf_sr_destroy_threads(bcf_srs_t *files);
...
+ *  @autoclose: close file along with reader or not, 1 - close, 0 - do not close
+ *  @fname: file name of @file_ptr
+ *
+ *  Returns 1 if the call succeeded, or 0 on error.
+ *
+ *  See also the bcf_srs_t data structure for parameters controlling
+ *  the reader's logic.
+ *  If fname is NULL, uses file_ptr->fn for fname
+ *  With fname as NULL, index file must be present in default location with
+ *  default name, as file_ptr->fn doesn't contain index information
```
jkbonfield, line 254: This feels problematic.  I can see that as `bcf_sr_add_reader` is public API then we can't easily change it (other than to make a new API), but `bcf_sr_add_hreader` is new so we probably should limit it straight off the bat.

I'd prefer the index filename to be an extra argument.  If NULL then it has the same caveats as above (as called by `bcf_sr_add_reader`), but if non-NULL then it specifies the index filename.  This means people explicitly using this API can correctly handle differing cases of filenames and indices.

Note this is complicated in VCF too due to differing conventions of indices and many index variants in use.  Eg VCF.gz may well be TBI, BAI or CSI.  (I've no idea why TBI exists for standard formats as it's not needed and just muddies the waters with formats within formats.)

We also won't have a tabix index for BCF, so tbx_index_load is inappropriate there.  Or is it just an historically misnamed function?  (Quite possible in this twisty maze of indices all alike!)

Edit: apparently so.  I don't understand what's going on in tabix index land, but the tbx component of this function name appears to be a total irrelevance regarding index format, but it is creating an tabix structure.   I don't understand what the point is of a tabix structure for a non-tabix index.  @pd3 care to elucidate what's going on here?  Why is tbx_t relevant for CSI indices.

jkbonfield: Having said that, bcftools uses `bcf_sr_add_reader` extensively (44 times), and apparently the inability to specify the index filename has never been a problem for it.  I assume just hunting around for .csi, .bai and .tbi is just widely understood behaviour.  It does handle other variants like `foo.bcf` and `foo.csi` (Picard style naming), but I'm not sure where that happens.

test/test-bcf-sr.c
```diff
@@ -133,13 +133,15 @@ int main(int argc, char *argv[])
         {"targets",required_argument,NULL,'t'},
         {"no-index",no_argument,NULL,1000},
         {"args",no_argument,NULL,1001},
+        {"usefptr",no_argument,NULL,'u'},
```
jkbonfield, line 136: Please add a line to the usage statement too.
Thanks

htslib/synced_bcf_reader.h
jkbonfield: Actually I think the filename handling is bugged anyway.

`bcftools view foo.bcf` works with either foo.bcf.csi or foo.csi.

However `bcftools view foo.vcf.gz` looks for foo.vcf.gz.csi and foo.vcf.csi.  It doesn't work with the filename suffix absent (ie foo.gz) and foo.csi.  However GATK appears to generate indices called foo.vcf.gz.tbi, so is different to how Picard handles BAM indices.  Nice to have consistency ;-)

jkbonfield:
The PR seems to work fine from what I can tell.  The additions to the test harness help this validation.  Thanks.

I added a few minor comments to address, but overall it solves the issue.

test/test-bcf-sr.c
vasudeva8: updated.

htslib/synced_bcf_reader.h
vasudeva8: Changed fname to idx name that index file can be explicitly passed.
if index path is NULL, tries to get index file from default path with default name, as earlier.

PR #1870, opened 2025-01-06
Typo in `vcf.h`

yuw444:
Below are my understandings:

1. `bcf_gt_phased` and `bcf_gt_unphased` encode `idx` to `val`
2. `bcf_gt_is_phased` and `bcf_gt_allele` decode `val` to phase info and `idx`

For instance, `idx = 0(ref), 1, 2, 3(alt)`

when phased(`|`), the corresponding values are
`val = 3, 5, 7, 9;`

when unphased (`\`), the corresponding values are
`val = 2, 4, 6, 8;`

According to [vcf: line 1168](https://github.com/samtools/htslib/tree/329e7943b7ba3f0af15b0eaa00a367a1ac15bd83/htslib), I believe `val` is what the author refers to.


pd3:
That's correct. Thank you

PR #1882, opened 2025-01-27
update to fail on duplicate LN tags, with different LN values

### LINKED ISSUE #1866: Inconsistency in parsing `@SQ` headers with multiple LN: and SN: tags
As noted in #1863, when reading `@SQ` headers `sam_hdr_create()` records the value for the [last `LN:` and `SN:`](https://github.com/samtools/htslib/blob/a662866807407021a0bebec995b10ae0e593c37a/sam.c#L1967) tags, while `refs_from_header()` uses [the result from `sam_hrecs_find_key()`](https://github.com/samtools/htslib/blob/a662866807407021a0bebec995b10ae0e593c37a/cram/cram_io.c#L2793) which is the [first](https://github.com/samtools/htslib/blob/a662866807407021a0bebec995b10ae0e593c37a/header.c#L2646) one.  This leads to a possible discrepancy in how such lines are interpreted by the SAM- and CRAM- reading code.

As such lines are invalid [according to the specification](https://samtools.github.io/hts-specs/SAMv1.pdf#subsection.1.3) ("Within each (non-`@CO`) header line, no field tag may appear more than once...") it might be best to reject such inputs by enforcing the one-tag rule.  Hopefully the number of files that would be rejected as a result will be very small, if any, and it will prevent any possibility of inconsistent interpretation.

vasudeva8:
Fixes #1866
Updated to use 1st LN tag value as in cram handling.
Accepts duplicate LN tags as long as they have same value.
If the values differ, returns failure.

sam_hdr_create updated for sam file checks and sam_hdr_sanitise for handling bam and cram.
sam_hdr_sanitise calls sam_hdr_fill_hrecs and in turn sam_hrecs_update_hashes, where duplicate check is made.

For cram, there is no added overhead as fill_hrecs is already in use.
For sam and bam, the sanitise update adds as overhead which is required to identify the duplicates.
  For sam, the check occurs during normal parsing in sam_hdr_create and sanitise update is redundant.
  For bam, fill_hrecs update is required to identify the duplicates. sanitise seem to be the best common place to have this.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1,6 +1,6 @@
 /*  sam.c -- SAM and BAM file I/O and manipulation.
 
-    Copyright (C) 2008-2010, 2012-2024 Genome Research Ltd.
+    Copyright (C) 2008-2010, 2012-2025 Genome Research Ltd.
     Copyright (C) 2010, 2012, 2013 Broad Institute.
 
     Author: Heng Li <lh3@sanger.ac.uk>
@@ -1911,12 +1911,6 @@ static sam_hdr_t *sam_hdr_sanitise(sam_hdr_t *h) {
         cp[h->l_text] = '\0';
     }
 
-     if (sam_hdr_fill_hrecs(h) < 0) {
-        //failed in parsing / validation
-        sam_hdr_destroy(h);
-        return NULL;
-    }
-
     return h;
 }
 
@@ -1972,13 +1966,13 @@ static sam_hdr_t *sam_hdr_create(htsFile* fp) {
                 } else {
                     if (strncmp(q, "LN:", 3) == 0) {
                         hts_pos_t tmp = strtoll(q + 3, (char**)&q, 10);
-                        if (ln != -1 && tmp != ln) {
-                            //duplicate LN tag with different value
-                            hts_log_error("111Header includes @SQ line \"%s\" with \
-duplicate LN: tag", sn);
+                        if (ln != -1 && ln != tmp) { //duplicate & different LN
+                            hts_log_error("Header includes @SQ line \"%s\" with"
+                                " multiple LN: tag with different values.", sn);
                             goto error;
+                        } else {
+                            ln = tmp;
                         }
-                        ln = tmp;
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


sam.c
```diff
@@ -1911,6 +1911,12 @@ static sam_hdr_t *sam_hdr_sanitise(sam_hdr_t *h) {
         cp[h->l_text] = '\0';
     }
 
+     if (sam_hdr_fill_hrecs(h) < 0) {
```
daviesrob, line 1914: I'm not sure if we want to call `sam_hdr_fill_hrecs()` here.  It will trigger a full parse of the header, which we may not want to do unless really needed.

sam.c
```diff
@@ -1964,8 +1970,16 @@ static sam_hdr_t *sam_hdr_create(htsFile* fp) {
                     strncpy(sn, q, r - q);
                     q = r;
                 } else {
-                    if (strncmp(q, "LN:", 3) == 0)
-                        ln = strtoll(q + 3, (char**)&q, 10);
+                    if (strncmp(q, "LN:", 3) == 0) {
+                        hts_pos_t tmp = strtoll(q + 3, (char**)&q, 10);
+                        if (ln != -1 && tmp != ln) {
+                            //duplicate LN tag with different value
+                            hts_log_error("111Header includes @SQ line \"%s\" with \
```
daviesrob, line 1977: This error message looks a bit odd...

vasudeva8: corrected

sam.c
vasudeva8: removed

daviesrob:
I thik you may have removed too much in the last update.  We still want to fail if SAM finds a header with two `LN:` tags.

vasudeva8:
If we have error return, then SAM and with some extra changes CRAM will fail in duplicate but BAM won't, as we will not invoke the hrec_fill due to overheads.
Is it not better to have consistent behaviour across formats, that all succeeds and just BAM won't warn?
As SAM uses the 1st LN value, the -ve value issue should not be there any more.

vasudeva8:
Updated to fail in case of duplicate LN tag with different values.

This change makes header read for sam, cram different from bam, where bam doesn't fail and sam, cram fails on header read api. This difference will be in library level but the library is mainly used in a utility (samtools) level where another operation like PG addition takes place, making the failure visible in bam as well and all formats have similar failure. (if --no-PG is used with bam it will not fail).

PR #1886, opened 2025-02-10
Make HAVE_ATTRIBUTE_TARGET check also check that SSSE3 intrinsics work

jmarshall:
We released a new pysam based on HTSlib et al 1.21 last week, and now #1838 has been re-reported as pysam-developers/pysam#1327. So that nobody gets more support requests, it would be good to fix it so that the _simd.c_ code is disabled for these ancient compilers.

Playing with the [Compiler Explorer](https://godbolt.org/), I confirmed that GCC ≤ 4.8.5 cannot compile _simd.c_, while GCC ≥ 4.9.0 can. I considered `HTS_GCC_AT_LEAST(4,9)` but realised that the existing `HAVE_SSSE3` appears to be exactly what is needed.

However further Compiler Exploration showed that GCC ≤ 4.8.5 can compile these intrinsics with `-mssse3` but not with merely `__attribute__((target("ssse3")))`, which fails to enable the intrinsics. Hence the right approach is to actually use some intrinsics in the configure test as @daviesrob suggested in https://github.com/samtools/htslib/issues/1838#issuecomment-2356380704. This PR does that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -309,7 +309,7 @@ config.h:
 	echo '#define HAVE_ATTRIBUTE_CONSTRUCTOR 1' >> $@
 	echo '#endif' >> $@
 	echo '#if (defined(__x86_64__) || defined(_M_X64))' >> $@
-	echo '#define HAVE_ATTRIBUTE_TARGET 1' >> $@
+	echo '#define HAVE_ATTRIBUTE_TARGET_SSSE3 1' >> $@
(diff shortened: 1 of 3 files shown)


daviesrob:
Thanks, I've confirmed that this works with the ancient gcc on Centos 7.  I'll also try to keep my centos 7 VM alive to check for any more regressions.

PR #1890, opened 2025-02-19
update to tabix with easy validation for a few file formats

vasudeva8:
Update to PR #1887 to make the behaviour controlled by an option.
The error exit code return is the default behaviour as in last PR.
A command line option, '-y/--easyparse' is added to have a relaxed parsing, to create output even with non-parsable data, skipping them, as earlier tabix did. Instead of an error log, a warning will be made in such cases.


vasudeva8:
Not required, current behaviour is correct.

PR #1891, opened 2025-02-25
Clarify bam_set1 parameter documentation.

cjw85:
The helper function `bam_set1()` is great for creating `bam1_t`s from scratch. However I fell over a foot gun in using it. I expected to be able to pass ASCII text as from a fastq file; instead to use the function correctly one first needs to subtract 33 from each char of the quality string.


whitwham:
Thank you.

PR #1899, opened 2025-03-24
Return errors instead of EOF after all I/O errors etc in `hts_itr_multi_next`/`sam_itr_next`/`sam_read1`/`vcf_parse`/`bcf_read`

jmarshall:
While investigating pysam-developers/pysam#1328, I noticed that in a number of cases `hts_itr_multi_next()` returns -1 after `iter->seek(…)` failure. -1 returned from this function (and from `sam_itr_next()`, which for multi iterators just returns this function's result) indicates a successfully reached EOF. Hence seek failure at these locations in the code results in `samtools` or other invoking code **silently omitting data** for multi-region queries.

A full audit of HTSlib functions that return ≥0 record / -1 EOF / ≤-2 error found a handful of other instances of incorrect -1 returns in `sam_read1()` and `vcf_parse()` and hence `bcf_read()`.

It also found a documentation error for `bgzf_getc()`, which does indeed distinguish between EOF and errors.

The usual reason for kstring functions to fail is memory allocation failure, which sets `errno` to ENOMEM. The audit also found cases in which these functions fail but do not set `errno`, which has been addressed. According to 6d927dfa3192492f3015b1bc9a0026f585e93acc, MinGW does not have EOVERFLOW so this works around its absence if necessary — in a somewhat involved way because this is a public header. (_hts_time_funcs.h_ has EOVERFLOW without guards, so perhaps this is out of date and EOVERFLOW is portable nowadays.)

### DIFF
--- a/hts.c
+++ b/hts.c
@@ -4291,7 +4291,7 @@ int hts_itr_multi_next(htsFile *fd, hts_itr_t *iter, void *r)
         if (iter->curr_off) { // seek to the start
             if (iter->seek(fp, iter->curr_off, SEEK_SET) < 0) {
                 hts_log_error("Seek at offset %" PRIu64 " failed.", iter->curr_off);
-                return -1;
+                return -2;
             }
             iter->curr_off = 0; // only seek once
         }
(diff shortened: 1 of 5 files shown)


jmarshall:
> (All of these functions are documented as returning -1 on EOF and <= -2 on error.)

This is a slight over-summary, as `vcf_parse()` does not document its return value (but studying the code shows it follows the same pattern as `bcf_read()`). `bcf_itr_next()` is also missing documentation.

PR #1903, opened 2025-04-02
Recognize T > A[chr15:12345[ breakend type

pd3:
When this is merged, this problem will go away https://github.com/samtools/bcftools/issues/2389

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -5142,13 +5142,11 @@ static void bcf_set_variant_type(const char *ref, const char *alt, bcf_variant_t
     }
     else if ( *r && !*a )
     {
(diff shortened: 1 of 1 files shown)


Patricie34:
Hi, I am encountering the same issues when using bcftools norm on manta vcfs. I understood that this is solved in the dev branch, but when it will be released? We are currently using bcftools version 1.20. Since we are implementing bcftools normalization in the nf-core/sarek pipeline we need to solve it asap. Does anyone know, how to fix it before this will be corrected in the new version of bcftools? Thanks

whitwham:
@Patricie34 We should be making a new release in the near future.

Issue #1780, opened 2024-05-06
Linking against Intel ISAL for faster deflate operations


pettyalex:
Hello,

I see that bgzf compression can optionally link against libdeflate for faster performance, but I was wondering if you've ever evaluated or considered linking against the Intel ISAL? It is BSD licensed, and offers the highest performance deflate implementation that I'm aware of, especially at lower compression ratios.

Perhaps it could be optionally linked, just like libdeflate is right now? Could it be preferred to libdeflate if both are available?

https://github.com/intel/isa-l

Benchmarks:
https://github.com/zlib-ng/zlib-ng/issues/1486
https://github.com/powturbo/TurboBench/issues/43

jkbonfield:
Looking at those benchmarks it seems like Intel have improved their performance.  I did evaluate this in the past and it was simply not a convincing win.

http://www.htslib.org/benchmarks/zlib.html

Although that was Intel's zlib rather than igzip specifically, but you would think it's the same technology in both?  Maybe not.

Profiling a `samtools view -1 -o /tmp/tmp.bam in.bam` command to see where the CPU time is spent, I see this:

```
  51.73%  samtools  libdeflate.so.0     [.] deflate_compress_fastest
  13.63%  samtools  libdeflate.so.0     [.] deflate_decompress_default
  10.43%  samtools  libdeflate.so.0     [.] deflate_flush_block
   9.95%  samtools  libdeflate.so.0     [.] crc32_x86_pclmul_avx
   2.36%  samtools  [kernel]            [k] 0xffffffff99800190
   2.06%  samtools  libc-2.27.so        [.] __memmove_sse2_unaligned_erms
   1.62%  samtools  libdeflate.so.0     [.] deflate_make_huffman_code
   1.21%  samtools  samtools            [.] bgzf_read
   1.02%  samtools  samtools            [.] bgzf_write
   0.95%  samtools  samtools            [.] bam_read1
```

Libdeflate is already fast for decoding, but according to the benchmarks may be around half the speed of ISA-L for encoding.  So with ISA-L outputting level 1 BAMs may be ~50% faster throughput.  We can't tell though if this holds true on the small block sizes bgzf uses without trying it out.

pettyalex:
I bring this up specifically while my group spends a lot of $ doing various bcftools tasks on cloud platforms that are spending most of their CPU time compressing output, I'm going to see if I can find some time to do this and bring you actual benchmarks on those small block sizes. 

[...]

PR #1908, opened 2025-04-29
ISA-L compression support evaluation.

rhpvorderman:
This is a quick test for re-evaluation of ISA-L support

see #1780 

I ran samtools view on the current repository using libdeflate:
```
$ ./samtools view -1 --threads 8 ~/test/HG002_20230424_1302_3H_PAO89685_2264ba8c_hac_simplex_downsampled.bam | pv > /dev/null
10.1GiB 0:00:44 [ 234MiB/s] 
```
And then using the below patch with ISA-L.
```
./samtools view -1 --threads 8 ~/test/HG002_20230424_1302_3H_PAO89685_2264ba8c_hac_simplex_downsampled.bam | pv > /dev/null
10.3GiB 0:00:26 [ 394MiB/s] [      
```

This is significantly faster (slightly less than 2X) with a marginally larger filesize.  And this is only the compression, I haven't tested decompression yet.

This needs to be properly integrated in the build system if this is going to be properly supported, so for now it is just hacked in to allow for benchmarks. (Install libisal-dev to compile on linux).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ AR     = ar
 RANLIB = ranlib
 
 # Default libraries to link if configure is not used
-htslib_default_libs = -lz -lm -lbz2 -llzma -lcurl
+htslib_default_libs = -lz -lm -lbz2 -llzma -lcurl -lisal
 
 CPPFLAGS =
 # TODO: make the 64-bit support for VCF optional via configure, for now add -DVCF_ALLOW_INT64
--- a/bgzf.c
+++ b/bgzf.c
@@ -609,6 +609,8 @@ static int _bgzf_compress_isal(uint8_t *dst, size_t *dlen, const void *src, size
         return -1;
     }
     *dlen = z->next_out - (uint8_t *)dst;
+    free(level_buf);
+    free(z);
     memcpy(dst, g_magic, BLOCK_HEADER_LENGTH); // the last two bytes are a place holder for the length of the block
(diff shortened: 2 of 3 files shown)


rhpvorderman:
ISA-L decompression is also slightly faster:
```

```
$ ./samtools view -u  ~/test/HG002_20230424_1302_3H_PAO89685_2264ba8c_hac_simplex_downsampled.bam | pv > /dev/null
16.2GiB 0:00:58 [ 284MiB/s] [     
```
./samtools view -u  ~/test/HG002_20230424_1302_3H_PAO89685_2264ba8c_hac_simplex_downsampled.bam | pv > /dev/null
16.2GiB 0:00:52 [ 315MiB/s] [                
```
Just 10% though and this is on a intel i7 6700K. This is not as clear cut as the compression.

rhpvorderman:
I can confirm that the same sort of results occur when using a ryzen 5 3600 (AMD Zen 2). 10% faster reading and almost twice as fast writing.

So it works equally well on both x86 vendors. Since we have a cluster with both AMD Epyc and Intel Xeon I always check if optimizations are not detrimental. 

I don't have any high performing ARM64 devices at my disposable to test. The latest release of ISA-L is available on conda (https://github.com/conda-forge/isa-l-feedstock) for ARM64 and MacOS ARM64 architectures. If anyone wants to test.
From some tests I have run on an arm64 processor it seems as ISA-L performs well, but that was on a single-board computer so that is not really representative of machines running bioinformatic work loads.

PR #1910, opened 2025-05-01
Update distdir target to include more recent source files

### LINKED ISSUE #1846: The htscodecs `version.h` is not built unless its .git is present; no instructions to prepare a working tarball
The htslib build process needs to build `htscodecs/htscodecs/version.h`, since it isn't shipped in the htscodecs repository. This is accomplished here:

https://github.com/samtools/htslib/blob/2ff207bd16bb3094843395450aa48936dd774ecb/Makefile#L578-L591

But if `htscodecs` doesn't have a `.git` folder (for example, because we're building from a tarball and not a source checkout), this doesn't do anything and the file is not created.

So if you want to prepare a buildable tarball of `htslib`, you need to make sure that `make htscodecs/htscodecs/version.h` has been run before any `.git` folders are removed.

For vg, we ship tarballs that include a vendored-in `htslib`. But it turns out our tarballs don't actually build, since we don't perform that step as part of our tarball packaging process.

So I went looking for the official instructions for preparing htslib to be tarballed, and I couldn't find them. The Makefile provides a `make distdir`, but that doesn't seem to include `htscodecs` at all, nor does it include `os/lzma_stub.h` which the build seems to require, so I'm not sure that it is actually in use. The actual htslib release tarballs include these, so they must not use (just) `make distdir`.

If tarballing up everything in the Git tree except the `.git` folders doesn't produce a directory tree that can build, there should be instructions in the project documentation on how to prepare the tree to be tarballed, or how to export a working tarball through the build system somehow.

I can by trial and error discover that I need a `make htscodecs/htscodecs/version.h`, but how will I know when new required generated files are added to the project?

jmarshall:
Update the wildcard to include source files and directories added since this was added in 2015 and last modified in 2017. Give up on omitting _**/README.md_ even though they are omitted from proper release tarballs, as we want to include _README.large_positions.md_.

In particular add the bundled parts of _htscodecs/**_ and ensure that _htscodecs/htscodecs/version.h_ has been built so is present in the distribution directory. (HTSlib's _version.h_ continues to be built via `version.sh` and omitted from the distribution directory.)

With this, `make distdir=../dist` produces a directory with the same files as the htslib-1.21 release tarball, modulo the addition of _README.md_ files and several expected test output files added on **develop** since the last release.

Fixes #1846.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -628,7 +628,7 @@ htscodecs/htscodecs:
 	@false
 
 # Build the htscodecs/htscodecs/version.h file if necessary
-htscodecs/htscodecs/version.h: force
+htscodecs/htscodecs/version.h: force | htscodecs/htscodecs
 	@if test -e $(srcdir)/htscodecs/.git && test -e $(srcdir)/htscodecs/configure.ac ; then \
 	  vers=`cd $(srcdir)/htscodecs && git describe --always --dirty --match 'v[0-9]\.[0-9]*'` && \
 	  case "$$vers" in \
--- a/Makefile
+++ b/Makefile
@@ -1050,11 +1050,11 @@ tags TAGS:
 # code with your program, this hook enables Automake-style "make dist"
 # for this subdirectory.  If you do bundle an htslib snapshot, please
 # add identifying information to $(PACKAGE_VERSION) as appropriate.
-# (The wildcards attempt to omit non-exported files (.git*, README.md,
+# (The wildcards attempt to omit non-exported files (.git, .gitignore,
 # etc) and other detritus that might be in the top-level directory.)
-distdir:
+distdir: htscodecs/htscodecs/version.h
 	@if [ -z "$(distdir)" ]; then echo "Please supply a distdir=DIR argument."; false; fi
-	tar -c *.[ch15] [ILMNRchtv]*[ELSbcekmnth] | (cd $(distdir) && tar -x)
+	tar -c *.[ch157] [I-R]*[ELSde] [cmors]*[bcemns4] [bhtv]*[bhknpt] htscodecs/*.md htscodecs/[ht]*/*[chp-t0-9] | (cd $(distdir) && tar -x)
 	+cd $(distdir) && $(MAKE) distclean
 
 force:


daviesrob:
Thanks.  This works OK on a fully checked out git repository, but I noticed that it gives slightly broken results if you've forgotten to recursively get the htscodecs submodule:
```
$ make distdir distdir=/tmp/htsdist
echo '# Default htscodecs.mk generated by Makefile' > htscodecs.mk
echo 'include $(HTSPREFIX)htscodecs_bundled.mk' >> htscodecs.mk
./hts_probe_cc.sh 'gcc' '-g -Wall -O2 -fvisibility=hidden ' '-fvisibility=hidden' >> htscodecs.mk
tar -c *.[ch157] [I-R]*[ELSde] [cmos]*[bcemns4] [bhtv]*[bhknpt] htscodecs/*.md htscodecs/[ht]*/*[chp-t0-9] | (cd /tmp/htsdist && tar -x)
tar: htscodecs/*.md: Cannot stat: No such file or directory
tar: htscodecs/[ht]*/*[chp-t0-9]: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
cd /tmp/htsdist && make distclean
make[1]: Entering directory '/tmp/htsdist'
rm -f test/*.tmp test/*.tmp.* test/faidx/*.tmp* \
               test/longrefs/*.tmp.* test/tabix/*.tmp.* \
               test/bgzf_boundaries/*.tmp.* test/*/FAIL* \
               header-exports.txt shlib-exports-so.txt
# .. other rms omitted ..
$ echo $?
0
```
So it looks like `make` succeeded (albeit noisily), but the resulting directory structure is incomplete.
The `htscodecs/htscodecs/version.h` dependency doesn't catch this case, as currently its rule triggers but doesn't do anything.  Ideally we'd want the `htscodecs/htscodecs` rule to kick in and tell you to add the submodule, but I'm having a bit of trouble getting it to work...

jmarshall:
I suppose the `tar -c` could write to a temporary file so its failing exit status isn't swallowed by the pipe, e.g.,

```
…
tar -cf $(distdir)/htslib.tmp.tar …crazy wildcards…
cd $(distdir) && tar -xf htslib.tmp.tar && rm -f htslib.tmp.tar
…
```

daviesrob:
Adding an order-only dependency on `htscodecs/htscodecs` to `htscodecs/htscodecs/version.h` seems to do the trick.  Trying this earlier I was having problems getting the rule to trigger, but it seems to be OK now:
```diff
--- a/Makefile
+++ b/Makefile
@@ -628,7 +628,7 @@ htscodecs/htscodecs:
        @false
 
 # Build the htscodecs/htscodecs/version.h file if necessary
-htscodecs/htscodecs/version.h: force
+htscodecs/htscodecs/version.h: force | htscodecs/htscodecs
        @if test -e $(srcdir)/htscodecs/.git && test -e $(srcdir)/htscodecs/configure.ac ; then \
          vers=`cd $(srcdir)/htscodecs && git describe --always --dirty --match 'v[0-9]\.[0-9]*'` && \
          case "$$vers" in \

```

```
$ make distdir distdir=/tmp/htsdist
echo '# Default htscodecs.mk generated by Makefile' > htscodecs.mk
echo 'include $(HTSPREFIX)htscodecs_bundled.mk' >> htscodecs.mk
./hts_probe_cc.sh 'gcc' '-g -Wall -O2 -fvisibility=hidden ' '-fvisibility=hidden' >> htscodecs.mk


Error: htscodecs submodule files not present for htslib.
Try running: 
    git submodule update --init --recursive
in the top-level htslib directory and then re-run make.


make: *** [Makefile:619: htscodecs/htscodecs] Error 1
```

Unfortunately this now needs to be updated to include the `ref_cache` subdirectory.  Sorry about that...

jmarshall:
> Unfortunately this now needs to be updated to include the `ref_cache` subdirectory. Sorry about that...

Done. Too easy — it only needed an `r` added to one regex…

I've left the order-only dependency or any equivalent for you to apply when you're happy with it.

daviesrob:
I've pushed a commit for the order-only dependency.  Testing shows that both `make distdir` and `make htscodecs/htscodecs/version.h` can now detect a missing submodule and print the message about how to get it.

PR #1911, opened 2025-05-06
Add ref-cache, a caching proxy for reference sequences

daviesrob:
ref-cache is a local server of reference sequences, for use when encoding or decoding CRAM files that use reference-based compression.  It can also act as a proxy for an upstream server (like the EBI's existing CRAM reference server) so that it can download and cache new references as they are needed.

ref-cache is only built for MacOS, Linux, and other UNIX-like platforms.  Windows is not supported due to the significant differences in the platform.  It is also only built if the configure script has been run.

As it's quite complex, and needs to deal with potentially untrusted inputs, ref-cache is built with somewhat stricter compiler options than the rest of HTSlib to try to reduce the number of lurking bugs.  `configure.ac` is updated to add tests for these options, along with the code to enable ref-cache.

ref-cache uses some APIs - notably `openat()`, `mkdirat()` etc. that first appeared in POSIX 200809.  This requires `_XOPEN_SOURCE` to be bumped from 600 to 700 to allow them to be used, which in turn deprecates `usleep()`, which is replaced by `nanosleep()`.

When built, ref-cache is tested by starting two servers, one pre-populated with some files, and a second with no starting files but with the first as its upstream.  If then attempts to pull references from one to another by viewing an example file with `test/test_view`, using the various servers as the reference source.  This showed up some issues on the CI tests, as they now actively use the `hfile_libcurl.so` plug-in, which was previously built but not really tested.  These are fixed by ensuring `HTS_PATH` is set before running `test/test.pl`, and by disabling the Address Sanitizer one-definition-rule check in `test/with-shlib.sh`, as it otherwise triggers when the static `test/test_view` program attempts to link in the shared library `libhts.so`.


jmarshall:
Nice — I think this will be an important piece of infrastructure.

One problem with going to `_XOPEN_SOURCE=700` is that POSIX issue 7 marked `gettimeofday()` as obsolescent which may make for trouble compiling _thread_pool.c_ on some particularly pedantic platforms.

ref_cache/upstream.c
```diff
@@ -0,0 +1,1197 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.  */
+
+#include "../config.h"
+
+#if defined(__APPLE__) && defined(__MACH__)
+// McOS headers hide CMSG_LEN() and CMSG_SPACE() definitions is these
```
jmarshall, line 28:
```suggestion
// macOS headers hide CMSG_LEN() and CMSG_SPACE() definitions if these
```

ref_cache/transaction.c
```diff
@@ -0,0 +1,828 @@
...
+    assert(count >= 0);
+
+    // Only read if there's data available and the buffer has a reasonable
+    // amount of space free (to avoid lots of short reads).
+    while (count > 0 && buffered < SEND_BUFFER_SIZE / 2) {
+        size_t len = (transact->buffer_in < transact->buffer_out
+                      ? transact->buffer_out - transact->buffer_in
+                      : SEND_BUFFER_SIZE - transact->buffer_in);
+        if ((off_t) len > count)
+            len = count;
```
jmarshall, line 550: I got a warning here which this cast silences, but whether it's really necessary or the right answer… 🤷 

```suggestion
            len = (size_t) count;
```

jkbonfield:
- I spotted a trivial memory leak if the CLI errors:

```
/tmp/_jkb/bin/ref-cache -d /tmp/_jkb/cache
Couldn't open directory /tmp/_jkb/cache: No such file or directory

==354505==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 320 byte(s) in 1 object(s) allocated from:
    #0 0x7f0e9437e730 in __interceptor_realloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:85
    #1 0x40ccaa in add_match_addr ref_cache/main.c:691

Direct leak of 48 byte(s) in 1 object(s) allocated from:
    #0 0x7f0e9437f2e7 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:77
    #1 0x40b0b9 in open_logs ref_cache/log_files.c:145

SUMMARY: AddressSanitizer: 368 byte(s) leaked in 2 allocation(s).
Aborted
```

An `mkdir` fixes the leak, so it's not an issue, and it's a one off rather than on-going leak-and-grow scenario so is largely irrelevant.

ref_cache/ref-cache.1
```diff
@@ -0,0 +1,229 @@
...
+mkdir logs
+ref-cache -b -d cached_refs -l logs -p 8080 -u https://www.ebi.ac.uk/ena/cram/md5/
+.EE
+To make SAMtools and HTSlib use the server,
+set its URL in the
+.B REF_PATH
+environment variable (note that colons should be doubled up in the URL,
+and you should substitute the hostname of your actual server).
+.EX
+REF_PATH=http:://myserver.example.com::8080/
```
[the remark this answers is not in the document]
jkbonfield: Although that said, I cannot explain why it is slower to redirect that way.  (Empty cache on localhost)

```
$ ./ref-cache -d /tmp/_jkb/cache -u http://seq22-os0000001:8000
...


$ REF_PATH=http:://localhost::8000 /usr/bin/time -f '%e seconds' samtools 
view ~/scratch/data/9827_2#49.1m.cram 1:600000-600000 -o /dev/null
2.33 seconds

$ REF_PATH=http:://localhost::8000 /usr/bin/time -f '%e seconds' samtools view ~/scratch/data/9827_2#49.1m.cram 1:600000-600000 -o /dev/null
0.35 seconds
```

We can see the difference between the server contacting upstream vs returning its local copy.  However if I go directly to upstream I also see 0.35 seconds.  That's expected.  What's not expected is the additional 2s to go from one local network host to the next.

It's not critical obviously if we also include the local directory in REF_PATH, but it is a little surprising to see the delay given we know how long the upstream copy of ref-cache takes to serve a request.

I assume that's the time taken to write the file?  Maybe it should return the data before writing it and do that bit asynchronously.  Plus md5sum it (command line md5sum is ~0.5s).

That's a wishlist item, not a major problem to fix before release.

jkbonfield (commented):
Ongoing still.  I haven't looked at the code yet, but part of me thinks as long as there's nothing glaring then this is an entirely new program so it can't break any existing infrastructure.

Although the Makefile changes and usleep things could, but don't appear to be from what I've tested so far.  We may need to test more systems though for that POSIX thing, like various BSDs and other unix flavours.

ref_cache/http_parser.c
```diff
@@ -0,0 +1,623 @@
...
+                return READ_ERROR;
+            }
+            short_read = 1;
+        } else {
+            short_read |= (size_t) res < to_read;
+            parser->in = (parser->in + res) & BUF_MASK;
+            parser->used += (unsigned int) res;
+        }
+    }
+    if (res > 0 && (parser->in + 1) < parser->out) {
```
jkbonfield, line 212: These two conditionals both fail when in+1 == out.

```
    if (parser->in >= parser->out) {
    ...
    if (res > 0 && (parser->in + 1) < parser->out) {
```

As a consequence, it can enter an infinite loop of read-more.

I triggered it by giving it a carefully crafted range request:

```
curl -v -s -r `perl -e 'print "x"x1009'` localhost:8000/f3814841f1939d3ca19072d9e89f3fd7
```

jkbonfield:
Fuzz testing it (https://github.com/JonCooperWorks/httpfuzz) is showing a few issues.  One of them is that it's possible to attempt to read from a client which has been closed.  We already have a flag indicating if the client is on the read or write list, so we can check that flag and also the `client->polled` pointer, as that can also be null at times after close.  (Otherwise it's something used internal to the pooled allocator, as it becomes its own free list.)

Line numbers are off due to extra debugging, but something like this:

```
@@ -535,11 +542,28 @@ static void do_reads(Clients *clients, const Options *opts, Poll_wrap *pw) {
     Client *client = NULL;
 
     while ((client = read_stack_pop(clients)) != NULL) {
+        if (!(client->flags & ON_READ_LIST) || !client->polled) {
+            if (opts->verbosity > 1)
+                fprintf(stderr, "Client is no longer being read from\n");
+            close_client(clients, client, pw);
+            continue;
+        }
+
```

jkbonfield:
Actually that's not sufficient.  We have to tackle the route cause of adding the same client multiple times to the read_stack, or if that's permitted, fix the removal code so it removes all instances of the client rather than the first one.  The above code works as far as fooling valgrind and address sanitizer, but only because we have our own memory pool.  If we subvert that it fails:

We can do that with:

```
#define pool_alloc(x) malloc(x->dsize)                                            
#define pool_free(p,x) free(x)                                                    
```

Then I get:

```
==3320777==ERROR: AddressSanitizer: heap-use-after-free on address 0x611000004790 at pc 0x0000004174cf bp 0x7ffc9e7217d0 sp 0x7ffc9e7217c0
READ of size 8 at 0x611000004790 thread T0
    #0 0x4174ce in read_stack_pop ref_cache/server.c:119
    #1 0x41b6b3 in do_reads ref_cache/server.c:548
    #2 0x41cbf9 in run_poll_loop ref_cache/server.c:817
    #3 0x40dd36 in make_new_child ref_cache/main.c:232
    #4 0x40f08d in run_server_population ref_cache/main.c:415
    #5 0x41330c in main ref_cache/main.c:1013
    #6 0x7fab3c4d8d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #7 0x7fab3c4d8e3f in __libc_start_main_impl ../csu/libc-start.c:392
    #8 0x4038d4 in _start (/nfs/users/nfs_j/jkb/work/samtools_master/htslib/ref_cache/ref-cache+0x4038d4)

0x611000004790 is located 16 bytes inside of 224-byte region [0x611000004780,0x611000004860)
freed by thread T0 here:
    #0 0x7fab3c83d470 in __interceptor_free ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:52
    #1 0x4181e0 in close_client ref_cache/server.c:209
    #2 0x41b594 in do_reads ref_cache/server.c:611
    #3 0x41cbf9 in run_poll_loop ref_cache/server.c:817
    #4 0x40dd36 in make_new_child ref_cache/main.c:232
    #5 0x40f08d in run_server_population ref_cache/main.c:415
    #6 0x41330c in main ref_cache/main.c:1013
    #7 0x7fab3c4d8d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58

previously allocated by thread T0 here:
    #0 0x7fab3c83e91f in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x417d13 in get_free_client ref_cache/server.c:171
    #2 0x4189fc in handle_incoming ref_cache/server.c:286
    #3 0x41ca35 in run_poll_loop ref_cache/server.c:790
    #4 0x40dd36 in make_new_child ref_cache/main.c:232
    #5 0x40f08d in run_server_population ref_cache/main.c:415
    #6 0x41330c in main ref_cache/main.c:1013
    #7 0x7fab3c4d8d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
```

It's basically the same bug I tried to fix above, but now it's the `read_stack_pop` function which is failing as the client has been freed already.

jkbonfield:
I think I see the problem.  `do_reads()` clears the flag via `client->flags &= ~ON_READ_LIST`.  However if it gets EOF we have:

```
            if ((client->flags & (READ_CLOSED|READ_DRAIN)) != 0) {                
                if (client->transact == NULL) {                                   
                    if (opts->verbosity > 1) {                                    
                        fprintf(stderr, "Closing fd #%d (EOF)\n", client->polled->fd);                                                                             
                    }                                                             
                    close_client(clients, client, pw);                            
                }                                                                 
```

Unfortunately `close_client` can't call `read_stack_remove` to remove the client because the `ON_READ_LIST` has now been deleted.

It's trivial to just make `read_stack_remove` a no-op when it's already been removed or when it's inappropriate (eg it's on the `can_write` instead of `can_read`) list:

```
+    fprintf(stderr, "read_stack_remove %p (polled %p)\n", client, client->polled);
     if (clients->can_read == client) {
         clients->can_read = client->next;
         if (clients->can_read != NULL) clients->can_read->prev = NULL;
     } else {
         Client *prev = client->prev;
         Client *next = client->next;
-        prev->next = next;
+        if (prev != NULL) prev->next = next;
         if (next != NULL) next->prev = prev;
     }
     client->prev = client->next = NULL;
```

Then we can call this always rather than checking the flag.  That may be easier than postponing the flag clear and remembering to clear it in the many locations we continue or break from the while loop in `do_reads()`.  We could do the same thing for `write_stack_remove` but I didn't test it.

jkbonfield:
Some of the changes above aren't necessary if we just make `read_stack_push` (/pop/remove, and write equivalents) hardened by checking for flags and setting / clearing flags.  This removes a source of error elsewhere, protects against double pushes, and simplifies code by removing duplication.  My aggregated changes to make it pass the fuzzer so far are:

```
diff --git a/ref_cache/server.c b/ref_cache/server.c
index c5729d4f..900c6063 100644
--- a/ref_cache/server.c
+++ b/ref_cache/server.c
@@ -50,6 +50,10 @@ DEALINGS IN THE SOFTWARE.  */
 #include "upstream.h"
 #include "../cram/pooled_alloc.h"
 
+// DEBUG of Client pool
+#define pool_alloc(x) malloc(x->dsize)
+#define pool_free(p,x) free(x)
+
 // May not have been defined, depending on compiler options
 #ifndef NI_MAXHOST
 #  define NI_MAXHOST 1025
@@ -101,6 +105,9 @@ int init_clients(Clients *clients) {
 }
 
 static inline void read_stack_push(Client *client, Client **stack) {
+    if (client->flags & ON_READ_LIST)
+        return; // already being processed
+    client->flags |= ON_READ_LIST;
     if (*stack != NULL) (*stack)->prev = client;
     client->next = *stack;
     client->prev = NULL;
@@ -114,6 +121,7 @@ static inline Client * read_stack_pop(Clients *clients) {
     clients->can_read = can_read->next;
     if (clients->can_read != NULL) clients->can_read->prev = NULL;
     can_read->prev = can_read->next = NULL;
+    can_read->flags &= ~ON_READ_LIST;
     return can_read;
 }
 
@@ -124,13 +132,17 @@ static inline void read_stack_remove(Clients *clients, Client *client) {
     } else {
         Client *prev = client->prev;
         Client *next = client->next;
-        prev->next = next;
+        if (prev != NULL) prev->next = next;
         if (next != NULL) next->prev = prev;
     }
     client->prev = client->next = NULL;
+    client->flags &= ~ON_READ_LIST;
 }
 
 static inline void write_stack_push(Client *client, Client **stack) {
+    if (client->flags & ON_WRITE_LIST)
+        return; // already being processed
+    client->flags |= ON_WRITE_LIST;
     if (*stack != NULL) (*stack)->prev_write = client;
     client->next_write = *stack;
     client->prev_write = NULL;
@@ -144,6 +156,7 @@ static inline Client * write_stack_pop(Clients *clients) {
     clients->can_write = can_write->next_write;
     if (clients->can_write != NULL) clients->can_write->prev_write = NULL;
     can_write->prev_write = can_write->next_write = NULL;
+    can_write->flags &= ~ON_WRITE_LIST;
     return can_write;
 }
 
@@ -154,9 +167,10 @@ static inline void write_stack_remove(Clients *clients, Client *client) {
     } else {
         Client *prev = client->prev_write;
         Client *next = client->next_write;
-        prev->next_write = next;
+        if (prev != NULL) prev->next_write = next;
         if (next != NULL) next->prev_write = prev;
     }
+    client->flags &= ~ON_WRITE_LIST;
     client->prev_write = client->next_write = NULL;
 }
 
@@ -175,15 +189,11 @@ static inline void close_client(Clients *clients, Client *client,
         client->polled = NULL;
     }
 
-    if ((client->flags & ON_READ_LIST) != 0) {
+    if ((client->flags & ON_READ_LIST) != 0)
         read_stack_remove(clients, client);
-        client->flags &= ~ON_READ_LIST;
-    }
 
-    if ((client->flags & ON_WRITE_LIST) != 0) {
+    if ((client->flags & ON_WRITE_LIST) != 0)
         write_stack_remove(clients, client);
-        client->flags &= ~ON_WRITE_LIST;
-    }
 
     cleanup_http_parser(&client->parser);
 
@@ -382,7 +392,6 @@ void queue_transaction_write(Transaction *transact, Client **write_stack) {
 
     /* Put onto write_stack */
     write_stack_push(client, write_stack);
-    client->flags |= ON_WRITE_LIST;
 }
 
 static Read_result eat_input(int fd) {
@@ -503,7 +512,6 @@ static void handle_client_events(Clients *clients, const Options *opts,
         /* Add to can_read list */
         if ((client->flags & READ_CLOSED) == 0) {
             read_stack_push(client, &clients->can_read);
-            client->flags |= ON_READ_LIST;
         } else {
             assert((client->flags & ON_READ_LIST) == 0);
         }
@@ -525,7 +533,6 @@ static void handle_client_events(Clients *clients, const Options *opts,
             && transaction_have_content(client->transact)) {
             /* Add to the can_write list */
             write_stack_push(client, &clients->can_write);
-            client->flags |= ON_WRITE_LIST;
         }
     }
 }
@@ -536,7 +543,6 @@ static void do_reads(Clients *clients, const Options *opts, Poll_wrap *pw) {
 
     while ((client = read_stack_pop(clients)) != NULL) {
         Read_result res;
-        client->flags &= ~ON_READ_LIST;
 
         if (opts->verbosity > 2) {
             fprintf(stderr, "Reading fd #%d\n", client->polled->fd);
@@ -569,7 +575,6 @@ static void do_reads(Clients *clients, const Options *opts, Poll_wrap *pw) {
                 fprintf(stderr, "fd #%d has more data...\n", client->polled->fd);
             }
             read_stack_push(client, &can_read);
-            client->flags |= ON_READ_LIST;
             break;
         case READ_EOF:
             client->flags |= READ_CLOSED;
@@ -581,6 +586,7 @@ static void do_reads(Clients *clients, const Options *opts, Poll_wrap *pw) {
                         fprintf(stderr, "Closing fd #%d (EOF)\n", client->polled->fd);
                     }
                     close_client(clients, client, pw);
+                    can_read = NULL;
                 }
             } else {
 #ifndef PW_HAVE_EDGE
@@ -591,6 +597,7 @@ static void do_reads(Clients *clients, const Options *opts, Poll_wrap *pw) {
                     fprintf(stderr, "Altering poll settings for fd#%d : %s\n",
                             client->polled->fd, strerror(errno));
                     close_client(clients, client, pw);
+                    can_read = NULL;
                 }
 #endif /* PW_HAVE_EDGE */
             }
@@ -600,6 +607,7 @@ static void do_reads(Clients *clients, const Options *opts, Poll_wrap *pw) {
                 fprintf(stderr, "Closing fd #%d (read error)\n", client->polled->fd);
             }
             close_client(clients, client, pw);
+            can_read = NULL;
             break;
         }
     }
@@ -614,8 +622,6 @@ static void do_writes(Clients *clients, const Options *opts, Poll_wrap *pw,
     while ((client = write_stack_pop(clients)) != NULL) {
         Write_result res;
 
-        client->flags &= ~ON_WRITE_LIST;
-
         if (opts->verbosity > 1) {
             fprintf(stderr, "Writing to fd #%d\n", client->polled->fd);
         }
@@ -674,7 +680,6 @@ static void do_writes(Clients *clients, const Options *opts, Poll_wrap *pw,
                 fprintf(stderr, "fd #%d can send more data...\n", client->polled->fd);
             }
             write_stack_push(client, &can_write);
-            client->flags |= ON_WRITE_LIST;
             break;
         case WRITE_EOF:
         case WRITE_ERROR:
```

(The first hunk of that patch is debugging only, to permit address-sanitizer to spot use-after-free bugs.)

jkbonfield:
I'm also not entirely sure there isn't a deadlock somewhere.  Sometimes my fuzzer stops and the server also stops.  Gdb shows it's doing an epoll, presumably waiting, but the fuzzer is also waiting for something too.

Server output with `-vvv`:

```
RECV'D: POST /uploadFile HTTP/1.1
parser_read_data returned 1
fd #313 has more data...
Events 0005 for fd #313
Reading fd #313
parser_read_data returned 1
fd #313 has more data...
Reading fd #313
parser_read_data returned 1
fd #313 has more data...
Reading fd #313
parser_read_data returned 1
fd #313 has more data...
Reading fd #313
parser_read_data returned 2
Writing to fd #313
transact_send_data returned 3
Shutting down fd #313
Events 0004 for fd #313
Events 0004 for fd #306
Events 0004 for fd #307
Events 0004 for fd #308
Events 0004 for fd #309
Events 0004 for fd #310
Events 0004 for fd #311
Events 0004 for fd #312
Events 0004 for fd #313
```

ref_cache/upstream.c
daviesrob: Typos fixed...

ref_cache/transaction.c
daviesrob: Yes, there's an annoying difference in the signed-ness of the parameters to `pread()`.  The cast is safe anyway, as we know in order to get here `count < len && len <= SEND_BUFFER_SIZE`.

ref_cache/main.c
```diff
@@ -0,0 +1,1029 @@
...
+        fprintf(stderr,
+                "Options:\n"
+                "  -b         Run in background as a daemon.\n"
+                "  -d <dir>   Directory for cached reference files.\n"
+                "  -h         Show help.\n"
+                "  -l <dir>   Directory for log files.  Log to stdout if not set and running in\n"
+                "             foreground.\n"
+                "  -L         Don't log.\n"
+                "  -n <1-4>   Number of server processes to run.\n"
+                "  -p <num>   Port number to listen on.\n"
```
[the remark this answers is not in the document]
daviesrob: They're now listed.

ref_cache/ref-cache.1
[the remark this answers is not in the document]
daviesrob: I've added text to the manual page showing how to do this.

ref_cache/http_parser.c
daviesrob: This bit has been rewritten to use `readv()`, to make it simpler.  It shouldn't get stuck any more.

jkbonfield:
Bug fixed in percent encoding on gcc-7, plus tidy up some signed vs unsigned conversion warnings.

```
diff --git a/htscodecs b/htscodecs
--- a/htscodecs
+++ b/htscodecs
@@ -1 +1 @@
-Subproject commit 51794289ac47455209c333182b6768f99a613948
+Subproject commit 51794289ac47455209c333182b6768f99a613948-dirty
diff --git a/ref_cache/log_files.c b/ref_cache/log_files.c
index 198a8e0c..ce3adea5 100644
--- a/ref_cache/log_files.c
+++ b/ref_cache/log_files.c
@@ -100,15 +100,15 @@ static int rotate_logs(Logfiles *logfiles, const Options *opts) {
     /* Remove stale logs */
     assert(opts->nlogs > 0);
     assert(logfiles->sz > opts->nlogs);
-    for (i = opts->nlogs - 1; i < logfiles->nlogs; i++) {
+    for (i = opts->nlogs - 1U; i < logfiles->nlogs; i++) {
         if (unlinkat(opts->log_dir_fd, logfiles->logs[i].name, 0) != 0) {
             fprintf(stderr, "Warning: Couldn't remove old log file %s/%s: %s\n",
                     opts->log_dir, logfiles->logs[i].name, strerror(errno));
         }
     }
 
-    if (logfiles->nlogs > (size_t) opts->nlogs - 1)
-        logfiles->nlogs = opts->nlogs - 1;
+    if (logfiles->nlogs > (size_t) opts->nlogs - 1U)
+        logfiles->nlogs = opts->nlogs - 1U;
 
     /* Put the new file at the front of the list */
     if (logfiles->nlogs > 0) {
diff --git a/ref_cache/main.c b/ref_cache/main.c
index b57cfa24..7dc6fd2f 100644
--- a/ref_cache/main.c
+++ b/ref_cache/main.c
@@ -101,10 +101,10 @@ static void change_name(char *name) {
 static int init_children(Options *opts) {
     int k;
 
-    upstream = malloc((opts->max_kids + 1) * sizeof(int));
+    upstream = malloc((opts->max_kids + 1U) * sizeof(int));
     if (upstream == NULL) return -1;
 
-    kids = malloc((opts->max_kids + 1) * sizeof(Child_proc));
+    kids = malloc((opts->max_kids + 1U) * sizeof(Child_proc));
     if (kids == NULL) return -1;
 
     for (k = 0; k <= opts->max_kids; k++) {
diff --git a/ref_cache/request_handler.c b/ref_cache/request_handler.c
index cb9191bb..24932d2d 100644
--- a/ref_cache/request_handler.c
+++ b/ref_cache/request_handler.c
@@ -77,9 +77,10 @@ static char * decode_uri(Http_Parser *parser) {
     for (in = out = uri; *in; in++) {
         if (*in == '/' && last == '/') continue;
         if (*in == '%'
-            && (d1 = hexval(*(in + 1)) >= 0) && (d2 = hexval(*(in + 2)) >= 0)) {
+            && ((d1 = hexval(*(in + 1))) >= 0)
+            && ((d2 = hexval(*(in + 2))) >= 0)) {
             char v = (char) (d1 << 4 | d2);
-            *in += 2;
+            in += 2;
             if (v == '/' && last == '/') continue;
             *out++ = last = v; continue;
         }
```

jkbonfield:
Also note for clang 3.7 we get this warning:

```
make CC="clang37" CFLAGS="-g -O3 -Werror -Wall -Wno-missing-field-initializers" -j -k
clang37 -g -O3 -Werror -Wall -Wno-missing-field-initializers -Wextra -Wformat -Wformat=2 -Wconversion -Wstrict-aliasing -fstack-protector-strong  -I. -I/nfs/users/nfs_j/jkb/ftp/compression/libdeflate -I/nfs/users/nfs_j/jkb/ftp/compression/isa-l/include/ -c -o ref_cache/upstream.o ref_cache/upstream.c
ref_cache/upstream.c:133:29: error: missing field 'msg_namelen' initializer [-Werror,-Wmissing-field-initializers]
    struct msghdr msg = { 0 };
                            ^
ref_cache/upstream.c:154:29: error: missing field 'msg_namelen' initializer [-Werror,-Wmissing-field-initializers]
    struct msghdr msg = { 0 };
                            ^
ref_cache/upstream.c:177:29: error: missing field 'msg_namelen' initializer [-Werror,-Wmissing-field-initializers]
    struct msghdr msg = { 0 };
                            ^
ref_cache/upstream.c:219:29: error: missing field 'msg_namelen' initializer [-Werror,-Wmissing-field-initializers]
    struct msghdr msg = { 0 };
```

I'm baffled why I couldn't silence it with `-Wno-missing-field-initializers`, but it could be replaced with a memset or we just ignore it and see if anyone complains.  By clang 7.0 it wasn't warning any more, given this is perfectly legal C and in common usage.

jkbonfield:
Icc annoyingly doesn't consider unknown options to be an error.  From config.log:

```
configure:6626: checking whether the compiler accepts -mbranch-protection=standard
configure:6644: icc -c  -Wall -g -O2 -fvisibility=hidden -Werror -mbranch-protection=standard -I/nfs/users/nfs_j/jkb/ftp/compression/libdeflate -I/nfs/users/nfs_j/jkb/ftp/compression/isa-l/include/ conftest.c >&5
icc: command line warning #10006: ignoring unknown option '-mbranch-protection=standard'
configure:6644: $? = 0
configure:6655: result: yes
```

The good news is because it doesn't return an error (even with `-Werror`) it also means the actual compile also won't error.  It just looks a bit untidy, but I can live with that given how many people use icc.

ref_cache/main.c
```diff
@@ -0,0 +1,1165 @@
...
+
+            if (addr->ai_family == AF_INET
+                && addr->ai_addrlen == sizeof(struct sockaddr_in)) {
+                struct sockaddr_in *sin = (struct sockaddr_in *) addr->ai_addr;
+                ma->family = sin->sin_family;
+                alen = sizeof(sin->sin_addr.s_addr);
+                addrp = (unsigned char *) &sin->sin_addr.s_addr;
+            } else if (addr->ai_family == AF_INET6
+                       && addr->ai_addrlen == sizeof(struct sockaddr_in6)) {
+                struct sockaddr_in6 *sin = (struct sockaddr_in6 *) addr->ai_addr;
```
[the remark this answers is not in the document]
jkbonfield: Bah,  Intel's clang derivative also fails on that server.c line:

```
ref_cache/server.c:283:38: error: incompatible pointer types passing 'struct sockaddr_storage *' to parameter of type 'const struct sockaddr *' [-Werror,-Wincompatible-pointer-types]
        if (check_addr_allowed(opts, &addr, addrlen) != 0) {
                                     ^~~~~
ref_cache/server.c:215:75: note: passing argument to parameter 'addr' here
static int check_addr_allowed(const Options *opts, const struct sockaddr *addr,
                                                                          ^
1 error generated.
```

Can we just omit all the extra warnings in the Makefile and enable them on a CI instance instead?  Ie have `REF_CACHE_EXTRA_C_FLAGS` at a reduced level but expand it in our github action for the ubuntu gcc build?  That avoids a huge amount of pain and ties the warnings to one sensible compiler instead.  We're just fighting compiler differences at this stage.

daviesrob:
These problems should all be fixed now.  I've also adjusted the configure script so that it only attempts to add the extra compiler options on `gcc`-like compilers, and to make the tests work better on `icc`.

jkbonfield:
I've verified it works well on Intel icc / clang.  Thanks.

Clang 7.0 gave bizarre error messages from the linker about x86 formats, but continued regardless (with exit 0) and produced a working binary.  So I'll just ignore that as an odd linker foible that doesn't affect us.  By clang 10 it had stopped doing this.

jmarshall:
Apologies for not making these command-line processing suggestions before this was merged.

I think it would be good if:

* `ref-cache` had long options as well;
* it had a `ref-cache --version` option (that would presumably report the HTSlib version number as `bgzip` et al do);
* it had a `ref-cache --help` option;
* if `ref-cache` without any arguments produced a full usage display (on stdout, with exit status 0) as per `-h` and the putative `--help`, as is the “htslib way” (at the moment it produces a one-line usage message).

daviesrob:
Yes, the differences in command-line processing are a sign that this was originally a different project that has been merged into HTSlib.  Making it a bit more native would be nice; I'll look into it for a future update.

PR #1912, opened 2025-05-08
Use the highest VCF version when merging headers

pd3:
See https://github.com/samtools/bcftools/issues/2395 and https://github.com/samtools/bcftools/pull/2404

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -1080,6 +1080,7 @@ int bcf_hdr_update_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec, const bcf_hrec_t *tmp)
     free(hrec->value);
     hrec->value = strdup(tmp->value);
     if ( !hrec->value ) return -1;
+    kh_val(aux->gen,k) = hrec;
 
     if (!strcmp(hrec->key,"fileformat")) {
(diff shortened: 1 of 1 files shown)


vcf.c
```diff
@@ -4699,6 +4700,16 @@ bcf_hdr_t *bcf_hdr_merge(bcf_hdr_t *dst, const bcf_hdr_t *src)
...
                 need_sync += res;
             }
+            else if ( !strcmp(src->hrec[i]->key,"fileformat") )
+            {
+                int ver_src = bcf_get_version(src,src->hrec[i]->value);
+                int ver_dst = bcf_get_version(dst,dst->hrec[j]->value);
+                if ( ver_src > ver_dst )
+                {
+                    res = bcf_hdr_set_version(dst,src->hrec[i]->value);
+                    need_sync += res;
```
jkbonfield, line 4710: `bcf_hdr_set_version` returns 0 on success and <0 on error (eg strdup allocation failure).  So it's inappropriate to do `need_sync += res` as at best it'll do nothing at all, or on error it'll even *remove* the need to syncronise.

We should change that to `need_sync = 1`.

PR #1916, opened 2025-05-22
Minor build fixes for ref-cache

jmarshall:
* Enable building ref-cache in a separate build directory
* Clean ref_cache test files
* Fix _ref_cache/upstream.o_ dependencies: it too depends only on _hts_defs.h_ amongst HTSlib headers


daviesrob:
Looks good, thanks.

PR #1927, opened 2025-06-11
Fix compilation error on old kernels

AnibalMG:
Tested on CentOS 7.x, kernel 3.10.0
Error:
ref_cache/server.c:237:13: error: 'const struct in6_addr' has no member named 's6_addr32'
  237 |         if (IN6_IS_ADDR_V4MAPPED(&addr->in6.sin6_addr)) {
      |             ^~~~~~~~~~~~~~~~~~~~


daviesrob:
Thanks for the patch.  This reminds me that I needed to finish my solution to this problem which is now PR #1928.  It takes the different approach of avoiding `IN6_IS_ADDR_V4MAPPED` completely, so there's no need to mess around with feature test macros (at least for this problem).

Note that if you do want to define something like `_GNU_SOURCE`, you need to wrap it in a `#if !defined()` check to ensure it isn't there already.  It shouldn't be normally, but it's not impossible as it may come in e.g. via a compiler option.

PR #1928, opened 2025-06-11
Fix compilation against older glibc / macOS SDK

daviesrob:
Work around problems caused by symbols being hidden when defining `_POSIX_C_SOURCE` or `_XOPEN_SOURCE`:

* The `IN6_IS_ADDR_V4MAPPED` macro fails to compile on glibc before 2.25.  Fix by avoiding it and using `memcmp()` to match the prefix it detects instead.
* macOS SDKs before 12.0 hid `O_DIRECTORY`.  Fix by defining `_DARWIN_C_SOURCE` to get it back.

Thanks to @jmarshall for spotting these problems, which affected Conda builds even on modern systems.

Replaces and closes #1927

### DIFF
--- a/ref_cache/main.c
+++ b/ref_cache/main.c
@@ -24,6 +24,12 @@ DEALINGS IN THE SOFTWARE.  */
 
 #include <config.h>
 
+#if defined(__APPLE__) && !defined(_DARWIN_C_SOURCE)
+// Work around older MacOS SDKs which hid O_DIRECTORY if _POSIX_C_SOURCE
+// or _XOPEN_SOURCE are defined.
+#define _DARWIN_C_SOURCE
+#endif
+
 #include <stdlib.h>
 #include <stdio.h>
 #include <unistd.h>
--- a/ref_cache/server.c
+++ b/ref_cache/server.c
@@ -233,8 +233,11 @@ static int check_addr_allowed(const Options *opts, sa_family_t family,
         start = 0;
         end = opts->first_ip6;
     } else if (family == AF_INET6 && addrlen == sizeof(struct sockaddr_in6)) {
+        const uint8_t v4mapped_prefix[12] = {
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff
+        };
         addrp = (const uint8_t *) &addr->in6.sin6_addr;
-        if (IN6_IS_ADDR_V4MAPPED(&addr->in6.sin6_addr)) {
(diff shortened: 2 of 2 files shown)


jmarshall:
```diff
        addrp = (const uint8_t *) &addr->in6.sin6_addr;
-       if (IN6_IS_ADDR_V4MAPPED(&addr->in6.sin6_addr)) {
+       if (memcmp(addrp, v4mapped_prefix, sizeof(v4mapped_prefix)) == 0) {
```

`.sin6_addr` is a `struct in6_addr`, and POSIX says of that struct:

> The <netinet/in.h> header shall define the `in6_addr` structure, which shall include at least the following member:
> `uint8_t s6_addr[16]`
>
> This array is used to contain a 128-bit IPv6 address, stored in network byte order.

I haven't seen any implementations in which `s6_addr` is not the first member, but in theory I guess that should be the following for maximum portability:

```c
addrp = (const uint8_t *) &addr->in6.sin6_addr.s6_addr;
```

I was going to say who knows how the address is encoded in `s6_addr`, but that last sentence quoted does tie it down. So your `memcmp` is in fact a portable approach.

But what happens if you later need to use some of the other `IN6_IS_ADDR_*` macros? It'll be a bit tedious to figure out the bit patterns for others too, and it's a shame to obscure the code for all platforms for the sake of a bug fixed eight years ago.

I'd probably suggest instead just using `IN6_IS_ADDR_V4MAPPED` and wrapping a work-around at the top of the file in

```c
…includes…
#include <unistd.h>
#ifdef __GLIBC_PREREQ
#if !__GLIBC_PREREQ(2, 25)
#define __USE_MISC  // or a different workaround
#endif
#endif
…includes…
#include <netdb.h>
#include <netinet/in.h>
…
```

so that only the old broken glibc headers are affected.

ref_cache/main.c
```diff
@@ -24,6 +24,12 @@ DEALINGS IN THE SOFTWARE.  */
 
 #include <config.h>
 
+#if defined(__APPLE__) && !defined(_DARWIN_C_SOURCE)
```
jmarshall, line 27: I'd love it if you could guard this with ` && __MACOS_SDK_VERSION < whatever`, but in typical Apple fashion their SDKs don't appear to define any SDK version macros. Le big sigh…

daviesrob:
I don't think I'll need to use any other `IN6_IS_ADDR_*` macros soon.  Should it become necessary I can look at other solutions, but hopefully it'll be far enough in the future that the problematic header files have dropped out of use.

Meanwhile, the current solution avoids having to `#define` any macros that may cause unintended side-effects.

PR #1931, opened 2025-06-24
Improved error handling in cram_compress_block2.

jkbonfield:
If we learn a method is good to use but that method subsequently fails to encode, we don't just fail in cram_compress_block2 any more. Instead we restart the trial / adaption technique again to learn what works instead.

This fixes an issue in CRAM encoding when we attempted to SIMD rANS encode a block of Ns, but after RLE is applied the resultant buffer was too small for SIMD to work.  (Also fixed in htscodecs)

Seen with long data which starts off with sequences, tuning soft-clip SC series to rans.  This then switches to long-read data with seq `*`, but still having soft-clips.  Due to an oddity of CRAM feature codes we still need to encode a sequence for the S CIGAR op, so this is all "N"s for seq `*` .  Ufortunately this caused rANS SIMD to fail.  (See https://github.com/samtools/htscodecs/pull/137)

### DIFF
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -3575,6 +3575,8 @@ static int cram_byte_array_stop_decode_char(cram_slice *slice, cram_codec *c,
 
     cp = (char *)b->data + b->idx;
     if (out) {
+        // FIXME: Add check for out overflow.  Maybe using *out_size
+        // as a maximum size?  Similarly for other variable sized decodes
         while ((ch = *cp) != (char)c->u.byte_array_stop.stop) {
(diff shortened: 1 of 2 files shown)


jkbonfield:
Fixed the threading issue by making the loop-detection explict with an extra argument rather than attempting to intuit it by looking at trial count numbers (which can be somewhat unpredictable when using many threads).

Also added some lock checks.

Finally, turned the `hts_log_info` into `hts_log_error` as although it's not a hard error (we can recover), it may be useful to know when it's happening as it *shouldn't* happen any more.

PR #1934, opened 2025-07-01
Make CRAM codecs with encoding E_BYTE_ARRAY do bounds checking on out.

jkbonfield:
The huffman code already had bounds checking, but cram_byte_array_stop_decode_char only checked for read-overflow on the CRAM block and not write-overflow on the destination.  That pointer is fine as we've preallocated it to be big enough *provided* the CRAM is internally consistent.

NB encoding E_BYTE_ARRAY_BLOCK is fine as these use block_append which dynamically growns the block.

This fixes a bug encountered while debugging an incorrectly produced CRAM file.  Note there may potentially be similar issues with other encoding types too, eg for E_BYTE used in `DS_QS`, but I haven't verified that.  Ideally we need a way to get the fuzzer to spot these automatically.  It's surprising it didn't, but possibly this is down to memory allocation growing too fast (to avoid too frequent reallocs), making the very first byte written given a significant amount of head-room and obfuscating buffer overruns.  That could be true [here](https://github.com/samtools/htslib/blob/develop/cram/cram_io.h#L226-L233).

I *think* it's fine though as it's only the unbounded things (byte array stop specifically) that really need checking.  Everything else is in loops that operate over a defined number of items anyway - eg the known sequence length.

### DIFF
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -2298,9 +2298,13 @@ int cram_decode_slice(cram_fd *fd, cram_container *c, cram_slice *s,
     // However it's likely that this also saves memory as own growth
     // factor (*=1.5) is never applied.
     {
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+        int qsize=0, nsize=0, q_id=0;
+#else
         int qsize, nsize, q_id;
         cram_decode_estimate_sizes(c->comp_hdr, s, &qsize, &nsize, &q_id);
         //fprintf(stderr, "qsize=%d nsize=%d\n", qsize, nsize);
+#endif
 
         if (qsize && (ds & CRAM_RL)) BLOCK_RESIZE_EXACT(s->seqs_blk, qsize+1);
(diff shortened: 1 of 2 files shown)


jkbonfield:
Eg we may consider this:

```
diff --git a/cram/cram_io.h b/cram/cram_io.h
index d2d583df..7a9e9c52 100644
--- a/cram/cram_io.h
+++ b/cram/cram_io.h
@@ -227,6 +227,12 @@ static inline int block_resize(cram_block *b, size_t len) {
     if (b->alloc > len)
         return 0;
 
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+    // Removal of extra padding causes many more reallocs, but detects
+    // more buffer overruns.
+    return block_resize_exact(b, len);
+#endif
+
     size_t alloc = b->alloc+800;
     alloc = MAX(alloc + (alloc>>2), len);
     return block_resize_exact(b, alloc);
```

However it's very slow due to excessive realloc calling.  It could perhaps uncover issues.  It's perhaps something to manually explore with a local fuzzer on develop to see if it could have spotted the issue this PR fixes.

Issue #1932, opened 2025-06-24
Prefixed genotype phasing


pd3:
As reported in bcftools https://github.com/samtools/bcftools/issues/2422#issuecomment-3000977572, there is a problem with the new prefixing of partially phased genotypes, likely introduced by https://github.com/samtools/htslib/commit/b14fffb4

I believe the intention was to explicitly prefix half-phased genotypes such as `/0/1|2`. The prefix is wrong in `/0|1`  (if anything, it should be `|0|1`, but no, it should not be prefixed in this case).

A copy-and-paste test case to reproduce the problem
```
printf '##fileformat=VCFv4.5\n##FORMAT=<ID=GT,Number=1,Type=String,Description="">\n##contig=<ID=chr6>\n#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT\taaa\nchr6\t160136611\t.\tC\tT\t.\t.\t.\tGT\t0|0\n' | bcftools view -o rmme1.bcf --write-index

printf '##fileformat=VCFv4.3\n##FORMAT=<ID=GT,Number=1,Type=String,Description="">\n##contig=<ID=chr6>\n#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT\tbbb\nchr6\t160136611\t.\tC\tT\t.\t.\t.\tGT\t0|1\n' | bcftools view -o rmme2.bcf --write-index

bcftools merge rmme1.bcf rmme2.bcf
```

daviesrob:
This is happening because the two files specify different VCF versions, and the VCF parser attempting to be backwards compatible with itself by not setting the is_phased bit on the first GT value if the file format is 4.3 or earlier.  Really we want to store one of three values - phased, unphased, or not set - but we can't do that in a single bit.  The consequence is that you need to take the format version into account when interpreting the data.

If you change your example files so that they're both 4.3 or both 4.5 then the merge works as expected and outputs `0|1`.  It's the mixture that causes the problem.

pd3:
I understand the versioning. However, this is a different topic, it's a bug: the genotype `/0|1` makes no sense, as it wouldn't `/0|1|2|3`. Since the three alleles are phased, the last one must be phased as well

PR #1938, opened 2025-07-22
converting bcf 1st phase data as in v44

vasudeva8:
Fix for #1932 
It was discussed earlier to keep the phasing data in VCF44 format internally as that easily resolves the issue mentioned (will necessitate some changes in bcftools as there are a few checks with v4.3 type phase values - in convert tests) with VCF checks.
Data needs a consistent binary representation, irrespective of VCF/BCF source, so the same conversion is needed for BCF data as well, from v4.x to v4.4 in terms of phasing values. This conversion is made on read, for bcf with version < v44. While writing, it is again converted that data is stored without the change made, if version < v44. This adds ~15% overhead it seems.
--removed the mention of a few changes that seemed required in bcftools, it worked fine without any such change with later checks--

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -1853,6 +1853,7 @@ static const char *get_type_name(int type) {
  *  @param ploidy - no. of phasing values per sample
  *  @param type - value type (one of BCF_BT_...)
  *  Returns 0 on success and 1 on failure
+ *  Update for haploids made only if it is not unknown (.)
  */
 static int updatephasing(uint8_t *p, uint8_t *end, uint8_t **q, int samples, int ploidy, int type)
 {
@@ -1875,7 +1876,7 @@ static int updatephasing(uint8_t *p, uint8_t *end, uint8_t **q, int samples, int
     case 1:
         // Trivial case - haploid data is phased by default
         for (j = 0; j < samples; ++j) {
-            *p |= 1;
+            if (*p) *p |= 1;    //only if not unknown (.)
             p += inc;
         }
         break;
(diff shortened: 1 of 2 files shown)


vcf.c
```diff
@@ -1814,6 +1814,135 @@ static void bcf_record_check_err(const bcf_hdr_t *hdr, bcf1_t *rec,
...
+        }
+        if (!anyunphased && al1 > 1) {  //no other unphased
+            /*set phased on read or make unphased on write as upto 4.3 1st
+            phasing is not described explicitly and has to be inferred*/
+            switch(type) {
+                case BCF_BT_INT8:
+                    *(uint8_t*)ptr1 += upd;
+                break;
+                case BCF_BT_INT16:
+                    *(uint16_t*)ptr1 += upd;
```
daviesrob, line 1873: This should use `i16_to_le()`, and the case below `i32_to_le()`.

vcf.c
```diff
@@ -1814,6 +1814,135 @@ static void bcf_record_check_err(const bcf_hdr_t *hdr, bcf1_t *rec,
...
+        int anyunphased = 0;
+        uint8_t *ptr1 = p;
+        int32_t al1 = 0, val;
+        for (k = 0; k < ploidy; ++k) { //in each ploidy
+            switch(type) {
+                case BCF_BT_INT8:
+                    val = *(uint8_t*)p;
+                break;
+                case BCF_BT_INT16:
+                    val = *(uint16_t*)p;
```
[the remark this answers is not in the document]
vasudeva8: Updated as suggested and the CI issue is fixed!

daviesrob (commented):
If `updatephasing()` is changed as suggested, all of HTSlib's tests still pass.

Most of bcftools' do as well, apart from `bcftools convert --haplegendsample` which starts incorrectly marking lots of genotypes as partially phased.  This is because the [`process_gt_to_hap()`](https://github.com/samtools/bcftools/blob/b6f490c104d95c009afb70a14976e0200928d0a1/convert.c#L823) and [`process_gt_to_hap2()`](https://github.com/samtools/bcftools/blob/b6f490c104d95c009afb70a14976e0200928d0a1/convert.c#L1004) functions currently assume the first phasing bit is always zero.  As they're likely broken for VCF4.4, they'll need to be fixed irrespective of this change.  Adjusting them to ignore the first phasing bit gets them working again, and after that all bcftools tests pass.

vcf.c
```diff
@@ -2300,6 +2442,11 @@ int bcf_write(htsFile *hfp, bcf_hdr_t *h, bcf1_t *v)
         return -1;
     }
 
+    if (update44phasing(h, v, 0)) { //reset phasing update made after read
+        hts_log_error("Failed to set prorper phasing at %s:%"PRIhts_pos"", bcf_seqname_safe(h,v), v->pos+1);
+        return -1;
+    }
+
```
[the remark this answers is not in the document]
vasudeva8: Removed.

synced_bcf_reader.c
```diff
@@ -708,6 +708,11 @@ static int _reader_fill_buffer(bcf_srs_t *files, bcf_sr_t *reader)
             ret = bcf_itr_next(reader->file, reader->itr, reader->buffer[reader->nbuffer+1]);
             if ( ret < -1 ) files->errnum = bcf_read_error;
             if ( ret < 0 ) break; // no more lines or an error
+            //set phasing of 1st value as in vcf v44
+            if ((ret = update44phasing(reader->header, reader->buffer[reader->nbuffer+1]))) {
```
pd3, line 712: I am curious how this affects speed?

vasudeva8: Initially it was observed to make ~15% overhead. On recheck, last update had more overhead and tweaked further. It shows ~13% overhead with bcftools view.

vasudeva8:
The BCF conversion is removed, to avoid the performance overhead, based on discussions.
1st phasing value in VCF, with version upto v4.3, are set based on rest of the values in the sample.
The API bcf_get_format_values modified to do the same conversion when the retrieved field is GT, limiting any overhead to such cases where GT is retrieved.

With this, BCF and VCF will have different internal binary values/representations.
When the bcf_get_format_values api is used, phasing data as per v4.4 is retrieved.
When the values are accessed directly, w/o the get_format api, user has to determine the phasing for 1st value based on phasing of other alleles.

daviesrob:
Rebased and tidied up...

PR #1939, opened 2025-07-28
Fix segfault on an empty valid MM tag.

### LINKED ISSUE #1936: bam_mods_at_next_pos() segfault for reverse read with degenerate MM tag value
The following file, attached as [MM-empty.sam](https://github.com/user-attachments/files/21287681/MM-empty.sam.txt) and distilled from pysam-developers/pysam#1355, contains an MM field with no modifications listed (which doesn't do much, but is allowed by the spec I believe):

```
@SQ	SN:chr1	LN:2000
r1	0	chr1	1000	20	6M	*	0	0	ATGCAT	abcdef	MM:Z:C+m;
r2	16	chr1	1000	20	6M	*	0	0	ATGCAT	abcdef	MM:Z:C+m;
```

When processing `r2`, which has the reverse flag set, this segfaults inside `bam_mods_at_next_pos()`, e.g.,

```
$ test/test_mod MM-empty.sam
0	A
1	T
2	G
3	C
4	A
5	T
---
Present: m.

===

0	A
1	T
2	G
3	C
4	A
5	T
---
Present: m.
Segmentation fault: 11
```

whitwham:
An MM tag with modifications but no bases (e.g. MM:Z:C+m;) would segfault on the reverse strand. Fixes #1936.

Note that an MM tag with multiple modifications without bases would also fail.

### DIFF
--- a/sam_mods.c
+++ b/sam_mods.c
@@ -501,6 +501,10 @@ int bam_mods_at_next_pos(const bam1_t *b, hts_base_mod_state *state,
         if (b->core.flag & BAM_FREVERSE) {
             // process MM list backwards
             char *cp;
+
+            if (state->MMend[i] == NULL)
+                return 0; // a valid but empty tag (with no other base mods)
+
             if (state->MMend[i]-1 < state->MM[i]) {
                 // Should be impossible to hit if coding is correct
                 hts_log_error("Assert failed while processing base modification states");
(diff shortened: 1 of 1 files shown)


daviesrob:
Yes, it does appear that `bam_mods_at_next_pos()` gets called with `state->seq_pos == b->core.l_qseq` even in the case where the tag has a non-empty list.  Luckily this is mostly harmless as the sequence data is usually followed by quality values so the resulting lookup returns bogus data instead of crashing.  As you note, catching it in `bam_next_basemod()` is probably the right solution.

PR #1942, opened 2025-08-04
For the VCF rlen calculation, only use SVLEN for DEL, DUP and CNV symbolic alleles

### LINKED ISSUE #1940: Error in rlen calculation from SVLEN
The new `get_rlen()` code https://github.com/samtools/htslib/blob/develop/vcf.c#L6168 has severe error, it misinterprets the SVLEN tag, ignoring its sign.

Consider the following VCF records
```
1	5	.	A	AACGTACGTACGTACGT	.	.	.
2	5	.	A	AACGTACGTACGTACGT	.	.	SVLEN=16
```
In both cases the affected reference length should be set to `rlen=1`. However, for the second case it is set to 17.

This issue was discovered in https://github.com/samtools/bcftools/issues/2439.

daviesrob:
PR #1897 updated the VCF rlen calculation in line with VCF 4.4, including taking account of INFO `SVLEN` for non-INS alleles.  This was not correct though as it included non-symbolic alleles if they (incorrectly) had a non-missing `SVLEN`; and other types of symbolic allele that do not affect the length with respect to the reference. In fact, only deletions can have an affect on rlen, so this changes the calculation to only use `SVLEN` for `<DEL>` alleles.

A bug is also fixed on big-endian platforms where INFO and FORMAT values were being accessed incorrectly.  These are always stored in little-endian order, so need to be converted to native ordering when read.  The tests are updated to include values greater than 128 to ensure that this works correctly.

Fixes #1940

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -6173,7 +6173,7 @@ static int64_t get_rlen(const bcf_hdr_t *h, bcf1_t *v)
     bcf_info_t *endinfo = NULL, *svleninfo = NULL, end_lcl, svlen_lcl;
     bcf_fmt_t *lenfmt = NULL, len_lcl;
 
-    //holds <DEL> allele status for the max no of alleles
+    //holds SVLEN allele status for the max no of alleles
     uint8_t svlenals[8192];
     //pos from info END, fmt LEN, info SVLEN
     hts_pos_t end = 0, end_fmtlen = 0, end_svlen = 0, hpos;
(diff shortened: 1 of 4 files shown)


daviesrob:
Following discussion and careful reading of the VCF specification, I've updated the PR so that it also uses `SVLEN` for `<DUP>`, and `<CNV>` (including `<CNV:TR>`) symbolic alleles.

daviesrob:
Now updated to add <INV> to the list of symbolic alleles where SVLEN should be used, along with all subtypes of CNV, DEL, DUP and INV.

vasudeva8 (approved):

vcf.c
```diff
@@ -6230,6 +6230,11 @@ static int64_t get_rlen(const bcf_hdr_t *h, bcf1_t *v)
         size = bcf_dec_size(f, &f, &type);
         f += size << bcf_type_shift[type];
     }
+
+    // Can skip SVLEN lookup if there are no <DEL> symbolic alleles
+    if (!use_svlen)
+        svlenid = -1;
+
```
pd3, line 6237: svlenid is initialized unnecessarily above if use_svlen==0, a table lookup can be avoided if placed here in an `else` statement

daviesrob: True, although I note that we could skip searching for ALT symbolic alleles if there's no `SVLEN` in the header.  That might be quicker, for files that don't use `SVLEN`.

PR #1945, opened 2025-08-14
Fix compilation failure on MacOS X 10.9 (and likely other very old platforms)

### LINKED ISSUE #1941: error: use of undeclared identifier 'AT_SYMLINK_NOFOLLOW'
htslib doesn't build on OS X (macOS) 10.9 and earlier anymore.

htslib 1.21 and earlier did build on such old versions of OS X (macOS), but 1.22.1 fails in this way:

```
ref_cache/log_files.c:84:22: warning: implicit declaration of function 'openat' is invalid in C99 [-Wimplicit-function-declaration]
            log_fd = openat(logfiles->log_dir_fd, name, O_WRONLY | O_CREAT | O_EXCL, 0644);
                     ^
ref_cache/log_files.c:101:9: warning: implicit declaration of function 'unlinkat' is invalid in C99 [-Wimplicit-function-declaration]
        unlinkat(logfiles->log_dir_fd, name, 0);
        ^
ref_cache/log_files.c:218:17: warning: implicit declaration of function 'fstatat' is invalid in C99 [-Wimplicit-function-declaration]
            if (fstatat(logfiles->log_dir_fd,
                ^
ref_cache/log_files.c:219:43: error: use of undeclared identifier 'AT_SYMLINK_NOFOLLOW'
                        ent->d_name, &st, AT_SYMLINK_NOFOLLOW) != 0) {
                                          ^
3 warnings and 1 error generated.
make: *** [ref_cache/log_files.o] Error 1
```

Here is a full build log: https://build.macports.org/builders/ports-10.9_x86_64-builder/builds/311518/steps/install-port/logs/stdio

1.22.1 does build on OS X (macOS) 10.10 and newer.

daviesrob:
Adds a configure check for `fstatat( , , , AT_SYMLINK_NOFOLLOW)` so that the `ref-cache` can be disabled automatically on platforms that lack this interface.  Also adjusts the position of the `BUILDING_SIMD_NIBBLE2BASE` guard in `simd.c` so that it doesn't try to include `immintrin.h` when the code that uses it is not going to be built.

Fixes #1941

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -666,6 +666,33 @@ Linux, FreeBSD, MacOS, and other BSD derivatives])
     ])
   ])
 
+# Check for a working fstatat interface
+AS_IF([test "$ref_cache" = enabled],[
+  AC_MSG_CHECKING([for fstatat])
+  AC_LINK_IFELSE([
+    AC_LANG_PROGRAM([
+      #include <fcntl.h>
+      #include <sys/stat.h>
+    ],[
+      struct stat statbuf;
+      return fstatat(AT_FDCWD, "/", &statbuf, AT_SYMLINK_NOFOLLOW);
+    ])
+  ], [
+    AC_MSG_RESULT([yes])
+  ], [
+    AC_MSG_RESULT([no])
+    ref_cache="disabled"
+    AS_IF([test "x$enable_ref_cache" = xcheck], [
+      AC_MSG_WARN([ref-cache not enabled: missing fstatat()])
+    ],[
+      MSG_ERROR([ref-cache not enabled
+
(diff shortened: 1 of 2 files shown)


daviesrob:
I've pushed an improved version of this with, as suggested, configure checks for the intrinsics header.  The file used has changed from `immintrin.h` to `x86intrin.h` as the latter is the one used by htscodecs.  While the latter isn't strictly the standard one, it is supported by gcc, clang, and recent Intel compilers.

I've also added a check for `__builtin_cpu_supports("ssse3")` on make-only builds so that they don't fall over if the compiler is too old to understand "ssse3" as a valid input for the builtin.

PR #1957, opened 2025-09-30
Switch from strtol to hts_str2uint in mod parsing

cjw85:
This saves around 10-15% in programs calling the modified base APIs.

### DIFF
--- a/sam_mods.c
+++ b/sam_mods.c
@@ -507,15 +507,22 @@ int bam_mods_at_next_pos(const bam1_t *b, hts_base_mod_state *state,
                 if (*cp == ',')
                     break;
             state->MMend[i] = cp;
-            if (cp != state->MM[i])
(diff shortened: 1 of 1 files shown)


sam_mods.c
```diff
@@ -507,15 +518,21 @@ int bam_mods_at_next_pos(const bam1_t *b, hts_base_mod_state *state,
                 if (*cp == ',')
                     break;
             state->MMend[i] = cp;
-            if (cp != state->MM[i])
-                state->MMcount[i] = strtol(cp+1, NULL, 10);
-            else
+            if (cp != state->MM[i]) {
```
jmarshall, line 521: And now there's no need to modify a bunch of surrounding lines by adding `{ … }`.

cjw85: Removed.

PR #1958, opened 2025-10-01
S3 Code Rewrite

whitwham:
Add code to read from S3 sources in the same style as the current writing code. Merge the s3_write.c code into hfile_s3.c to put all the S3 code in the same file.

This should simplify the S3 code and give us a little more flexibility in how we use it.

Still need to remove the old writing code (file removal and other modifications).


hfile_s3.c
```diff
@@ -1113,330 +1074,1281 @@ static int write_authorisation_callback(void *auth, char *request, kstring_t *co
...
+    // Check 0 <= origin+offset < fp->file_size carefully, avoiding overflow
+    if ((offset < 0)? origin + offset < 0
+                : (fp->file_size >= 0 && offset > fp->file_size - origin)) {
+        errno = EINVAL;
+        return -1;
+    }
+
+    fp->keep_going = 1;
+    fp->last_read = origin + offset; // origin is really only useful if we can make the other modes work
+    ks_clear(&fp->buffer); // resetting fp->buffer triggers a new remote read
```
daviesrob, line 1958: It would be nice to work out if the destination is within the range of the existing buffer, and reuse it if that's the case.  It might allow some upstream requests to be avoided.

hfile_s3.c
```diff
@@ -1113,330 +1074,1281 @@ static int write_authorisation_callback(void *auth, char *request, kstring_t *co
...
+    return ret;
+}
+
+
+static ssize_t s3_read(hFILE *fpv, void *bufferv, size_t nbytes) {
+    hFILE_s3 *fp = (hFILE_s3 *)fpv;
+    char *buffer = (char *)bufferv;
+    size_t read = 0;
+    static size_t total = 0;
+    static size_t gets  = 0;
```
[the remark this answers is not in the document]
whitwham: All those static variables were testing code.  I meant to take them out.

hfile_s3.c
```diff
@@ -1113,160 +1075,171 @@ static int write_authorisation_callback(void *auth, char *request, kstring_t *co
...
+*/
+static char *stristr(char *haystack, char *needle) {
+
+    while (*haystack) {
+        char *h = haystack;
+        char *n = needle;
+
+        while (toupper(*h) == toupper(*n)) {
+            h++, n++;
+            if (!h || !n) break;
```
daviesrob: I think you wanted this here:
```suggestion
            if (!*h || !*n) break;
```

PR #1973, opened 2025-11-25
Don't drop in-flight decode jobs when seeking in cram_next_slice()

daviesrob:
`cram_next_slice()` with a region set tries to skip over containers with contents that don't overlap the region of interest.  For files with a mixture of read lengths, it's possible that some decode jobs have been queued before a later container is skipped.  This makes calling `cram_seek()` hazardous as it drops any in-flight jobs before updating the file position.  Instead, call `hseek()` directly so already-queued decode jobs are retained.

Fixes samtools/samtools#2285 (CRAM region queries miss reads when using additional threads)

### DIFF
--- a/test/index3.sam
+++ b/test/index3.sam
@@ -1,0 +1,17 @@
+@HD	VN:1.6	SO:coordinate
+@SQ	SN:CHROMOSOME_I	LN:1009800	M5:8ede36131e0dbf3417807e48f77f3ebd
+@CO	CRAM container skipping test - very long reads followed by short ones.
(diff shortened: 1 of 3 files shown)


jkbonfield:
I tested samtools with it, and also tested the issues reported in https://github.com/samtools/htslib/pull/655 which gave rise to the initial addition of the `cram_drain_rqueue` call.  That has a convenient test which I repeated on a clone of 1.6 to be sure I had a case which broke.  The new code doesn't recreate that problem.  

Looks good.  Thanks

PR #1981, opened 2026-01-12
Added extra overrun detection in BYTE_ARRAY_LEN decoder.

jkbonfield:
Although the input specifies the maximum size of 'out' in *out_side (which is subsequently replaced by the actual size), this wasn't checked for when calling byte_array_len.len_codec->decode to fetch the length field.

The obvious fix did trigger other problems however.  We've been routinely calling this with incorrect parameters when decoding aux tags.  Eg XA:i encoded as XAs (short) would specify out_size of 1, as we're attempting to decode a single item.  This is the standard way all of our decoders work - indicating the number of things to decode rather than their byte lengths.  However by definition this is a "byte array" so we need to compensate for that before calling the decode function.  It wasn't required before (as we didn't check).

### DIFF
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -3357,14 +3357,19 @@ int cram_byte_array_len_decode(cram_slice *slice, cram_codec *c,
     int32_t len = 0, one = 1;
     int r;
 
-    r = c->u.byte_array_len.len_codec->decode(slice, c->u.byte_array_len.len_codec,
-                                              in, (char *)&len, &one);
-    //printf("ByteArray Len=%d\n", len);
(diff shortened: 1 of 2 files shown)


daviesrob:
It looks like there's an implicit assumption here that `B`-type aux tags are stored in external blocks.  I don't think that has to be the case as BYTE_ARRAY_LEN could, in theory, use HUFFMAN as its value codec.  In practice our HUFFMAN decoder does not support output to byte arrays, so the assumption is true but there could be valid files that we cannot read.  As this problem goes beyond the issue being addressed by this PR I'll raise it separately.

PR #1986, opened 2026-02-05
added prefetch and minor updates to speedup pileup

vasudeva8:
Added prefetch to improve pileup performance.
Updated overlap_remove to reduce hash search.

this is for samtools mpileup speed up (PR samtools#2301)

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -6011,7 +6011,7 @@ const bam_pileup1_t *bam_plp64_next(bam_plp_t iter, int *_tid, hts_pos_t *_pos, 
         lbnode_t **pptr = &iter->head;
         while (*pptr != iter->tail) {
             if ((*pptr)->next)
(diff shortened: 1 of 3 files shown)


sam.c
```diff
@@ -6007,6 +6010,8 @@ const bam_pileup1_t *bam_plp64_next(bam_plp_t iter, int *_tid, hts_pos_t *_pos,
         // write iter->plp at iter->pos
         lbnode_t **pptr = &iter->head;
         while (*pptr != iter->tail) {
+            if ((*pptr)->next)
+                __builtin_prefetch((*pptr)->next);
```
jkbonfield: I think maybe we should have an `hts_prefetch` defined somewhere with compiler checks.  `__builtin_prefetch` is a gcc thing, albeit supported on many other platforms and compilers too.

I'll push a new commit containing autoconf changes.

PR #1987, opened 2026-02-12
Add retry/resilience for libcurl HTTP backend

gileshall:
## Summary

- Add automatic retry with exponential backoff for transient HTTP errors (429, 500, 502, 503, 504) and network failures (connection drops, timeouts, partial transfers)
- Add stall detection via `CURLOPT_LOW_SPEED_LIMIT` / `CURLOPT_LOW_SPEED_TIME` to abort and retry transfers that stop making progress
- All retry behavior is configurable via environment variables (`HTS_RETRY_MAX`, `HTS_RETRY_DELAY`, `HTS_RETRY_MAX_DELAY`, `HTS_LOW_SPEED_LIMIT`, `HTS_LOW_SPEED_TIME`) with sensible defaults
- Permanent errors (401, 403, 404) are never retried
- Add comprehensive test suite with a Python mock HTTP server exercising all retry scenarios

## Motivation

When streaming files over HTTP/HTTPS (including GCS `gs://` URLs), htslib currently gives up immediately on any transient network failure. This is problematic for long-running genomics pipelines where occasional 503s, throttling (429), or brief network interruptions are expected. The existing `restart_from_position()` function already supports reconnecting at a byte offset but was only used for seeking, never for error recovery.

## Implementation

**Error classification** (`is_retryable()`): Classifies CURLcodes and HTTP status codes as retryable or permanent.

**Open retries**: `libcurl_open()` retries the initial connection with exponential backoff when the first attempt fails with a retryable error.

**Read retries**: `libcurl_read()` handles mid-transfer failures by returning partial data with a `needs_reconnect` flag, then reconnecting on the next read call. Zero-data failures retry inline.

**Stream position tracking**: `stream_pos` field tracks the current byte offset for accurate reconnection via `restart_from_position()`.

**Configuration refresh**: `refresh_retry_config()` re-reads environment variables at each `libcurl_open()` call, allowing per-request tuning.

## Test plan

- [x] New `test/test_hfile_libcurl` passes all 7 test cases:
  - Normal transfer (baseline)
  - 503 retry succeeds
  - 429 retry succeeds  
  - Connection drop mid-transfer retry
  - 404 not retried (permanent error)
  - Retry exhaustion (fails after max attempts)
  - Retry disabled (`HTS_RETRY_MAX=0`)
- [x] Existing `make check` tests still pass
- [x] Test gracefully skips when python3 or HTTP support unavailable

### DIFF
--- a/hfile_libcurl.c
+++ b/hfile_libcurl.c
@@ -240,6 +240,12 @@ static int is_retryable(CURL *easy, CURLcode err)
     case CURLE_OPERATION_TIMEDOUT:
     case CURLE_GOT_NOTHING:
     case CURLE_SSL_CONNECT_ERROR:
+#ifdef CURLE_HTTP2
+    case CURLE_HTTP2:
+#endif
+#ifdef CURLE_HTTP2_STREAM
+    case CURLE_HTTP2_STREAM:
+#endif
         return 1;
 
     case CURLE_HTTP_RETURNED_ERROR: {


whitwham:
Thank you for your contribution.  Sorry it took so long to merge.

PR #1992, opened 2026-03-26
Move misplaced NULL pointer checks

daviesrob:
Found using Coccinelle.  No problems have been reported due to these, but it is good form to have them in the right place.


jkbonfield:
I'm guessing these are cases that couldn't happen due to error checking elsewhere, given if they did it'd have died horribly.  However yes, it's good for it to make sense!  Thanks.

PR #1993, opened 2026-03-30
Cram explicit tlen

jkbonfield:
- Fix CRAM v4.0 CRAM_FLAG_EXPLICIT_TLEN usage.  We needed to check the `explicit_tlen` variable as well as the cram flags, which led to removal of items from the cram stats and potentially changing data by erroneously believing the values to be constant.
- Fix c->ref_end for embedded references when bizarrely asking to generate a consensus for embedding while also telling it what the external reference is called.  A nonsensical set of options, but they should work.

Both of these show up in io_lib if I switch to using htslib for CRAM reading and writing.

### DIFF
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -1503,19 +1503,19 @@ int next_cigar_op(uint32_t *cigar, uint32_t ncigar, int *skip, int *spos,
 // Ensure ref and hist are large enough.
 static inline int extend_ref(char **ref, uint32_t (**hist)[5], hts_pos_t pos,
                              hts_pos_t ref_start, hts_pos_t *ref_end,
-                             hts_pos_t *max_pos) {
-    if (*max_pos < pos)
-        *max_pos = pos;
+                             hts_pos_t *ref_end_alloc) {
+    if (*ref_end < pos)
+        *ref_end = pos;
     if (pos < ref_start)
         return -1;
-    if (pos < *ref_end)
+    if (pos < *ref_end_alloc)
         return 0;
 
     // realloc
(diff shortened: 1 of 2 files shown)


cram/cram_encode.c
```diff
@@ -1502,7 +1502,10 @@ int next_cigar_op(uint32_t *cigar, uint32_t ncigar, int *skip, int *spos,
 
 // Ensure ref and hist are large enough.
 static inline int extend_ref(char **ref, uint32_t (**hist)[5], hts_pos_t pos,
-                             hts_pos_t ref_start, hts_pos_t *ref_end) {
+                             hts_pos_t ref_start, hts_pos_t *ref_end,
+                             hts_pos_t *max_pos) {
```
daviesrob: It might be an idea to rename some of these variables.  Currently `max_pos` is what ends up in `c->ref_end`, and `ref_end` is the point where the oversized allocated buffer ends.  This leads to confusion elsewhere, for example [the consensus calculation](https://github.com/jkbonfield/htslib/blob/37f55227c89247d893e82a501e86e5de64511d2f/cram/cram_encode.c#L1762) hasn't been updated so it runs over the whole buffer instead of just the bit that was filled in.

I think it might be better to make the new variable track the allocated size and change `ref_end` to work like `max_pos` does here.  That would mean a few more changes here, and reverting the change in line [1778](https://github.com/samtools/htslib/pull/1993/changes#diff-c2154d53506fa922a43b72ee1c8eb06e566840743e340590096f193897d0127eR1778), but the benefit would be that `ref_end` would start working more like it does elsewhere for external references.

PR #1994, opened 2026-03-30
Ensure indirect function calls have the correct type.

daviesrob:
Some HTSlib interfaces (notably `kgetline()`, `kgetline2()`, `hts_itr_querys()` and `hts_parse_region()`) have function callbacks that are intended to be generic, so the function signatures include a `void *` for data to be passed in.  E.g. for `kgetline()` this would be the `FILE *` or `hFILE *` to read.  Historically they have often been called with a function that has an explicit type (e.g. `FILE *` for `fgets()` used with `kgetline`) on the assumption that casting it to `void *` is harmless. Unfortunately doing this is strictly undefined behaviour and will fail in some conditions - for example if compiling with control flow integrity turned on.  Recent versions of clang will also call this out when using undefined behaviour sanitizer.  The solution is to create wrapper functions that do take a `void *`, and call the intended function with suitable casts.

Some new interfaces are also added so that the messy implementation details can be hidden away, and the inputs can be better type checked. Hopefully this should also allow the wrapper functions to be inlined away.

 * `hget_kstr()` to read lines from an `hFILE *` to a `kstring`.

 * `kgetline()` to read lines from a `FILE *` to a `kstring`.

 * `tbx_itr_querys1()` to create a tabix iterator on a region. Used to reimplement the existing `tbx_itr_querys()` macro.

 * `bcf_itr_querys1()` to create a region iterator an a bcf file Used to reimplement the existing `bcf_itr_querys()` macro.

### DIFF
--- a/htslib/hfile.h
+++ b/htslib/hfile.h
@@ -211,6 +211,19 @@ kstring's `kgetline()` to read arbitrarily-long lines into a _kstring_t_.
 HTSLIB_EXPORT
 char *hgets(char *buffer, int size, hFILE *fp) HTS_RESULT_USED;
 
+/// Read a line from a stream and append it to a kstring
+/** @param kstr    The destination kstring
+    @param fp      The file stream
+    @return  0 on success; EOF on end of file or if an error occurred.
+    @since   1.24
+
+Reads a "\n"- or "\r\n"- terminated line from fp into kstr.
+The line read is appended without its terminator and 0 is returned;
+EOF is returned at end of file or on error.
+*/
+HTSLIB_EXPORT
+int khgetline(struct kstring_t *kstr, hFILE *fp) HTS_RESULT_USED;
+
 /// Peek at characters to be read without removing them from buffers
 /** @param fp      The file stream
     @param buffer  The buffer to which the peeked bytes will be written
--- a/faidx.c
+++ b/faidx.c
@@ -1022,7 +1022,7 @@ const char *fai_parse_region(const faidx_t *fai, const char *s,
                              int *tid, hts_pos_t *beg, hts_pos_t *end,
                              int flags)
 {
-    return hts_parse_region(s, tid, beg, end, (hts_name2id_f)fai_name2id, (void *)fai, flags);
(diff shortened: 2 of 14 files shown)


htslib/hfile.h
```diff
@@ -211,6 +211,19 @@ kstring's `kgetline()` to read arbitrarily-long lines into a _kstring_t_.
...
+    @param fp      The file stream
+    @return  0 on success; EOF on end of file or if an error occurred.
+    @since   1.24
+
+Reads a "\n"- or "\r\n"- terminated line from fp into kstr.
+The line read is appended without its terminator and 0 is returned;
+EOF is returned at end of file or on error.
+*/
+HTSLIB_EXPORT
+int hget_kstr(struct kstring_t *kstr, hFILE *fp) HTS_RESULT_USED;
```
jkbonfield: Why `hget_kstr` and not `khgetline`?

We have a new `kfgetline` that works via `fgets`, so `khgetline` using `hgets` seems like the more memorable function name.  Either that or rename 'kfgetline' to `fget_kstr`, but the getline ones seem more appropriate.

jkbonfield (commented):
Looks OK and I can confirm the errors when built using `clang21 -fsanitize=kcfi` go away with this PR.  (I'm assuming there are samtools and bcftools updates coming too.)

My only query is on the function naming.

Note: you also have a typo in your commit message.  

"kgetline() to read lines from a FILE * to a kstring" should read
"kfgetline() to read lines from a FILE * to a kstring"

htslib/hfile.h
daviesrob: The name was a modification of `hgets()` above, which itself derives from `fgets()`.  `kgetline()` of course derived from the POSIX `getline()` function. I guess I could make it `khgetline()`, although having two functions that differ by one letter could be its own annoyance...

daviesrob: Actually, the return value better matches that for `getline()`, so I'll change it.

daviesrob:
`hget_kstr()` has been renamed as `khgetline()` and the commit message fixed.

Issue #1949, opened 2025-09-02
tabix performance


pd3:
It was reported offline by Kees Albers that the VCF index performs poorly on very large data. In fact, when chatGPT was asked to write a bgzip parser and then the VCF was queried directly using a simple binary search algorithm, this solution was much much faster than using tabix.

The test data (available offline)
- 7.4 GB in size and contains 6807 variants for ~500K samples
- the tabix index size is 180 bytes

Test machine
- 8 cores, 16 GB of RAM, SSD disk

Tabix index performance
- 86 seconds

Direct binary search 
- 0.17 seconds

[varseek.txt](https://github.com/user-attachments/files/22093155/varseek.txt)

PR #1997, opened 2026-04-06
Add multi-region iterator support for tabix

### LINKED ISSUE #1913: Implement multi-region iterators for tabix
As noted in https://github.com/samtools/htslib/issues/1470#issuecomment-1181508413, HTSlib has tabix iterators, and it has multi-region iterators for BAM and CRAM, but it does not have multi-region tabix iterators.

The important thing about multi-region iterators, and why they can't be easily replicated with repeated single-region queries, is that they resolve [overlapping regions](https://github.com/samtools/htslib/issues/785#issuecomment-433869384) and [results present in multiple regions](https://github.com/samtools/htslib/issues/785#issuecomment-464135842) for you, efficiently. The algorithm required to deduplicate results when you only have a single-region query involves a lot of fiddly parsing of the results and reasoning about when you can forget the deduplication information, and even then might run into trouble when the underlying file contains distinct duplicate entries that you want to see as many times as they actually occur.

Since htslib supports tabix for GAF files containing reads as of @jmonlong's work in https://github.com/samtools/htslib/pull/1763, the lack of multi-region tabix support means there's no good way to query reads from multiple regions from a tabix-indexed GAF, without getting duplicate results. A similar issue comes up with tabix-indexed SAM, which @jkbonfield talks about wanting to use for some reason in https://github.com/samtools/htslib/issues/1902#issuecomment-2771803259.

The tabix iterators are here, and they're all macros around the generic `hts_itr_*` functions:
https://github.com/samtools/htslib/blob/a4dde3f93c75dff09e9a1ddaf0db2e54c91ac5e3/htslib/tbx.h#L59-L63

But to make a multi-region iterator, you need to define an `hts_itr_multi_query_func`, the requirements/expected behavior of which are not documented:
https://github.com/samtools/htslib/blob/a4dde3f93c75dff09e9a1ddaf0db2e54c91ac5e3/htslib/hts.h#L1322-L1344

You also need to obtain an `hts_seek_func`, an `hts_tell_func`, and an `hts_readrec_func`. For tabix there's only the `tbx_readrec` function defined right now.

To implement tabix multi-iterators, someone needs to:
- Figure out what the `hts_itr_multi_query_func` functions are supposed to do
- Figure out if `hts_itr_multi_tabix` is a thing that can be implemented, or whether different implementations for the different tabix-indexable files (`hts_itr_multi_vcf`, `hts_itr_multi_gaf`, `tbx_itr_multi_sam`, etc.) are required.
- Implement the query function(s), or make the documentation sufficient that library users can implement them for their own tabix-indexed file types.
- Implement the seek and tell functions for tabix, or make it clearer that the user is intended to write these themselves.

HeyangQin:
## Summary

Closes #1913 (and addresses the multi-region query aspect of #1949).

This adds `tbx_itr_regions()` and `tbx_itr_multi_next()`, giving tabix-indexed files (VCF, BED, GFF, GAF, etc.) the same multi-region query capability that BAM/CRAM already have via `sam_itr_regions()`.

The key feature is **automatic deduplication**: when query regions overlap, each record is returned exactly once. This is critical for use cases like "query all known cancer genes" where gene regions may be adjacent or overlapping.

## Design

The core insight is that tabix and BAM share the same binning index format (TBI/CSI ↔ BAI/CSI), so `hts_itr_multi_bam` works directly as the query function for tabix — no new query function needed.

The one interface gap was that `hts_itr_multi_next()` hardcodes `htsFile*` as the data argument to `readrec`, but `tbx_readrec` expects `tbx_t*`. This is resolved by adding a `void *readrec_data` field to `hts_itr_t`:

- When set (by `tbx_itr_regions`), it is passed to `readrec` instead of `fd`
- When NULL (default, from `calloc`), existing BAM/CRAM behaviour is preserved
- One field addition, three line changes in `hts_itr_multi_next` — fully backwards-compatible

## Usage

```c
// Parse regions (merges overlaps automatically)
int n_reg;
hts_reglist_t *reglist = hts_reglist_create(regions, n_regions,
    &n_reg, tbx, (hts_name2id_f)(tbx_name2id));

// Create multi-region iterator
hts_itr_t *itr = tbx_itr_regions(tbx, reglist, n_reg);

// Iterate — each record returned at most once
kstring_t s = {0, 0, NULL};
while (tbx_itr_multi_next(fp, itr, &s) >= 0) {
    // process s.s
}
tbx_itr_destroy(itr);
```

## Changes

| File | Change |
|------|--------|
| `htslib/hts.h` | Add `void *readrec_data` to `hts_itr_t` |
| `hts.c` | Use `readrec_data` when set in `hts_itr_multi_next` (3 call sites) |
| `htslib/tbx.h` | Add `tbx_itr_regions()` declaration, `tbx_itr_multi_next()` macro |
| `tbx.c` | Implement `tbx_itr_regions()`, `tbx_pseek()`, `tbx_ptell()` |
| `htslib.map` | Export `tbx_itr_regions` |

## Test plan

- [x] htslib builds cleanly with no new warnings
- [x] Tested with a VCF containing 7 variants across 2 chromosomes:
  - Single-region query: correct record count
  - Non-overlapping multi-region: records from both regions returned
  - Overlapping regions (chr1:100-300 + chr1:200-400): 4 unique records, no duplicates
  - Cross-chromosome with overlap (3 regions, 2 overlapping): 7 unique records
- [x] samtools builds and passes markdup tests against updated htslib
- [x] Existing BAM/CRAM multi-region behaviour unaffected (readrec_data defaults to NULL)

### DIFF
--- a/htslib/hts.h
+++ b/htslib/hts.h
@@ -886,6 +886,7 @@ typedef struct hts_itr_t {
     hts_readrec_func *readrec;
     hts_seek_func *seek;
     hts_tell_func *tell;
+    void *readrec_data; // If set, passed to readrec instead of fd
     struct {
         int n, m;
         int *a;
--- a/hts.c
+++ b/hts.c
@@ -4320,6 +4320,7 @@ int hts_itr_next(BGZF *fp, hts_itr_t *iter, void *r, void *data)
 int hts_itr_multi_next(htsFile *fd, hts_itr_t *iter, void *r)
 {
     void *fp;
+    void *readrec_data;
     int ret, tid, i, cr, ci;
     hts_pos_t beg, end;
     hts_reglist_t *found_reg;
@@ -4332,6 +4333,8 @@ int hts_itr_multi_next(htsFile *fd, hts_itr_t *iter, void *r)
         fp = fd->fp.bgzf;
     }
 
+    readrec_data = iter->readrec_data ? iter->readrec_data : fd;
+
     if (iter->read_rest) {
         if (iter->curr_off) { // seek to the start
             if (iter->seek(fp, iter->curr_off, SEEK_SET) < 0) {
@@ -4341,7 +4344,7 @@ int hts_itr_multi_next(htsFile *fd, hts_itr_t *iter, void *r)
             iter->curr_off = 0; // only seek once
         }
 
-        ret = iter->readrec(fp, fd, r, &tid, &beg, &end);
+        ret = iter->readrec(fp, readrec_data, r, &tid, &beg, &end);
         if (ret < 0)
             iter->finished = 1;
 
@@ -4425,7 +4428,7 @@ int hts_itr_multi_next(htsFile *fd, hts_itr_t *iter, void *r)
                     // contain a few mapped reads, so scroll
                     // forward until finding the first unmapped read.
                     do {
-                        ret = iter->readrec(fp, fd, r, &tid, &beg, &end);
+                        ret = iter->readrec(fp, readrec_data, r, &tid, &beg, &end);
                     } while (tid >= 0 && ret >=0);
 
                     if (ret < 0)
@@ -4531,7 +4534,7 @@ int hts_itr_multi_next(htsFile *fd, hts_itr_t *iter, void *r)
             }
         }
 
-        ret = iter->readrec(fp, fd, r, &tid, &beg, &end);
+        ret = iter->readrec(fp, readrec_data, r, &tid, &beg, &end);
         if (ret < 0) {
             if (iter->is_cram && cram_eof(fp)) {
                 // Skip to end of range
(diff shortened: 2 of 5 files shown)


htslib/hts.h
```diff
@@ -886,6 +886,7 @@ typedef struct hts_itr_t {
     hts_readrec_func *readrec;
     hts_seek_func *seek;
     hts_tell_func *tell;
+    void *readrec_data; // If set, passed to readrec instead of fd
```
daviesrob, line 889: This changes the public `hts_itr_t` structure, so would require an ABI bump if we merged it. Ideally we'd like to avoid that.

The reason `readrec_data` is here is that the second parameter to `tbx_readrec()` is a pointer to the `tbx_t` struct, which is not like the way it's used in `hts_itr_multi_next()` where the second parameter is an `htsFile`.

A way around this might be to smuggle the data in using `htsFile::state`.  It would requite a new `tbx_readrec()` function though, which knew to look there for the `tbx_t` struct.

PR #1998, opened 2026-04-14
fix(performance): Use bounded HTTP Range requests for indexed BAM queries

TechIsCool:
## Problem

When reading remote BAM files with an index, htslib seeks to each chunk's start offset but issues unbounded Range requests. The server advertises gigabytes of Content-Length even though we only need kilobytes:

 | Seek | Unbounded Request | Server Advertises | Actual Data Required |
  |------|-------------------|-------------------|----------------------|
  | chr1 | `bytes=8224425-` | 17.2 GB | 141 KB |
  | chr2 | `bytes=1631423494-` | 15.6 GB | 123 KB |
  | chr7 | `bytes=7287649006-` | 9.9 GB | 167 KB |

The client terminates early, but "early termination" isn't free - data in flight still transfers. We have also found that being specific about what is needed improves S3 responsiveness.

## Solution

The BAM index already contains chunk end offsets. Pass them through to the HTTP layer:

```
hts_itr_next()
  → bgzf_seek_limit(fp, chunk.start, SEEK_SET, chunk.end)  // NEW: chunk.end
    → hfile_set_readahead_limit(fp->fp, compressed_limit)
      → CURLOPT_RANGE "bytes=X-Y" instead of CURLOPT_RESUME_FROM_LARGE
```

```bash
samtools view --verbosity 10 -X <bam> <bai> <regions> 2>&1 | grep "Range:"

# Before: Range: bytes=7287649006-
# After:  Range: bytes=7287649006-7287816262
```

## EC2 Benchmark

(35 MB/s bandwidth)
**Environment:** EC2 m8azn.medium (up to 25 Gbps bandwidth), us-east-1  
**Test file:** `s3://1000genomes/.../NA12878.mapped.ILLUMINA.bwa.CEU.exome.20121211.bam` (17.3 GB)  
**Measurement:** Wall clock time + actual wire transfer via `/sys/class/net/<iface>/statistics/rx_bytes`

It appears S3 optimizes resource allocation for bounded requests, leading to much faster responses. The time improvement exceeds bandwidth savings, suggesting that S3 can serve bounded requests more efficiently.

| Query | Unbounded | Bounded | Bandwidth | Time |
|-------|-----------|---------|-----------|------|
| 1 region | 1.04 MB, 0.67s | 0.72 MB, 0.38s | 31% less | **1.8x faster** |
| 5 regions | 1.94 MB, 1.33s | 1.28 MB, 0.45s | 34% less | **2.9x faster** |
| 10 regions | 3.88 MB, 2.19s | 2.55 MB, 1.11s | 34% less | **2.0x faster** |
| chr22 (275 MB) | 275.8 MB, 7.17s | 275.2 MB, 5.07s | ~same | **1.4x faster** |

#### Per-Request Comparison (5 regions)

| Seek | Unbounded (before) | Bounded (after) |
|------|-------------------|-----------------|
| | `Range: bytes=X-` | `Range: bytes=X-Y` |
| chr1 | Content-Length: 17.2 GB | Content-Length: 141 KB |
| chr2 | Content-Length: 15.6 GB | Content-Length: 123 KB |
| chr3 | Content-Length: 14.3 GB | Content-Length: 155 KB |
| chr5 | Content-Length: 12.2 GB | Content-Length: 122 KB |
| chr7 | Content-Length: 9.9 GB | Content-Length: 167 KB |
| **Total Advertised** | **69.2 GB** | **708 KB** |

## Local Benchmark 

(2 MB/s bandwidth)
**Environment:** MacOS M1 (up to 100Mbps), California  
**Test file:** `s3://1000genomes/.../NA12878.mapped.ILLUMINA.bwa.CEU.exome.20121211.bam` (17.3 GB)  

| Regions | Unbounded | Bounded | Speedup |
|---------|-----------|---------|---------|
| 1 region | 2.6s | 2.5s | 1.04x |
| 5 regions | 7.0s | 4.5s | 1.55x |
| 10 regions | 13.0s | 7.5s | 1.74x |

## Reproduction

```bash
# Measure wall time and wire transfer (on EC2/Linux)
IFACE=$(ls /sys/class/net/ | grep -v lo | head -1)
BEFORE=$(cat /sys/class/net/$IFACE/statistics/rx_bytes)
START=$(date +%s.%N)
samtools view --verbosity 10 -X \
  https://s3.amazonaws.com/1000genomes/phase3/data/NA12878/exome_alignment/NA12878.mapped.ILLUMINA.bwa.CEU.exome.20121211.bam \
  https://s3.amazonaws.com/1000genomes/phase3/data/NA12878/exome_alignment/NA12878.mapped.ILLUMINA.bwa.CEU.exome.20121211.bam.bai \
  1:1000000-1000100 2:5000000-5000100 3:10000000-10000100 \
  5:50000000-50000100 7:117188547-117188800 >/dev/null 2>&1
END=$(date +%s.%N)
AFTER=$(cat /sys/class/net/$IFACE/statistics/rx_bytes)
echo "Time: $(echo "$END - $START" | bc)s, Bytes: $((AFTER - BEFORE))"
```

### DIFF
--- a/bgzf.c
+++ b/bgzf.c
@@ -2257,6 +2257,32 @@ int64_t bgzf_seek(BGZF* fp, int64_t pos, int where)
     return bgzf_seek_common(fp, pos >> 16, pos & 0xFFFF);
 }
 
+int64_t bgzf_seek_limit(BGZF* fp, int64_t pos, int where, int64_t limit)
+{
+    if (fp->is_write || where != SEEK_SET || fp->is_gzip) {
+        fp->errcode |= BGZF_ERR_MISUSE;
+        return -1;
+    }
+
+    fp->seeked = pos;
+
+    // Perform the seek first - bgzf_seek_common calls hseek which clears readahead_limit
+    int64_t ret = bgzf_seek_common(fp, pos >> 16, pos & 0xFFFF);
+    if (ret < 0)
+        return ret;
+
+    // Set readahead limit hint AFTER seek (hseek clears it, so must be set after)
+    // This enables bounded Range requests for remote backends
+    if (limit > 0 && fp->fp) {
+        off_t compressed_limit = limit >> 16;
+        // Add some buffer for BGZF block overhead (~64KB worst case)
+        compressed_limit += 65536;
+        hfile_set_readahead_limit(fp->fp, compressed_limit);
+    }
+
+    return ret;
+}
+
 int bgzf_is_bgzf(const char *fn)
 {
     uint8_t buf[16];
--- a/hfile.c
+++ b/hfile.c
@@ -126,6 +126,7 @@ hFILE *hfile_init(size_t struct_size, const char *mode, size_t capacity)
     fp->limit = &fp->buffer[capacity];
 
     fp->offset = 0;
+    fp->readahead_limit = 0;
     fp->at_eof = 0;
     fp->mobile = 1;
     fp->readonly = (strchr(mode, 'r') && ! strchr(mode, '+'));
@@ -149,6 +150,7 @@ hFILE *hfile_init_fixed(size_t struct_size, const char *mode,
     fp->limit = &fp->buffer[buf_size];
 
     fp->offset = 0;
+    fp->readahead_limit = 0;
     fp->at_eof = 1;
     fp->mobile = 0;
     fp->readonly = (strchr(mode, 'r') && ! strchr(mode, '+'));
@@ -484,7 +486,13 @@ off_t hseek(hFILE *fp, off_t offset, int whence)
     fp->at_eof = 0;
 
     fp->offset = pos;
+    fp->readahead_limit = 0;  // Clear hint after seek
     return pos;
+}
+
+void hfile_set_readahead_limit(hFILE *fp, off_t limit)
+{
+    fp->readahead_limit = limit;
 }
 
 int hclose(hFILE *fp)
--- a/hfile_libcurl.c
+++ b/hfile_libcurl.c
@@ -1138,9 +1138,8 @@ static int restart_from_position(hFILE_libcurl *fp, off_t pos) {
     int update_headers = 0;
     int save_errno = 0;
 
-    // TODO If we seem to be doing random access, use CURLOPT_RANGE to do
-    // limited reads (e.g. about a BAM block!) so seeking can reuse the
-    // existing connection more often.
+    // When readahead_limit is set (via hfile_set_readahead_limit from BAM index),
+    // use CURLOPT_RANGE for bounded reads instead of reading to EOF.
 
     // Get new headers from the callback (if defined).  This changes the
     // headers in fp before it gets duplicated, but they should be have been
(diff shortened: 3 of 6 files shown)


daviesrob:
This changes the public `hFILE` structure, so it would require an ABI bump if merged.  It would be good to avoid that if possible.

This problem has already been fixed in the `s3` plug-in, albeit in a different way.  If you try using `s3://` URLs instead of `https://`, you'll find that the amount of data requested is much smaller, and the time taken is similar to what you report for your solution.

TechIsCool:
Thanks for the review.

I am unsure if the PR could exist without impacting ABI?

For context we use `https://` URLs because our workflow relies on pre-signed S3 URLs - the `s3://` protocol requires IAM credentials which aren't available in our environment. I looked into what you suggested by reading up on `hfile_s3.c` and it appears to use a fixed 1MB `part_size`. I'm happy to extend this PR to pass `readahead_limit` through the s3 plugin as well. 

## Benchmarks

I tested `s3://` vs `https://` on EC2 (m8azn.medium, us-east-1). The `s3://` plugin uses a fixed 1MB `part_size`, which leads to more data transfer and more HTTP requests for larger queries:

| Region | S3 | HTTPS | S3 GETs | HTTPS GETs |
|--------|-----|-------|---------|------------|
| 1:1000000-1000100 | 0.62s, 3.26 MB | 0.27s, 705 KB | 4 | 4 |
| 2:5000000-5000100 | 0.28s, 3.25 MB | 0.39s, 683 KB | 4 | 4 |
| 3:10000000-10000100 | 0.22s, 3.25 MB | 0.33s, 726 KB | 4 | 4 |
| 5:50000000-50000100 | 0.30s, 3.25 MB | 0.29s, 686 KB | 4 | 4 |
| 7:117188547-117188800 | 0.29s, 3.25 MB | 0.37s, 712 KB | 4 | 4 |
| 22:20000000-25000000 | 11.5s, ~42 MB | 6.7s, ~42 MB | 44 | 4 |

For small regions, **GET** counts are similar but S3 transfers ~5x more data (fixed 1MB chunks). For larger regions, the 1MB chunking causes 44 separate requests vs 1 bounded request.


## Reproduction

```bash
IFACE=$(ls /sys/class/net/ | grep -v lo | head -1)
REGIONS="1:1000000-1000100 2:5000000-5000100 3:10000000-10000100 5:50000000-50000100 7:117188547-117188800"
BAM="1000genomes/phase3/data/NA12878/exome_alignment/NA12878.mapped.ILLUMINA.bwa.CEU.exome.20121211.bam"

test_region() {
    local bam=$1 bai=$2 region=$3
    local before=$(cat /sys/class/net/$IFACE/statistics/rx_bytes)
    local start=$(date +%s.%N)
    samtools view -X "$bam" "$bai" "$region" >/dev/null 2>&1
    local elapsed=$(echo "$(date +%s.%N) - $start" | bc)
    echo "$region: ${elapsed}s, $(($(cat /sys/class/net/$IFACE/statistics/rx_bytes) - before)) bytes"
}

echo "=== S3 ==="
for r in $REGIONS; do test_region "s3://$BAM" "s3://$BAM.bai" "$r"; done

echo "=== HTTPS ==="
for r in $REGIONS; do test_region "https://s3.amazonaws.com/$BAM" "https://s3.amazonaws.com/$BAM.bai" "$r"; done
```

---

**Additional finding:** On slower links, `s3://` can show lower latency than `https://` for small queries due to connection reuse. The s3 plugin reuses connections while `hfile_libcurl.c` creates new connections on each seek.

Adding `CURL_LOCK_DATA_CONNECT` to the share object in `hfile_libcurl.c` brings HTTPS latency in line with S3 while maintaining the bandwidth savings. This could be a separate improvement.

daviesrob:
I can see why you'd want to use `https://` if you only have pre-signed URLs.

I think it may be possible to do something like this while keeping the ABI backwards-compatible, but it would need some care.  The best place to put the new interface would be in an extended `hFILE_backend` structure, but you'd also need a flag to say that it's safe to call as unfortunately `hFILE_backend` doesn't itself have any way of announcing extensions.  The flag could be added to the hFILE bit field as it has plenty of spare space, but it would need to be set explicitly by `hopen()` (and hopen-like functions) as there's currently no guarantee about how `hfile_init()` called by a plug-in would set it.  `hopen()` could find out how to set the flag by checking the `priority` in the plug-in's `hFILE_scheme_handler`, which thankfully does support backwards-compatible extensions.

PR #2000, opened 2026-04-24
Add Github actions builds for Linux and Mac OS

daviesrob:
As Cirrus CI is closing down, tests need to migrate elsewhere.  As Windows tests already run using Github actions, using it for Linux and Mac OS builds is the easiest solution.

Adds Github actions workflows for Linux (specifically ubuntu as it's the only flavour supported natively), and Mac OS.

Updates actions/checkout on the Windows build to the latest release.  The reference to it is also changed from a tag name to a sha hash to reduce the risk of supply-chain attack.

Fixes some issues with the existing tests found while checking the new builds:

* Adjusts `HTS_PROG_CC_WARNINGS` and `HTS_PROG_CC_WERROR` to remove a directory left behind by the Mac OS compiler, and to redirect compiler messages to the log file.

* Remove quotes around `$CC` invocations in `HTS_PROG_CC_WARNINGS` and `HTS_PROG_CC_WERROR`.  `AC_PROG_CC` may add flags after the C compiler path to `$CC`, so it needs to be used unquoted so that field splitting can be done on the value.

* Makes `test_hfile_libcurl` work when `--enable-plugins` is used.

The Cirrus CI tests are left enabled for now so we can run both side-by-side for a bit.  They can be removed once we're confident that the new ones are working.  Unfortunately will will lose the Rock Linux test when we turn Cirrus off, although it may be possible to do something with containers to get it back again.

### DIFF
--- a/m4/hts_prog_cc_warnings.m4
+++ b/m4/hts_prog_cc_warnings.m4
@@ -65,75 +65,76 @@ dnl an option that includes a hash sign...
         # Tests for flags to enable C compiler warnings
         # GCC compatible
         AS_IF([test "x$GCC" = "xyes" &&
-               "$CC" -c -Wall conftest.c > /dev/null 2>&1 &&
+               $CC -c -Wall conftest.c >&AS_MESSAGE_LOG_FD 2>&1 &&
                test -f conftest.o],[dnl
           AS_IF([test "x$ansi" = "x"],
                 [hts_cv_prog_cc_warnings="-Wall"],
                 [hts_cv_prog_cc_warnings="-Wall -ansi -pedantic"])
         ],
         # Sun Studio or Solaris C compiler
-        ["$CC" -V 2>&1 | $GREP -i -E "WorkShop|Sun C" > /dev/null 2>&1 &&
-         "$CC" -c -v -Xc conftest.c > /dev/null 2>&1 &&
+        [$CC -V 2>&1 | $GREP -i -E "WorkShop|Sun C" >&AS_MESSAGE_LOG_FD 2>&1 &&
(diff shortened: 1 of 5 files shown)


m4/hts_prog_cc_warnings.m4
```diff
@@ -65,75 +65,76 @@ dnl an option that includes a hash sign...
...
+         $CC -c -fullwarn -ansi -ansiE conftest.c >&AS_MESSAGE_LOG_FD 2>&1 &&
          test -f conftest.o],[dnl
            AS_IF([test "x$ansi" = "x"],
                  [hts_cv_prog_cc_warnings="-fullwarn"],
                  [hts_cv_prog_cc_warnings="-fullwarn -ansi -ansiE"])
           ],
         # HP-UX C compiler
-        [what "$CC" 2>&1 | $GREP -i "HP C Compiler" > /dev/null 2>&1 &&
-         "$CC" -c -Aa +w1 conftest.c > /dev/null 2>&1 &&
+        [what "$CC" 2>&1 | $GREP -i "HP C Compiler" >&AS_MESSAGE_LOG_FD 2>&1 &&
```
jkbonfield, line 107: I don't know the hpux `what` command, but I assume it's SCCS.   In that case quoting would also fail if CC is a compund command with arguments.  I wouldn't expect it to work without a fully qualified pathname either.   Although frankly I've no idea if anyone uses hp-ux these days!

SCCS what takes multiple files:  `what file ...`.  So removing quotes would essentially check "gcc" and "-some-arg" filenames, permitting the grep to then work.  I suspect the quoting needs to go too, but I have no way of testing it.

jkbonfield (approved):

PR #2001, opened 2026-05-01
Add opt-in planned parser for VCF FORMAT fields

jhl-oai:
## Summary

**TL;DR:** A fast-path VCF FORMAT parser for well-defined FORMAT fields results in large speedups for GT-heavy VCF parsing: on real 1000G chr22 GT-only input, `test_view` improves from 25.08s to 10.27s (2.44x), and downstream `bcftools view -Ob` and `bcftools query` improves by 2-2.56x user time across `--threads 0/2/4`.  On more general mixed FORMAT cohort inputs, the benefit is smaller but consistent: about 1.07-1.14x on GATK-output-like conversion/filter and reordered likelihood workloads, with unsupported shapes falling back to parity.

This PR adds an opt-in planned FORMAT parser for VCF input, enabled with:

```sh
HTS_VCF_FORMAT_PLAN=1
```

The default remains the existing generic FORMAT parser.  When enabled, htslib compiles the FORMAT column into a cached per-tag plan and uses a composable executor for supported rows.  If any compile-time or row-local invariant is not met, the whole FORMAT column falls back to the generic parser so output remains byte-identical.

`HTS_VCF_FORMAT_PLAN` and `HTS_VCF_FORMAT_PLAN_STATS` are temporary review/development controls.  They are useful for validating this branch and measuring fallback behavior, but are not intended to become long-term public API; they should be removed or replaced before making the planned parser unconditional/default behavior.

## Motivation

VCF FORMAT parsing is a hot path for large cohort files because each record may carry thousands of per-sample fields.  The generic parser has to support every legal and malformed row shape, but many real datasets repeatedly use a small set of well-described FORMAT patterns.  This PR targets those common cases without requiring exact FORMAT-string kernels. When a row cannot be handled by the planner, it falls back to the generic parser. 

The planned parser covers the shapes typically seen in real data; in particular, we evaluate: GT-only rows from 1000G-style genotype files, raw variant calls from the NYGC 3202 sample 1000 genomes resequencing effort, GT plus integer vectors such as `AD` and `PL`, scalar fields such as `DP` and `GQ`, allele-count-dependent `Number=A/R/G` fields, reordered FORMAT columns, and common string annotations such as phase-set/filter fields.  These are enough to accelerate the rows that dominate sample-rich real data while keeping uncommon or hard-to-match cases on the generic parser.

The fallback model is the main safety property: unsupported schemas, unusually wide strings, malformed fields, and string-plus-float layouts that do not benefit from planning are handled by the existing parser.  That lets the optimization cover many common production cases while preserving the flexibility required to support the VCF specification. 

## Added Functionality

This PR adds an alternate FORMAT parsing path for VCF text records. Instead of matching complete FORMAT strings, it builds a compact per-tag description from the FORMAT column and the VCF header: header id, declared type, cardinality model, and whether the field needs row-local width measurement. That description is cached on the header, so repeated FORMAT layouts can skip repeated tokenization and metadata lookup.

At read time, the planner resolves record-dependent widths such as `Number=A`, `Number=R`, `Number=G`, bounded `Number=.` vectors, and string fields. The executor then parses the sample columns from those descriptors and writes the same packed BCF FORMAT representation produced by the existing parser. Because this is tag/type based rather than exact-string based, layouts such as `GT:AD`, `GT:AD:DP:PL`, reordered fields, and supported supersets use the same machinery.

The planned path is conservative, such that it only attempts to handle common cohort-VCF value classes, including simple diploid genotypes, integer and float scalars/vectors, bounded strings, and sample subsetting via `bcf_hdr_set_samples()`.

If the header is unsynchronized, a tag is undefined or duplicated, a row has an unsupported shape, a value cannot be parsed in the supported subset, or row-local validation fails, the record rolls back and the existing generic parser handles the whole FORMAT column. `HTS_VCF_FORMAT_PLAN_STATS=1` only exposes temporary review diagnostics such as planned rows and fallback reasons.

The implementation is organized into a few internal pieces: 

- opt-in controls and diagnostics, 
- a header-owned FORMAT plan cache, 
- a tag/type plan compiler, row-local width resolution, 
- per-type value parsers, 
- and an executor that writes packed BCF FORMAT data or rolls back to the existing parser. 

Most of the complexity is inn preserving byte-identical BCF encoding, selected-sample handling, and safe fallback. The remaining changes are focused test fixtures, a cache invalidation test, and small `test_view`/test harness hooks.

## Test Coverage

```sh
make test/test_view test/test_format_plan_cache
test/test_format_plan_cache
perl test/test.pl -F test_vcf_format_plan
git diff --check
```

`test_vcf_format_plan` runs 21 planner cases and byte-compares planned output against the generic parser and disabled-control mode.

Important cases covered by tests:

- empty FORMAT tokens such as `GT::DP`;
- empty float FORMAT subfields;
- empty first-position string fields;
- malformed sample counts;
- repeated wide GT values;
- selected-sample inclusion/exclusion;
- header metadata changes after cached plans were compiled;
- row-local fallback after partial planned parsing.

## `test_view` Measurements on datasets

Measurements below compare:

```sh
HTS_VCF_FORMAT_PLAN=0 test/test_view -b -l 0 <input>
HTS_VCF_FORMAT_PLAN=1 test/test_view -b -l 0 <input>
```

All measured outputs compared byte-identical to baseline.  Times are user time, reported as mean +/- standard error across five sequential runs.

Machine for the local timings: Apple M4 Max, 14 CPU cores, 36 GiB RAM, macOS 26.4.1 (25E253).  These are single-process `test/test_view` runs with level-0 BCF output.

Corpus inputs in this table:

All timed corpus inputs are BGZF-compressed VCF text files (`.vcf.gz`), not BCF inputs.  The timing command writes level-0 BCF output, so the comparison is focused on VCF text parsing plus normal BCF encoding.

- `1000G chr22 full GT`: real public [1000 Genomes Phase 3 integrated chromosome 22 genotype VCF](https://ftp.1000genomes.ebi.ac.uk/vol1/ftp/release/20130502/ALL.chr22.phase3_shapeit2_mvncall_integrated_v5b.20130502.genotypes.vcf.gz) from the 20130502 release, stored as `.vcf.gz`.  It is dominated by many-sample `GT` FORMAT rows, so it represents the high-volume genotype-only case where text parsing overhead is very visible.
- `CCDG 10k`: real `.vcf.gz` 10,000-record slice from the 3,202-sample 1000G high-coverage chromosome 22 [recalibrated variants VCF](https://ftp.1000genomes.ebi.ac.uk/vol1/ftp/data_collections/1000G_2504_high_coverage/working/20201028_3202_raw_GT_with_annot/20201028_CCDG_14151_B01_GRM_WGS_2020-08-05_chr22.recalibrated_variants.vcf.gz) (`20201028_CCDG_14151_B01_GRM_WGS_2020-08-05_chr22.recalibrated_variants.vcf.gz`).  It represents cohort likelihood-style FORMAT rows with `GT`, depth vectors, likelihood vectors, scalar fields, and phase strings.
- `gnomAD v4.1 exomes sites chr22`: real public [gnomAD v4.1 exomes sites](https://storage.googleapis.com/gcp-public-data--gnomad/release/4.1/vcf/exomes/gnomad.exomes.v4.1.sites.chr22.vcf.bgz) `.vcf.gz` slice on chromosome 22.  This is a sites-only, INFO-heavy control with no FORMAT/sample columns, so it checks that enabling the planner does not penalize non-FORMAT VCF parsing.
- `Large reordered likelihood`: generated synthetic `.vcf.gz` 20,000-record, 2,048-sample workload based on CCDG-like likelihood fields, but with FORMAT columns reordered.  It checks that the implementation optimizes by tag metadata rather than by exact FORMAT string.
- `Mixed row-local fallbacks`: generated synthetic `.vcf.gz` 12,000-record, 2,048-sample likelihood workload with row-local unsupported or wrong-width values.  It exercises the record-local fallback and vector-width handling paths.
- `Large float/string negative`: generated synthetic `.vcf.gz` 16,000-record, 2,048-sample float-vector plus string workload.  It guards the conservative support boundary for schemas that should stay on the generic parser.
- `Two-string float negative`: generated synthetic `.vcf.gz` 12,000-record, 2,048-sample two-string plus float workload.  It is another negative-control shape where planning should not make parsing slower.

| Input | Baseline | Planned | Speedup | Planner hits / fallback | Notes |
|---|---:|---:|---:|---:|---|
| 1000G chr22 full GT | 25.078 +/- 0.225 s | 10.272 +/- 0.087 s | 2.44x | 1,103,547 / 0 | Largest win; sample-rich GT-only rows stay fully planned. |
| CCDG 10k | 2.562 +/- 0.006 s | 2.384 +/- 0.004 s | 1.07x | 9,861 / 139 | Real cohort likelihood-style rows; long string rows fall back record-locally. |
| gnomAD v4.1 exomes sites chr22 | 0.400 +/- 0.000 s | 0.402 +/- 0.002 s | 1.00x | n/a, no FORMAT | Sites-only `.vcf.gz` control; planner does not affect non-FORMAT rows. |
| Large reordered likelihood | 2.902 +/- 0.004 s | 2.556 +/- 0.008 s | 1.14x | 20,000 / 0 | Shows tag-level composition handles reordered FORMAT layouts. |
| Mixed row-local fallbacks | 2.144 +/- 0.005 s | 1.912 +/- 0.005 s | 1.12x | 12,000 / 0 | Exercises variable row widths while preserving planned execution. |
| Large float/string negative | 2.856 +/- 0.009 s | 2.862 +/- 0.009 s | 1.00x | 0 / 16,000 | Shows no regression when generic fallback is triggered. |
| Two-string float negative | 2.238 +/- 0.002 s | 2.230 +/- 0.000 s | 1.00x | 0 / 12,000 | Shows no regression when generic fallback is triggered. |

Key points:

- The clearest win is many-sample GT parsing.
- Integer-heavy likelihood rows see smaller but consistent gains.
- String-plus-float layouts that previously regressed now stay on the generic parser and remain near parity.
- CCDG-style long phase-set strings fall back by record, not by poisoning the cached FORMAT schema.

## Notes For Review

- The parser is opt-in only; unset or unknown `HTS_VCF_FORMAT_PLAN` values use the generic parser.
- `HTS_VCF_FORMAT_PLAN` and `HTS_VCF_FORMAT_PLAN_STATS` are temporary branch controls and should not be treated as stable user-facing configuration.

## bcftools Measurements

These runs use a bcftools binary built with this version of htslib and compare `HTS_VCF_FORMAT_PLAN=0` against `HTS_VCF_FORMAT_PLAN=1`.  Outputs are streamed through `cksum` to verify byte equivalence.

The CCDG 10k input is the same 10,000-record, 3,202-sample `.vcf.gz` slice used above.  The 1000G GT-only input is the full chromosome 22 Phase 3 genotype `.vcf.gz` with 2,504 samples.  Times below are mean seconds across five sequential runs, with error bars omitted for readability.  Each cell is `baseline -> planned (speedup)`.  `--threads` applies to the `bcftools view` command shapes only; `bcftools query` rows are unthreaded and therefore use `threads=0`.  

Command key:

- `view_bcf`: convert all samples to BCF with `bcftools view --no-version -Ob -l 0 [--threads N] -o - INPUT`.
- `view_keep2_bcf`: convert only `HG00096,HG00097` to BCF with `bcftools view --no-version -s HG00096,HG00097 -Ob -l 0 [--threads N] -o - INPUT`.
- `view_sites_bcf`: drop all FORMAT/sample data and write sites-only BCF with `bcftools view --no-version -G -Ob -l 0 [--threads N] -o - INPUT`.
- `query_gt_keep2`: print GT for `HG00096,HG00097` with `bcftools query -s HG00096,HG00097 -f '%CHROM\t%POS[\t%GT]\n' INPUT`.
- `query_gt_all`: print GT for every sample with `bcftools query -f '%CHROM\t%POS[\t%GT]\n' INPUT`.
- `filter_gt_keep2`: select `HG00096,HG00097`, filter on alternate genotypes, and write BCF with `bcftools view --no-version -s HG00096,HG00097 -i 'GT="alt"' -Ob -l 0 [--threads N] -o - INPUT`.

| Dataset | Command | User t0 | User t2 | User t4 | Real t0 | Real t2 | Real t4 |
|---|---|---:|---:|---:|---:|---:|---:|
| CCDG 10k | `view_bcf` | 2.398 -> 2.200 (1.09x) | 2.908 -> 2.712 (1.07x) | 2.918 -> 2.720 (1.07x) | 3.914 -> 3.718 (1.05x) | 3.116 -> 2.932 (1.06x) | 3.110 -> 2.932 (1.06x) |
| CCDG 10k | `view_keep2_bcf` | 2.516 -> 2.314 (1.09x) | 2.672 -> 2.474 (1.08x) | 2.684 -> 2.476 (1.08x) | 2.574 -> 2.366 (1.09x) | 2.274 -> 2.074 (1.10x) | 2.284 -> 2.074 (1.10x) |
| CCDG 10k | `view_sites_bcf` | 2.512 -> 2.304 (1.09x) | 2.648 -> 2.448 (1.08x) | 2.650 -> 2.450 (1.08x) | 2.564 -> 2.350 (1.09x) | 2.244 -> 2.044 (1.10x) | 2.244 -> 2.042 (1.10x) |
| CCDG 10k | `query_gt_keep2` | 0.640 -> 0.410 (1.56x) | n/a | n/a | 0.670 -> 0.438 (1.53x) | n/a | n/a |
| CCDG 10k | `query_gt_all` | 2.880 -> 2.664 (1.08x) | n/a | n/a | 2.954 -> 2.728 (1.08x) | n/a | n/a |
| CCDG 10k | `filter_gt_keep2` | 2.790 -> 2.580 (1.08x) | 2.940 -> 2.732 (1.08x) | 2.944 -> 2.740 (1.07x) | 2.846 -> 2.638 (1.08x) | 2.544 -> 2.340 (1.09x) | 2.546 -> 2.350 (1.08x) |
| 1000G chr22 full GT | `view_bcf` | 26.384 -> 10.310 (2.56x) | 28.112 -> 11.952 (2.35x) | 28.202 -> 11.878 (2.37x) | 32.446 -> 15.942 (2.04x) | 25.744 -> 12.344 (2.09x) | 25.782 -> 12.328 (2.09x) |
| 1000G chr22 full GT | `view_keep2_bcf` | 26.514 -> 10.454 (2.54x) | 27.956 -> 11.956 (2.34x) | 28.030 -> 11.812 (2.37x) | 27.170 -> 11.236 (2.42x) | 25.938 -> 9.688 (2.68x) | 25.918 -> 9.522 (2.72x) |
| 1000G chr22 full GT | `view_sites_bcf` | 25.734 -> 9.780 (2.63x) | 27.238 -> 11.338 (2.40x) | 27.260 -> 11.276 (2.42x) | 26.246 -> 10.108 (2.60x) | 25.192 -> 9.076 (2.78x) | 25.284 -> 9.008 (2.81x) |
| 1000G chr22 full GT | `query_gt_keep2` | 5.708 -> 2.856 (2.00x) | n/a | n/a | 5.836 -> 2.946 (1.98x) | n/a | n/a |
| 1000G chr22 full GT | `query_gt_all` | 57.454 -> 41.258 (1.39x) | n/a | n/a | 58.538 -> 42.248 (1.39x) | n/a | n/a |
| 1000G chr22 full GT | `filter_gt_keep2` | 47.540 -> 31.272 (1.52x) | 49.904 -> 33.490 (1.49x) | 49.658 -> 33.718 (1.47x) | 48.006 -> 31.612 (1.52x) | 47.874 -> 31.302 (1.53x) | 47.638 -> 31.642 (1.51x) |

The largest effect is for GT-heavy VCF parsing, especially sample-rich rows where FORMAT decoding dominates.  The 1000G GT-only workload shows the clearest effect: 2.35-2.56x user-time speedups for full BCF conversion, roughly 2.0x for two-sample GT query, and roughly 1.5x for GT-based filtering, with wall-clock gains maintained under `--threads 2` and `--threads 4`.  Mixed FORMAT inputs such as CCDG 10k still get smaller opportunistic gains around 1.07-1.10x on conversion/filter paths.


jhl-oai:
Opened this upstream PR prematurely; moving the review to my fork for now: https://github.com/jhl-oai/htslib-public-fork/pull/2

jkbonfield:
I haven't looked in detail, but I see lots of functions with similar names that look to be hand tuned for specific layouts of data.  Is this the case?  I think if so it's maybe over-optimised.  None the less, I can definitely see how there would be considerably faster ways of implementing the existing ABI/API with a more performance oriented design.

Long term I'd like an alternative API to VCF (and maybe BCF, but tbh I'm less certain of the merit of it given very low support out in the wild) that is more function query oriented than presenting data structures.  If we go down that road there are clear wins to just-in-time parsing and intelligent field splitting.  However such as thing has to be API breaking, or at least a new additional API, as the existing "API" is fundamentally slow by design, forcing full parse and type conversion of the entire VCF record.

I started work on such a thing a couple years ago, but this was then interrupted by another project and I've never had time to continue the exploration.  However I just pushed it up to github as some of the ideas may be interesting or useful.

https://github.com/jkbonfield/vcf_parse

Specifically it has VCF as the in memory object rather than BCF.  The VCF is tokenised as required.  Each token is also parsed only as required.  So a VCF record is a series of columns, which are potentially a series of lists, which may be key=value tuples or other lists.  The data types are polymorphic, so it starts off as a string, but can then morph to a list of strings if we ask for a specific column, but that can then change to an integer if we want to query it numerically, or a tuple, etc.  The key to the speed is only parsing what has been requested.  This makes things such as filtering MUCH faster as most of the time we're writing out the entire VCF record and there's no need to convert back and forth.

PR #2008, opened 2026-05-12
Make faidx work with very long (>4 Gbyte!) lines

### LINKED ISSUE #2331: samtools faidx incorrrent sequence extraction when scaffold length is very long (>2.1Gb)
#### Are you using the latest version of [samtools](https://github.com/samtools/samtools/releases/latest) and [HTSlib](https://github.com/samtools/htslib/releases/latest)? If not, please specify.
Yes, version 1.23.1

 #### Please describe your environment.
Linux 5.14.0-362.8.1.el9_3.x86_64
x86_64
gcc (GCC) 11.4.1 20230605 (Red Hat 11.4.1-2)

#### Please specify the steps taken to generate the issue, the command you are running and the relevant output.
I ran `samtools faidx yahs_hap1.fasta` and the largest scaffold is over 8G. 
```
scaffold_1.H1	8419821385	17649969988	4124854089	4124854090
scaffold_2.H1	4408564030	12694507082	113596734	113596735
```
When I try to check the last few base pairs, it extract the new line symbol ">". 
`samtools faidx yahs_hap1.fasta "scaffold_1.H1:8419821382-8419821385"`
Here is the output:
```
>scaffold_1.H1:8419821382-8419821385
CA
>
```
In addition, when I try to split the scaffold at the gaps, it does not extract the sequence correctly. For example: 
After running `seqtk gap yahs_hap1.fasta > yahs_hap1_gaps.bed` , this is one gap location:
```
scaffold_2.H1	3085955352	3085955552
```
The output of running `samtools faidx yahs_hap1.fasta "scaffold_2.H1:3085955352-3085955552"`
```
>scaffold_2.H1:3085955352-3085955552
NNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNCCTCTCGTCCCTTTCTCCCCCCCTCTC
```
It seems that the location is shifted somehow.

daviesrob:
Although faidx should support very long references, writing one longer than 4Gbases on a single line broke it because it used a `uint32_t` field to store the line length.

To make it work with such inputs, `faidx1_t::line_blen` is increased in size to `uint64_t` so the correct length can be stored.  To avoid having to do the same for `faidx1_t::line_len`, which would make each entry quite a bit bigger for a fairly rare use-case, that field is changed so that it stores the number of bytes to be skipped at the end of each line instead of the full length. As this value will usually only be 1 or 2, a `uint32_t` is plenty big enough for it.  Combined with the fact that the original structure had a four-byte hole in it (between `line_blen` and `len`), it's possible to store the longer line lengths while keeping `faidx1_t` exactly the same size as it had before.

Fixes samtools/samtools#2331

### DIFF
--- a/faidx.c
+++ b/faidx.c
@@ -69,10 +69,13 @@ static inline int bgzf_getc_(BGZF *fp) {
 
 typedef struct {
     int id; // faidx_t->name[id] is for this struct.
-    uint32_t line_len, line_blen;
-    uint64_t len;
-    uint64_t seq_offset;
-    uint64_t qual_offset;
+    uint32_t line_extra;  // Bytes to skip at the end of each line
+                          // Usually 1 for \n or 2 for \r\n ending
+    uint64_t line_blen;   // Number of bases (or quality values) on each line
+                          // Total line length = line_blen + line_extra
+    uint64_t len;         // Length of entire sequence
+    uint64_t seq_offset;  // File offset to start of sequence
+    uint64_t qual_offset; // File offset to start of quality values (if fastq)
 } faidx1_t;
 KHASH_MAP_INIT_STR(s, faidx1_t)
 
@@ -91,10 +94,24 @@ static int fai_name2id(void *v, const char *ref)
     return k == kh_end(fai->hash) ? -1 : kh_val(fai->hash, k).id;
 }
 
-static inline int fai_insert_index(faidx_t *idx, const char *name, uint64_t len, uint32_t line_len, uint32_t line_blen, uint64_t seq_offset, uint64_t qual_offset)
+static inline int fai_insert_index(faidx_t *idx, const char *name, uint64_t len, uint64_t line_len, uint64_t line_blen, uint64_t seq_offset, uint64_t qual_offset)
 {
     if (!name) {
-        hts_log_error("Malformed line");
+        hts_log_error("Malformed line: no name");
+        return -1;
+    }
+    if (line_len > HTS_POS_MAX || line_blen > HTS_POS_MAX) {
+        hts_log_error("Malformed line: length out of bounds");
+        return -1;
+    }
+    if (line_len < line_blen) {
+        hts_log_error("Malformed line: width (%"PRIu64") less than base count (%"PRIu64")",
+                      line_len, line_blen);
+        return -1;
+    }
+    if (line_len - line_blen > UINT32_MAX) {
+        hts_log_error("Malformed line: difference between width (%"PRIu64") and base count (%"PRIu64") too large",
+                      line_len, line_blen);
         return -1;
     }
 
@@ -121,7 +138,7 @@ static inline int fai_insert_index(faidx_t *idx, const char *name, uint64_t len,
     v->id = idx->n;
     idx->name[idx->n++] = name_key;
     v->len = len;
-    v->line_len = line_len;
+    v->line_extra = (uint32_t) (line_len - line_blen);
     v->line_blen = line_blen;
     v->seq_offset = seq_offset;
     v->qual_offset = qual_offset;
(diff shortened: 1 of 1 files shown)


faidx.c
```diff
@@ -393,7 +407,7 @@ static faidx_t *fai_read(hFILE *fp, const char *fname, int format)
     if (!buf) goto fail;
 
     while ((l = hgetln(buf, 0x10000, fp)) > 0) {
-        uint32_t line_len, line_blen, n;
+        uint64_t line_len, line_blen, n;
```
jmarshall: It doesn't affect the behaviour, but this is a good opportunity to make `n` a plain `int`.

daviesrob: Agreed, it's now an `int`.

faidx.c
```diff
@@ -840,7 +858,7 @@ hts_pos_t fai_line_length(const faidx_t *fai, const char *str)
     if (fai_get_val(fai, str, &len, &val, &beg, &end))
         return -1;
     else
-        return val.line_blen;
+        return (hts_pos_t) (val.line_blen <= HTS_POS_MAX ? val.line_blen : HTS_POS_MAX);
```
jkbonfield: This should return -1 rather than using saturated maths and returning HTS_POS_MAX, as providing the wrong value here will then cause a calculation of the offset to read from to be incorrect in anything that calls this function.

However that said, it's a bit of a technicality and a moot point.

1. [Samtools promptly turns -1 into HTS_POS_MAX](https://github.com/samtools/samtools/blob/develop/faidx.c#L269-L270) anyway.  Wrongly, causing the subsequent file offset to be incorrect.
2. We can't have files that big as it's more storage than anyone has!  So we'd fail at a different point making it somewhat moot.

jkbonfield:
I managed to break it with a malformed fai file:

```
foo	5000000000	5	-5000000000	-5000000000
```

```
$ samtools faidx long.fa foo:4999999999-
==1329386==ERROR: AddressSanitizer: negative-size-param: (size=-9999999998)
    #0 0x7f8997e84bad in memcpy /tmp/gcc-16.1.0/libsanitizer/sanitizer_common/sanitizer_common_interceptors_memintrinsics.inc:117
    #1 0x00000079a2a2 in bgzf_read_small (/nfs/users/nfs_j/jkb/work/samtools_master/samtools/samtools+0x79a2a2)
    #2 0x00000079d566 in fai_retrieve (/nfs/users/nfs_j/jkb/work/samtools_master/samtools/samtools+0x79d566)
    #3 0x00000079d937 in fai_fetch64 (/nfs/users/nfs_j/jkb/work/samtools_master/samtools/samtools+0x79d937)
    #4 0x000000582293 in write_output /nfs/users/nfs_j/jkb/work/samtools_master/samtools/faidx.c:272
    #5 0x0000005860b1 in faidx_core /nfs/users/nfs_j/jkb/work/samtools_master/samtools/faidx.c:640
    #6 0x0000005866db in faidx_main /nfs/users/nfs_j/jkb/work/samtools_master/samtools/faidx.c:686
    #7 0x000000502dac in main /nfs/users/nfs_j/jkb/work/samtools_master/samtools/bamtk.c:254
    #8 0x7f8996e68d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #9 0x7f8996e68e3f in __libc_start_main_impl ../csu/libc-start.c:392
    #10 0x000000408794 in _start (/nfs/users/nfs_j/jkb/work/samtools_master/samtools/samtools+0x408794)

0x7f89957ad800 is located 0 bytes inside of 131072-byte region [0x7f89957ad800,0x7f89957cd800)
allocated by thread T0 here:
    #0 0x7f8997e871df in malloc /tmp/gcc-16.1.0/libsanitizer/asan/asan_malloc_linux.cpp:67
    #1 0x0000007952e7 in bgzf_read_init /nfs/users/nfs_j/jkb/work/samtools_master/htslib/bgzf.c:394

SUMMARY: AddressSanitizer: negative-size-param (/nfs/users/nfs_j/jkb/work/samtools_master/samtools/samtools+0x79a2a2) in bgzf_read_small
==1329386==ABORTING
```

One problem here is perhaps the logic in `bgzf_read_small` as the cast makes over-sized queries pass the `<` test:

```
154	    if ((ssize_t)length < fp->block_length - fp->block_offset) {
155	        // Short cut the common and easy mode
156	        memcpy((uint8_t *)data,
157	               (uint8_t *)fp->uncompressed_block + fp->block_offset,
158	               length);
```

`length` here is a very large unsigned value, but `-9999999998` after the cast.  Obviously it needs fixing upstream of this, but it does expose a potential weakness in other pieces of code that may be able to trigger the same case.

The faidx.c `firstline_len = line_len - beg % val->line_blen` line has `firstline_len` as signed and `line_len` as unsigned, which triggers an overflow, but it ought to have been weeded out in the initial setting of `val->line_blen` and `val->line_extra` in `fai_insert_index`.

Eg

```
diff --git a/faidx.c b/faidx.c
index f2539b67..7fb0ad98 100644
--- a/faidx.c
+++ b/faidx.c
@@ -95,6 +95,11 @@ static int fai_name2id(void *v, const char *ref)
 
 static inline int fai_insert_index(faidx_t *idx, const char *name, uint64_t len, uint64_t line_len, uint64_t line_blen, uint64_t seq_offset, uint64_t qual_offset)
 {
+    if ((hts_pos_t)line_len < 0 || (hts_pos_t)line_blen < 0) {
+        hts_log_error("Malformed line: line length is out of bounds");
+        return -1;
+    }
+
     if (!name) {
         hts_log_error("Malformed line: no name");
         return -1;
```

jkbonfield:
Also as per our new policy, you need to sign off your commit.  We clearly need a CI test that can identify this so unsigned PRs fail as it'll help spot these asap and remove the need for us to manually check each commit.

faidx.c
daviesrob: This should now return -1

daviesrob:
Extra check has been added, and I've also rebased to get rid of a conflict.

PR #2010, opened 2026-05-15
Add missing ref_cache/misc.o dependency line

daviesrob:
Somehow this got missed when ref-cache was first added.


daviesrob:
Rebased with fix for conflict...

PR #2011, opened 2026-05-16
`bcf_sr_set_regions`: opt-in fastpath for dense single-base BEDs (samtools/bcftools#2557)

carstenerickson:
## Summary

`bcf_sr_set_regions(file=1)` runs one `tbx_itr_queryi()` per BED entry, which is 300-500× slower than a sequential scan at SNP-panel sizes — [samtools/bcftools#2557](https://github.com/samtools/bcftools/issues/2557) was a production case where 84M single-base regions against a 23 GB VCF did not complete in 11+ hours of 100% CPU. The streaming-targets path already handles this workload at near-baseline speed; this PR adds auto-routing to it.

## Design

New `BCF_SR_AUTO_TARGETS_FROM_REGIONS` opt (end of `bcf_sr_opt_t`, ABI-safe; default off). When set, `bcf_sr_set_regions()` runs a sniffer that accepts only when ALL hold:

- The path is a regular local file (FIFO/stdin/URL skip the sniff so we never consume the prefix before `bcf_sr_regions_init()` reopens).
- First 256 non-comment entries are all single-base (`END − START == 1`).
- Average intra-chromosome inter-entry distance < 10 kbp.

On accept, `regions_overlap` is copied to `targets_overlap` (value semantics are identical) and `bcf_sr_set_targets()` takes over. On reject (sparse / wide / malformed / too small / non-local), the existing seek-per-region path runs unchanged.

The opt-in gate is the safety mechanism: successful promotion populates `readers->targets`, so a subsequent `bcf_sr_set_targets()` fails — callers must explicitly accept that trade-off.

## Measurements

bcftools 1.23.1 + this PR; single-base BED at 10 bp avg spacing, matching VCF; macOS arm64:

|     N | seek-per-region | fastpath | speedup |
|------:|----------------:|---------:|--------:|
|  100K |        13.3 s   | 0.19 s   |    70× |
|    1M |       156.6 s   | 0.55 s   |   285× |
|    5M |         705 s   | 2.76 s   |   255× |

Production (84M entries × 23 GB VCF) extrapolates from hours to minutes.

## Test plan

New `test_bcf_sr_regions_fastpath` (`test/test.pl`) diffs three genuinely different code paths producing identical output on a 300-entry single-base fixture:

- `-R FILE` slow path (opt off)
- `-R FILE --auto-targets-from-regions` fastpath (opt + sniffer)
- `-T FILE` existing streaming-targets path

`test-bcf-sr` gains matching `-R`/`-T` file-mode flags and the `--auto-targets-from-regions` flag. All 359 existing htslib tests pass.

## Notes

The matching bcftools-side wiring is at samtools/bcftools#2561; without it, `bcftools view -R FILE` keeps the slow path. This PR is strictly the htslib half.

Tracks samtools/bcftools#2557.


daviesrob:
It looks like this is trying to make `bcftools` work better for cases where `-T` to select regions would have worked much faster then `-R`.  While useful, I think a better solution would be to make the synced reader use the multi-region iterator, which may be a bit tricky given how the synced reader currently works, but should hopefully be possible.  The advantage is that it should avoid the inefficiency of starting new iterators for each region, while still allowing jumps over uninteresting parts of the data.

I'm currently having a good look at the synced reader for other reasons, so I'll see if I can work out how hard it would be to add multi-region support while I'm in there.

Incidentally, we've recently added some policy around developer sign-off and the use of AI coding tools.  Please see [CONTRIBUTING.md](https://github.com/samtools/htslib/blob/develop/CONTRIBUTING.md) for details.

carstenerickson:
Closing in favor of Rob's planned multi region work on the synced reader, which is the better long term fix. The reproducer and benchmark numbers are preserved at samtools/bcftools#2557 if useful as a test case. Happy to test the multi region implementation against the same workload once something lands. Closing samtools/bcftools#2561 as well since it is just the bcftools side wiring for the opt added here.

Re. CONTRIBUTING.md sign off and AI policy, noted for any future PR.

PR #2006, opened 2026-05-11
Add wrappers for malloc, realloc with a calloc-like interface

daviesrob:
Help avoid bugs due to integer wrap-around when calculating sizes to pass to `malloc()` or `realloc()`.  `calloc()` is better in this respect as it takes two parameters (number and size) and catches overflows when multiplying them together.  The new interfaces provide similar functions for `malloc()` and `realloc()`, along with additional ones to handle cases that commonly occur in the code base.

All the functions are static inlines, so it should be possible for the optimiser to simplify the code in the cases where overflow cannot occur (for example, due to the size of the data types actually being passed in).

The interfaces are:

* Saturating arithmetic

  Primitives used to build the later functions.

  - `hts_add_sat2(a, b)` returns `(a + b)` or `SIZE_MAX` on overflow

  - `hts_add_sat3(a, b, c)` returns `(a + b + c)` or `SIZE_MAX` on overflow

  - `hts_prod_sat2(a, b)` returns `(a * b)` or `SIZE_MAX` on overflow

* Wrappers around `malloc()`

  - `hts_malloc(size_t size)`
    For use with the saturating arithmetic functions above.  Catches over-large allocations before they get to `malloc()`.  Prevents spurious gcc warnings about very large allocations, and allows the optimiser to convert the overflow cases to an early exit.

  - `hts_malloc_p(size_t a, size_t b)`
    Replaces `malloc(a * b)`

  - `hts_malloc_ps(size_t sz, size_t a, size_t b)`
    Replaces `malloc(sz * (a + b))`

  - `hts_malloc_pse(size_t sz, size_t a, size_t b, size_t extra)`
    Replaces `malloc(sz * (a + b) + extra)`

* Wrappers around `calloc()`

  - `hts_calloc(size_t size, size_t num)`
    For use with the saturating arithmetic functions above.  Catches over-large allocations before they get to `calloc()`.  Prevents spurious gcc warnings about very large allocations, and allows the optimiser to convert the overflow cases to an early exit.

  - `hts_calloc_ps(size_t sz, size_t a, size_t b)`
    Replaces `calloc(a + b, sz)`

  - `hts_calloc_pse(size_t sz, size_t a, size_t b, size_t extra)`
    Replaces `calloc(sz * (a + b) + extra, 1)`

* Wrappers around `realloc()`

  - `hts_realloc(void *orig, size_t size)`
[...]

PR #2015, opened 2026-05-20
Parallel cram2bam

jkbonfield:
The main thread does a lot of work in `cram_to_bam` and `bam_set1`.  Io_lib's model for this was to put this code into the worker threads.  I thought I'd done that here too (I have for the writer already), but apparently not.  This PR fixes that.

A profile of the main thread:

```
$ ./test/test_view.dev -B -@32  ~/scratch/data/novaseq.10m.cram & perf record -t $! 2>/dev/null;perf report -n 2>/dev/null| head -20 | egrep -v '^#'
[1] 3634487
[1]+  Done                    ./test/test_view.dev -B -@32 ~/scratch/data/novaseq.10m.cram
    49.75%          2100  test_view.dev  test_view.dev      [.] bam_set1
     9.51%           407  test_view.dev  test_view.dev      [.] cram_to_bam
     5.93%           266  test_view.dev  libc.so.6          [.] __memmove_avx512_unaligned_erms
     3.13%           129  test_view.dev  libc.so.6          [.] __strncpy_evex
     2.50%            92  test_view.dev  [unknown]          [k] 0xffffffff82cb9b7e
     2.37%            97  test_view.dev  test_view.dev      [.] cram_get_seq
     2.25%            92  test_view.dev  test_view.dev      [.] sam_read1
     2.11%            92  test_view.dev  test_view.dev      [.] bam_cigar2rqlens
     1.65%            64  test_view.dev  test_view.dev      [.] cram_get_bam_seq
```

With this PR:

```
@ qpg-gpu-01[samtools.../htslib]; ./test/test_view.cram2bam -B -@32  ~/scratch/data/novaseq.10m.cram & perf record -t $! 2>/dev/null;perf report -n 2>/dev/null| head -20 | egrep -v '^#'
[1] 3634842
[1]+  Done                    ./test/test_view.cram2bam -B -@32 ~/scratch/data/novaseq.10m.cram
    15.65%           101  test_view.cram2  test_view.cram2bam  [.] cram_get_bam_seq
    11.31%            65  test_view.cram2  libc.so.6           [.] _int_free
     6.69%            38  test_view.cram2  [unknown]           [k] 0xffffffff829076fb
     5.17%            34  test_view.cram2  test_view.cram2bam  [.] cram_get_seq
     5.12%            29  test_view.cram2  [unknown]           [k] 0xffffffff8291f048
     4.73%            32  test_view.cram2  test_view.cram2bam  [.] sam_read1
     4.24%            24  test_view.cram2  [unknown]           [k] 0xffffffff828d0d41
     4.23%            24  test_view.cram2  [unknown]           [k] 0xffffffff8292a32a
     3.35%            19  test_view.cram2  [unknown]           [k] 0xffffffff82929c09
```

Wall clock benchmarks with 32 cores show a doubling in throughput (making it slightly faster than BAM, but that's due to the same main-thread excessive work problem there too):

```
$ for i in `seq 1 10`;do /usr/bin/time -f '%e %U %S'  ./test/test_view.dev 
-B -@32  ~/scratch/data/novaseq.10m.cram;done
1.14 10.34 0.76
1.06 9.13 0.69
1.11 9.42 0.72
1.12 10.20 0.62
1.00 8.72 0.57
1.23 8.85 0.77
1.46 9.13 0.79
1.17 8.67 0.63
1.30 8.73 0.69
0.99 9.64 0.54


$ for i in `seq 1 10`;do /usr/bin/time -f '%e %U %S'  ./test/test_view.cram2bam -B -@32  ~/scratch/data/novaseq.10m.cram;done
0.50 11.83 0.98
0.50 10.32 0.82
0.45 10.13 0.82
0.42 10.10 0.78
0.50 10.19 0.87
0.47 10.07 0.81
0.48 10.15 0.98
0.49 10.03 1.00
0.51 10.06 0.85
0.49 10.28 0.76
```

### DIFF
--- a/cram/cram_decode.c
+++ b/cram/cram_decode.c
@@ -2380,7 +2380,7 @@ static int bam_size(sam_hrecs_t *bfd, cram_fd *fd, cram_record *cr) {
  *
  * Note memory for these is in a single malloc.  Hence compute upfront the
  * memory size of each record prior to conversion.
- * 
+ *
  * Returns 0 on success,
  *        -1 on failure
  */
@@ -2405,6 +2405,9 @@ static int bulk_cram_to_bam(sam_hrecs_t *bfd, cram_fd *fd, cram_slice *s) {
                 return -1;
             memset(&bl->bams[bl->nbams], 0,
                    (s->hdr->num_records - bl->nbams) * sizeof(*bl->bams));
+            int i;
+            for (i = bl->nbams; i < s->hdr->num_records; i++)
+                bam_set_mempolicy(&bl->bams[i], BAM_USER_OWNS_STRUCT);
             bl->nbams = s->hdr->num_records;
         }
     } else {
@@ -2414,20 +2417,20 @@ static int bulk_cram_to_bam(sam_hrecs_t *bfd, cram_fd *fd, cram_slice *s) {
             return -1;
         bl->nbams = s->hdr->num_records;
         bl->next = NULL;
-        bl->bams = calloc(s->hdr->num_records, sizeof(bam_seq_t *));
+        bl->bams = calloc(s->hdr->num_records, sizeof(*bl->bams));
         if (!bl->bams) {
             free(bl);
             return -1;
         }
+        int i;
+        for (i = 0; i < s->hdr->num_records; i++)
+            bam_set_mempolicy(&bl->bams[i], BAM_USER_OWNS_STRUCT);
     }
     s->bl = bl;
 
     for (i = 0; i < s->hdr->num_records; i++) {
         int sz = bam_size(bfd, fd, &s->crecs[i]);
-        if (!s->bl->bams[i])
-            if (!(s->bl->bams[i] = bam_init1()))
-                return -1;
-        realloc_bam_data(s->bl->bams[i], sz);
+        realloc_bam_data(&s->bl->bams[i], sz);
     }
 
     for (i = 0; i < s->hdr->num_records; i++) {
@@ -3210,7 +3213,7 @@ int cram_decode_slice_mt(cram_fd *fd, cram_container *c, cram_slice *s,
  *         -1 on failure.
  */
 int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
-                cram_record *cr, int rec, bam_seq_t **bam) {
+                cram_record *cr, int rec, bam_seq_t *bam) {
     int ret, rg_len;
     char name_a[1024], *name;
     int name_len;
@@ -3277,7 +3280,7 @@ int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
         qual = NULL;
     }
 
-    ret = bam_set1(*bam,
+    ret = bam_set1(bam,
                    name_len, name,
                    cr->flags, cr->ref_id, cr->apos - 1, cr->mqual,
                    cr->ncigar, &s->cigar[cr->cigar],
@@ -3288,13 +3291,13 @@ int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
         return ret;
     }
 
-    aux = (char *)bam_aux(*bam);
+    aux = (char *)bam_aux(bam);
 
     /* Auxiliary strings */
     if (cr->aux_size != 0) {
         memcpy(aux, BLOCK_DATA(s->aux_blk) + cr->aux, cr->aux_size);
         aux += cr->aux_size;
-        (*bam)->l_data += cr->aux_size;
+        bam->l_data += cr->aux_size;
     }
 
     /* RG:Z: */
@@ -3304,10 +3307,10 @@ int cram_to_bam(sam_hdr_t *sh, cram_fd *fd, cram_slice *s,
         memcpy(aux, bfd->rg[cr->rg].name, len);
         aux += len;
         *aux++ = 0;
-        (*bam)->l_data += rg_len;
+        bam->l_data += rg_len;
     }
 
-    return (*bam)->l_data;
+    return bam->l_data;
 }
 
 /*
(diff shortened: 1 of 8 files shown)


cram/cram_decode.c
```diff
@@ -2339,6 +2342,104 @@ static int cram_decode_tlen(cram_fd *fd, cram_container *c, cram_slice *s,
...
+        }
+        int i;
+        for (i = 0; i < s->hdr->num_records; i++)
+            bam_set_mempolicy(&bl->bams[i], BAM_USER_OWNS_STRUCT);
+    }
+    s->bl = bl;
+
+    for (i = 0; i < s->hdr->num_records; i++) {
+        int sz = bam_size(bfd, fd, &s->crecs[i]);
+        realloc_bam_data(&s->bl->bams[i], sz);
```
daviesrob: Is this needed?  `bam_set1()` (via `cram2bam()`) would calculate the size and do the same `realloc_bam_data()` call, so this is just moving that work a bit earlier (and duplicating some of it).

cram/cram_decode.c
```diff
@@ -2339,6 +2342,104 @@ static int cram_decode_tlen(cram_fd *fd, cram_container *c, cram_slice *s,
...
+    if (fd->bl) {
+        bl = fd->bl;
+        fd->bl = fd->bl->next;
+    }
+    pthread_mutex_unlock(&fd->bam_list_lock);
+
+    if (bl) {
+        // Reuse an old bam list, possibly growing it
+        if (s->hdr->num_records > bl->nbams) {
+            bl->bams = realloc(bl->bams, s->hdr->num_records *
```
daviesrob: As #2006 has been merged, it might be good to rebase this so it can take advantage of the `hts_realloc_p()` interface.

daviesrob (commented):
Looks like this does a good job of moving more work into the thread pool.  Would it be worth doing the CG processing there too?  I think it would just need `bam_tag2cigar()` to be made non-static so that it can be called from `cram/cram_decode.c`.

cram/cram_decode.c
jkbonfield: Agreed.  It was probably there because it's how io_lib previously implemented this, forgetting some of that code had already been ported over (eg bam_set1).

cram/cram_decode.c
jkbonfield: Done here and the new realloc in cram_encode.c too.

jkbonfield:
> Looks like this does a good job of moving more work into the thread pool. Would it be worth doing the CG processing there too? I think it would just need bam_tag2cigar() to be made non-static so that it can be called from cram/cram_decode.c.

Possibly, but I doubt this code is called very often.  It's only an issue on ultra-long ONT reads, and maybe not even there it's not that often any more due to improved quality.  If we moved it, we'd still need a flag to say whether or not it's been done already, so that's changed.

I think I'd need to find some test data to evaluate.  (Most of mine got lost on lustre.)  It may be an unnecessary optimisation though.

What's more important appears to be careful malloc handling.  As we add more threads, we add more bam-list allocations.  These appear to cause malloc heap thrashing.  GNU malloc is *particularly* badly performing here.  TCMalloc is double the speed at high thread count as it has minimal slow down with more threads.  GNU malloc performs badly as it gives back memory and then has to reallocate it again.  Defining `MALLOC_MMAP_MAX_=0 MALLOC_TOP_PAD_=100000000` solves this, but in theory it's not meant to be giving the memory back anyway as this is what the bam-list bit was meant to solve.  Maybe the memory thrashing is somewhere else.  I need to diagnose that.

(Io_lib just used mallopt to work around gnu malloc performance, but it's not ideal in a library)

```
@ [samtools.../htslib]; for i in `seq 1 5`;do /usr/bin/time -f '%e %U %S'  ./test/test_view -B -@64
  ~/scratch/data/novaseq.10m.31.cram;done
3.27 28.82 6.30
3.17 28.65 6.35
3.30 28.32 6.44
3.32 28.42 6.67
3.27 29.00 6.53

@ [samtools.../htslib]; for i in `seq 1 5`;do MALLOC_MMAP_MAX_=0 MALLOC_TOP_PAD_=100000000 /usr/bin/time -f '%e %U %S'  ./test/test_view -B -@64  ~/scratch/data/novaseq.10m.31.cram;done
1.47 28.74 2.88
1.49 29.27 2.91
1.46 28.18 2.71
1.48 28.72 2.71
1.48 28.13 2.79

@ [samtools.../htslib]; for i in `seq 1 5`;do LD_PRELOAD=~/scratch/opt/lib/libtcmalloc.so /usr/bin/time -f '%e %U %S'  ./test/test_view -B -@64  ~/scratch/data/novaseq.10m.31.cram;done
1.66 35.33 3.25
1.61 34.49 3.24
1.61 34.72 3.07
1.64 34.43 3.29
1.65 34.65 3.35
```

(Note this machine has 16 cores each with 2 hyperthreads, so this is just a stress test of memory by over-committing on bam-list arrays.)

Edit:  It's not the bam_list allocations, but the cram record one:

```
    if (!(s->crecs = hts_malloc_p(sizeof(*s->crecs), s->hdr->num_records)))     
        return -1;                                                              
```

This is probably another PR, but I'll investigate

jkbonfield:
I have ideas for reducing malloc pressure, but it may come later depending on how long this sits in review (as it'll need to be based on top of these commits).

I've fixed the reallocs and removed needless code. I've rebased it too.  I'll take a look at the `bam_tag2cigar` change as it may make the interface easier so perhaps it's worth it for that reason alone.

jkbonfield:
I can squash when you wish, but figured it's helpful for review to not do that just now so you can see the new commits first.  (Or squash it yourself if you wish.)

The bam_tag2cigar was quite simple to do.

cram/cram_decode.c
```diff
@@ -53,6 +53,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "../htslib/hts.h"
 #include "../htslib/hts_alloc.h"
 #include "../htslib/hfile.h"
+#include "../sam_internal.h" // realloc_bam_data()
```
daviesrob: This is needed for `bam_tag2cigar()` now, rather than `realloc_bam_data()`.

cram/cram_decode.c
```diff
@@ -2339,6 +2343,66 @@ static int cram_decode_tlen(cram_fd *fd, cram_container *c, cram_slice *s,
...
+    if (fd->bl) {
+        bl = fd->bl;
+        fd->bl = fd->bl->next;
+    }
+    pthread_mutex_unlock(&fd->bam_list_lock);
+
+    if (bl) {
+        // Reuse an old bam list, possibly growing it
+        if (s->hdr->num_records > bl->nbams) {
+            bl->bams = hts_realloc_p(bl->bams, s->hdr->num_records,
```
daviesrob: There could be a memory leak here if `hts_realloc_p()` fails.

cram/cram_encode.c
```diff
@@ -4186,29 +4187,43 @@ int cram_put_bam_seq(cram_fd *fd, bam_seq_t *b) {
         /* First time through, allocate a set of bam pointers */
         pthread_mutex_lock(&fd->bam_list_lock);
         if (fd->bl) {
-            spare_bams *spare = fd->bl;
+            bam_list *spare = fd->bl;
+            if (c->max_c_rec > spare->nbams) {
+                if (!(spare->bams =
+                      hts_realloc_p(spare->bams, c->max_c_rec,
```
daviesrob: Also a possible leak here.

daviesrob (commented):
Looks OK, apart from a couple of potential leaks on `realloc()`.

cram/cram_decode.c
jkbonfield: Done and pushed

cram/cram_decode.c
jkbonfield: done

cram/cram_encode.c
jkbonfield: done

jkbonfield:
I went through all the reallocs in cram_encode.c and cram_decode.c and found a few more, so I fixed those too.
Now pushed up.

I also have another branch that creates a cram_list struct to go along the bam_list struct and reuses it between cram slices.
However it doesn't really seem to help much and we still need mallopt envs to coerce GNU malloc to not be so rubbish! (Or use tcmalloc).  Frustrating.  So for now I'm not adding that extra complexity.

I'm also now finding it hard to find any actual speed benefit on our modern machines (but it is faster on the old seq4 node).   It's still provably faster by profiling the main thread, but something else is slowing it up now.  Such are the vagaries of I/O and shared CPUs, but at least it can be faster when the conditions are favourable.

PR #2018, opened 2026-05-22
Add chunked hFILE input scheme

fabwa:
## Summary

This adds a built-in `chunked:` hFILE input scheme for reading byte-split binary streams such as BAMs whose raw bytes have been stored in ordered chunks.

The scheme takes a manifest file containing one chunk path per line, ignores blank/comment lines, resolves relative chunk paths against the manifest directory, and exposes the chunks as one seekable logical file. BGZF/BAM readers can consume the concatenated byte stream without samtools-specific wrapping, and normal BAM indexes can be built and used against the logical chunked input.

Example:

```sh
samtools view chunked:chunks.fofn
samtools index chunked:chunks.fofn
samtools view chunked:chunks.fofn chr1:1000-2000
```

`chunks.fofn` should list the raw BAM byte chunks in order.

## Notes

Chunk files must be seekable so HTSlib can determine chunk sizes and translate logical BAM offsets to the right chunk and intra-chunk offset.

For local manifests, default index names are derived from the manifest path, so `samtools index chunked:chunks.fofn` writes `chunks.fofn.bai` and `samtools view chunked:chunks.fofn region` can discover it.

## Tests

```sh
make -j4
REF_PATH=: ./test.pl -F test_view
test/hfile
test/test_introspection
git diff --check
```

Also manually tested manifest-relative chunk names with blank/comment lines in both normal and `-@4` threaded reads.

### DIFF
--- a/hfile.c
+++ b/hfile.c
@@ -1111,6 +1111,7 @@ static int init_add_plugin(void *obj, int (*init)(struct hFILE_plugin *),
 static int load_hfile_plugins(void)
 {
     static const struct hFILE_scheme_handler
+        chunked = { hopen_chunked_manifest, hfile_always_local, "built-in", 80 },
         data = { hopen_mem, hfile_always_local, "built-in", 80 },
         file = { hopen_fd_fileuri, hfile_always_local, "built-in", 80 },
         preload = { hopen_preload, is_preload_url_remote, "built-in", 80 };
@@ -1119,6 +1120,7 @@ static int load_hfile_plugins(void)
     if (schemes == NULL)
         return -1;
 
+    hfile_add_scheme_handler("chunked", &chunked);
     hfile_add_scheme_handler("data", &data);
     hfile_add_scheme_handler("file", &file);
     hfile_add_scheme_handler("preload", &preload);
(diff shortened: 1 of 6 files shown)


whitwham:
Hi @fabwa, thank you for your contribution.  Do you have an actual use case for the chunked reading?

Also we have just added a [contribution policy](https://github.com/samtools/htslib/blob/develop/CONTRIBUTING.md), with sections on developer sign-off and the use of AI coding tools.

fabwa:
> Hi @fabwa, thank you for your contribution. Do you have an actual use case for the chunked reading?
> 
> Also we have just added a [contribution policy](https://github.com/samtools/htslib/blob/develop/CONTRIBUTING.md), with sections on developer sign-off and the use of AI coding tools.

Thanks for pointing out the policy.
As BAMs can grow quite large, they may have to be spread over multiple devices.
I faced this case in my home lab - the commit helped my case.

whitwham:
Hello @fabwa, after some discussion we have decided not to merge your PR.  We felt it was not quite what we wanted in the core library.  If you want to you could alter it to fit into the [plugins](https://github.com/samtools/htslib-plugins) repository.  This is where we put things that might be useful but do not want in the main code base.

PR #2020, opened 2026-05-26
Remove CRAM v4.0 support.

jkbonfield:
The reasoning for this is that it is highly complex and has had a number of security issues due to insufficient stress testing.  Given the likelihood of CRAM4 ever seeing the light of day, it makes no sense to expose users to the risk of future bugs in this code without any practical benefit.

There are some good ideas still in CRAM v4, but the community support just wasn't there and it's languished as an experimental format for years.


daviesrob:
Rebased...

PR #2027, opened 2026-06-01
Remove a divide-by-zero bug in cram_xpack_decode_char

jkbonfield:
When xpack.nbits is zero we attempt a divide by zero in the argument passing to hts_unpack().  We now check for this case in the initialisation function, but only when nvals > 1 otherwise nbits becomes irrelevant due to it being a pre-determined constant instead.

With thanks to Team Atlanta.

Co-authored-by: Team Atlanta


cram/cram_codecs.c
daviesrob: Return value needs to be checked here.
It also looks like the same thing needs to be done for `cram_xrle_decode_expand_char()`

jkbonfield: Done and pushed.  Even the xdelta ones just for neatness, despite their guaranteed failure.  (It's all a big NOP anyway as it's about to be culled and it wasn't fully implemented in older releases either.)

PR #2034, opened 2026-06-12
News updates for a summer 2026 release.

whitwham:
Updates for HTSlib.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -41,6 +41,15 @@ Updates
   (PR #2030, fixes #1930.  Requested by Adam Novak.
    Replaces the alternative PRs #1997 and #2022)
 
+* Remove references to the non-existant FAI_CACHE.
+  (PR #2033, fixes #2032.  Reported by John Marshall)
+
+* Improve hfile_s3.c error handling.
+  (PR #2036.  Thanks to John Marshall)
+
+* In VCF, improve the "not defined in the header" messages.
+  (PR #2007)
+
 
 Build Changes
 -------------
--- a/NEWS
+++ b/NEWS
@@ -29,7 +29,7 @@ Updates
   broke it because it used a uint32_t field to store the line length.
   (PR #2008. fixes samtools/samtools#2331.  Reported by Ying Chen)
 
-* Parallel cram2bam.  Add more of the work into worker threads to speed
+* Parallel cram2bam.  Move more of the work into worker threads to speed
   everything up.
   (PR #2015)
 
@@ -64,7 +64,7 @@ Build Changes
 Bug fixes
 ---------
 
-* Fix a read buffer overflow and improve SQ LN field checking.
+* Fix a read buffer overflow and improve SQ LN field checking in CRAM encode.
   (PR #1999 fixes oss-fuzz issue 499447432)
 
 * A number of small fixes, often removing unused code or correcting misplaced
--- a/NEWS
+++ b/NEWS
@@ -44,7 +44,8 @@ Updates
 * Remove references to the non-existant FAI_CACHE.
   (PR #2033, fixes #2032.  Reported by John Marshall)
 
-* Improve hfile_s3.c error handling.
+* The underlying causes of errors when accessing files in S3 buckets are now
+  reported more accurately.
   (PR #2036.  Thanks to John Marshall)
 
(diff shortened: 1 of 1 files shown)


whitwham:
This will remain in draft until we are ready for a release.

NEWS
```diff
@@ -1,6 +1,101 @@
...
+
+* Bounds-check refid when loading a CRAM .crai index.
+  (PR #2029.  Thanks to Sidhartha Kumar)
+
+
+Documentation updates
+---------------------
+
+* Clarify how defaults work when building an index with tabix.
+  (PR #2002, addresses #1995. Query made by dariober)
```
jmarshall:
```suggestion
  (PR #2002, addresses #1995. Query made by Dario Beraldi)
```

whitwham: I tend to use the name set in GitHub (which is dariober).  Do you think Dario would be okay with using their real name?

jmarshall: I would think he would be (FWIW I used to work with him). His full name already appears in _samtools/NEWS.md_… @dariober?

dariober: Sure, go ahead with whichever you prefer - thanks for asking!

whitwham: Then I will use the full name.

NEWS
```diff
@@ -1,6 +1,101 @@
...
+* Add wrappers for malloc, realloc with a calloc-like interface.  This helps
+  avoid bugs due to integer wrap-around when calculating memory sizes.
+  (PR #2006)
+
+* Make faidx work with very long (>4 Gbyte!) lines. Although faidx should
+  support very long references, writing one longer than 4Gbases on a single line
+  broke it because it used a uint32_t field to store the line length.
+  (PR #2008. fixes samtools/samtools#2331.  Reported by Ying Chen)
+
+* Parallel cram2bam.  Add more of the work into worker threads to speed
```
jkbonfield: Technically it's "move more of the work", otherwise we're not really gaining a speed up. :-)

However it's a very minor point so don't bother fixing it unless making other changes.

NEWS
```diff
@@ -1,6 +1,101 @@
...
+  (PR #2000)
+
+* Ensure PACKAGE_VERSION is set in the Makefile.
+  (PR #2038.  See also samtools/samtools#2337)
+
+Bug fixes
+---------
+
+* Fix a read buffer overflow and improve SQ LN field checking.
+  (PR #1999 fixes oss-fuzz issue 499447432)
```
jkbonfield: Note this isn't a change from one release to the next since the error it is fixing was added after the previous release, so it's invisible to the users and IMO doesn't warrant a NEWS item.  (If we do wish to go with it, then it should be stated that it's a fix to something that appeared elsewhere within this release.  It also would need to say what it was fixing - CRAM encoding)

whitwham: It was an external issue (oss-fuzz in this case) so I added it to the news.  I always prefer to acknowledge outside help.  CRAM encoding I can mention.

NEWS
```diff
@@ -1,6 +1,101 @@
...
+  (PR #2024)
+
+* Add tbx and bcf multi-region iterators.
+  (PR #2030, fixes #1930.  Requested by Adam Novak.
+   Replaces the alternative PRs #1997 and #2022)
+
+* Remove references to the non-existant FAI_CACHE.
+  (PR #2033, fixes #2032.  Reported by John Marshall)
+
+* Improve hfile_s3.c error handling.
```
jmarshall: A more user-oriented way to say this might be something like:

> The underlying causes of errors when accessing files in S3 buckets are now reported more accurately.

NEWS
jkbonfield: While it was found by an external user, they were fuzzing develop, not a release so people reading it don't need to feel concerned if they're sticking to official releases.  We ought to at least clarify that.

(Although tbh there are plenty of other things they *SHOULD* be concerned over so it's a moot point!)

whitwham: Done.

jkbonfield:
Are we ready for a squash and merge now? Happy with the changes.

whitwham:
> Are we ready for a squash and merge now? Happy with the changes.

I'm done of nothing else turns up.

PR #2036, opened 2026-06-16
Improve hfile_s3.c error handling

jmarshall:
The rewritten S3 backend does not translate libcurl error values into errno values; in some circumstances it returns an error indication but does not set `errno` at all, leaving it set to whatever it was set to previously. So for example a malformatted URL results in a spurious ENOENT error message:

```
$ htsfile s3:expect_invalid_argument
htsfile: can't open "s3:expect_invalid_argument": No such file or directory
```

This led to a [bioconda package build test failure](https://github.com/bioconda/bioconda-recipes/actions/runs/24994788442/job/73189088762#step:10:2324) (bioconda/bioconda-recipes#64884):

```
running run_test.py
    pysam.AlignmentFile('s3:expect_invalid_argument')
  File "pysam/libcalignmentfile.pyx", line 752, in pysam.libcalignmentfile.AlignmentFile.__cinit__
  File "pysam/libcalignmentfile.pyx", line 952, in pysam.libcalignmentfile.AlignmentFile._open
FileNotFoundError: [Errno 2] Could not open alignment file: No such file or directory: 's3:expect_invalid_argument'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/conda-bld/pysam_1777293322095/test_tmp/run_test.py", line 15, in <module>
    assert e.errno == errno.EINVAL
AssertionError
```

because the test was written to validate the expected particular `errno` value that the libcurl backend produces. This particular test has since been rewritten in a more robust way, but nonetheless _hfile_s3.c_ should enable meaningful error messages.

Also requests for buckets which you don't have permission to access or non-existent buckets produce unhelpful errors:

```
$ htsfile s3://bucket/foo.sam
htsfile: can't open "s3://bucket/foo.sam": Resource temporarily unavailable
$ htsfile s3://bork-asdfghjk/foo.sam  # Nonexistent at the time of writing
htsfile: can't open "s3://bork-asdfghjk/foo.sam": Resource temporarily unavailable
```

With `-vvvv` you can see that `errno` is being set appropriately from the 403 or 404 response, but it is overwritten with a spurious EAGAIN `errno` value before it is returned to the caller.

Hence this PR:

Libcurl API functions mostly do not set `errno`, so set it explicitly based on libcurl API error codes. Copy `easy_errno()` from _hfile_libcurl.c_, alongside the previously copied `http_status_errno()`.

Also preserve `errno` in `s3_read_open()`/`s3_write_open()`'s error cleanup. Otherwise it will often be changed to EAGAIN by `curl_easy_cleanup()`, which is called during `cleanup_local()`.

### DIFF
--- a/hfile_s3.c
+++ b/hfile_s3.c
@@ -984,7 +984,7 @@ static int update_time(s3_auth_data *ad, time_t now) {
         }
 
         if (strftime(ad->date_short, 9, "%Y%m%d", tm) != 8) {
-            return -1;;
+            return -1;
         }
 
         ad->date_html.l = 0;
@@ -1242,6 +1242,9 @@ static int report_s3_error(kstring_t *body, long resp_code) {
 
 static int http_status_errno(int status)
 {
+    if (hts_verbose >= HTS_LOG_INFO)
+        fprintf(stderr, "hfile_s3: setting errno from HTTP status code %d\n", status);
+
     if (status >= 500)
         switch (status) {
         case 501: return ENOSYS;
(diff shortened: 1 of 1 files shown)


hfile_s3.c
```diff
@@ -2461,7 +2593,8 @@ int PLUGIN_GLOBAL(hfile_plugin_init,_s3)(struct hFILE_plugin *self) {
     err = curl_global_init(CURL_GLOBAL_ALL);
 
     if (err != CURLE_OK) {
-        // look at putting in an errno here
+        // Set a suitably catastrophic error code
+        errno = ENETDOWN;
```
jmarshall, line 2597: `curl_global_init()` always returns CURLE_FAILED_INIT on error, so simply translate that never-happens into a suitably catastrophic `errno` value.

hfile_s3.c
```diff
@@ -1645,7 +1748,10 @@ static ssize_t s3_write(hFILE *fpv, const void *bufferv, size_t nbytes) {
 
             cret = curl_easy_getinfo(fp->curl, CURLINFO_RESPONSE_CODE, &response_code);
 
-            if (cret != CURLE_OK || response_code > 200) {
+            if (cret != CURLE_OK) {
```
jmarshall, line 1751: In the unlikely event `curl_easy_getinfo()` fails, `response_code` will usually not have been written to. So separating these cases avoids reading an uninitialised variable in `http_status_errno(response_code)` below.

whitwham:
Thank you.

PR #2038, opened 2026-06-16
Ensure PACKAGE_VERSION is set in the Makefile

daviesrob:
This variable is set by expanding `$(shell version.sh)`.  As this is not in a normal Makefile rule, failing to run `version.sh` just results in an empty variable.  To check that it really was set, add a check to ensure it's not empty in the rule that builds the `version.h` file.

See also samtools/samtools#2337.  Failing here is unlikely to result in broken files, but we should probably catch it anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,7 +167,13 @@ MACH_O_CURRENT_VERSION = 3.1.23
 version.h: $(if $(wildcard version.h),$(if $(findstring "$(PACKAGE_VERSION)",$(shell cat version.h)),,force))
 
 version.h:
-	echo '#define HTS_VERSION_TEXT "$(PACKAGE_VERSION)"' > $@
+	@if test 'x$(PACKAGE_VERSION)' != x ; then \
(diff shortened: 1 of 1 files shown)


jmarshall:
Because PACKAGE_VERSION is constructed using GNU Make-specific functions, there's probably no downside to a GNU Make-specific check. E.g., how about something like

```diff
 PACKAGE_VERSION := $(shell $(srcdir)/version.sh)
+ifeq "$(PACKAGE_VERSION)" ""
+$(error PACKAGE_VERSION is not defined.  Check version.sh works)
+endif
```

Which admittedly differs from your version in not letting you avoid the check by manually making _version.h_. Which is either a good or a bad thing.

daviesrob:
True, but I preferred a more traditional solution, although it did get a bit more complicated by hiding the if statement from the `make` output.  I did also ponder running `version.sh` in the make rule so its failure would be detected automatically, but I think the bit that quietly detects if `version.h` needs a rebuild is useful.

PR #2039, opened 2026-06-19
OSS-Fuzz: Add new fuzzer targets fai processing

arthurscchan:
This PR adds a new OSS-Fuzz fuzzer targets fai processing, together with the build fix in OSS-Fuzz (https://github.com/google/oss-fuzz/pull/15778) to enable the new fuzzer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,6 +100,7 @@ BUILT_TEST_PROGRAMS = \
 	test/test-vcf-sweep \
 	test/test-bcf-sr \
 	test/fuzz/hts_open_fuzzer.o \
+	test/fuzz/fai_build_fuzzer.o \
 	test/test-bcf-translate \
 	test/test-parse-reg \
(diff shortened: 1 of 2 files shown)


daviesrob:
Thanks for this.  I've been trying it out locally (via AFL++) and it rapidly generates lots of test cases.  One problem I've noticed though is (when run via AFL++) it can leave temporary files behind.  I suspect this happens with AFL decides to time out a test, as it kills the process, preventing the harness from running its clean-up code.  Possibly it could be worked around by calling `fai_build3()` directly instead of via `fai_load()`, and using some Linux-isms like the ability to pass unlinked but still open file descriptors via  `/proc/self/fd` (probably OK as long as we check for the directory first)?

I also noticed that you've left the copyright as "Genome Research Ltd."  That should probably be "ADA Logics Ltd." instead.

PR #2051, opened 2026-07-13
GCS overriding

PeterDowdy:
I added a `HTS_GCS_HOST` variable that, when set, mocks the server for GCS. This matches the approach used for S3 mocking. I did this because I'm building a spark library on top of htslib and would like to be able to do a full local integration test. Is this something you're interested in?

I ran my own integration test on fake-gcs-server, but it doesn't seem like there's an integration test for S3 against an S3 mock (like moto). Do you want me to add one to both?

I am also interested in adding Azure storage support.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,12 @@
 Noteworthy changes in release a.b
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* New environment variable HTS_GCS_HOST, analogous to the existing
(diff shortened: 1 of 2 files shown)


whitwham:
> I ran my own integration test on fake-gcs-server, but it doesn't seem like there's an integration test for S3 against an S3 mock (like moto). Do you want me to add one to both?

I would certainly be interested in looking at anything you come up with.

> I am also interested in adding Azure storage support.

We have very little access to Azure storage so this is not something that we are going to do ourselves.

whitwham:
Merged, thank you.

PR #2052, opened 2026-07-17
Use multi iterators in the synced reader

### LINKED ISSUE #2557: `view -R`: pathological performance with large single-base region sets (84M regions, 11+ hours unfinished)
## Summary

`bcftools view -R sites.bed.gz target.vcf.gz` is 350-400× slower than a sequential scan when the BED contains many densely-clustered single-base regions — a common pattern in PRS / ancestry pipelines (HGDP+1kGP, AADR 1240k, PGS Catalog). A production run on 84M such regions against a 23 GB VCF did not complete in 11+ hours of 100% CPU before we killed it.

## Reproducer

```bash
python3 gen_reproducer.py --regions 1000000 --prefix synth1m
bgzip -f synth1m.bed && tabix -p bed synth1m.bed.gz
bgzip -f synth1m.vcf && tabix -p vcf synth1m.vcf.gz
bcftools view -R synth1m.bed.gz synth1m.vcf.gz > /dev/null
```

<details>
<summary><code>gen_reproducer.py</code> (N densely-clustered single-base positions, written to both BED and VCF)</summary>

```python
import argparse, random

# Approx GRCh38 autosome lengths (Mbp) — used as relative weights only.
CHROM_LEN_MBP = [248,242,198,190,181,170,159,145,138,133,
                 135,133,114,107,101, 90, 83, 80, 58, 64, 46, 50]

def gen_positions(n_total, avg_spacing, seed=42):
    rng = random.Random(seed)
    rate = 1.0 / avg_spacing
    total = sum(CHROM_LEN_MBP)
    for chrom_i, mbp in enumerate(CHROM_LEN_MBP, start=1):
        length, share = mbp * 1_000_000, n_total * mbp // total
        pos = 0
        for _ in range(share):
            pos += max(1, int(rng.expovariate(rate)))
            if pos >= length: break
            yield str(chrom_i), pos

ap = argparse.ArgumentParser()
ap.add_argument("--regions", type=int, default=1_000_000)
ap.add_argument("--avg-spacing", type=int, default=10)
ap.add_argument("--prefix", default="synth")
args = ap.parse_args()
positions = list(gen_positions(args.regions, args.avg_spacing))

with open(f"{args.prefix}.bed", "w") as f:
    for chrom, pos in positions:
        f.write(f"{chrom}\t{pos-1}\t{pos}\n")

with open(f"{args.prefix}.vcf", "w") as f:
    f.write("##fileformat=VCFv4.2\n")
    for i, mbp in enumerate(CHROM_LEN_MBP, start=1):
        f.write(f"##contig=<ID={i},length={mbp * 1_000_000}>\n")
    f.write('##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">\n')
    f.write("#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\tFORMAT\tS1\tS2\tS3\n")
    for chrom, pos in positions:
        f.write(f"{chrom}\t{pos}\t.\tA\tG\t100\tPASS\t.\tGT\t0/0\t0/0\t0/0\n")
```
</details>

## Measurements (bcftools 1.23.1, htslib 1.23.1, macOS arm64)

N = 1,000,000 regions = 1,000,000 records:

| run | real | × baseline |
|---|---|---|
| `view -Ou vcf.gz > /dev/null` (baseline) | **0.45 s** | 1× |
| `view -R bed.gz` (default `--regions-overlap 1`) | **156.6 s** | **348×** |
| `view -R bed.gz --regions-overlap 0` | 193.1 s | 429× |
| `view -R bed.gz --regions-overlap 2` | 239.7 s | 533× |
| `view -T bed.gz` (streaming, `--targets-overlap 0`) | **1.34 s** | **3.0×** |
| `view -T bed.gz --targets-overlap 1` | 1.68 s | 3.7× |
| `view -T bed.gz --targets-overlap 2` | 1.08 s | 2.4× |

At N=5M: baseline 1.79 s, `view -R` default 705.5 s (394×). The 84M production case extrapolates to ≥ 12,000 s (~3.3 h) of iterator setup alone, before any meaningful work.

Two takeaways: `--regions-overlap` value is irrelevant to the bottleneck; `-T` (streaming) handles the identical workload at near-baseline speed.

## Root cause (htslib `synced_bcf_reader.c`)

`_readers_next_region()` calls `_reader_seek()` per BED region, which invokes a fresh `tbx_itr_queryi()`. For N regions, N iterator setup+teardown cycles. With 84M dense 1bp regions, that's 84M iterator queries when ~22 (one per autosome) would suffice.

```c
// synced_bcf_reader.c:595 — runs N times for N BED entries
static int _readers_next_region(bcf_srs_t *files) {
    ...
    if ( bcf_sr_regions_next(files->regions)<0 ) return -1;
    for (i=0; i<files->nreaders; i++)
        _reader_seek(&files->readers[i],            // ← fresh tbx_itr_queryi
                     files->regions->seq_names[files->regions->iseq],
                     files->regions->start, files->regions->end);
    return 0;
}
```

## Suggested fix

Filing as a feature request — semantics are correct, just impractically slow at this scale.

**Preferred:** region coalescing in `_readers_next_region`. If the next BED region starts within K bp of the current iterator position, keep streaming through the existing iterator and let the per-record overlap check at line 733 filter records in gaps. Collapses 84M iterator queries → ~22, generalizes to any dense-region workload. K tunable, default ~64 KB ≈ one BGZF block.

**Alternative:** opt-in `--regions-stream` flag forcing in-memory regidx + top-to-bottom VCF stream (semantically equivalent to `-T` but inheriting `-R`'s default overlap mode). Lowest-risk; preserves current behavior.

The existing `-T` path is the correctness baseline (1.34 s vs 0.45 s at 1M — see table); a regression test built on the reproducer above is straightforward.

---
bcftools 1.23.1 / htslib 1.23.1 / macOS arm64; reproducible on Linux x86_64 per the production observation above.

daviesrob:
Adds `bcf_itr_regions()` and `tbx_itr_regions()` interfaces as BCF and tabix equivalents to `sam_itr_regions()`.

Uses the new interfaces to add multi-iterator support to the BCF/VCF synced reader.  This makes it considerably faster when index jumping over a large number of regions by reducing the amount of work done looking up index entries.  Note that this only affects regions supplied via `bcf_sr_set_regions()` which works using iterators over the input files.  The similar `bcf_sr_set_targets()` function works as a filter and does not use iterators.

The locations supplied to `bcf_sr_set_regions()` can either be an in-memory list, or read from a file (which can itself be indexed with `tabix`).  This change should enable multi-iterators for both of these cases.

There are a few changes to behaviour as a result of this change, which unfortunately can't be completely hidden from callers due to the visibility of the synced reader structures in the public headers. 

* The `bcf_sr_regions_t::start` and `bcf_sr_regions_t::end` fields don't track the location of the current output records as accurately as before because the iterators no longer stop at the end of each listed region.
* `bcf_sr_regions_t::regs` may be non-NULL when reading regions from a tabix indexed file.
* More storage is needed when reading regions from a tabix indexed file, as regions are now handled in chromosome-sized chunks instead of individually.

Additionally this fixes a bug where setting the `BCF_SR_REGIONS_OVERLAP` option to 2 ("true variant overlap") with `bcf_sr_set_opt()` could result in records being returned out of order. This could happen if the variant in the record occurred after the end of the first region it overlapped, but also hit a later region.

Fixes samtools/bcftools#2557

### DIFF
--- a/htslib/vcf.h
+++ b/htslib/vcf.h
@@ -1353,6 +1353,25 @@ overlaps more than one region, it will only be returned once.
     hts_itr_t *bcf_itr_regarray(const hts_idx_t *idx, bcf_hdr_t *hdr,
                                 char **regarray, unsigned int regcount);
 
+/// Create a multi-region iterator from a hts_reglist_t regions list
+/** @param idx       Index
+    @param hdr       Header
+    @param reglist   Array of regions to iterate over
+    @param regcount  Number of items in reglist
+
+Each @p reglist entry should have the reference name in the `reg` field, an
+array of regions for that reference in `intervals` and the number of items
+in `intervals` should be stored in `count`.  No other fields need to be filled
+in.
+
+The iterator will return all reads overlapping the given regions.  If a read
+overlaps more than one region, it will only be returned once.
+ */
+
+    HTSLIB_EXPORT
+    hts_itr_t *bcf_itr_regions(const hts_idx_t *idx, bcf_hdr_t *hdr,
+                               hts_reglist_t *reglist, unsigned int regcount);
+
     static inline int bcf_itr_next(htsFile *htsfp, hts_itr_t *itr, void *r) {
         if (!htsfp->is_bgzf) {
             hts_log_error("Only bgzf compressed files can be used with iterators");
--- a/synced_bcf_reader.c
+++ b/synced_bcf_reader.c
@@ -589,21 +589,24 @@ static void decr_multi_running(region_t *reg)
     reg->creg = reg->multi_end;
 }
 
-void bcf_sr_remove_reader(bcf_srs_t *files, int i)
+void bcf_sr_remove_reader(bcf_srs_t *files, int reader_idx)
 {
     assert( !files->samples );  // not ready for this yet
     int *autoclose = BCF_SR_AUX(files)->closefile;
 
-    if (i < 0 || i >= files->nreaders)
+    if (reader_idx < 0 || reader_idx >= files->nreaders)
         return;
 
-    bcf_sr_sort_remove_reader(files, &BCF_SR_AUX(files)->sort, i);
-    int had_itr = bcf_sr_destroy1(&files->readers[i], autoclose[i]);
-    if ( i+1 < files->nreaders )
+    bcf_sr_sort_remove_reader(files, &BCF_SR_AUX(files)->sort, reader_idx);
+    int had_itr = bcf_sr_destroy1(&files->readers[reader_idx], autoclose[reader_idx]);
+    if ( reader_idx+1 < files->nreaders )
     {
-        memmove(&files->readers[i], &files->readers[i+1], (files->nreaders-i-1)*sizeof(bcf_sr_t));
-        memmove(&files->has_line[i], &files->has_line[i+1], (files->nreaders-i-1)*sizeof(int));
-        memmove(&autoclose[i], &autoclose[i+1], (files->nreaders-i-1)*sizeof(int));
+        memmove(&files->readers[reader_idx], &files->readers[reader_idx+1],
+                (files->nreaders-reader_idx-1)*sizeof(bcf_sr_t));
+        memmove(&files->has_line[reader_idx], &files->has_line[reader_idx+1],
+                (files->nreaders-reader_idx-1)*sizeof(int));
+        memmove(&autoclose[reader_idx], &autoclose[reader_idx+1],
+                (files->nreaders-reader_idx-1)*sizeof(int));
     }
     if (files->regions && files->regions->regs)
     {
@@ -613,9 +616,9 @@ void bcf_sr_remove_reader(bcf_srs_t *files, int i)
             region_t *reg = &files->regions->regs[seq];
             if (reg->multi_curr)
             {
-                if (i + 1 < files->nreaders)
-                    memmove(&reg->multi_curr[i], &reg->multi_curr[i + 1],
-                            (files->nreaders-i-1)*sizeof(*reg->multi_curr));
+                if (reader_idx + 1 < files->nreaders)
+                    memmove(&reg->multi_curr[reader_idx], &reg->multi_curr[reader_idx + 1],
+                            (files->nreaders-reader_idx-1)*sizeof(*reg->multi_curr));
                 if (had_itr)
                     decr_multi_running(reg);
             }
--- a/synced_bcf_reader.c
+++ b/synced_bcf_reader.c
@@ -62,6 +62,9 @@ typedef struct bcf_sr_region_t
 {
     region1_t *regs;            // regions will sorted and merged, redundant records marked for skipping have start>end
     int nregs, mregs, creg;     // creg: the current active region
+    int multi_end;              // end point for multi iterators
+    int multi_running;          // number of running multi iterators
+    int *multi_curr;            // current region for multi iterators
 }
 region_t;
 
@@ -516,10 +519,10 @@ bcf_srs_t *bcf_sr_init(void)
     return files;
 }
 
-static void bcf_sr_destroy1(bcf_sr_t *reader, int closefile)
+static int bcf_sr_destroy1(bcf_sr_t *reader, int closefile)
 {
     if (!reader)
-        return;
+        return 0;
     free(reader->fname);
     if ( reader->tbx_idx ) tbx_destroy(reader->tbx_idx);
     if ( reader->bcf_idx ) hts_idx_destroy(reader->bcf_idx);
@@ -528,13 +531,15 @@ static void bcf_sr_destroy1(bcf_sr_t *reader, int closefile)
         if (hts_close(reader->file) < 0)
             hts_log_error("Error on closing %s", reader->fname);
     }
-    if ( reader->itr ) tbx_itr_destroy(reader->itr);
+    int had_itr = reader->itr != NULL;
+    if ( had_itr ) tbx_itr_destroy(reader->itr);
     int j;
     for (j=0; j<reader->mbuffer; j++)
         bcf_destroy1(reader->buffer[j]);
     free(reader->buffer);
     free(reader->samples);
     free(reader->filter_ids);
+    return had_itr;
 }
 
 void bcf_sr_destroy(bcf_srs_t *files)
@@ -561,6 +566,21 @@ void bcf_sr_destroy(bcf_srs_t *files)
     free(files);
 }
 
+// Deal with completed multi iterators.  On closing the last free the
+// array tracking which region each iterator has reached and reset the
+// current region to all files to where the multi iterators finished
+static void decr_multi_running(region_t *reg)
+{
+    if (!reg->multi_curr)
+        return; // Not using multi iterators
+    if (--reg->multi_running >= 1)
+        return; // At least one iterator still running
+    // All multi iterators finished, tidy up ready to move on to the next
+    free(reg->multi_curr);
+    reg->multi_curr = NULL;
+    reg->creg = reg->multi_end;
+}
+
 void bcf_sr_remove_reader(bcf_srs_t *files, int i)
 {
     assert( !files->samples );  // not ready for this yet
@@ -570,12 +590,28 @@ void bcf_sr_remove_reader(bcf_srs_t *files, int i)
         return;
 
     bcf_sr_sort_remove_reader(files, &BCF_SR_AUX(files)->sort, i);
-    bcf_sr_destroy1(&files->readers[i], autoclose[i]);
+    int had_itr = bcf_sr_destroy1(&files->readers[i], autoclose[i]);
     if ( i+1 < files->nreaders )
     {
         memmove(&files->readers[i], &files->readers[i+1], (files->nreaders-i-1)*sizeof(bcf_sr_t));
         memmove(&files->has_line[i], &files->has_line[i+1], (files->nreaders-i-1)*sizeof(int));
         memmove(&autoclose[i], &autoclose[i+1], (files->nreaders-i-1)*sizeof(int));
+    }
+    if (files->regions && files->regions->regs)
+    {
+        int seq;
+        for (seq = 0; seq < files->regions->nseqs; seq++)
+        {
+            region_t *reg = &files->regions->regs[seq];
+            if (reg->multi_curr)
+            {
+                if (i + 1 < files->nreaders)
+                    memmove(&reg->multi_curr[i], &reg->multi_curr[i + 1],
+                            (files->nreaders-i-1)*sizeof(*reg->multi_curr));
+                if (had_itr)
+                    decr_multi_running(reg);
+            }
+        }
     }
     files->nreaders--;
 }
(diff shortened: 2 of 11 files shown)


jkbonfield:
How does this
> The bcf_sr_regions_t::start and bcf_sr_regions_t::end fields don't track the location of the current output records as accurately as before because the iterators no longer stop at the end of each listed region.

Interact with the frameshifts plugin here:
https://github.com/samtools/bcftools/blob/develop/plugins/frameshifts.c#L125-L145

It explicitly checks start and end for each region.  The comment is a bit vague about tracking as accurately as before, but it sounds like this should now fail.  I'll explore more in testing.

Edit: it's a different interface and is apparently not interacting with the changes, although this is very unclear initially.

htslib/vcf.h
```diff
@@ -1353,6 +1353,25 @@ overlaps more than one region, it will only be returned once.
     hts_itr_t *bcf_itr_regarray(const hts_idx_t *idx, bcf_hdr_t *hdr,
                                 char **regarray, unsigned int regcount);
 
+/// Crate a multi-region iterator from a hts_reglist_t regions list
```
jkbonfield: Typo: "Create"

synced_bcf_reader.c
```diff
@@ -570,13 +598,29 @@ void bcf_sr_remove_reader(bcf_srs_t *files, int i)
...
+        int seq;
+        for (seq = 0; seq < files->regions->nseqs; seq++)
+        {
+            region_t *reg = &files->regions->regs[seq];
+            if (reg->multi_curr)
+            {
+                if (i + 1 < files->nreaders)
+                    memmove(&reg->multi_curr[i], &reg->multi_curr[i + 1],
+                            (files->nreaders-i-1)*sizeof(*reg->multi_curr));
+                if (had_itr)
```
jkbonfield: Any chance we can fix the function declaration to use something other than `int i`.  Looking at the review I couldn't figure out what was happening, because my instinct was this is a loop variable yet we only have `seq` for that.  I later realised `i` is passed in as an argument!

Something far more descriptive would be useful, especially given it has no documentation or comment above the function.  It doesn't need to be long or overly verbose, just not something most commonly associated with trivial local looping.

synced_bcf_reader.c
```diff
@@ -659,6 +703,124 @@ static int _reader_seek(bcf_sr_t *reader, const char *seq, hts_pos_t start, hts_
...
+        reglist->intervals = hts_malloc_p(sizeof(*reglist->intervals),
+                                          reglist_in->count);
+        if (!reglist->intervals)
+            goto memfail;
+        memcpy(reglist->intervals, reglist_in->intervals,
+               sizeof(*reglist->intervals) * reglist_in->count);
+    } else {
+        // Use the one passed in
+        reglist = reglist_in;
+    }
```
jkbonfield, line 829: This still leaves `reglist->reg` as a copy to the old pointer.  Does it matter if we don't completely copy `reglist_in` and have a shared pointer?

I believe not, as it's just a cached copy of `regions->seq_names[regions->iseq]` anyway, but thought I'd double check.

synced_bcf_reader.c
```diff
@@ -733,13 +906,59 @@ static void _set_variant_boundaries(bcf1_t *rec, hts_pos_t *beg, hts_pos_t *end)
...
 }
 
+static int forward_check_overlap(region_t *reg, int idx,
+                                 hts_pos_t beg, hts_pos_t end)
+{
+    do {
+        ++idx;
+    } while ( idx < reg->nregs
+              && ( beg > reg->regs[idx].end
+                   || reg->regs[idx].start > reg->regs[idx].end) );
```
jkbonfield, line 952: How is it possible for `reg[idx].start > reg[idx].end`?  This feels like corrupted data, as each region should always have end >= start.

Is this a typo for something else?

Edit: no - it's documented in region1_t as:

```
    hts_pos_t start, end;   // records are marked for skipping have start>end
```

This probably needs a comment as the code simply looks wrong.

synced_bcf_reader.c
```diff
@@ -659,6 +703,124 @@ static int _reader_seek(bcf_sr_t *reader, const char *seq, hts_pos_t start, hts_
...
+        return NULL;
+    }
+
+    reglist->reg = regions->seq_names[regions->iseq];
+    reglist->count = 0;
+
+    int32_t i;
+    for (i = reg->creg; i < reg->nregs; i++) {
+        if (reg->regs[i].start > reg->regs[i].end)
+            continue;  // Marked to skip by merge_regions
```
jkbonfield: I cannot find a function named `merge_regions`.

After trial and error I found `regions_merge` instead.  This and the other mention need renaming.  It's not relating to changes in this PR, but I was attempting to figure out why we checked start > end and discovered these comments in the process.

synced_bcf_reader.c
```diff
@@ -733,13 +906,59 @@ static void _set_variant_boundaries(bcf1_t *rec, hts_pos_t *beg, hts_pos_t *end)
...
+                               hts_pos_t pos, hts_pos_t beg, hts_pos_t end)
+{
+    int midx = reg->multi_curr[reader_idx];
+    // Scan to find the next region ending after pos
+    while ( midx < reg->nregs )
+    {
+        if ( pos <= reg->regs[midx].end )
+            break;
+        do { // Advance, also avoiding regions marked to skip by merge_regions()
+            ++midx;
```
jkbonfield, line 968: Fix to `regions_merge()`.

synced_bcf_reader.c
```diff
@@ -733,13 +906,59 @@ static void _set_variant_boundaries(bcf1_t *rec, hts_pos_t *beg, hts_pos_t *end)
...
+    // If not, need to scan forward to see if it hits a later one
+    return forward_check_overlap(reg, midx, beg, end);
+}
+
 /*
  *  _reader_fill_buffer() - buffers all records with the same coordinate
  *  returns 0 on success, -1 on failure (also sets files->errnum)
  */
-static int _reader_fill_buffer(bcf_srs_t *files, bcf_sr_t *reader)
+static int _reader_fill_buffer(bcf_srs_t *files, int reader_idx)
```
jkbonfield: If we're modifying internal function declarations, then should we also make them C specification compliant by removing the leading underscores?

jkbonfield (commented):
Mostly minor issues.  Typos and queries.  Quite a few of my problems with it stem from the original code.  I needed to figure out what was going on in order to understand the changes, and that's far from clear due to the opaqueness and lack of adequate commenting.

Comparing against bcftools/plugins/frameshifts.c, and our conversation earlier:

Initially it was unclear why frameshifts isn't bugged given the change in behaviour of bcf_sr_regions_t::start/end.  However I believe it's the difference between bcf_sr_next_line vs bcf_sr_regions_overlap. Although it's unclear how bcf_sr_regions_next also differs here.  I think it falls into the same camp as bcf_sr_regions_overlap, which is that it doesn't use a multi-region iterator.

It looks like bcf_sr_next_line operates on file(s), while bcf_sr_regions_next operates on region lists instead (agnostic of associated files). The documentation doesn't adequately explain why it's beneficial when processing a single file to add it to an array of incoming synced-readers and iterate that way instead of using the manual approach.  It probably should add a comment however stating the benefits of doing multi-region work when using the file based interfaces.  (Plus maybe a comment on bcf_sr_regions_overlap as to why it's *not* using a file, which is a bit confusing given the naming convention.)

Also "bcf_sr_regions_next" claims to return the next interval, while "bcf_sr_regions_overlap" checks a specific location, presumably (but unstated) updating bcf_sr_regions_t::start/end itself if the queried region does infact overlap.   This needs documenting, as bcf_sr_regions_next currently does.

That said, if bcf_sr_regions_next isn't updating start/end correctly in some circumstances, then those also need explicitly mentioning in the documentation too.  Or rather I suspect it's setting them correctly, but the regions may not be the same ones given in the input file due to region-merging?

jkbonfield:
I still need to do some integration testing.  The above was simply a review by reading the code, rather than stress testing it.  That's up next.  I'll put on my evil-user hat! :-)

jkbonfield:
If we have a bed file that contains no overlap, eg our VCF is chr20 only and our bed file is for chr1 only, then there's a difference, albeit in warning only.

Dev prints up nothing, while this PR prints up:

```
[W::hts_itr_regions] Region 'chr1' specifies an unknown reference name. Continue anyway
```

This is incorrect, as chr1 is infact in the header contig lines.

jkbonfield:
I see vcfstats.c uses `bcf_sr_next_line()` , and hence speeds up with the multi-region iterator.  However it also has a `do_indel_stats` function which [accesses start and end directly](https://github.com/samtools/bcftools/blob/develop/vcfstats.c#L776-L789).

I thought I managed to get a difference in output, but it was just me forgetting to grep out the header lines.

Nonetheless,  it is running considerably faster in this PR, implying that the multi-region iterator is in use.  So I wonder in what scenarios the start / end are incorrect and whether I'm just getting lucky on this data set.

It may be that the field it's accessing is the underlying bed iterator rather than the region iterator.  It's all a bit convoluted!

jkbonfield:
The following is a problematic BED file as start == end, and start is zero based so it should always be less than end.

```
chr1    3336336 3336338                                                                                  
chr1    3336345 3336345                                                                                  
chr1    3336345 3336353                                                                                  
```

Nonetheless, a zero-sized region should not invalidate the next line, and our handling of this case is a regression.

```
for i in dev pr;do echo === $i ===;./bcftools.$i view -R _2.bed /nfs/users/nfs_j/jkb/scratch/data/dragen-novaseq.pcr-free.20x/HG002.novaseq.pcr-free.20x.vcf.gz|egrep -v '^#'|cut -f 1-5;done
=== dev ===
chr1	3336344	.	ATG	A
=== pr ===
```

Although `chr1    3336345 3336353` would appear to be `chr1:3336346-3336353` in 1-based coordinations, as this is a 2bp deletion it spans that region and should be found.

If I remove either of the first two lines, then the 3rd line now reports the entry in both tools.

It's marginal though, so maybe this is an off-by-one error lurking in there somewhere.

jkbonfield:
The above bug is fixed by this:

```
diff --git a/synced_bcf_reader.c b/synced_bcf_reader.c
index 6b3f193f..a05ef0f3 100644
--- a/synced_bcf_reader.c
+++ b/synced_bcf_reader.c
@@ -936,9 +936,12 @@ static int multi_check_overlap(region_t *reg, int reader_idx,
     if ( midx >= reg->nregs )
         return 0;  // pos is after all regions
 
-    // Skip records starting inside previous region or
+    // Skip records starting inside previous (non-skipped) region or
     // ending before the current one starts
-    hts_pos_t prev_end = midx > 0 ? reg->regs[midx - 1].end : -1;
+    int pidx = midx-1;
+    while (pidx >= 0 && reg->regs[pidx].start > reg->regs[pidx].end)
+        pidx--;
+    hts_pos_t prev_end = pidx >= 0 ? reg->regs[pidx].end : -1;
     if ( beg <= prev_end || end < reg->regs[midx].start )
         return 0;
```

The problem is prev_end was accessing a record which had been skipped due to it being zero length.  The general skip strategy (start>end, which it will be for zero length data due to the nature of BED) fixes it.

synced_bcf_reader.c
daviesrob: No, it shouldn't matter.  The pointer needs to be valid when calling `tbx_itr_regions()` / `bcf_itr_regions()` but after that it isn't used.  Ownership of the string pointed to is also expected to be elsewhere.

htslib/vcf.h
daviesrob: Typo fixed.

synced_bcf_reader.c
daviesrob: It's now `reader_idx` and some documentation has been added to the header.

synced_bcf_reader.c
daviesrob: Comment added.

synced_bcf_reader.c
daviesrob: Changed to `regions_merge()`, along with the other instances mentioned.

synced_bcf_reader.c
daviesrob: Done.

daviesrob:
The bug fix has been included and the "unknown reference" messages have been suppressed.

The accessing of `bcf_sr_regions_t::start` and `bcf_sr_regions_t::end` in `bcftools stats` and `bcftools plugin frameshifts` shouldn't be a problem.  In both cases the `bcf_sr_regions_t` structure is created directly, and is being used via `bcf_sr_regions_overlap()` rather than to make an iterator.  That usage is unaffected by this change.

synced_bcf_reader.c
pd3: I checked: this is required as a consequence of `regions_merge()` which can produce redundant regions. It is cheaper to mark as redundant (by making them invalid) then memmove the structures around.

pd3:
image

I tested on a 24 GB VCF (vcf.gz) and 282,615 regions (tabix-indexed) using the command.
```
bcftools query -f '%CHROM\t%POS\t%REF\t%ALT' file.vcf.gz -R regions.txt.gz
```
The new version is consistently 4.5x faster.

The only difference between the outputs is that the original outputs duplicate regions.

Thank you for this!

jkbonfield:
Test:

@claude review

daviesrob:
Rebased with various fixes squashed in, and an extra test for the `regions-overlap` option.

PR #2056, opened 2026-07-24
Add option to strip UR tags from CRAM

whitwham:
Add a function to the header manipulation code to remove all tags of a certain name from one type of header entry.

Add a cram format option to remove the UR tags from the SQ lines in the header. There is also a small fix for options that should of had an uppercase equivalent.

Part of the solution to samtools/samtools#2327 with a further PR to add an option to `samtools view`.

### DIFF
--- a/header.c
+++ b/header.c
@@ -3206,17 +3206,12 @@ int sam_hdr_remove_tag_all(sam_hdr_t *bh, const char *type, const char *tag_id) 
     t1 = t2 = kh_val(hrecs->h, k);
 
     do {
-        sam_hrec_tag_t *tag;
-        for (tag = t1->tag; tag; tag = tag->next) {
-            if (tag->str[0] == tag_id[0] && tag->str[1] == tag_id[1]) {
-                if ((ret = sam_hrecs_remove_key(hrecs, t1, tag_id)) > 0) {
-                    deleted++;
-                    break;
-                } else {
-                    return -1;
-                }
-            }
+        if ((ret = sam_hrecs_remove_key(hrecs, t1, tag_id)) == 1) {
+            deleted++;
+        } else if (ret == -1) {
+            return -1;
         }
+
         t1 = t1->next;
     } while (t1 != t2);
 
--- a/header.c
+++ b/header.c
@@ -3174,3 +3174,51 @@ enum sam_group_order sam_hrecs_group_order(sam_hrecs_t *hrecs) {
 
     return go;
 }
+
+
+/*
+ *
+ * Removes all the instances of a tag.
+ * Returns number removed (can be 0) or -1 on a fail.
+ *
+ */
+int sam_hdr_remove_tag_all(sam_hdr_t *bh, const char *type, const char *tag_id) {
+    sam_hrecs_t *hrecs;
+    sam_hrec_type_t *t1, *t2;
+    khint_t k;
+    int ret = 0;
+    int deleted = 0;
+
+    if (!bh || !type || !tag_id)
+        return -1;
+
+    if (!(hrecs = bh->hrecs)) {
+        if (sam_hdr_fill_hrecs(bh) != 0)
+            return -1;
+        hrecs = bh->hrecs;
+    }
+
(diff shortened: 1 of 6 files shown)


header.c
```diff
@@ -3174,3 +3174,51 @@ enum sam_group_order sam_hrecs_group_order(sam_hrecs_t *hrecs) {
...
+                    deleted++;
+                    break;
+                } else {
+                    return -1;
+                }
+            }
+        }
+        t1 = t1->next;
+    } while (t1 != t2);
+
```
jkbonfield: The `sam_hrec_type_t` loop (t1) makes sense as it's going through all the SQ records.  I'm less convinced with the `for (tag = t1->tag; tag; tag = tag->next) {` loop as we should only ever have one UR tag per SQ line.  This is required by the spec ("Within each (non-@CO) header line, no field tag may appear more than once "), but I think we only check some specific tags such as "ID".

Note that `sam_hrecs_remove_key` calls `tag = sam_hrecs_find_key(type, key, &prev);` which also has the same loop in it to find the tag, so even if we did have multiple duplicate tags, we'd end up in an O(N^2) loop here.  Not that it matters much given we'll have 0 or 1 only in reality so it's probably a non-issue.

Edit: sorry I didn't notice the `break`, so it stops on the first occurrence of the tag key.  In that case it may as well remove the inner loop and call `sam_hrecs_remove_key` already searches for the tag key itself.

header.c
```diff
@@ -3174,3 +3174,51 @@ enum sam_group_order sam_hrecs_group_order(sam_hrecs_t *hrecs) {
...
 }
+
+
+/*
+ *
+ * Removes all the instances of a tag.
+ * Returns number removed (can be 0) or -1 on a fail.
+ *
+ */
+int sam_hdr_remove_tag_all(sam_hdr_t *bh, const char *type, const char *tag_id) {
```
jkbonfield, line 3185: I'm unsure on the naming of the function.  It's probably fine as is, but just being cautious.

While the SAM spec refers to "tags", and described as "TAG:VALUE", other functions in header.c use "key" in the naming.  Eg `sam_hrecs_find_key` and `sam_hrecs_remove_key` when searching for specific tag names (ie key in the more traditional KEY:VALUE nomenclature).  Although we also have `sam_hdr_remove_tag_id` and `sam_hrecs_free_tags` which uses tags instead.  Oh for consistency...

I don't know what's best, but we may want to consider it more as this is a public API.  @daviesrob looked at the overall external header API naming more recently than me.

hts.c
```diff
@@ -1170,11 +1170,11 @@ int hts_opt_add(hts_opt **opts, const char *c_arg) {
         o->opt = CRAM_OPT_PREFIX, o->val.s = val;
 
     else if (strcmp(o->arg, "store_md") == 0 ||
-             strcmp(o->arg, "store_md") == 0)
+             strcmp(o->arg, "STORE_MD") == 0)
```
jkbonfield, line 1173: Good catch.  Thanks :)

jkbonfield (commented):
I haven't compiled it and kicked it in anger yet, but code wise it's OK.

Some minor points for discussion, but I can be persuaded either way.  Just musings.

jkbonfield:
One other comment: maybe we need to update one of the existing tests that writes cram to just add this as an option to exercise the code path.  Eg adding `-o remove_ur` to this line (test.pl:781 ish):

```
testv $opts, "./test_view $tv_args -t $ref -C -o VERSION=3.0 $bam > $cram";
```


We don't need to explicitly read it back and validate UR has been removed - just make sure the code is executed basically.  (Although GitHub copilot did suggest a function in test/sam.c to call the function and then go through the header API to verify the UR tag is now absent, which also works.)

header.c
whitwham: Yes, `sam_hrecs_remove_key`  does all the work the inner loop does.  I will get rid of the inner loop.

header.c
whitwham: I tried to name it after the public headers in **sam.h**, but I am not entirely happy with the current name either.

PR #2059, opened 2026-07-29
Add extra bgzip error checking

daviesrob:
Add more error checking to the bgzip application, mostly for memory allocations, but it will also more directly report read failures (previously these relied on `hclose()` returning an error after a read failure).  For brevity, reporting is done using the existing `error()` function, which is now also annotated with `HTS_NORETURN`.

Changes an instance of `strcpy()`/`strcat()` to the less controversial `snprintf()`.

So far I've resisted splitting the rather large `main()` function up, but it is quite tempting.

### DIFF
--- a/bgzip.c
+++ b/bgzip.c
@@ -39,6 +39,7 @@
 #include "htslib/bgzf.h"
 #include "htslib/hts.h"
 #include "htslib/hts_alloc.h"
+#include "htslib/hts_defs.h"
 #include "htslib/hfile.h"
 
 #ifdef _WIN32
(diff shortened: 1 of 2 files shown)


bgzip.c
```diff
@@ -349,14 +352,17 @@ int main(int argc, char **argv)
...
-                    strcpy(name, argv[optind]);
-                    strcat(name, ".gz");
+                    if (!fp)
+                        error("[bgzip] Couldn't open BGZF stream to stdout: %s\n",
+                              strerror(errno));
+                } else {
+                    size_t len = strlen(argv[optind]);
+                    char *name = hts_malloc_ps(sizeof(*name), len, 5);
+                    if (!name)
+                        error("[bgzip] Couldn't allocate space for output file name");
```
jkbonfield: `error()` requires a newline, unlike `hts_log_error`.  Maybe it should be changed, but easier to to fix it here.

jkbonfield (commented):
All looks fine bar a missing newline.

bgzip.c
daviesrob: Missing newline should now be present.

PR #2069, opened 2026-08-14
Apply CURL_CA_BUNDLE environment variable in the S3 backend too

jmarshall:
Refactor the repeated trivial CURLOPT setting into a subroutine, and additionally reinstate setting CURLOPT_CAINFO via the curl(1) tool's CURL_CA_BUNDLE environment variable, which the S3 code previously listened to via _hfile_libcurl.c_ code.

### DIFF
--- a/hfile_libcurl.c
+++ b/hfile_libcurl.c
@@ -1704,7 +1704,7 @@ int PLUGIN_GLOBAL(hfile_plugin_init,_libcurl)(struct hFILE_plugin *self)
     CURLSHcode errsh;
 
     err = curl_global_init(CURL_GLOBAL_ALL);
-    if (err != CURLE_OK) { errno = easy_errno(NULL, err); return -1; }
+    if (err != CURLE_OK) { errno = ENETDOWN; return -1; }
 
     curl.share = curl_share_init();
     if (curl.share == NULL) { curl_global_cleanup(); errno = EIO; return -1; }
--- a/hfile_s3.c
+++ b/hfile_s3.c
@@ -1452,6 +1452,20 @@ error:
     return headers;
 }
 
+static CURLcode set_common_easy_options(hFILE_s3 *fp) {
+    CURLcode err = CURLE_OK;
+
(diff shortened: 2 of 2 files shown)


jmarshall:
Added a second one-liner to make _hfile_libcurl.c_ handle `curl_global_init()` failure the same way as _hfile_s3.c_. These days `easy_errno()` uses `curl_easy_strerror()` in a log message, which actually caused a crash in conjunction with some upcoming pysam changes!

daviesrob:
Thanks.  Out of interest, how did `curl_easy_strerror()` fail?

jmarshall:
I suspected that might pique your interest… :smile:

Python wheels like to bundle the more exotic system libraries to make themselves portable across Linux distros. For a constellation of reasons (not least that this bundling is IMHO misguided), I'm changing pysam to omit libcurl and libcrypto — and the dozen others they drag in — from its wheels by dynamically loading them. (Sadly even building _hfile_libcurl.so_ et al as htslib plugins wasn't enough to defeat the bundling machinery.)

So I have a wrapper around `curl_global_init()` that dlopens libcurl and libcrypto and sets up a table of function pointers for all the library routines that _hfile\_*.c_ are going to call. But when that wrapper fails and returns CURLE_FAILED_INIT (because it was dlopening `libcrypto.so.3` on a platform that only had `libcrypto.so.1.1`… d'oh!), _hfile_libcurl.c_'s subsequent error handling called `curl_easy_strerror` which at that point was a function pointer that was NULL… :boom:

So technically it wasn't the **real** `curl_easy_strerror` that actually failed… but the point remains that the libcurl documentation says not to do this.

(BTW I've since refactored it to set up libcurl and libcrypto separately.)

This means that I won't be merging pysam-developers/pysam#1415 so pysam doesn't have an urgent need for the first commit of this PR after all. However it's still good to reinstate the CURL_CA_BUNDLE functionality for S3.

daviesrob:
Ah, yes.  I thought it might be something related to library loading.

PR #2070, opened 2026-08-17
Protect against high memory usage in malformed CRAM files.

jkbonfield:
Replication the 10,000 block limit in slices.  We already had it for allocating Content IDs (which is 1 per block), but oddly not for the block pointers themselves.

Also add a similar check for the container header too.

These won't limit valid or real-world CRAM usage.

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -3632,6 +3632,15 @@ cram_container *cram_read_container(cram_fd *fd) {
         return NULL;
     }
 #endif
+    // We already have an upper limit of 10,000 blocks per slice to prevent
+    // bad data using excessive memory.  We could in theory have many slices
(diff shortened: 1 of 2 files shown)


cram/cram_io.c
```diff
@@ -3632,6 +3632,15 @@ cram_container *cram_read_container(cram_fd *fd) {
         return NULL;
     }
 #endif
+    // We already have an upper limit of 10,000 blocks per slice to prevent
+    // bad data using excessive memory.  We could in theory have many slices
+    // per container, but in practice we only ever stick to 1 in modern
+    // implementations.  Nonetheless, let's use a limit here too on
+    // landmarks (offsets relative to here for block of
```
daviesrob: The end of this comment has been chopped off somehow...

PR #2072, opened 2026-08-18
Resolve cigar2 overflow

jkbonfield:
This PR is does two things.

1.  Fixes overflow bugs with long consecutive indels.  Eg 250000D x10.  We now basically fail as it represents something we cannot store in p->indel due to size limitations.  It shouldn't ever happen, but I was wary of it being a vector for curious bugs given it could turn deletions into insertions and vice versa.  This doesn't seem to trigger really bad behaviour with our code, but it could potentially elsewhere.

2. Refactor the resolve_cigar2 function, and try and speed it up somewhat to mitigate the performance hit of step 1 (which is surprisingly high given it's just swapping int for int64 essentially).

On short read data, this is around 8% slower (clang) and 11% slower (gcc).
On long read data it's 2% slower (clang) and 5% slower (gcc).

Note those timings are `resolve_cigar2` only, measured using `-fno-inline` and `perf record` to check function counters.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -5404,86 +5404,111 @@ static inline int resolve_cigar2(bam_pileup1_t *p, hts_pos_t pos, cstate_t *s)
     bam1_core_t *c = &b->core;
     uint32_t *cigar = bam_get_cigar(b);
     int k;
+
     // determine the current CIGAR operation
     //fprintf(stderr, "%s\tpos=%ld\tend=%ld\t(%d,%ld,%ld)\n", bam_get_qname(b), pos, s->end, s->k, s->x, s->y);
     if (s->k == -1) { // never processed
         p->qpos = 0;
         if (c->n_cigar == 1) { // just one operation, save a loop
-            if (_cop(cigar[0]) == BAM_CMATCH || _cop(cigar[0]) == BAM_CEQUAL || _cop(cigar[0]) == BAM_CDIFF) s->k = 0, s->x = c->pos, s->y = 0;
-        } else { // find the first match or deletion
+            int op = _cop(cigar[0]);
+            if (op == BAM_CMATCH || op == BAM_CEQUAL || op == BAM_CDIFF)
+                s->k = 0, s->x = c->pos, s->y = 0;
+        } else { // find the first reference base (match, mismatch or deletion)
             for (k = 0, s->x = c->pos, s->y = 0; k < c->n_cigar; ++k) {
                 int op = _cop(cigar[k]);
                 int l = _cln(cigar[k]);
-                if (op == BAM_CMATCH || op == BAM_CDEL || op == BAM_CREF_SKIP ||
-                    op == BAM_CEQUAL || op == BAM_CDIFF) break;
-                else if (op == BAM_CINS || op == BAM_CSOFT_CLIP) s->y += l;
+                if (bam_cigar_type(op) & 2)      // consumes ref
+                    break;
+                else if (bam_cigar_type(op) & 1) // consumes query
+                    s->y += l;
             }
             assert(k < c->n_cigar);
             s->k = k;
         }
+
     } else { // the read has been processed before
-        int op, l = _cln(cigar[s->k]);
+        int op = _cop(cigar[s->k]);
+        int l  = _cln(cigar[s->k]);
         if (pos - s->x >= l) { // jump to the next operation
-            assert(s->k < c->n_cigar); // otherwise a bug: this function should not be called in this case
-            op = _cop(cigar[s->k+1]);
-            if (op == BAM_CMATCH || op == BAM_CDEL || op == BAM_CREF_SKIP || op == BAM_CEQUAL || op == BAM_CDIFF) { // jump to the next without a loop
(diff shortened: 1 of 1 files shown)


jkbonfield:
Claude review caught something more for me to explore:

_Overflow check doesn't cover the full qpos computation — moderate, but relevant given the fix's purpose_

```
if (MEX & (1<<op)) {
    p->qpos = s->y + (pos - s->x);
}
...
int ret = 1;
if (s->y > INT_MAX)
    ret = 0;
```

_p->qpos (an int32_t) is assigned s->y + (pos - s->x), but the overflow guard only checks s->y itself. If s->y is, say, INT_MAX - 100 and (pos - s->x) is larger than 100 (which can happen for a single large CIGAR op, since op lengths can be up to ~2^28), the sum can exceed INT_MAX and silently wrap when stored into the int32_t field — while ret stays 1 (success), so the corrupted qpos is not caught and gets returned to the caller as valid._

jkbonfield:
The overflow reported above, while true, can only occur when we have more than 2^31 bases in a read as the operations are on query position rather than reference position.

Specifically, in `p->qpos = s->y + (pos - s->x)`, `pos - s->x` is how far through a cigar operation we are (max 2^28) and it's within a clause requiring M, = or X cigar op.  Also `s->y` is query bases, so similarly needs M=X or also I/S.  We could add a check here, but doing so slows up this function (by about 8% measured) and it's impossible to get here.

SAM and BAM both validate that `b->core.l_qseq` is >= 0  (for some baffling reason it's signed in our code despite being labelled as uint32_t in the specification, but that was probably a recent spec edit).   CRAM technically can break this, but the weird type conversion going on in that code is likely to explode in memory failures as it gets cast to `size_t` for some of the allocations so negatives would fail too.  (I'll make that more explicit in a separate PR.)

So all the ways of breaking this code in an overflow are already guarded against elsewhere.  Given the performance hit, I've replaced the check with a comment.

jkbonfield:
Other Claude complaints on reviewing the PR:

_Pileup fields are written before the overflow check runs. In resolve_cigar2, p->qpos, p->is_del, p->is_refskip, p->cigar_ind (and possibly a truncated p->indel) are all set using the 64-bit s->y/indel values before the final if (s->y > INT_MAX) ret = 0; … return ret; check. Today this is harmless because the sole caller (bam_plp64_next) discards the entire iteration and returns NULL on ret == 0, but it means the function can leave a half-populated/garbage bam_pileup1_t behind on failure — fragile if another caller is ever added that doesn't immediately bail on a 0 return._

Correct, but it's deliberate.  It's what we would have had before, and I didn't want to change that output (ie bug compatible) but the return value is now correctly indicating failure and not updating those fields could also lead to bugs if someone doesn't check the return value, such as infinite loops.  Whether that decision is correct or not is debatable, and I can see it from both viewpoints.

Also note this function is static inline, so in theory no one else should be calling it unless they're subverting the API protection.

sam.c
```diff
@@ -5393,102 +5385,145 @@ static inline void mp_free(mempool_t *mp, lbnode_t *p)
...
         if (pos - s->x >= l) { // jump to the next operation
-            assert(s->k < c->n_cigar); // otherwise a bug: this function should not be called in this case
-            op = _cop(cigar[s->k+1]);
-            if (op == BAM_CMATCH || op == BAM_CDEL || op == BAM_CREF_SKIP || op == BAM_CEQUAL || op == BAM_CDIFF) { // jump to the next without a loop
-              if (_cop(cigar[s->k]) == BAM_CMATCH|| _cop(cigar[s->k]) == BAM_CEQUAL || _cop(cigar[s->k]) == BAM_CDIFF) s->y += l;
-                s->x += l;
+            // otherwise a bug: this function should not be called in this case
+            assert(s->k+1 < c->n_cigar);
+
+            // Common bit between M=X and M=XDN
```
daviesrob: I think this comment only makes sense if you know what the previous code looked like.  Maybe change it to something like "Update cstate_t ref and query positions for the next CIGAR op".

sam.c
```diff
@@ -5393,102 +5385,145 @@ static inline void mp_free(mempool_t *mp, lbnode_t *p)
...
-                // Within a deletion (the 2D in 1D2D) we keep p->indel=0
+                // Within a deletion (the 2D in 1D2D) we keep indel=0
                 // and rely on is_del=1 as we would for 3D.
-                p->indel = -(int)l2;
+                indel = -l2;
                 for (k = s->k+2; k < c->n_cigar; ++k) {
-                    op2 = _cop(cigar[k]); l2 = _cln(cigar[k]);
-                    if (op2 == BAM_CDEL) p->indel -= l2;
-                    else break;
+                    const int op2 = _cop(cigar[k]);
```
daviesrob, line 5475: This shadows the outer `op2`.  Maybe rename to `op_k`?
(The same goes for the `BAM_CINS` and `BAM_CPAD` loops below.)

daviesrob (commented):
Mostly looks OK, apart from a bit of variable shadowing in the rewrite.

PR #2073, opened 2026-08-19
Add validation for l_qseq >= 0 in bam_set1

jkbonfield:
Minimal extra protection against malformed CRAMs.

It shouldn't have any major impact except for faster and more reliable failure.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -559,6 +559,14 @@ int bam_set1(bam1_t *bam,
     }
 
     // validate parameters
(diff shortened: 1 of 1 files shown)


sam.c
```diff
@@ -559,6 +559,13 @@ int bam_set1(bam1_t *bam,
     }
 
     // validate parameters
+    if ((int)l_seq < 0 || l_seq > INT32_MAX) {
```
daviesrob: You only need the second of these tests.

PR #2074, opened 2026-08-19
Vcf header fixes

### LINKED ISSUE #2065: four defects in VCF header parsing
Found while deriving the VCF structured-metadata (`##KEY=<...>`) grammar for an independent parser. Each
was observed behaviourally against the released binary, then traced to a specific line in `vcf.c`, then
re-confirmed. They are independent and can be split into separate issues; Issue 1 is the one with real
data-loss consequences.

- **Version:** bcftools 1.23.1 / htslib 1.23.1
- **Source:** tag `1.23.1`, commit `d1a2a873552a4fb8c30b3a77620c5f8bef7a7143`
- All line references are `vcf.c` at that commit.

Every reproduction below is a complete file. `##contig=<ID=9,length=5>` is included as a control so it is
visible that the rest of the header parses normally.

---

### Summary

If `length=` cannot be parsed as a non-negative integer, `bcf_hdr_register_hrec` returns 0 and the whole
`##contig` declaration is destroyed. **Nothing is printed** — not a warning, not an error. The contig
simply is not in the header any more. This includes the case where the value is a perfectly valid number
that merely happens to be quoted.

### Reproduction

```vcf
##fileformat=VCFv4.3
##contig=<ID=2,length=xyz>
##contig=<ID=9,length=5>
```

```console
$ bcftools view -h in.vcf 2>&1 >/dev/null      # stderr
                                                # (empty)
$ bcftools view -h in.vcf 2>/dev/null | grep contig
##contig=<ID=9,length=5>
```

Contig `2` is gone; stderr is empty. Identical silent loss for:

| `length=` value | stderr | contig kept? |
|---|---|---|
| `xyz` | *(empty)* | **no** |
| `-5` | *(empty)* | **no** |
| `"2000"` — a **valid** number, quoted | *(empty)* | **no** |
| `2000` (control) | *(empty)* | yes |

The quoted case is the most surprising: `length="2000"` is rejected purely for being quoted, and the
rejection costs the entire declaration rather than just the attribute.

### Expected

A diagnostic at minimum. Ideally the declaration is kept without a length (as happens when `length` is
absent entirely, which is fine), rather than discarded.

### Mechanism

`vcf.c:845-850` — `strtoll`; if no digits were consumed (`end == vals[i]`) or the result is negative,
`return 0`. `bcf_hdr_add_hrec` then destroys the hrec at `vcf.c:1141`. No logging on that path.

### Why this is worse than it looks

**It is inconsistent with every neighbouring interpreted key**, all of which warn and keep going:

| key | malformed-value behaviour |
|---|---|
| `IDX` | `hts_log_warning("Error parsing the IDX tag, skipping")` (`:882`, `:923`) |
| `Type` | `The type "%s" is not supported, assuming "String"` (`:936`) |
| `Number` | `... has no Number defined. Assuming '.'` (`:966`) |
| **`length`** | **silent, and the declaration is destroyed** |

And the only downstream signal is misleading: records referencing that contig later produce
`[W::vcf_parse] Contig '2' is not defined in the header`, which points at the records rather than at the
malformed header line that actually caused it. `fix_chromosome` (`:3743-3760`) may then silently
synthesise a replacement `##contig=<ID=2>`, so a round-trip can appear to *almost* work while having
dropped the length.

---

### Summary

htslib injects `##FILTER=<ID=PASS,Description="All filters passed">` before parsing the file (`:1425`).
A `PASS` declaration in the file is therefore always a duplicate, and duplicates are dropped first-wins
and silently — so the file's own `Description` is discarded with no diagnostic, and `bcftools view -h`
reports wording the file never contained.

### Reproduction

```vcf
##fileformat=VCFv4.3
##FILTER=<ID=PASS,Description="MY OWN TEXT">
##contig=<ID=9,length=5>
```

```console
$ bcftools view -h in.vcf 2>&1 >/dev/null       # stderr
                                                 # (empty)
$ bcftools view -h in.vcf 2>/dev/null | grep FILTER
```

### Expected

Injecting a default `PASS` is clearly intentional and correct — the spec requires it to exist, and a file
that omits it should still get one. The issue is the **silent override** of a file that *did* declare it.
Either the file's own declaration should win (it is more specific than a library default), or the
override should be reported. Silently rewriting a header line the file explicitly supplied is surprising
in a tool used for round-tripping.

### Mechanism

The default is added at `:1425`, before the file's own header lines are parsed. The file's `PASS` then
hits the duplicate path (`BCF_HL_STR`-with-ID, `:1164-1179`), which `return 0`s and destroys the hrec at
`:1141`, with no logging.

### Secondary consequence, for anyone measuring htslib behaviour

`PASS` is unusable as a probe for duplicate-declaration handling: it is *always* already duplicated, so
any experiment using it reports first-wins regardless of what the duplicate logic does. Using a
non-reserved ID (e.g. `##FILTER=<ID=q10,...>` twice) isolates the real behaviour.

---

### Summary

A `[`-delimited attribute value with no closing `]` is the **only** header malformation that is fatal to
the whole file — every other one drops a single line and continues. The accompanying message is
`strerror(errno)`, which has nothing to do with the failure.

### Reproduction

```vcf
##fileformat=VCFv4.3
##contig=<ID=2,attr=[a,length=2000>
##contig=<ID=9,length=5>
```

```console
$ bcftools view -h in.vcf; echo "exit=$?"
[E::bcf_hdr_parse_line] Missing ']' in header line "##contig=<ID=2,attr=[a,length=2000>"
[E::bcf_hdr_parse] Could not parse header line: No such file or directory
Failed to read from in.vcf: could not parse header
exit=255
```

`No such file or directory` is `errno` left over from something unrelated; the file exists and was read.

### Expected

Two things. The second message should not report a stale `errno` for a parse failure that has nothing to
do with I/O. And the severity is worth reconsidering: a malformed `[...]` value costs the entire file,
whereas a malformed quoted value costs one declaration. The first message is good — it names the problem
and quotes the line — so the fatality may simply be unintended.

### Mechanism

`vcf.c:748` — `goto fail` on a missing `]`, which sets `*len = -1` (`:778-781`). `bcf_hdr_parse` treats
`-1` as fatal and logs `Could not parse header line: %s` with `strerror(errno)` (`:1441-1444`). Only two
things reach `fail`: allocation failure (where `errno` *is* meaningful) and this.

---

### Summary

That warning formats `line` with a bare `%s`. `line` points into the full, NUL-terminated header buffer,
so the message runs from the offending line to the end of the header instead of stopping at the line.

### Reproduction

```vcf
##fileformat=VCFv4.3
##INFO=<ID=YY,Number=1,Type=String,Description="abc>
##FILTER=<ID=q10,Description="Quality below 10">
##contig=<ID=9,length=5>
```

```console
$ bcftools view -h in.vcf 2>&1 >/dev/null
[W::bcf_hdr_parse_line] Incomplete header line, trying to proceed anyway:
	[##INFO=<ID=YY,Number=1,Type=String,Description="abc>
##FILTER=<ID=q10,Description="Quality below 10">
##contig=<ID=9,length=5>
]
	[10]
```

Three unrelated lines are quoted as though they were part of the offending one. On a real file with a
large header this is a very large diagnostic.

### Expected

Bound the output to the line, as the sibling diagnostics already do.

### Mechanism

`vcf.c:761-762` uses a raw `%s`, while `Dropped trailing junk` (`:770`) and `Could not parse the header
line` (`:788`) both use `hts_strprint` with an explicit length. The trailing `[%d]` field is `q[0]`,
which by the loop's exit condition can only ever be `0` or `10` — so it is always either `0` or `10` and
conveys nothing.

---

## Reproduction environment

- bcftools 1.23.1 / htslib 1.23.1, stock release build.
- Every input above is a complete file, byte-verified against the intended payload before running
  (shell quoting silently corrupts backslashes), with a known-good control line in each file so that
  "everything vanished" and "the harness is broken" can be told apart.
- Behaviour was established against the binary first; `vcf.c` line references were then read at tag
  `1.23.1` to confirm the mechanism, and the predictions from that reading were re-measured against the
  binary before writing this up.

Notes: issue was drafted with AI.

jkbonfield:
Fixes #2065

### DIFF
--- a/vcf.c
+++ b/vcf.c
@@ -1002,8 +1002,26 @@ static int bcf_hdr_register_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec)
     if ( k != kh_end(d) )
     {
         // already present
+        int is_pass = strcmp(str, "PASS")==0;
         free(str);
-        if ( kh_val(d, k).hrec[info&0xf] ) return 0;
+        if ( kh_val(d, k).hrec[info&0xf] ) {
+            // Handle a duplicate ##FILTER=<ID=PASS,...> is possible as the
+            // user is permitted to add one, but we always interject our own
+            // after the ##fileformat line.
+            //
+            // I'm certain hdr->hrec[1] will always be our automatic
+            // PASS, but check anyway just incase.
+            if (is_pass && (info & 0xf) == BCF_HL_FLT &&
+                hdr->nhrec > 1 && hdr->hrec[1]->type == BCF_HL_FLT) {
+                if (idx == -1 && hrec_add_idx(hrec, kh_val(d, k).id) < 0) {
+                    return -1;
+                }
+                bcf_hrec_t hrec_pass = *hdr->hrec[1];
+                *hdr->hrec[1] = *hrec;
+                *hrec = hrec_pass;
+            }
+            return 0;
+        }
         kh_val(d, k).info[info&0xf] = info;
         kh_val(d, k).hrec[info&0xf] = hrec;
         if ( idx==-1 ) {
--- a/vcf.c
+++ b/vcf.c
@@ -656,14 +656,18 @@ bcf_hrec_t *bcf_hdr_parse_line(const bcf_hdr_t *h, const char *line, int *len)
 {
     bcf_hrec_t *hrec = NULL;
     const char *p = line;
+    char *reason = "unknown";
     if (p[0] != '#' || p[1] != '#') { *len = 0; return NULL; }
     p += 2;
 
     const char *q = p;
     while ( *q && *q!='=' && *q != '\n' ) q++;
     ptrdiff_t n = q-p;
-    if ( *q!='=' || !n ) // wrong format
+    if ( *q!='=' || !n ) {
+        // wrong format
+        reason = "missing '='";
         goto malformed_line;
+    }
 
     hrec = (bcf_hrec_t*) calloc(1,sizeof(bcf_hrec_t));
     if (!hrec) { *len = -1; return NULL; }
@@ -702,8 +706,10 @@ bcf_hrec_t *bcf_hdr_parse_line(const bcf_hdr_t *h, const char *line, int *len)
         n = q-p;
         int m = 0;
         while ( *q && *q==' ' ) { q++; m++; }
-        if ( *q!='=' || !n )
+        if ( *q!='=' || !n ) {
+            reason = "missing '='";
             goto malformed_line;
+        }
 
         if (bcf_hrec_add_key(hrec, p, q-p-m) < 0) goto fail;
         p = ++q;
@@ -742,11 +748,8 @@ bcf_hrec_t *bcf_hdr_parse_line(const bcf_hdr_t *h, const char *line, int *len)
                 q++;
                 quoted = 0;
             } else {
-                char buffer[320];
-                hts_log_error("Missing ']' in header line %s",
-                              hts_strprint(buffer, sizeof(buffer), '"',
-                                           line, q-line));
-                goto fail;
+                reason = "missing ']'";
+                goto malformed_line;
             }
         }
         while ( r > p && r[-1] == ' ' ) r--;
@@ -785,7 +788,7 @@ bcf_hrec_t *bcf_hdr_parse_line(const bcf_hdr_t *h, const char *line, int *len)
     {
         char buffer[320];
         while ( *q && *q!='\n' ) q++;  // Ensure *len includes full line
-        hts_log_error("Could not parse the header line: %s",
+        hts_log_error("Could not parse the header line (%s): %s", reason,
                       hts_strprint(buffer, sizeof(buffer),
                                    '"', line, q - line));
         *len = q - line + (*q ? 1 : 0);
--- a/vcf.c
+++ b/vcf.c
@@ -762,8 +762,10 @@ bcf_hrec_t *bcf_hdr_parse_line(const bcf_hdr_t *h, const char *line, int *len)
             q++;
         }
     }
-    if ( nopen )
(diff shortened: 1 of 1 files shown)


vcf.c
```diff
@@ -846,9 +851,16 @@ static int bcf_hdr_register_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec)
...
         if ( i<0 ) len = 0;
         else {
-            char *end = hrec->vals[i];
-            len = strtoll(hrec->vals[i], &end, 10);
-            if (end == hrec->vals[i] || len < 0) return 0;
+            char *start = hrec->vals[i] + (*hrec->vals[i] == '"'), *end;
+            len = strtoll(start, &end, 10);
+            if (start == end || len < 0) {
+                int id_i = bcf_hrec_find_key(hrec,"ID");
+                if ( i<0 ) return 0;
```
daviesrob: Should this be `id_i`?

jkbonfield: After a bit of head scratching, yes.  Too many `i`s floating around in this code!

vcf.c
```diff
@@ -982,16 +993,35 @@ static int bcf_hdr_register_hrec(bcf_hdr_t *hdr, bcf_hrec_t *hrec)
...
 
     vdict_t *d = (vdict_t*)hdr->dict[BCF_DT_ID];
-    k = kh_get(vdict, d, str);
+    k = kh_get(vdict, d, id_copy);
     if ( k != kh_end(d) )
     {
         // already present
-        free(str);
-        if ( kh_val(d, k).hrec[info&0xf] ) return 0;
+        int is_pass = strcmp(id_copy, "PASS")==0;
```
daviesrob, line 1004: This test could be moved inside the next if statement, as `is_pass` is only used there so there's no point in doing the `strcmp()` unless the record looks like a duplicate.

jkbonfield: It's only there so we don't have two places to free id_copy.

PR #2076, opened 2026-08-20
hts_expr fuzzer plus related bug fixes

jkbonfield:
The fuzzer may be a bit rough around the edges perhaps, but it's not a standard build target and it's already found some issues.

- Fixed a divide by zero bug in modulo arithmetic on non-zero values that convert to zero after integer conversion
- Fixed several small memory leaks in returning from malformed expression


test/fuzz/fuzz_expr.c
```diff
@@ -0,0 +1,96 @@
...
+static bam1_t *b = NULL;
+static sam_hdr_t *hdr = NULL;
+static samFile *in = NULL;
+
+// Possible race condition, but don't particularly care for the fuzzer.
+// We could also construct a bam using bam_set1().
+static void init_bam(void) {
+    puts("INIT_BAM START");
+    opened = 1;
+    hFILE *memfile = hopen("mem:", "rb:", SAM_DAT, strlen(SAM_DAT));
```
daviesrob: This could be simplified by using a `data:` URL, see for example https://github.com/samtools/htslib/blob/0290336036f83bdd4132b232c1e6a32a2cbe346c/test/test-vcf-api.c#L570

test/fuzz/fuzz_expr.c
```diff
@@ -0,0 +1,96 @@
...
+    char expr[8192];
+    size_t len = strnlen((char *)data, size);
+    if (len > 8191)
+        len = 8191;
+    memcpy(expr, data, len);
+    expr[len] = 0;
+
+    hts_filter_t *filt = hts_filter_init(expr);
+    if (!filt)
+        return -1;
```
daviesrob: OSS-fuzz now discourages returning -1 from `LLVMFuzzerTestOneInput`.  See [commit 7e8d53bc](https://github.com/google/oss-fuzz/commit/7e8d53bc9974afa6823c2aa25571e661d133d980) - the explanation in the commit message expands a bit beyond the updated text.  It doesn't really explain why this may lead to a faulty harness (apart from the Honggfuzz compatibility issue), but I guess that not remembering invalid inputs may lead to the fuzzer endlessly rediscovering them, which could slow it down.

daviesrob (commented):
This adds rather a lot of files in the `corpus.expr` directory, making the entire project over 20% bigger.  As most users don't need this, and it will make `git` operations slower for them, would it be better to put them all into a single file?  It would be fairly simple to add a bit of script to the Makefile to expand them out into a directory when you build the fuzzer.

jkbonfield:
> This adds rather a lot of files in the `corpus.expr` directory, making the entire project over 20% bigger. As most users don't need this, and it will make `git` operations slower for them, would it be better to put them all into a single file? It would be fairly simple to add a bit of script to the Makefile to expand them out into a directory when you build the fuzzer.

I make it around 5% bigger in disk space, but a minuscule difference in download size due to the many small files causing bigger disk blocks.

I'll add a Makefile rule though.

jkbonfield:
Now pushed a new commit, to be squashed into 27a067af033164e575af66a26048ee66e142a082.

- Replace the test/fuzz/corpus.expr/* with a tar.gz file
- Return 0 instead of -1 from LLVMFuzzerTestOneInput
- Use data: instead of mem:

jkbonfield:
I've done some rebasing and the first commit is the fuzzing harness, so you can check that out and run it to see all the horrors.

The following commits have been partially squashed together, but now contain the various classes of bug fixes covered.  I could move them to a new PR if you wish, but I don't think it matters hugely as the fuzz harness is pretty independent and not built by default so it's not going to break things if we have to fix or improve it later on.

Makefile
```diff
@@ -741,6 +741,25 @@ test/hts_endian: test/hts_endian.o
...
+	done < test/fuzz/corpus.expr.dat
+
+# Recommended build process:
+#   make CC=clang21 test/fuzz/fuzz_expr
+# Comment out the sub-make below for full rebuild
+test/fuzz/fuzz_expr: test/fuzz/fuzz_expr.c test/fuzz/corpus.expr
+	# make clean
+	# make CFLAGS="-fsanitize=address,undefined,fuzzer -g -O3" libhts.a -j8
+	$(CC) -fsanitize=address,undefined,fuzzer $(CFLAGS) $(TARGET_CFLAGS) $(ALL_CPPFLAGS) -o $@ test/fuzz/fuzz_expr.c $(LDFLAGS) libhts.a $(LIBS) -lpthread
+	@echo "Run with: (cd test/fuzz; ASAN_OPTIONS=allow_addr2line=1:abort_on_error=1:detect_leaks=1 ./fuzz_expr corpus.expr)"
```
daviesrob:
```suggestion
	@echo "Run with: ASAN_OPTIONS=allow_addr2line=1:abort_on_error=1:detect_leaks=1 ./test/fuzz/fuzz_expr -dict=$(srcdir)/test/fuzz/corpus.expr.dict test/fuzz/corpus.expr 2>&1 | cat -v"
```

Makefile
```diff
@@ -741,6 +741,25 @@ test/hts_endian: test/hts_endian.o
...
+	    n=`expr $$n + 1`; \
+	done < test/fuzz/corpus.expr.dat
+
+# Recommended build process:
+#   make CC=clang21 test/fuzz/fuzz_expr
+# Comment out the sub-make below for full rebuild
+test/fuzz/fuzz_expr: test/fuzz/fuzz_expr.c test/fuzz/corpus.expr
+	# make clean
+	# make CFLAGS="-fsanitize=address,undefined,fuzzer -g -O3" libhts.a -j8
+	$(CC) -fsanitize=address,undefined,fuzzer $(CFLAGS) $(TARGET_CFLAGS) $(ALL_CPPFLAGS) -o $@ test/fuzz/fuzz_expr.c $(LDFLAGS) libhts.a $(LIBS) -lpthread
```
daviesrob: It would be useful to make this work with out-of-tree builds.
```suggestion
test/fuzz/corpus.expr:
	mkdir test/fuzz/corpus.expr
	while read -r e; \
	do \
	    echo -n "$$e" > test/fuzz/corpus.expr/`printf "e.%03d" $$n`; \
	    n=`expr $$n + 1`; \
	done < $(srcdir)/test/fuzz/corpus.expr.dat

# Recommended build process:
#   make CC=clang21 test/fuzz/fuzz_expr
# Uncomment the sub-make below for full rebuild
test/fuzz/fuzz_expr: $(srcdir)/test/fuzz/fuzz_expr.c test/fuzz/corpus.expr
	# make clean
	# make CFLAGS="-fsanitize=address,undefined,fuzzer -g -O3" libhts.a -j8
	$(CC) -fsanitize=address,undefined,fuzzer $(CFLAGS) $(TARGET_CFLAGS) $(ALL_CPPFLAGS) -o $@ $(srcdir)/test/fuzz/fuzz_expr.c $(LDFLAGS) libhts.a $(LIBS) -lpthread
```

daviesrob (commented):
Some suggestions for the Makefile.  It might be an idea to add some cleaning rules too (maybe also adding one for the existing fuzzer, which seems to have been missed out so far...)

PR #2078, opened 2026-08-24
Add more memory allocation checks in header API functions

daviesrob:
* Check `kputsn()` works in `sam_hrecs_remove_ref_altnames()`. This function can now return an error, so also add checks where it gets called.

* Check `kputc()` and `kputsn()` worked in `sam_hdr_build_from_sam_file()`.

### DIFF
--- a/header.c
+++ b/header.c
@@ -113,17 +113,20 @@ static int sam_hrecs_add_ref_altnames(sam_hrecs_t *hrecs, int nref, const char *
     return 0;
 }
 
-static void sam_hrecs_remove_ref_altnames(sam_hrecs_t *hrecs, int expected, const char *list) {
+static int sam_hrecs_remove_ref_altnames(sam_hrecs_t *hrecs, int expected, const char *list) {
     const char *token, *sn;
     ks_tokaux_t aux;
     kstring_t str = KS_INITIALIZE;
 
     if (expected < 0 || expected >= hrecs->nref)
-        return;
+        return 0;
     sn = hrecs->ref[expected].name;
 
     for (token = kstrtok(list, ",", &aux); token; token = kstrtok(NULL, NULL, &aux)) {
-        kputsn(token, aux.p - token, ks_clear(&str));
+        if (kputsn(token, aux.p - token, ks_clear(&str)) < 0) {
+            free(str.s);
+            return -1;
+        }
         khint_t k = kh_get(m_s2i, hrecs->ref_hash, str.s);
         if (k != kh_end(hrecs->ref_hash)
             && kh_val(hrecs->ref_hash, k) == expected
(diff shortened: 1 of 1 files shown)


header.c
```diff
@@ -440,11 +444,13 @@ static int sam_hrecs_remove_hash_entry(sam_hrecs_t *hrecs, khint32_t type, sam_h
...
                 int idx = kh_val(hrecs->ref_hash, k);
+                if (altnames) {
+                    if (sam_hrecs_remove_ref_altnames(hrecs, idx, altnames) < 0)
+                        return -1;
+                }
                 if (idx + 1 < hrecs->nref)
                     memmove(&hrecs->ref[idx], &hrecs->ref[idx+1],
                             sizeof(sam_hrec_sq_t)*(hrecs->nref - idx - 1));
-                if (altnames)
-                    sam_hrecs_remove_ref_altnames(hrecs, idx, altnames);
```
jkbonfield, line 447: This moves the `sam_hrecs_remove_ref_altnames` call to before the `memmove`, which feels correct.  However it's a change of behaviour that's not documented in the commit.

I initially thought maybe it's because it has no impact, but we're searching for the `idx` matching the key (via kh_get/val).  This then means that `hrecs->ref[idx]` should match `key`.  Eg removing chrX from chr22, chrX, chrY would yield ref[idx] == "chrX", and after the memmove it would be == "chrY", leading to a failed removal call.  (I haven't verified it, but that's my understanding looking at the code).

So I agree with the new code, by proof-reading only rather than testing it in anger, but it needs documenting in the commit message.

jkbonfield: I asked both copilot and claude (before making my statement above, so they couldn't be biased into politely agreeing with me) and both confirmed this is indeed the bug fix I think it is.

daviesrob: It looks like I inadvertently fixed another bug, although one with little impact in practice.  Passing `idx` to `sam_hrecs_remove_ref_altnames()` was done presumably to stop it from removing the name listed in the `@SQ SN` tag from the list.  As it would get the wrong name due to everything being shuffled by the `memmove()` this wouldn't have worked, but the hash entry for `@SQ SN` is also deleted on the very next line here so the outcome is just that we've tried to do it twice (which isn't a problem when it's just calling `kh_del()` as the key entry is not touched).  As the incorrect name that `sam_hrecs_remove_ref_altnames()` looks for shouldn't have been in the `AN` list either (else one name would point to two `@SQ` entries) it should by the time it's done have removed all of the entries that it should have, and none that it shouldn't.

Passing the index for the current record makes more sense for [where `sam_hrecs_remove_ref_altnames()` is called](https://github.com/samtools/htslib/blob/88ae6591e771e5cd103d17161975a2e1ede88154/header.c#L3049) by `sam_hrecs_remove_key()` as in that case we only want to remove the items in the `AN` list.  Here we're blowing everything away so it didn't matter.

daviesrob: I've updated the commit message with a comment about the coincidental bug fix.

PR #2080, opened 2026-08-26
Make check_copyright.pl better at skipping auto-generated files

daviesrob:
Fixes an annoyance when you have build directories with alternate configurations as subdirectories of the main htslib one.  Adjusts the Makefile so auto-generated files get comments saying they were automatically built, and makes `check_copyright.pl` pass any files that include these comments.


jkbonfield:
A sufficient improvement for the current problem, however given this check only exists for the consistency of developers and not end-users, we could recode it to work off git clones and only validate files that are checked in.  (There is zero point in running this on a extracted tarball as it was already checked on tar creation.)

This would fix several other issues, such as having arbitrary temporary files in a development tree that trips up the maintainer checks when run locally. (I often fall foul of this which is why I've given up using it in local checkouts and just wait for the github CI to spot issues.)

PR #2082, opened 2026-08-27
Cram embed_ref=2 and huffman optimisation

jkbonfield:
- Put an upper limit on the number of cram huffman symbols to prevent malformed data consuming a lot of ram and cpu.
- Prevent excessive memory and/or cpu usage in embed_ref=2 for extremely sparse data or huge deletion / ref-skip CIGAR ops.
- Plus a trivial compiler warning fix for fuzz builds.

### DIFF
--- a/cram/cram_codecs.c
+++ b/cram/cram_codecs.c
@@ -1245,16 +1245,11 @@ cram_codec *cram_huffman_decode_init(cram_block_compression_hdr *hdr,
         hts_log_error("Invalid number of symbols in huffman stream");
         return NULL;
     }
-    if (ncodes >= SIZE_MAX / sizeof(*codes)) {
-        errno = ENOMEM;
+    if (ncodes >= 1000000) {
+        errno = ERANGE;
         return NULL;
     }
-#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
-    if (ncodes > FUZZ_ALLOC_LIMIT / sizeof(*codes)) {
-        errno = ENOMEM;
-        return NULL;
-    }
-#endif
+
     h = calloc(1, sizeof(*h));
     if (!h)
         return NULL;
--- a/cram/cram_encode.c
+++ b/cram/cram_encode.c
@@ -1433,8 +1433,11 @@ int next_cigar_op(uint32_t *cigar, uint32_t ncigar, int *skip, int *spos,
     return *cig_op;
 }
 
+typedef uint8_t hist_t;
+#define MAX_HIST_T UINT8_MAX
+
 // Ensure ref and hist are large enough.
-static inline int extend_ref(char **ref, uint32_t (**hist)[5], hts_pos_t pos,
+static inline int extend_ref(char **ref, hist_t (**hist)[6], hts_pos_t pos,
                              hts_pos_t ref_start, hts_pos_t *ref_end,
                              hts_pos_t *ref_end_alloc) {
     if (*ref_end < pos && pos < *ref_end_alloc)
@@ -1454,7 +1457,7 @@ static inline int extend_ref(char **ref, uint32_t (**hist)[5], hts_pos_t pos,
     // Refuse to work on excessively large blocks.
     // We'll just switch to referenceless encoding, which is probably better
     // here as this must be very sparse data anyway.
-    if (new_end - ref_start > UINT_MAX/sizeof(**hist)/2)
+    if (new_end - ref_start > UINT_MAX/sizeof(**hist)/8)
         return -2;
 
     char *tmp = realloc(*ref, new_end-ref_start+1);
@@ -1462,11 +1465,12 @@ static inline int extend_ref(char **ref, uint32_t (**hist)[5], hts_pos_t pos,
         return -1;
     *ref = tmp;
 
-    uint32_t (*tmp5)[5] = hts_realloc_p(**hist, sizeof(**hist),
-                                        new_end - ref_start);
-    if (!tmp5)
+    // hist[6] = {A C G T N total}
+    hist_t (*tmp_hist)[6] = hts_realloc_p(**hist, sizeof(**hist),
+                                          new_end - ref_start);
+    if (!tmp_hist)
         return -1;
-    *hist = tmp5;
+    *hist = tmp_hist;
     *ref_end_alloc = new_end;
 
     // initialise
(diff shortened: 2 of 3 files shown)


cram/cram_encode.c
```diff
@@ -1434,7 +1434,9 @@ int next_cigar_op(uint32_t *cigar, uint32_t ncigar, int *skip, int *spos,
 }
 
 // Ensure ref and hist are large enough.
-static inline int extend_ref(char **ref, uint32_t (**hist)[5], hts_pos_t pos,
+typedef uint8_t hist_t;
+#define MAX_HIST_T (uint32_t)((1LL << (8*sizeof(hist_t)))-1)
```
daviesrob: This is over-complicated, especially as `hist_t` is now `uint8_t`.
```suggestion
#define MAX_HIST_T UINT8_MAX
```

cram/cram_encode.c
```diff
@@ -3967,8 +3986,8 @@ int cram_put_bam_seq(cram_fd *fd, bam_seq_t *b) {
          * The multi_seq var here refers to our intention for the next slice.
          * This slice has already been encoded so we output as-is.
          */
-        if (fd->multi_seq == -1 && c->curr_rec < c->max_rec/4+10 &&
-            fd->last_slice && fd->last_slice < c->max_rec/4+10 &&
+        if (fd->multi_seq == -1 && c->curr_rec+10 < c->max_rec/4+10 &&
+            fd->last_slice && fd->last_slice+10 < c->max_rec/4+10 &&
```
daviesrob: Could we remove the `+10` on both sides of the comparison?

cram/cram_codecs.c
daviesrob: Strictly this now does nothing, but it might be useful to keep in case `FUZZ_ALLOC_LIMIT` is set to something smaller.

cram/cram_encode.c
jkbonfield: Agreed.  It's better than the alternative I was thinking of anyway with const uint32_t instead.

cram/cram_encode.c
jkbonfield: As it stood before, it passed the test on tiny containers of 3 records and switched to multi-seq mode.

It's lost in the mists of time[1] as to why I had the +10 there before, but it was wrong anyway.  I *think* I was trying to do a small sample correction, but that's generally (a+const)/(b+const) rather than one-sided.   The small constant basically means we don't take these decisions to switch mode unless there is a significant number of alignment records, which I wish to keep intact.

[1] Ie it's maybe in io_lib's history somewhere, but it predates the code arriving in htslib.

jkbonfield: Found it.  It arrived in 2013. :-)

https://github.com/jkbonfield/io_lib/commit/ecec3e61432e61ead772bcd00ff2030ede921c8b

There was a lot of code though as it was the creation of the multi_seq mode so the finer nuances weren't documented,

daviesrob: It may have made sense when the +10 was only on one side, but now it's on both and `c->curr_rec+10 < c->max_rec/4+10` is the same as `c->curr_rec < c->max_rec/4` (ignoring the possibility of overflow).

jkbonfield: Doh sorry, implementation failure.  It's meant to be `(c->curr_rec+10) < (c->max_rec+10)/4` I think for a small sample size correction.  What I had before vaguely made sense if it was on the left side (or a -10 on the right obviously), but I'm not sure quite what I was thinking as it was 13 years ago.

Or put another way (if it was floating point):  `(c->curr_rec+10.0)/(c->max_rec+10.0) < 0.25`.

jkbonfield: Basically compare this:

```
$ perl -e '$N=20;for ($i=0;$i<$N;$i+=$N/10) {print $i/$N,"\t",($i+10)/($N
+10),"\n"}'
0	0.333333333333333
0.1	0.4
0.2	0.466666666666667
0.3	0.533333333333333
0.4	0.6
0.5	0.666666666666667
0.6	0.733333333333333
0.7	0.8
0.8	0.866666666666667
0.9	0.933333333333333

$ perl -e '$N=2000;for ($i=0;$i<$N;$i+=$N/10) {print $i/$N,"\t",($i+10)/($N+10),"\n"}'
0	0.00497512437810945
0.1	0.104477611940299
0.2	0.203980099502488
0.3	0.303482587064677
0.4	0.402985074626866
0.5	0.502487562189055
0.6	0.601990049751244
0.7	0.701492537313433
0.8	0.800995024875622
0.9	0.900497512437811
```

If we wanted to trigger behaviour when we drop below, say, 50% capacity then on a large block that's basically curr/max at 0.5.  On a small block we need a more extreme ratio of 0.2ish, to compensate for error from a low sample size.

The +10 is entirely arbitrary of course, but the principle is sound.  Unlike the original (or amended) code!

daviesrob: Well, that makes more sense, although the rewritten version is really just `c->curr_rec+7.5 < c->max_rec/4`.  And to avoid the floating point, you'd probably go for either 7 or 8 on the left.

daviesrob: Apologies for incorrect original version of that, due to mathematical ineptitude on my part...

jkbonfield: I now have `c->curr_rec+8 < (c->max_rec+8)/4` which can only succeed when `max_rec >= 32`.  I'm not sure if this is the correct ratio, but it's sufficient I think to act as a starting point.

All of this is heuristics anyway in how best to optimise the CRAM file and when to switch into multi-ref mode.
The intention was simply if you're consistently producing CRAM containers which are a 1/4tr of their requested size, in terms of the number of alignments you've put into them containing the same reference, then switch out of single-ref mode into multi-ref mode (and use the RI data series to indicate which reference is which instead of a constant value in the slice header).

The purpose was for unsorted data basically, where we were originally creating thousands of small containers with 1 or 2 reads in as the chances of have many reads for the same reference together is slim.  We just don't want to switch to this mode inappropriately, but the bar is deliberately quite high.

jkbonfield: Agreed on it basically being able to be simplified, but it obscures the derivation.  (X+c)/(N+c) is a standard mechanism for small sample size correction in significance tests, but factoring it out just hides this.

It'll all be reordered by the compiler anyway I expect so let's stick with something clearer.  Feel free to add a comment explaining the origins of it if you wish, although it's documented here for anyone who wishes to delve into git blame.

PR #2085, opened 2026-09-07
Simplify crai reading and zlib_mem_inflate()

daviesrob:
Use BGZF interface in place of `hFILE` in `cram_index_load()` so it no longer has to include its own decompression code.  This allows `zlib_mem_inflate()` to be simplified as the only remaining caller both passes in the uncompressed size and rejects data that does not decompress to the size given, so logic for growing the output buffer can be removed.

Fixes a bug reported in #2081 where `zlib_mem_inflate()` could get stuck if the uncompressed size passed to it was 1, but the decompressed data needed more space.

Fixes a bug in `cram_index_load()` where the `hFILE` structure could be leaked if an error was detected when reading the index file.

Adds a quick test to `cram_uncompress_block()` so that GZIP blocks claiming an impossible compression ratio can be rejected without trying to decompress them first.

Closes #2081

### DIFF
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -1340,6 +1340,13 @@ int cram_uncompress_block(cram_block *b) {
         return 0;
 
     case GZIP:
+        if (b->uncomp_size / 2048 > b->comp_size) {
+            // The maximum compression ratio of gzip is 1032
+            // (LZ match of 258 bytes encoded in a 2bit huffman code)
+            // so catch blocks that claim to be wildly over this.
+            hts_log_error("GZIP cram block has impossibly large compression ratio");
+            return -1;
+        }
         uncomp_size = b->uncomp_size;
         uncomp = zlib_mem_inflate((char *)b->data, b->comp_size, &uncomp_size);
 
--- a/cram/cram_io.c
+++ b/cram/cram_io.c
@@ -827,39 +827,26 @@ int int32_put_blk(cram_block *b, int32_t val) {
 
 // Named the same as the version that uses zlib as we always use libdeflate for
 // decompression when available.
-char *zlib_mem_inflate(char *cdata, size_t csize, size_t *size) {
+static char *zlib_mem_inflate(char *cdata, size_t csize, size_t *size) {
     struct libdeflate_decompressor *z = libdeflate_alloc_decompressor();
     if (!z) {
         hts_log_error("Call to libdeflate_alloc_decompressor failed");
         return NULL;
     }
 
-    uint8_t *data = NULL, *new_data;
-    if (!*size)
-        *size = csize*2;
-    for(;;) {
-        new_data = realloc(data, *size);
-        if (!new_data) {
-            hts_log_error("Memory allocation failure");
-            goto fail;
-        }
-        data = new_data;
+    assert (*size > 0);
+    uint8_t *data = malloc(*size);
+    if (!data) {
+        hts_log_error("Memory allocation failure");
+        goto fail;
+    }
 
-        int ret = libdeflate_gzip_decompress(z, cdata, csize, data, *size, size);
 
-        // Auto grow output buffer size if needed and try again.
-        // Fortunately for all bar one call of this we know the size already.
-        if (ret == LIBDEFLATE_INSUFFICIENT_SPACE) {
-            (*size) *= 1.5;
-            continue;
-        }
+    int ret = libdeflate_gzip_decompress(z, cdata, csize, data, *size, size);
 
-        if (ret != LIBDEFLATE_SUCCESS) {
-            hts_log_error("Inflate operation failed: %d", ret);
-            goto fail;
-        } else {
-            break;
-        }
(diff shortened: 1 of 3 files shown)


cram/cram_io.c
```diff
@@ -827,39 +827,26 @@ int int32_put_blk(cram_block *b, int32_t val) {
...
-    if (!*size)
-        *size = csize*2;
-    for(;;) {
-        new_data = realloc(data, *size);
-        if (!new_data) {
-            hts_log_error("Memory allocation failure");
-            goto fail;
-        }
-        data = new_data;
+    assert (*size > 0);
```
jkbonfield, line 837: size comes from b->comp_size which is read from disk.  We shouldn't be using assert of user-data derived fields.

We could just return NULL instead and the standard error handling will pick it up.

cram/cram_io.c
```diff
@@ -919,23 +906,27 @@ static char *libdeflate_deflate(char *data, size_t size, size_t *cdata_size,
 char *zlib_mem_inflate(char *cdata, size_t csize, size_t *size) {
     z_stream s;
     unsigned char *data = NULL; /* Uncompressed output */
-    int data_alloc = 0;
     int err;
 
-    /* Starting point at uncompressed size, and scale after that */
-    data = malloc(data_alloc = csize*1.2+100);
+    assert(*size > 0);
```
jkbonfield, line 911: As above, make this a return NULL instead.

cram/cram_io.c
```diff
@@ -919,23 +906,27 @@ static char *libdeflate_deflate(char *data, size_t size, size_t *cdata_size,
...
+    // These should always be true due to type of cram_block::comp_size
+    // and cram_block::uncomp_size
+    assert(*size < UINT_MAX);
+    assert(csize < UINT_MAX);
```
jkbonfield, line 915: Although it's broken data, size and csize could be negative.  (I've no idea why the original CRAM format permitted this, but I'm assuming because it was developed in Java initially which has no unsigned types.)

Size is already checked above, but `csize < 0` should be checked somewhere and return NULL again.

cram/cram_io.c
```diff
@@ -919,23 +906,27 @@ static char *libdeflate_deflate(char *data, size_t size, size_t *cdata_size,
...
     s.zfree  = Z_NULL;
+    s.msg    = Z_NULL;
     s.opaque = Z_NULL;
     s.next_in  = (unsigned char *)cdata;
-    s.avail_in = csize;
+    s.avail_in = (uInt) csize;
     s.total_in = 0;
     s.next_out  = data;
-    s.avail_out = data_alloc;
+    s.avail_out = (uInt) *size;
```
jkbonfield, line 929: Unnecessary casts.  We should have already checked signed vs unsigned sizes.

jkbonfield (changes requested):

cram/cram_io.c
daviesrob: Currently `*size` can never be zero due to [this check](https://github.com/daviesrob/htslib/blob/e12b88515df87a5463e98d434b25965b648cb2f4/cram/cram_io.c#L1331).  This assert is in place as a guard against future changes that may break that invariant.

cram/cram_io.c
daviesrob: Similarly this can't happen at the moment.  It exists to prevent future accidents.

cram/cram_io.c
daviesrob: cram_read_block() currently [ensures `csize >= 0`](https://github.com/daviesrob/htslib/blob/e12b88515df87a5463e98d434b25965b648cb2f4/cram/cram_io.c#L1188), so this should again never trigger.  Or is it possible to get a block in a way where this might not be true?

Note that [this very old assert](https://github.com/daviesrob/htslib/blob/e12b88515df87a5463e98d434b25965b648cb2f4/cram/cram_io.c#L1336) in `cram_uncompress_block()` does the same check for `uncomp_size`.  Also `csize` and `*size` here are `size_t` so cannot be negative.

cram/cram_io.c
daviesrob: The casts are to suppress compiler warnings about narrowing conversions, should they be in use.

We should really be better at adding casts like this, as they act as a handy warning about places where overflow bugs might be lurking.

cram/cram_io.c
jkbonfield: Ok I see that now, although if there's a check elsewhere then the assert seems pointless.  As you say it's to guard against something later breaking, but a check and return NULL is an equally valid guard.  I just feel uneasy about using assert for data directly decoded from a file, even if it's currently not possible to trigger.

cram/cram_io.c
jkbonfield: No, that's probably sufficient.  However if we're adding more checks to this function then it's probably sensible to add both upper and lower bounds ones so it's trivial to read and understand without having to do a deep dive into all code path that led us here.

That said, I failed to spot they're size_t and not ssize_t, so ignore me. :-)

PR #2095, opened 2026-09-22
Hts opt add validation

jkbonfield:
Adds validation to options specified using hts_opt_add or cram_set_voption.

Mostly this is overflow detection (eg cache_size * 1024) or silly inappropriate values (which often got ignored).  We now issue more errors for malformed configurations.

To consider: instead of validation min to max with hard errors, should we simply cap at min/max and return the min/maximum extent and issue a warning instead?  I suspect the number of people affected either way is close to zero.


hts.c
```diff
@@ -1014,6 +1014,49 @@ scan_keyword(const char *str, char delim, char *buf, size_t buflen)
...
+// If *val is >= it's also the function return value, for ease of use,
+// otherwise we return 0.
+static int get_int_kmg(const char *str, int min_val, int max_val,
+                       int kmg, int *val) {
+    char *endptr;
+    int64_t l = strtol(str, &endptr, 10);
+
+    if (kmg) {
+        switch (*endptr) {
+        case 'g': case 'G': l *= 1024; // fall through
```
daviesrob: This could still overflow on platforms where `long` is a 64-bit type.  Easiest solution would be to put this in a separate variable (say `mult`), then check:
```
if (l < min_val / mult || l > max_val / mult) {
    hts_log_error("Value of out range: expected >= %d and <= %d",
                  min_val, max_val);
    return -1;
}
```

I'd also make `l` a `long` to match the return type of `strtol()`.

hts.c
```diff
@@ -1047,182 +1090,217 @@ int hts_opt_add(hts_opt **opts, const char *c_arg) {
         *val++ = '\0';
 
     if (strcmp(o->arg, "decode_md") == 0 ||
-        strcmp(o->arg, "DECODE_MD") == 0)
-        o->opt = CRAM_OPT_DECODE_MD, o->val.i = atoi(val);
+        strcmp(o->arg, "DECODE_MD") == 0) {
+        o->opt = CRAM_OPT_DECODE_MD;
+        if (get_int(val, -1, 1, &o->val.i) < 0)
```
daviesrob: Why is `min_val` -1 here?  [The manual page](https://www.htslib.org/doc/samtools.html#GLOBAL_COMMAND_OPTIONS) says this should be 0 or 1.

See also `multi_seq_per_slice`, and `embed_ref` - although for that one the documentation is out of date as we don't say what 2 does.

hts.c
```diff
@@ -1047,182 +1090,217 @@ int hts_opt_add(hts_opt **opts, const char *c_arg) {
...
+        if (get_int(val, -1, 2, &o->val.i) < 0)
+            goto err;
 
-    else if (strcmp(o->arg, "no_ref") == 0 ||
-             strcmp(o->arg, "NO_REF") == 0)
-        o->opt = CRAM_OPT_NO_REF, o->val.i = atoi(val);
+    } else if (strcmp(o->arg, "no_ref") == 0 ||
+               strcmp(o->arg, "NO_REF") == 0) {
+        o->opt = CRAM_OPT_NO_REF;
+        if (get_int(val, 0, 1, &o->val.i) < 0)
```
daviesrob: For these binary flags, any non-zero value would have worked in the past.  Should we still allow that?  I note that we do document these as 0 or 1, so this version does enforce what the manual page says.

hts.c
jkbonfield: I chose int64_t as long could be 32-bit, although realistically I doubt we support 32-bit platforms properly anymore given we don't CI test them.  However if we use two variables it doesn't matter.

What I meant to do was add was a min_val/max_val check before the multiplication, which then makes the 64-bit multiplication valid and we can check post-multiply.  (I started with it there and incorrectly moved it rather than duplicated it.)  Your division approach is perhaps simpler.  Thanks

hts.c
jkbonfield: I'd noted the embed_ref=2 omission already, but the documentation is a separate PR as it's in another repo so it'll appear in another PR.

I think the -1 is probably a cut and paste error.  It's used in some other places (despite no documentation) to indicate the default behaviour or as an "auto" mode.  `multi_seq_per_slice` falls into that category.

hts.c
jkbonfield: I did consider that, but also checked the man page and see this as simply enforcing what we'd previously stated.  We could relax it if you cared.

The alternative approach, which is what I alluded to in my commit, is to not make these checks an error but as clipping with a warning.  So >1 becomes 1 and <0 becomes 0.

hts.c
jkbonfield: `if (l < min_val / mult || l > max_val / mult) {` has all sorts of rounding errors given we have integer divisions.  Although I don't think we'd ever hit those cases, I don't like it as it's a hidden "feature" that may bite us if we ever reuse this function elsewhere.  Possibly it doesn't matter and they cancel out, but I'm uneasy with it as the logic is not obvious to me.

This is just argument checking, so I don't care about efficiency and will just do what I originally intended here and check before and after multiplying.

daviesrob: All the multiplications and divisions here are powers of 2, so effectively bit-shifts.  So it's fairly easy in this case to see that the edge cases will work out the way intended.

hts.c
daviesrob: We'll see if anyone complains.

PR #2096, opened 2026-09-22
Add Google OSS-fuzz zips

jkbonfield:
For hts_open_fuzzer we add hts_open_fuzzer_seed_corpus.zip with SAM, BAM, CRAM, VCF, BCF and FASTA files.  The bgzf files are uncompressed (naked), so the fuzzer has a chance of decoding them without hitting checksum issues.

For the expression fuzzer, we add fuzz_expr_seed_corpus.zip too.  This is just a zip version of the corpus.expr.dat file, but it's easier here as it makes the OSS fuzz build recipe simpler.  We could change our Makefile to use unzip instead of a read loop, but it's a tiny file anyway so a bit of redundancy isn't a big problem.

See https://github.com/jkbonfield/oss-fuzz/tree/htslib-seed%2Bdict/projects/htslib for a pending change to OSS-fuzz, once this PR lands.  It's somewhat untested, but I don't know how to validate it locally.


daviesrob:
All the cram files look like they're version 3.0.  Including a few 3.1 files might be a good idea.

It also looks like there's quite a few nearly-identical SAM files.  I expect some could be pruned as I doubt they are contributing much to the discovered code paths.

